PR #18, opened 2015-05-10
Syncs to Bay (5/10/2015)

Neerti:


Neerti:
To explain the last commit, exodus-3 has been an issue for merge conflicts for awhile, and we will be using a different map in the near future anyways, so just deleting it will save time.

PR #26, opened 2015-06-27
Traits

SinTwo:
Adds the framework for traits system.


Neerti:
Travis is borked atm so ignore that.
Might be best to tell Ala bout this before merging, since it seems to be a major feature.

Neerti:
After massive delays, I've reviewed this, and there seems to be a lot of copypasta code w/ the skill system.  I also don't really see the point of adding this system.
Some of the items (IE the engi jumpskirt) seems to be contained inside a single .dmi, which isn't ideal and should be w/ the rest of the jumpsuits, as an example.

PR #35, opened 2015-07-07
Adding the Outpost Z-level

SledgeLP:
Adding of the map it's self, along with changing some names in the shuttle controller file and adding to the outpost areas needed to have the map work.


Neerti:
Everything looks fine, :+1: and all that.

PR #41, opened 2015-07-08
Removes the virus event.

Neerti:
It will be sad to see the event that makes half of the crew spend half of the round inside virology go.  The event that was so fun that it happened three times in a single round.


Spookerton:
:+1:

Neerti:
Ala has declined to sign off on this, soooooo oh well.

PR #46, opened 2015-07-21
Makes map merger clean colony maps.  Band-aid fix for now.

Neerti:
Hopefully we can steal /tg/'s at a later point, which doesn't make assumptions about the map name or how many z-levels exist.

Also you'll get a few errors about how it can't find a few files, that's because the maps for those don't exist yet, so it's nothing to worry about.

-edit-
Bay's KP's map merge is apparently superior to their normal one, sooooo I stole it.


Woodratt:
Ideal

PR #57, opened 2015-08-02
Increases security officer slots to four.

Neerti:


Neerti:
Ala approved, rejoice.

Issue #78, opened 2015-08-03
[Redman's Map] No air alarm outside HoP's office


Neerti:

zaredman:
Fixed in #85

Issue #76, opened 2015-08-03
(REDS MAP) Xenoflora has no air


SledgeLP:
on the left room of xenoflora, it lacks air, but not the other side.

zaredman:
Hole under drying rack.

zaredman:
Fixed in #85

Issue #75, opened 2015-08-03
Tempest Huxley is the only crewmember (Red's map)


serenemelody:
Latejoiners or anybody but him are not on the manifest.

Neerti:
Think it's related to the 'everyone spawns in space' due to Red forgetting to add the title screen.

zaredman:
Reference #71 and #88

Issue #74, opened 2015-08-03
Generators can not meet high-load demand.


zaredman:
Ship running 500,000W load from virtual blackout. Generators can only get half that at best.

zaredman:
Did a thing in #85

PR #84, opened 2015-08-03
Map fixing bash MK-1

zaredman:
Fixes maybe assuming I did it right:
#69 #70 #71 #72 #73 #74 #75 #76 #78 

Also adds a drill to the mining area to get that sweet sweet mining stuff.


zaredman:
Forgot a thing. To be reposted.

Issue #73, opened 2015-08-03
Uranium Fuel Stores are x1 stacks.


zaredman:
Fix levels. Amount var.

Neerti:
Change the 'amount' variable to suit your needs.

zaredman:
Fixed in #85

Issue #72, opened 2015-08-03
Missing power wire under door.


zaredman:
![delete](https://cloud.githubusercontent.com/assets/8041697/9031111/60a073ea-397b-11e5-97b0-0d866e91d90e.PNG)

zaredman:
So apparently #85 does not fix that. To be sorted on the next bash.

Issue #71, opened 2015-08-03
Missing title screen causes everyone to spawn in space?


zaredman:
AAAAAAAAAAAAAAAAAAAA

zaredman:
Mostly fixed in #85 but new issue in #88 Reference that one for updates.

Issue #70, opened 2015-08-03
Teleporter orientation needs to be rotated.


zaredman:
Does not function in current rotation.

zaredman:
Fixed in #85

PR #85, opened 2015-08-03
Map fixing bash MK-1-BRAVO

zaredman:
Repost of #84 with crates I forgot.

Fixes maybe assuming I did it right:
#69 #70 #71 #72 #73 #74 #75 #76 #78

Also adds a drill to the mining area to get that sweet sweet mining stuff.


Neerti:
Not sure if you used the map merger but since you're the only one working on your map, it doesn't matter /too/ much.

PR #124, opened 2015-08-05
Corrects medical access

Mark9013100:
Adds maintenance access to the CMO, considering the paramedic role has it.

Gives the medical doctor role maintenance and EVA access, which they had prior before paramedic was introduced.

Geneticist, chemist, and psychiatrist lose access to each other's areas. But geneticist and chemist get maintenance access in return.

### DIFF
--- a/code/game/jobs/job/medical.dm
+++ b/code/game/jobs/job/medical.dm
@@ -13,10 +13,10 @@
 	req_admin_notify = 1
 	access = list(access_medical, access_medical_equip, access_morgue, access_genetics, access_heads,
 			access_chemistry, access_virology, access_cmo, access_surgery, access_RC_announce,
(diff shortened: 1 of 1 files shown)


zaredman:
Any reason why chem/psych/gen got maintnence access? Or was it requested by @alabasterTempest

Mark9013100:
They have maintenance access since I believe at one point I think they had it before paramedic was introduced.

zaredman:
I'll be honest - and pls dunt kill me AT - As much as the idea of back-alley psych and chemlabs are awesome, not sure if maint access is a good idea. D:

zaredman:
Update. No to the added maint access to chem/psych/gen per @alabasterTempest

Mark9013100:
Right, I'll throw a new PR up soon-ish.

PR #130, opened 2015-08-08
Makes windows require only one sheet per window.

Neerti:


zaredman:
General idea: Allow liquids/raw nutriment but not solid. How hard would it be do to that?

PR #136, opened 2015-08-08
Minor ERT/Central changes

Mark9013100:
Changes the access names of Central Command.
Fixes a tiny oops with the security softcaps.
Adds an ERT equipment locker to replace the locker, not mapped in because map merger is abit wonky for me currently.
Gives the combat/ERT turtleneck suit sensor options. Doesn't change the tactical/mercenary one.


Neerti:
:+1:

PR #182, opened 2015-08-30
Removes the restriction on wearing/taking off clothing

Mark9013100:
Comments out the lines that prevent clothing from being worn/taken off.


Neerti:
If I'm looking at this correctly, this only removes the proc definition and not the checks when you try to put something on/off, meaning it would complain of a nonexistant proc if you tried to put your clothes on/off?

Am I mistaken?

Mark9013100:
Nope, wasn't done, will close and make a new PR.

PR #211, opened 2015-09-14
[DMN] [WIP] Adds several new combat-related changes.

Neerti:
Loads of changes here, I might miss a few things...

# Attackmodes

Some weapons have different 'attack modes', which allow you to use a weapon in a different way.  For example, stabbing with an esword instead of slashing, or focusing on blocking with it.  Changing your attack mode has different benefits and drawbacks.

To change your attackmode, you need to have a weapon in your active hand which has alternate attackmodes.  If you have one, you will see a new button on your HUD.  Click on it to change the mode.
Image of attackmodes

# Staggering and Balance

When you are hit by certain weapons, generally heavy and blunt, you might be staggered backwards.  If this occurs too much in a short period of time, you'll be floored for a short period of time, of which you'll be immune to staggering while downed and for a short period of time when you get up.

Humans have a 'balance' of three points that must be exhaused before they can be stunned.  Note that traditional stuns from sources like flashes and stun batons don't care for balance.  Other mobs also have balance, with more or less points available.

Some sources of staggering are unarmed attacks, shield bashing, heavy thrown items, crossbow bolts, baseball bats, and some weapon's alternate attackmodes, that generally trade damage for knockback.

Image of balance and stagger

# Poise

Some heavy armor will protect you from being staggered or knocked down from stagger, in the form of poise, which I totally did not rip off from Dark Souls.  (I did)

Sources of poise include riot armor, the captain's suit, all voidsuits except for the emergency voidsuit, RIGs, and magboots, but only if they're toggled on.

# Armor Changes

Armor has been changed significantly, which makes weaker armor more relevant and people wearing very good armor not be nigh-invulnerable.

The old armor system involved two dice rolls whenever you got hit.  For example, if you were hit by a melee attack and your armor's melee protection was 40, you had a 40% chance of negating half of the damage.  If you stopped half of it, you would roll again, with the same 40% chance of fully negating the damage.  This meant that weak armor that had low protection values did almost nothing, while very strong armor often resulting in only taking damage rarely.

The new system involves a blend of reliable damage soaking and some luck involved.  When you are hit, it uses the armor protection as a base amount to protect you from, then it does a roll between +25% and -25% of that base protection, and adds it to the base protection.  The result of that is how much damage is reduced from the attack.

For example, if you are hit by an esword in the chest while wearing armor that has 50 melee protection, the base protection is 50, then RNG can make the final protection be between 37.5 and 62.5 percent.  The damage you'd take could range from 11.25 to 18.75, instead of 0, 15, or 30.


Mark9013100:
SS13: Prepare to Die Edition

Neerti:
Now I can poisebreak spiders and beat them to death.

Neerti:
Gonna try to salvage this onto the dev2 branch.

PR #222, opened 2015-09-18
Improves helmets and fixes shields

Mark9013100:
Removes the very old helmet(including riot and SWAT) sprites. New regular helmet sprites are still in the files, but unused.

Gives the SWAT, riot, and regular helmet ausop's sprites, which are darker, and look better with the armor vests.
The visors of the regular helmet and riot helmet can now be adjusted, uses the same code as flipping caps.

Fixes the un-activated e-shield not being able to go on ears.
The un-activated telescopic shield can no longer be placed on the back.
Minor edit to the regular balaclava sprite.

Alternative headsets renamed to bowman headsets. Excluding the syndicate one.

Sprites in action:
http://imgur.com/a/jlcqz


Neerti:
I think removing the ability to hold closed eshields in your ear was intentional, but it was hilarious so why not.

We can always change it later if need be.

PR #232, opened 2015-10-01
Minor edits to the sec lockers

Mark9013100:
Removes the flare mapped on top of the security lockers, removes the armbands/holsters mapped on top of the security wardrobe.

Adds armbands, holsters, and randomized backpacks to the security wardrobe.

Dunno why the other map files had a line added to them, but nothing came up when I tested this.


Mark9013100:
Closing since this edits the current map.

PR #237, opened 2015-10-11
[FIX] Makes Deckard's Gun Usable

Datraen:
Prior to fix, Deckard's gun was not usable.
Woodrat told me that it was someone's custom item's sprite, but if they don't play any more on Bay, we might as well have it available for anyone that's interested.


Neerti:
Since the last 'dev is broken fiasco' was caused by a huge screwup involving me trying to sync changes made on master to dev, could you resubmit this targeting dev2?

PR #287, opened 2015-11-05
[BOTANY] Plant Tweaks

Datraen:
Tweaks plants to create a more diverse selection, different environment constants.


Mark9013100:
Is acetone and hydrazine even in the code anymore? Thought those were reverted by Neerti.

Datraen:
Right
I was lazy and copy-pasted
Will remove them.

PR #288, opened 2015-11-06
Minor ERT edits, 11/5/2015

Mark9013100:
Edits the ERT's fluff, also raises the max cap back to six(bay lowered it to 5 for whatever reason), small-ish edits to special operations.


Mark9013100:
Bleh, closing since another map update was done.

PR #309, opened 2015-11-20
Ports the new Character Setup + Fixes socks (AGAIN)

Neerti:
Now with less bald thirty year old men.

Also fixes character setup icon not updating.


Neerti:
This also fixes the character setup icon not updating properly.

PR #314, opened 2015-11-23
Refactors holidays slightly

Neerti:
Refactors holiday system slightly.  It's now a list based system, so you can have multiple holidays on the same day, in case you wanted to do that.
Removes the dumb holidays.
Adds more real life and fictional holidays, and adds a blurb for all of them.
Taj holidays added w/ Serene's blessing, and warble holidays sourced from existing skrell lore.


Datraen:
Thanksgiving should be updated to a similar system to Fri the 13th, as a 29 > x > 20 on day and searching for Thursday. Black Friday for the day after on a similar check.

PR #355, opened 2015-11-29
Moderator Logs

SinTwo:
Adds ability for mods to see attack logs and such.


code/modules/reagents/Chemistry-Logging.dm
```diff
@@ -18,7 +18,7 @@
 	set name = "Show Chemical Reactions"
 	set category = "Admin"
 
-	if(!check_rights(R_ADMIN))
+	if(!check_rights(R_ADMIN|R_MOD))
```
Neerti: Should this be || ?

Datraen: For some reason it's | and not || in dm, should work as intended as coded.

Neerti: Well the DM reference lied to me.

PR #358, opened 2015-11-30
Adds dev accesses

Datraen:
Devs now have access to:
msay
ahelps
jumping to things in msay.

Yay being able to interact with things devs can see.
Also makes moderators moderators, instead of admins


Datraen:
Last commit was because git was insisting I merge master into my branch.

Issue #338, opened 2015-11-28
NanoUI maps are incorrect


EmperorJon:
The NanoUI maps are offset a reasonable amount and the windows seem to be too small to contain the entire map of the Northern Star.

This means that camera/suit sensors maps are almost unusable because they're extremely confusing to navigate.

GinjaNinja32:
Not fixed.

EmperorJon:
This is fine.

PR #361, opened 2015-11-30
Fixes the map for NanoUI. 

EmperorJon:
It's now 1200x1200px not 1024x1024px due to our dimensions.

Necessary changes to nanomapgen.dm and various image files to generate new nanoui maps have not been committed. These changes were required because our station's map is 300x300 tiles, not 255x255 tiles like the Exodus.

Changes were needed to layout_default.css to ensure the map shows properly in the UI at 300x300.
This should work fine for now, but other changes can be pushed through to bring the tools in line with our map's size if necessary.


EmperorJon:
Oh, right, fixes #338

PR #381, opened 2015-12-02
Fixes hemostat surgery

EmperorJon:
Changes the arbitrary failure return for surgery from 2 to -1. This is because ORGAN_BLEEDING in damage_organs.dm is now 2 so some statements evaluate to 2, which causes bleeding clamp surgery to fail as it triggered the "This is an error that's already been dealt with" condition coincidentally.

Surgeries which report failures (like trying to stick a brain in someone who's already got one) now throw back a -1 not a 2.


Zuhayr:
Would be better practice to #define SURGERY_FAILURE -1 somewhere and use that instead of a magic number, while you're already changing all those lines.

EmperorJon:
Indeed, I was considering putting it in the damage defines with the rest to ensure that it's not one of the same, but as Psi's done it in a surgery define already that's solved.

EmperorJon:
Back-picked Psi's changes from https://github.com/Baystation12/Baystation12/issues/11560. :)

polaris.dme
```diff
@@ -5,6 +5,72 @@
...
+#define FILE_DIR "sound/items"
+#define FILE_DIR "sound/machines"
+#define FILE_DIR "sound/mecha"
+#define FILE_DIR "sound/misc"
+#define FILE_DIR "sound/music"
+#define FILE_DIR "sound/piano"
+#define FILE_DIR "sound/violin"
+#define FILE_DIR "sound/voice"
+#define FILE_DIR "sound/voice/Serithi"
+#define FILE_DIR "sound/weapons"
```
PsiOmegaDelta: Interesting additions to the dme.

EmperorJon: Good job noticing that. As I'm sure you can tell this is all rather a learning experience for me.

Neerti: @EmperorJon look at this: http://polaris.spacetechnology.net/title/Getting_Your_Pull_Accepted/#3._Do_not_automatically_add_FILE_DIR.

PR #398, opened 2015-12-02
Makes round-end last five minutes instead of one.

Neerti:
Also adds THE FINAL COUNTDOWN in the chat log, so you can have your five minutes of RP on Central Command.

If the station gets nuked, the round ends in one minute, as it did before.


GinjaNinja32:
byond lets you do `"[number] minute\s"` instead of the not-as-nice-to-read `"[number] [number == 1 ? "minute" : "minutes"]"` :P

PR #410, opened 2015-12-03
Allows you to define a base turf for an area

EmperorJon:
Overrides the z-level's base turf in unsorted.dm.

It will first check to see if the turf is set, and if not it will still use the z-level default.
This means we can mark areas as having a base of space to override the z-level base of asteroid that we're using.
As a result, we can ensure the bombs, shuttles moving, etc. leave behind an space turf by defining shuttles, solars/"external" areas as using space as a base.
Areas will need to be updated to check with base_turf = /turf/space and references to the old get_base_turf proc updated so all calls (bombs, Nar'Sie, deconstruction, etc.) have the correct effect, but it's tired and I want to commit this now so I'm committed to finishing it tomorrow.


EmperorJon:
"but it's tired"

Yep.

PR #419, opened 2015-12-04
Fixes some /tg/ ported floor sprites

Sabess:
Does what it says on the tin. Making this PR was a fucking headache so I /may/ of fucked it, but that shouldn't be the case. But yes, I have tested this (at least the ones that are actually coded, there's a lot of sprites in there that you can't spawn or acquire in game at all) and it's working, so here's hoping nothing exploded in between the testing and this PR.


Neerti:
This appears to be in order.

PR #430, opened 2015-12-05
Fix for record problems in #220

### LINKED ISSUE #220: Records are FUBAR'd.
Hard to explain, but across all records consoles, all details(aside from stuff like age, male, race), will be switched around. IE, Crewmember A will have his records, Crewmember B, C, and D will have A's records. This also applies with criminal/arrest status as well. For example, crewmember A will be set to released, and this'll apply to B, C, and D as well.

Reproduction:
Have a decent amount of people with records set up.
Start viewing records and see who has who's records.
Attempt to set someone's criminal status(not sure if medical status or whatever it's called is bugged as well) and look at a security record's console.

MagmaRam:
Because of the limitations of BYOND's bitwise operators (which affect both the old and new implementations of num2hex()), you can't use numbers larger than 2^16 with that proc.

I cut the number of digits down to 4 and changed the maximum on the RNG to match. If you want to restore the IDs to six digits, you'd have to use decimal numbers, bypass num2hex() entirely, or rewrite num2hex() in a way that supports larger numbers than 2^16.

Fixes #220.

### DIFF
--- a/code/datums/datacore.dm
+++ b/code/datums/datacore.dm
@@ -166,7 +166,7 @@
 	if(H.mind && !player_is_antag(H.mind, only_offstation_roles = 1))
 		var/assignment = GetAssignment(H)
 
-		var/id = add_zero(num2hex(rand(1, 1.6777215E7)), 6)	//this was the best they could come up with? A large random number? *sigh*
+		var/id = add_zero(num2hex(rand(1, 65535)), 4)	//no point generating higher numbers because of the limitations of num2hex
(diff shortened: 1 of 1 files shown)


Datraen:
Looks good to me.

MagmaRam:
Yes. Both auto-made and player-made records will run the full gamut of values from 0001 to FFFF.

Datraen:
Alright, yeah. Looks good, makes sense.

PR #437, opened 2015-12-05
Adds ability to smash drinking bottles on things other than people

Yoshax:
Now you can smash a bottle on the bar before your bar fight and threateningly wave it around!


code/modules/reagents/reagent_containers/food/drinks/bottle.dm
```diff
@@ -73,6 +73,19 @@
...
+/obj/item/weapon/reagent_containers/food/drinks/bottle/verb/smash_bottle()
+	set name = "Smash Bottle"
+	set category = "Object"
+
+	var/list/things_to_smash_on = list()
+	for(var/atom/A in range (1, usr))
+		if(A.density && usr.Adjacent(A) && !istype(A, /mob))
+			things_to_smash_on += A
+
+	var/choice = input("Select what you want to smash the bottle on.") as null|anything in things_to_smash_on
```
PsiOmegaDelta: The user may cancel the input which leaves choice as null, need to  add
```
if(!choice)
    return
```

You also need to recheck if the object is still dense and that the user is adjacent after the input, as both states may have changed while the user was considering his options, thus:
```
if(!(choice.density && usr.Adjacent(choice)))
    // Insert options message to user here, about choice not being available anymore.
    return
```

code/modules/reagents/reagent_containers/food/drinks/bottle.dm
```diff
@@ -73,6 +73,19 @@
...
+	usr.put_in_hands(src.smash(usr.loc, choice))
+	usr.visible_message("<span class='danger'>[usr] smashed the bottle on [choice]!</span>")
```
PsiOmegaDelta: "<span class='danger'>\The [usr] smashed the bottle on \the [choice]!</span>"
You may also wish to add a custom message to the user itself, as a second argument.

Neerti:
I can't see any glaring issues from this so :+1:

PR #439, opened 2015-12-05
Adds an "emo fringe" hairstyle

EmperorJon:
Just emo long without the long bits. I quite like it so whatever.


Spookerton:
Oh!
So when do you want your sexual favors?

PR #444, opened 2015-12-05
Adds cellphones!

Neerti:
Details in the changelog.  Rushing this because server updates.

Adds the communicator, a (loadout only atm) that allows you and someone off station (a ghost) to talk ICly, so long as telecomms stays on.  The character that the ghost uses is that ghost's currently loaded character in character setup.

To use, find and register a communicator by clicking on it, then wait for a ghost to send you a request with the Call Communicator verb.  If a ghost sends you one, then you can 'pick up' the call, and then you can talk!

The ghost will be placed into the device itself as a new mob, with only the ability to talk and emote (it has a video feed).  The call can end if one of the users hands up, or the radio experiences issues.


Spacemanspark:
That's... really fucking cool, actually.

PR #463, opened 2015-12-07
Added more reskins for the M1911 and some other stuff.

Sabess:
Added more reskins for the M1911 all of which have different sprites when empty. Added this functionality for the default M1911 as well as the default MK58s. Thanks to Zuhayr for helping me with some of this, and also... whoever made some of these sprites? I got them from a repository, mostly.

USP sprite is edited from an old deagle sprite by me, P08 brown is edited from a Luger sprite that was in gun.dmi but isn't even in the code at all, empty mag gun sprites for the VP78 and MK58s also by me.

Most of the sprites (will be) on here.
http://polaris.spacetechnology.net/forums/viewtopic.php?f=16&t=153


code/modules/projectiles/guns/projectile/pistol.dm
```diff
@@ -8,10 +9,80 @@
...
 	load_method = MAGAZINE
 
+//For some reason you have to add one of these "if else" statements on the M1911 reskins or the code breaks.
+//If you're not working with two different sprites, just set them to the same icon_state.
+
 /obj/item/weapon/gun/projectile/colt/detective
 	desc = "A Martian recreation of an old Terran pistol. Uses .45 rounds."
 	magazine_type = /obj/item/ammo_magazine/c45m/rubber
 
+/obj/item/weapon/gun/projectile/colt/detective/update_icon()
```
Neerti: Why are there seven separate update_icon()s that appear to be for the exact same type?  Assuming that this works, a much better way to have this work would be to first make all the icon_states with empty mags have '-e' at the end, then you can simply do icon_state = "[unique_reskin]-e" if the gun is empty, with one update_icon(), and if the gun is not empty, then set icon_state = unique_reskin .

code/modules/projectiles/gun.dm
```diff
@@ -309,6 +309,14 @@
 
 	return !P.launch_from_gun(target, user, src, target_zone, x_offset, y_offset)
 
+//aprt of reskins that have two sprites, no touch please
+/obj/item/weapon/gun/projectile/colt/detective/attack_hand(var/mob/living/user)
```
Neerti: This would probably be better as a simple object verb, and put in the same file as the detective guns.  The bonus is that once you reskin your gun, you can remove the verb from the weapon so you can't do it again, as verbs are actually contained in a list named verbs.

Neerti: I'm actually dumb, it's already an object verb.

code/modules/projectiles/guns/projectile/pistol.dm
Neerti: Here's (untested) code you can use to make this much much cleaner, tho you may need to rename a few things in the .DMI
https://gist.github.com/Neerti/58a272e7150ed227d889

zaredman:
Updated your OP on this @Sabess so you were linking to your post and not "CLICK HERE TO EDIT THAT POST"

Sabess:
...Pfff. Oops. Anyway, this is fixed as per Neerti's suggestion.

mwerezak:
Mk58s were Raptor's I believe.

PR #496, opened 2015-12-09
1.2.6 Bar Sign and Power tweaks

Woodratt:
- Created prescription Huds for medical and security. Two exist in the medical exam room, two in the security restroom.
- Boot Knives Bar Sign
- Tactical light in the warehouse replaced with a more low speed option
- Adjusted Atmos SMES to input from round start
- Adjusted Main SMES to start with about 70% power and a 700k Output.


Neerti:
:+1:

PR #504, opened 2015-12-10
Polaris fixes

GinjaNinja32:
fix #501, #477, #424, #389, #482, #447

PR #505, opened 2015-12-10
Ports Fix for Ninja Camera Teleport Exploit

### LINKED ISSUE #501: Ninja can teleport to turfs visible on camera console
Not tested with laptops. Massive oversight that allows a ninja to use their "within vision" turf jumping ability to jump even between Z-levels if necessary. Discovered when a ninja teleported from security to the mining outpost.

Hubblenaut:
Fixes #501


GinjaNinja32:
#504 has this already.

Hubblenaut:
Ninja'd.

PR #524, opened 2015-12-11
Job age refactor.

Zuhayr:
DNM for testing. Implements the ability to restrict jobs to minimum ages.

Ages are placeholders, if you disagree with them please comment.

### DIFF
--- a/code/game/jobs/job/assistant.dm
+++ b/code/game/jobs/job/assistant.dm
@@ -12,8 +12,6 @@
 	access = list()			//See /datum/job/assistant/get_access()
 	minimal_access = list()	//See /datum/job/assistant/get_access()
 	alt_titles = list("Technical Assistant","Medical Intern","Research Assistant","Visitor", "Resident")
-	minimum_character_age = null
-	ideal_character_age = null
 
 /datum/job/assistant/equip(var/mob/living/carbon/human/H)
 	if(!H)	return 0
--- a/code/game/jobs/job/assistant.dm
+++ b/code/game/jobs/job/assistant.dm
@@ -12,6 +12,8 @@
 	access = list()			//See /datum/job/assistant/get_access()
 	minimal_access = list()	//See /datum/job/assistant/get_access()
 	alt_titles = list("Technical Assistant","Medical Intern","Research Assistant","Visitor", "Resident")
+	minimum_character_age = null
+	ideal_character_age = null
 
 /datum/job/assistant/equip(var/mob/living/carbon/human/H)
 	if(!H)	return 0
--- a/code/game/jobs/job/captain.dm
+++ b/code/game/jobs/job/captain.dm
@@ -18,6 +18,11 @@ var/datum/announcement/minor/captain_announcement = new(do_newscast = 1)
 	minimal_access = list() 	//See get_access()
 	minimal_player_age = 14
 	economic_modifier = 20
+
+	minimum_character_age = 30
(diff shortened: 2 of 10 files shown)


Hubblenaut:
Out of curiosity, why does the QM have a higher minimum age than the HoP?
Also, what's with the HoS?

EmperorJon:
Whilst I dislike 20 year old Captains as much as the next guy, I can see this suggestion being an extreme salt mine. 

What does interest me though is the idea of prioritising older people in the job-roll at round start. I've seen a few times where you have characters in their forties or fifties lose out on their head-slot to a 26-year-old or being assigned it, which always amuses me slightly. I imagine it would need a lot of discussion and I know for a fact that people will disagree, but personally I like that idea.

sabiram:
35 is going to restrict a lot of people who presently play. I think 25 would be entirely reasonable.  The shift we play doesn't have to be the best of the best every single time.

zaredman:
Honest opinions. 

Command? 30s. 
Captain? 30s. Maybe 35. 

Non-Command Security? 18. Or 21.
Warden? 21. Or 25.

Science? 18. 

Civilian? 18.

Engineering? 18.

Yoshax:
I don't see the need for it really, it seems unnecessary. If overly young characters are a problem it can be dealt with by staff, it doesn't need to be hard coded to prevent it. Some people can play it well, and with good reasons and suchlike, they shouldn't be punished for a few people who just play lol 18 year old doctors/captains. I think it's fine as is.

Hubblenaut:
> 35 is going to restrict a lot of people who presently play. I think 25 would be entirely reasonable. The shift we play doesn't have to be the best of the best every single time.

25 seems very young for command roles

sabiram:
25 may not be the number we want but 35 will be pretty restrictive. At 25 you can have a degree and months, if not a few years, experience in your field. Maybe they've entered a NanoTrasen managerial training program. Everyone has to start somewhere, and this is a safe station, attached to a colony, no less. 

I agree with Yoshax that the code should work for the players. I would consider the reasons why there aren't many older characters being played before imposing restrictions (not that 35 is old, anyway).

PsiOmegaDelta:
Head positions are actually using age to give characters of a certain age a greater likelihood of being selected for the role, as seen [here](https://github.com/Baystation12/Baystation12/blob/master/code/game/jobs/job_controller.dm#L130).
The code for filling roles is a bit of a mess, but perhaps these new age limits for roles could be utilized in a similar way.

zaredman:
Hi server admin here on two servers. 

Having a hard-coded "MUST BE THIS OLD TO BE" thing would be fantastic as there are really more important things to do than have Staff deal with 18 y/o new Cappy. 

25 is hella young for a Captain of an entire, complex, and expensive spacecraft/settlement/whatever. Most Merchant Marine cappies are 30 at the youngest iirc.

Zuhayr:
They are checked in this code @PsiOmegaDelta, you could finnagle the behavior described in this thread by making the min 17 and ideal 70.

PsiOmegaDelta:
Ah, so I see.

Zuhayr:
This system will now simply prioritize certain ages for certain jobs. If minimum ages are desirable down the track, they can be added very easily.

sabiram:
How heavy is the weighting? Will an older character always get the slot over a younger?

Zuhayr:
No. It's the same weighting that exists currently. You can look at the specific values in the code.

Zuhayr:
Yes or no?

Issue #532, opened 2015-12-11
Lattices built in space create sand tiles


Woodratt:
As seen in the title. Placing down lattices on space turfs turn them into asteroid turfs

PR #533, opened 2015-12-11
Space's default turf is space. 

EmperorJon:
We'll no longer get sand when we place down lattices. You'd think that was pretty clear and easy to remember to do, right?

Wrong. Apparently.


EmperorJon:
Fixes #532.

PR #538, opened 2015-12-12
Telecrystal Refactoring

Datraen:
Refactors Telecrystals into individual pools, adds a new command to allow for admins to modify telecrystal levels.

Things that don't work:
Uplink Implant does NOT have a separate pool. Will look into another time. Price adjusted down and noted.

EDIT: Oh, and the command is logged and sent to all admins/mods.

### DIFF
--- a/code/modules/admin/admin.dm
+++ b/code/modules/admin/admin.dm
@@ -1326,3 +1326,19 @@ proc/admin_notice(var/message, var/rights)
 	message_admins("[key_name(usr)] attempting to force mode latespawn.")
 	ticker.mode.next_spawn = 0
 	ticker.mode.try_latespawn()
+	
+/datum/admins/proc/set_tcrystals(mob/living/carbon/human/H as mob)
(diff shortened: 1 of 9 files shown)


code/modules/admin/admin.dm
```diff
@@ -1326,3 +1326,19 @@ proc/admin_notice(var/message, var/rights)
...
 	ticker.mode.try_latespawn()
+	
+/datum/admins/proc/set_tcrystals(mob/living/carbon/human/H as mob)
+	set category = "Debug"
+	set name = "Set Telecrystals"
+	set desc = "Allows admins to change telecrystals of a user."
+	
+	var/crystals
+	
+	if(R_ADMIN)
```
GinjaNinja32: what

Datraen: I'd like to mention that Neerti caught it prior to server merge. I was also kind of out of it at the time.

PR #541, opened 2015-12-12
Adds a detective belt and other edits for security

Mark9013100:
Adds the Forensic utility belt for detectives, can be found in the detective's closet and from the forensic investigation equipment crate.
Gives black gloves to security officers, to help avoid contaminating evidence with their fingerprints.
Increases the amount of station bounced radios, flashlights, and crowbars in the brig.


Minijar:
Why does a forensic guy need a belt? They have a crime scene kit to carry everything.

Mark9013100:
The belt is more convenient than the kit, having to carry around a briefcase everywhere isn't so practical.

NortonDK:
The kit can only whats in it from the beginning, some more slots in it would be nice!

Neerti:
This will also help stop certain people from foaming at the mouth because a detective grabbed a secbelt, since now that slot's occupied by the forensics one.

PR #543, opened 2015-12-12
Fixes admin check in new set_telecrystal proc.

Neerti:


Datraen:
It worked perfectly fine while testing as +HOST...

PR #547, opened 2015-12-13
Adds set_tcrystals() to hidden adminverb list.

Neerti:
As requested by Datraen.


Datraen:
Thanks, currently out.

PR #548, opened 2015-12-13
HEADPHONES.

EmperorJon:
DO THEY WORK PROPERLY NOW? OF COURSE!

Okay, now that's all sorted, Zuhayr's sprites, so kudos.
They cost 1 point in the loadout menu. Work just like earmuffs with added verb functionality to toggle cool animation. That's all.


Minijar:
Sooo they don't play music? Might be some cool functionality, to let them have the same functionality as the jukebox music, but only have the wearer able to hear the selected music.

EmperorJon:
I don't really see the need for that, given that if you want to listen to music just play music... all they do is have a cute animated musical note feature so you can stand around looking like you're listening to music. I could also add functionality whereby when you turn them on it asks you what you're listening to, which it then labels them as, so people who see you wearing it or are nearby can see what you're listening to and pretend they can hear it, etc.

PR #555, opened 2015-12-13
Makes table slams a bit more robust.

Dibasic:
Table slams are not viable in fights because they take many more actions and do very little damage. Slamming people's faces into tables is cool and should happen more in fights that are near tables.

Increased table slam damage from 8 to 15. Increased weaken chance from 15% to 25%. Increased chance of embedding shards from 1% to 5%.

The DPS of fighting this way is still probably about half of just punching, and is much harder, and if you can tableslam you can table them and stomp their face in anyhow, so this isn't gonna change the meta or anything.

### DIFF
--- a/code/modules/tables/interactions.dm
+++ b/code/modules/tables/interactions.dm
@@ -84,8 +84,8 @@
 				return
 			if (G.state < 2)
 				if(user.a_intent == I_HURT)
-					if (prob(15))	M.Weaken(5)
-					M.apply_damage(8,def_zone = "head")
+					if (prob(25))	M.Weaken(5)
+					M.apply_damage(15,def_zone = "head")
 					visible_message("<span class='danger'>[G.assailant] slams [G.affecting]'s face against \the [src]!</span>")
 					if(material)
 						playsound(loc, material.tableslam_noise, 50, 1)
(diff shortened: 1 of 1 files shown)


Dibasic:
http://polaris.spacetechnology.net/forums/viewtopic.php?f=18&t=225&p=1221#p1221

Zuhayr:
I feel like it would be better to derive these values from the material used for the table.

EmperorJon:
I'm agreeing with Zuh here. It would be nice to see more sturdy tables have more force, so slamming someone into a soft plastic table negligible when compared to slamming their face into a marble counter. >:D

Zuhayr:
Is this going to be updated?

Dibasic:
If I can figure out how to check the material of a table, I'll implement
it that way. I think the idea is cool, but I'm not sure how to do it.

mwerezak:
Checking the material of a table is fairly simple, just get the material instance and look at it's integrity var or whatever it was called.

Zuhayr:
The table has a reference to the material datum on it already, you'd just have to check src.material.integrity or src.material.hardness.

GinjaNinja32:
`material` might be null (if it's just a frame), check that first

Zuhayr:
PR is now a week stale, I'm going to close it for the time being. When you update it feel free to reopen.

PR #563, opened 2015-12-14
Adds Brass Knuckles

Sabess:
Does what it says on the tin. The idea is that they don't have an in-hand sprite on purpose, so that they're slightly more stealthy than a lot of melee weapons. They do less damage than any other melee weapon that's craftable from metal sheets, however. Bats, butterfly knives and etc do more damage, so.


code/modules/materials/material_recipes.dm
```diff
@@ -40,7 +40,7 @@
 		new/datum/stack_recipe("purple comfy chair", /obj/structure/bed/chair/comfy/purp, 2, one_per_turf = 1, on_floor = 1), \
 		new/datum/stack_recipe("green comfy chair", /obj/structure/bed/chair/comfy/green, 2, one_per_turf = 1, on_floor = 1), \
 		))
-
+	recipes += new/datum/stack_recipe("brass knuckles", /obj/item/weapon/brassknuckles, 1, time = 10, one_per_turf = 0, on_floor = 1)
```
GinjaNinja32: `on_floor` does not make sense for this recipe, it's for things that need to be built on a floor (e.g. chairs, racks, closets).

Yoshax:
You should make them like bats and let them be made out of anything.

code/modules/materials/material_recipes.dm
Sabess: Bats and etc use it, so.

GinjaNinja32: all it does is force you to be on a floor when building it, it's literally pointless for bats and other items you can pick up

Sabess: It should probably be fixed all round, then. :v

Hubblenaut:
"A pair of brass knuckles" is kind of duplicative to what it will say on examination already.

Also needs the CONDUCT flag, matter, origin technology and maybe a few more attack verby than only 'punched'.

GinjaNinja32:
they should likely be plural too, "That's a brass knuckles" vs "That's some brass knuckles"

Sabess:
If they weren't at least very very hard to notice, they'd be quite literally useless. As far as I know, bats and etc don't have origin tech, attack verbs are completely flavour and most things don't even have them, making them modular would make them OP (afaik a uranium bat, for example, does more damage than a steel one) annnd I'm not really sure what a conduct flag is 'cause I've never seen anything use that. Plural thing I can fix.

Hubblenaut:
Other objects missing some properties for fully incorporating them into the game mechanics is not a reason to add more of them, really.

The conduct flag will let the game know that your item conducts electricity, important so people can't abuse its damage value to break through electrified grids.

Zuhayr:
If you make these a material-derived weapon then the system will take care of everything concerning conductivity etc. for you.

code/modules/materials/material_recipes.dm
Hubblenaut: Weird place to put it, btw. Why between the comfy chairs and station constructs? Would probably make more sense at the top.

code/defines/obj/weapon.dm
```diff
@@ -119,6 +119,15 @@
 	else
 		..()
 
+
+/obj/item/weapon/brassknuckles
```
Hubblenaut: You are putting this definition inbetween two conceiled cane-procs. D:

Hubblenaut:
What is brassknuckles1111 for?

Sabess:
Left it there if anyone wanted to use the sprite for anything (if they wanted black knuckles for whatever reason), just called "1111" because it's extremely unlikely anybody would accidentally use that with anything.

Hubblenaut:
Put the definition into /code/game/objects/items/weapons/material/misc.dm and make it a subtype of /obj/item/weapon/material, then delete the flag var again so the system can handle it for you.

Sabess:
It's not intended as a material weapon. :I

Hubblenaut:
You're forging it out of a steel sheet, so by definition it is a material weapon - a weapon made of a material. As you currently have it, your ecipe will only show up on steel sheets.

Sabess:
But that doesn't mean it's intended to work for every single material.

Hubblenaut:
As I said, it will only show up on steel sheets.

Sabess:
Exactly. Which is the point. Because it's not intended to work for every single material.

Hubblenaut:
Soooo where is the issue?

Sabess:
As far as I'm aware, making them a material weapon would give them properties relevant to being a material. As in, they'd be called "steel brass knuckles", possibly be modular (which isn't intended but I'm not sure if that's different code), change color, etc. Unless I'm mistaken and you're asking for something else.

Hubblenaut:
Having it be a subtype of weapons/material is important so the game will actually handle your item as the material it is made of. This is especially helpful when making changes to the material properties or upcoming features like the melting of metal items, burning of wood, ect. Yes, the item would be called "steel brass knuckles", because they are, after all, made of steel. If you want to imply that they are actually brass, you could add brass as material and make it available somewhere - with the current system that can be done in a matter of a minute. If you don't want it being called that way because it looks weird (which I agree with to some point), overwrite the set_material proc to not set a name (the snowflake way) or better yet, have the system check per variable if the name should be overwritten.
Set applies_material_color = 0 so it will use the icon as is.

Datraen:
I've discussed this with Sabess and it seems that it is not intended to have material properties with concerns of the brass knuckles being rather over powered if it was made of certain materials. Creating a material for the sole purpose of making the brass knuckles seems to be a bit much for what I understand is desired as an end product.

While I personally wouldn't mind seeing material implementation of the brass knuckles, I can understand the concerns. Creating the subtypes and balancing them would be extra work for something that is not desired in the first place.

Hubblenaut:
> [...] with concerns of the brass knuckles being rather over powered if it was made of certain materials

I thought it should only be made of steel?

Hubblenaut:
Before you work yourself out any more here, let me do this for you.

Datraen:
Yeah, it should be restricted to metal.

Never mind my ramblings, I'm responding while studying for a final.

Hubblenaut:
#600

Sabess:
...What you're doing with that is "I disagree with this, let me do it for you". Not how it should be.

Hubblenaut:
Hm? What is wrong with it?

zaredman:
Throwing a DNM flag on these until (1) Y'all figure out which one is the actual one and (2) @Neerti reviews this for QC purposes.

Sabess:
Updooted.

Hubblenaut:
Yeap! You got it, that looks good. Will close my PR then.
Don't forget though, it should be 'gender = PLURAL', not 'gender = "plural"'

If you don't want the item to show as 'steel brass knuckles'; I suggest the following:
In material_weapons.dmi:

'var/set_material_name = 1'

    /obj/item/weapon/material/proc/set_material(var/new_material)
	material = get_material_by_name(new_material)
	if(!material)
		qdel(src)
	else
		health = round(material.integrity/10)
		if(set_material_name)
			name = "[material.display_name] [initial(name)]"
		if(applies_material_colour)
			color = material.icon_colour
		if(material.products_need_process())
			processing_objects |= src
		update_force()

Sabess:
Huh. Thought I fixed the gender thing. Evidently it didn't save properly. It still shows up as just "brass knuckles" in my experience, but I'll test it again and just change it to "knuckle dusters" with the black version of the sprite if it shows up differently on another test.

Zuhayr:
:+1:

Issue #512, opened 2015-12-10
Flavour text is blanked out on attempting to edit


sabiram:
On click, the box pops up to allow you to edit, but the box is empty. Expected behavior would retain the previously entered text, like employment/medical records presently do.

SinTwo:
Fixed! Except for the general field.

Zuhayr:
Fixed or not fixed, Sin?

EmperorJon:
I can confirm that this is not fixed. I'll take a look into it now actually.

PR #568, opened 2015-12-14
Fixes flavour text

EmperorJon:
Removes the else to properly for the switch statement, meaning that it no longer wipes your flavour text/thinks you're opening the general slot ever time.


EmperorJon:
Fixes #512.

PR #569, opened 2015-12-14
Communicators: Reloaded: Revenge of the NanoUI

Neerti:
See commit log/changelog for details.  The files are also documented to high heaven and back.


Neerti:
Datrean reviewed it and saw no problems soooo here we go!

Datraen:
Didn't see any problems at a glance., didn't do an in-depth look at all of it. Just as a warning.

PR #573, opened 2015-12-15
Corrections to custom loadout

Mark9013100:
Places restrictions on the large sunglasses and prescription sunglasses.
Places medical role restrictions on the EMT labcoat.
Comments out the boot knife in custom loadout, weapons shouldn't be in custom loadout.


Datraen:
Just going to let the check complete before merging.

PR #580, opened 2015-12-15
Fixes #552

### LINKED ISSUE #552: Floor Tile placing sets air to zero automatically
We just exposed one of the asteroid pockets of vacuum to air, and we pressurized it. At the point when we began putting down lattice and plating, every time we placed the initial plating on top of the lattice, the air on the tile went to zero. We had already pressurized the room, but the air on each tile set to zero each time we added plating, quickly depressurizing the room.

Hubblenaut:
Fixes #552


GinjaNinja32:
if it's just in the title it doesn't link :<

fixes #552

PR #581, opened 2015-12-15
unix2dos polaris.dme

GinjaNinja32:


zaredman:
For us idiots -- What exactly is this, @GinjaNinja32 ?

Spookerton:
It changes line endings in the project from \n to \r\n, because mixed line endings are bad, and convention in baycode and thus what, like, 90% of existing code is \r\n

PsiOmegaDelta:
It's also the case that map merge tool outputs Windows newlines in both *nix and Windows systems, which can lead to fairly major diffs if someone commits maps with Unix line endings.

GinjaNinja32:
it changes polaris.dme from \n to \r\n because DM/BYOND save \r\n and every time I go to commit it I have to dos2unix it to get git to accept that I barely changed it.

PR #597, opened 2015-12-16
Prevents blob from growing over asteroid walls

### LINKED ISSUE #584: Green blobs actually don't destroy rock walls but grow in/through them.
https://gyazo.com/d2db4343e044a00984b10043c6d0922b

Hubblenaut:
Fixes #584

### DIFF
--- a/code/modules/blob/blob.dm
+++ b/code/modules/blob/blob.dm
@@ -54,7 +54,7 @@
 	update_icon()
 
 /obj/effect/blob/proc/expand(var/turf/T)
-	if(istype(T, /turf/unsimulated/) || istype(T, /turf/space))
+	if(istype(T, /turf/unsimulated/) || istype(T, /turf/space) || (istype(T, /turf/simulated/mineral) && T.density))
 		return
 	if(istype(T, /turf/simulated/wall))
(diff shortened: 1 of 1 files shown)


code/modules/blob/blob.dm
```diff
@@ -54,7 +54,7 @@
 	update_icon()
 
 /obj/effect/blob/proc/expand(var/turf/T)
-	if(istype(T, /turf/unsimulated/) || istype(T, /turf/space))
+	if(istype(T, /turf/unsimulated/) || istype(T, /turf/space) || istype(T, /turf/unsimulated/mineral))
```
Techhead0: Shouldn't `istype(T, /turf/unsimulated/)` return true for `/turf/unsimulated/mineral` anyway?

Zuhayr: Yes.

Hubblenaut: My mistake, this is supposed to be simulated, not unsimulated.

Zuhayr:
You will need to cast to /turf/simulated/mineral and check density, a change was merged that uses the same turf for all kinds of asteroid, with density determining if it's a wall or not.

PR #600, opened 2015-12-16
Adds Brass Knuckles v2.0

Hubblenaut:
See #563 
Incorporates knuckles into material system.
Available materials are brass (created from steel sheets) and gold.

### DIFF
--- a/code/modules/materials/material_recipes.dm
+++ b/code/modules/materials/material_recipes.dm
@@ -25,6 +25,8 @@
 
 /material/steel/generate_recipes()
 	..()
+	recipes += new/datum/stack_recipe("brass knuckles", /obj/item/weapon/material/knuckles, 2, time = 10, one_per_turf = 0, on_floor = 1, supplied_material = "brass")
+
 	recipes += new/datum/stack_recipe_list("office chairs",list( \
 		new/datum/stack_recipe("dark office chair", /obj/structure/bed/chair/office/dark, 5, one_per_turf = 1, on_floor = 1), \
 		new/datum/stack_recipe("light office chair", /obj/structure/bed/chair/office/light, 5, one_per_turf = 1, on_floor = 1) \
@@ -130,3 +132,7 @@
 		new/datum/stack_recipe("white folder", /obj/item/weapon/folder/white), \
 		new/datum/stack_recipe("yellow folder", /obj/item/weapon/folder/yellow), \
 		))
+
+/material/gold/generate_recipes()
+	..()
+	recipes += new/datum/stack_recipe("gold knuckles", /obj/item/weapon/material/knuckles, 2, time = 10, one_per_turf = 0, on_floor = 1, supplied_material = "gold")
--- a/code/game/objects/items/weapons/material/misc.dm
+++ b/code/game/objects/items/weapons/material/misc.dm
@@ -75,3 +75,14 @@
 	slot_flags = SLOT_BACK
 	origin_tech = "materials=2;combat=2"
 	attack_verb = list("chopped", "sliced", "cut", "reaped")
+
+/obj/item/weapon/material/knuckles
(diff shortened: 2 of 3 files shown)


zaredman:
What is the difference between this and #563 ?

Sabess:
They're a material item in this. Pull whichever you prefer, I guess; I'll refrain from posting my list of problems with this.

Hubblenaut:
This is free of things left to be fixed and fully incorporated into the material system like it should be.
Sabess, I'm not sure where your problem lies. This is exactly what you have intended to add with the addition that it is embedded in our mechanics.

Sabess:
It's not what I intended. It's what /you/ intended. I'm not going to go into more detail because I'd rather not in a public setting, because I have no gentle way of saying it.

Hubblenaut:
So much for trying to be helpful, I guess

code/modules/materials/material_recipes.dm
```diff
@@ -25,6 +25,8 @@
 
 /material/steel/generate_recipes()
 	..()
+	recipes += new/datum/stack_recipe("brass knuckles", /obj/item/weapon/material/knuckles, 2, time = 10, one_per_turf = 0, on_floor = 1, supplied_material = "brass")
```
Zuhayr: Making brass knuckles out of steel makes no sense.

Zuhayr:
It's really rude to just outright steamroll over someone's work like that when they have objections to the principle, @Hubblenaut, and it's disingenous to hide behind 'we' when it was specifically you making the decision to do it.

Sabess:
Pretty much that, Zuh, though you said it a lot nicer than I would have. 
Anyway, objectively, the code here may actually be a bit nicer (assuming the force divisor makes it to the originally intended damage). 
A lot of people have been saying they'd rather this be an autolathe thing. So I'll probably just make it an autolathe thing instead.

EmperorJon:
There's an argument that we can already carry around lots of metal and magically make chairs, so why not brass knuckles. Personally I find that rather silly and whilst I can accept it probably won't change I don't particularly want to see more magical item creation like that, especially not items which are intended purely to be weapons. That's why I'd prefer them being autolathe-produced, or even hacked autolathe only, although for balance purposes that's probably asking for too much for something that isn't in practice particularly useful.

Hubblenaut:
I apologize, Sabess, my intention was not to steamroll your work, but to help you out. I adhered to the item properties you've given in your branch, the sole difference here is that the item is a child of weapon/material. That is important for having the item inherit material properties. It is however only available in brass and, as I found to be a nice addition, gold. If you're not comfortable with this however, I will close this PR and leave you to it.


> Making specifically 'brass' knuckles out of steel makes no sense

That is an issue, yes. But an object made of pure brass coming out of the autolathe wouldn't make much more sense either. I'm indifferent to where this is created - forming knifes and spoons out of sheets with your bare hands is the same problem that could be adressed by a complete overhaul of production processes in the future.

Sabess:
Having an item restricted to "hacked" is basically just a formality; it doesn't really change anything. That said, if we put tac knives under hacked, then yeah, ti should go under hacked.

zaredman:
>my intention was not to steamroll your work, but to help you out. 

Now, call me new to Github, but isn't there a way to highlight/comment/edit code without throwing up a new PR out-right? I can see how this is causing agitation.

Also yea, seconding the 'hacked autolathe' one. Keep all weapons to that please.

Hubblenaut:
Since we're at it, we should probably move all the weapons - forks, knife, spoons, blade ect - to the autolathe. If I'm not mistaken though, it can't choose from different materials to build yet.

Techhead0:
If the only reason for making them out of brass is because 'steel brass knuckles' sounds weird, why not just call them 'steel knuckledusters' instead?

PR #602, opened 2015-12-16
Colors on-mob sprites

Hubblenaut:
Colors the on-mob overlays for every slot but eyes, ears and pockets.
Material items will now no longer show white on characters.


Zuhayr:
bleh I was hoping someone would rewrite the entire file before this happened. w/e. :+1:

Hubblenaut:
I almost thought so. :P

PR #615, opened 2015-12-18
[DNM] Replaces the standard .45 pistols with the detective's.

Mark9013100:
Now everyone in security can have their own 
All security roles can change the name of the vintage .45 pistol.
Replaces the standard .45 pistols with the vintage .45 pistols.
Massive thanks to Neerti for help with the code.


Sabess:
Reskins for all.

Mark9013100:
Ja ja.

Minijar:
I don't think nano would allow their officers to run about with whatever firearm they want. This should remain the detective's gimmick.

Mark9013100:
I don't see the harm in allowing officers to have something a little unique.

Mark9013100:
And considering I still don't know to rebase worth a damn, I'll be opening the PR again within the next hour or so.

Sabess:
Personally I don't see why they would care if the gun loaded the correct caliber. Also, not everything needs to be realistic, so.

PR #616, opened 2015-12-18
Replaces the .45 pistol with the detective's.

Mark9013100:
Link to previous PR: https://github.com/PolarisSS13/Polaris/pull/615


Hubblenaut:
Why is the window on that spot on one million health points..?

GinjaNinja32:
Someone probably copied a centcom window without realising.

Mark9013100:
Just tested that specific window, broke normally when I shot it with a energy gun, since I'm sure windows rely on the maxhealth var instead of the health var.

Sabess:
Question. Would this replace them in cargo crates as well? Might as well do so if not. Pretty easy to do.

Mark9013100:
Yes, it does the replace the standard pistols in cargo.

Datraen:
What's up with the map changes? I'm fine with line 1594, but the others?
Keep in mind, this is a question coming from someone who's unfamiliar with map related things in BYOND.

GinjaNinja32:
Looks like whoever added those particular turfs missed the starting `/`, which DM *always* saves.

Mark9013100:
Welp, merge conflicts, closing and reopening soon-ish.

PR #619, opened 2015-12-19
Shorts

SinTwo:


EmperorJon:
If males wear the shorts apparently you get 1 pixel of dingdong hanging out. This might wanna be looked into. :P

Hubblenaut:
Probably just the new fashion

Sabess:
That's my faaaaaaault I'll fix it

PR #621, opened 2015-12-19
More clothes and general sprite additions.

SinTwo:


code/modules/client/preferences_gear.dm
```diff
@@ -419,6 +433,36 @@ var/global/list/gear_datums = list()
...
+/datum/gear/croptop
+	display_name = "croptop, NT"
+	path = /obj/item/clothing/under/croptop
+	slot = slot_w_uniform
+	cost = 1
+
+/datum/gear/croptop/grey
+	display_name = "croptop, grey"
+	path = /obj/item/clothing/under/croptop/grey
+	slot = slot_w_uniform
```
Zuhayr: You're inheriting this from /datum/gear/croptop along with cost, don't need to redefine it for the following children.

SinTwo:
...I knew that.. but I've just been doing it because...

Datraen:
It's a bad habit to copy over inherited variables.

PR #628, opened 2015-12-20
Stops sparks from trying to qdel twice.

### LINKED ISSUE #460: "obj/effect/effect/sparks was unable to be GC'd and was deleted"
http://puu.sh/lLOFE/68028ac859.png

Investigation needs to be done as this /may/ be the cause of all the lag.

Datraen:
Fixes #460.

### DIFF
--- a/code/game/objects/effects/effect_system.dm
+++ b/code/game/objects/effects/effect_system.dm
@@ -155,6 +155,8 @@ steam.start() -- spawns the effect
 				for(i=0, i<pick(1,2,3), i++)
 					sleep(5)
 					step(sparks,direction)
+				spawn(20)
+					src.total_sparks--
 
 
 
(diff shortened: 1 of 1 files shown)


code/game/objects/effects/effect_system.dm
```diff
@@ -155,10 +155,6 @@ steam.start() -- spawns the effect
 				for(i=0, i<pick(1,2,3), i++)
 					sleep(5)
 					step(sparks,direction)
-				spawn(20)
-					if(sparks)
-						qdel(sparks)
-					src.total_sparks--
```
Neerti: You might want to throw this up on the spawn(20), in-case it does something important.

Datraen: Yeah, got overzealous, will resolve in a moment.

Datraen:
Fixed all outstanding issues, merging.

PR #629, opened 2015-12-20
Fixes for the asteroid turf post-changes.

Zuhayr:
- RCDs work again.
- Lattice/floor tiles work again.
- Icons update properly.

TODO: work out why atmos isn't spreading into them.


Zuhayr:
If someone wiser in the ways of ZAS could look into why atmos doesn't spread into these turfs when they're not dense, please let me know.

PR #635, opened 2015-12-20
Removes Resomi.

Zuhayr:
Their lore has been denied and as WrongEnd does not wish them to be used and neither do the staff, I've removed them entirely. This does not break any of the features that were introduced with them like the complex holder icon generation.


WrongEnd:
:+1:

PR #644, opened 2015-12-22
move smartfridge to better location

GinjaNinja32:
This position means the NanoUI window doesn't auto-close if you try to use the left dispenser.


Zuhayr:
:+1:

PR #645, opened 2015-12-22
Allowed grenades to be stuffed into mouths and primed.

Zuhayr:


Zuhayr:
Tested, works fine.

PsiOmegaDelta:
Not sure if Polaris actually has the proc and since you're using the equipment slot it may not even matter but if you want to be extra sure the grenade can be stuffed into someone's mouth you may want to use:
```/mob/living/carbon/human/can_eat(var/food, var/feedback = 1)```
```/mob/living/carbon/human/can_force_feed(var/feeder, var/food, var/feedback = 1)```
Checks if the target even has a mouth (would fail for IPCs for example, but which you're removing), if it's blocked, gives feedback if desired, and all that.

Zuhayr:
It's not literal eating, to be honest, and it will be niche enough that I think just having the slot should qualify you for it.

Hubblenaut:
Since we're on the topic of grenades, flash grenades should probably do a bit of burn damage for when it exploded in hand/mouth.

Zuhayr:
Outside of scope.

PR #659, opened 2015-12-24
[HELLA DNM] Sanity for full prosthetics reversion.

Zuhayr:
This is just up in case a revert has to happen and save files tweak out. Hopefully won't need to be merged.


Zuhayr:
Shouldn't be needed now, will leave the branch around in case an emergency revert has to happen.

PR #670, opened 2015-12-26
Fix for brains ending up popping out of cryo consoles.

Zuhayr:


code/game/machinery/cryopod.dm
```diff
@@ -368,7 +368,7 @@
 	qdel(R.mmi)
 	for(var/obj/item/I in R.module) // the tools the borg has; metal, glass, guns etc
 		for(var/obj/item/O in I) // the things inside the tools, if anything; mainly for janiborg trash bags
-			O.loc = R
+			O.forceMOve(R)
```
GinjaNinja32: `forceMOve`

Zuhayr: ... why didn't my final compile pick this up.

PR #678, opened 2015-12-27
Applying Harpy's prosthetic limb damage fix.

Zuhayr:
This may mean that synthetics need several general buffs as four egun shots to the chest will kill them.


mwerezak:
See https://github.com/Baystation12/Baystation12/pull/11760. You may also want the blindness fix.

Zuhayr:
I removed 'optics' as a concept so it shouldn't be necessary.

Hubblenaut:
Four laser shots kill a normal human as well

PR #691, opened 2015-12-29
Added gear harnesses to the loadout.

Zuhayr:
It's a uniform item that has no body coverage and serves only as something that full synths/dionaea can wear in the uniform slot so they can wear belts/IDs.


Neerti:
:+1: with travis.

PR #692, opened 2015-12-29
Nerfs the Unathi knife, adds the harness to loadout.

Zuhayr:
This needs admin approval before merge since Neerti was not sure if it would be desirable or not.
The knives are weapons, but they are also exceptionally weak ones, and the item has a pretty high cost. You can get a more effective weapon by taking hatchet or kitchen knife from the cafeteria.

### DIFF
--- a/code/game/objects/items/weapons/material/misc.dm
+++ b/code/game/objects/items/weapons/material/misc.dm
@@ -38,6 +38,8 @@
 	icon = 'icons/obj/weapons.dmi'
 	icon_state = "unathiknife"
 	attack_verb = list("ripped", "torn", "cut")
+	force_divisor = 0.15
+	thrown_force_divisor = 0.15
 
 /obj/item/weapon/material/hatchet/tacknife
 	name = "tactical knife"
--- a/code/modules/client/preferences_gear.dm
+++ b/code/modules/client/preferences_gear.dm
@@ -926,6 +926,13 @@ var/global/list/gear_datums = list()
 	display_name = "scarf, striped blue"
 	path = /obj/item/clothing/accessory/stripedbluescarf
 
+/datum/gear/unathi_harness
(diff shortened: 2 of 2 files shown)


ParadoxSpace:
I got a pretty dire need for this. They are whitelisted exclusively for unathi usage, right?

Zuhayr:
````
+	whitelisted = "Unathi"
````

Neerti:
I wouldn't mind adding this if the harness started empty, since I don't believe the custom loadout should be the place to get weapons, no matter how weak they are, since it also begs the question of 'how did this get pass customs' etc.

Yoshax:
Same reason Sikhs are allowed to have Kirpans, or Scots in full outfit are allowed Sgian-dubhs. Religious or cultural reasons. But that would probably be Sledge's call.

ParadoxSpace:
Yoshax does have a point about kirpans. It's a traditional thing for Sikhs. Even in the shower they take them. It's simply cultural for Unathi to be able to bring their dueling knives. They aren't dangerous, neither are Kirpans or Sgian-dubhs.

sabiram:
Sikhs have extreme reservations about actually using their knife, even in defense. I wonder if the same could be said of unathi.

ParadoxSpace:
I assumed it would only be used in honorable or utility contexts like actually dueling, cutting food, self defense, etc.

Zuhayr:
For clarity the knives are wooden with teeth set into them, akin to http://australianmuseum.net.au/uploads/images/19319/ih000359_big.jpg - they're not going to be something that poses a credible death threat to anyone wearing anything heavier than a T-shirt.

Zuhayr:
Additionally without the knives this is pointless. It adds two extremely specific storage slots (knives only) and a barely visible accessory icon.

Zuhayr:
As discussed, closing this for now.

PR #698, opened 2015-12-30
Changes to Confused Moving

Hubblenaut:
Instead of complete random cardinal movements while confused, the chance of the direction being randomly picked is now at 75% for running (and driving) and 25% for walking. This is to ensure people have a chance of getting into the general direction they want to.
Bumping into an immobile object while confused and running has a 50% chance of making you visibly run into the obstacle. Induces minor (5 points of) damage.

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -123,22 +123,26 @@ default behaviour is:
 		now_pushing = 0
 		spawn(0)
 			..()
-			if (!istype(AM, /atom/movable))
+			if (!istype(AM, /atom/movable) || AM.anchored)
+				if(confused && prob(50) && m_intent=="run")
+					Paralyse(1)
+					playsound(loc, "punch", 25, 1, -1)
+					visible_message("<span class='warning'>[src] [pick("ran", "slammed")] into \the [AM]!</span>")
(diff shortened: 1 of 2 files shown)


Hubblenaut:
I have no clue what Git is doing to me here. Stand by.

code/modules/mob/living/living.dm
```diff
@@ -123,22 +123,26 @@ default behaviour is:
 		now_pushing = 0
 		spawn(0)
 			..()
-			if (!istype(AM, /atom/movable))
+			if (!istype(AM, /atom/movable) || AM.anchored)
+				if(confused && prob(50) && m_intent=="run")
+					Paralyse(1)
+					playsound(loc, "punch", 25, 1, -1)
+					visible_message("<span class='warning'>[src] [pick("ran", "slammed", "waltzed")] into \the [AM]!</span>")
```
Zuhayr: .. waltzed?

Hubblenaut: Not a word? Correct me, I'm using a translator quite often.

Zuhayr: A waltz is a dance, it's a pretty odd term to describe someone running headlong into an obstacle.

PR #700, opened 2015-12-30
Medical storage items changes and additions

Yoshax:
* Gives medical black webbing vests same as security
  * Allows medical jobs to take the black webbing vests as custom loadout
* Adds a surgical belt, can hold all surgical items, but nothing else
  * Adds the surgical belt to medical equipment
  * Adds the surgical belt to surgical crate
  * Adds surgical belt to ert medical
* Adds smaller emergency anesthetic tanks
  * Adds small anesthetic tanks to anesthetic lockers
  * Adds small anesthetic tanks to surigcal crates
* Replaces antag surgical kit with improves fully loaded surgical belt


sabiram:
Why do paramedics need surgical belts? Surely the surgeons would?

Yoshax:
Surgeons have their tools on tables. They're not leaving the operation theatre. But sometimes there's not surgeons, or the people who are the paramedics are surgeons, or whatnot. Maybe the dude's full of stuff and you can't move him, etc. If a surgeon wants one they're free to get one.

comma:
It's still odd that tools for speciality are stored in area for people who MIGHT use them instead on rare occasion when filling in for empty medbay. Aren't paramedics supposed to be 'run to scene, stabilize, get to medbay for full treatment' as opposed to full MD in different scrubs (since they are different job and all)? Putting this in their lockers is basically saying 'yeah you are also surgeons'

Yoshax:
Well, it made more sense to give them to the paramedics, because they're mobile, whilst people inside mobile aren't. But, I'll move them for everyone to access.

Neerti:
After an internal discussion with staff, we've gone to the consensus that this proposal is not something that is really needed or one that is wanted for Polaris.

Part of the reason for this is that I feel it adds excess bloat (powercreep might be the better term but I'm not sure) to medical, as from personal experience and medical players I've consulted, medical isn't usually hurting for slots.

Adding a surgery belt also introduces a style of play that we'd rather avoid, of doing surgery 'in the field', especially as an EMT/paramedic.  It also adds a risk of someone going into surgery and taking all the tools, and then losing them for whatever reason (dying, cryoing out, etc), and it promotes the idea that doing surgery in the field is acceptable and optimal, instead of the preexisting dynamic of EMTs/paramedics stabilizing the patient, and bringing them to Medbay to receive proper treatment.

The addition of a mini anesthetic tank seems both redundant and also at risk of contributing to the style of play that we wish to avoid mentioned above.

PR #701, opened 2015-12-30
Removes actually seen mentions of the word retard

Yoshax:
Folks don't like it, and get offended.

### DIFF
--- a/code/game/objects/structures/signs.dm
+++ b/code/game/objects/structures/signs.dm
@@ -138,7 +138,7 @@
 
 /obj/structure/sign/kiddieplaque
 	name = "\improper AI developers plaque"
-	desc = "Next to the extremely long list of names and job titles, there is a drawing of a little child. The child appears to be retarded. Beneath the image, someone has scratched the word \"PACKETS\""
+	desc = "Next to the extremely long list of names and job titles, there is a drawing of a little child. The child appears to have learning difficulties. Beneath the image, someone has scratched the word \"PACKETS\""
 	icon_state = "kiddieplaque"
 
 /obj/structure/sign/atmosplaque
--- a/code/game/objects/structures/signs.dm
+++ b/code/game/objects/structures/signs.dm
@@ -138,7 +138,7 @@
 
 /obj/structure/sign/kiddieplaque
 	name = "\improper AI developers plaque"
-	desc = "Next to the extremely long list of names and job titles, there is a drawing of a little child. The child appears to have learning difficulties. Beneath the image, someone has scratched the word \"PACKETS\""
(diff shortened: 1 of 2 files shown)


Mark9013100:
For the plaque, I understand, but mental retardation is a real condition and I don't see the need for replacing it with soft language.

Minijar:
Gotta agree with Mark here. Retardation is the correct medical term, using anything else would be worse if anything. People just need to stop being offended by everything.

Hubblenaut:
I literally just got bwoinked for using the word retardation in that context. Honestly, if it's not outright used as an insult, the usual usage of it in scientific contexts are okay.
Retardation means nothing but 'backward' and is commonly used in non-insulting context. It's a normal word that doesn't hold any insulting meaning if not used as directly as one.

sabiram:
Remove it from the sign, don't remove it from any medical context.

Woodratt:
"I literally just got bwoinked for using the word retardation in that context."

You were bonked because you were using it in a joking fashion, not as a medical terminology.

Hubblenaut:
I don't like this debate about political correctness. In my opinion, as long as people don't outright insult each other, things are fine. The people here mean nothing harmful when they say 'retarded'. I'll opt out here.

Yoshax:
I couldn't really care either way. I just made the PR because some folks were complaining. If you wanna change it, merge it, if not, close it :)

Zuhayr:
http://www.medscape.com/viewarticle/782769

It hasn't been a medical term since 2013 at the very latest, please stop using that argument.

'intellectual disability' is generally used for inherited conditions, I would suggest 'a brain injury' or 'an acquired brain injury' considering it's physical damage to the grey matter.

Zuhayr:
Spoke to a friend of mine who actually works in healthcare in the US, the current terminology at her workplace is 'traumatic brain injury' for the actual injury and 'developmentally disabled' for the state. Do what you will with this information.

Zuhayr:
:+1:

code/game/objects/structures/signs.dm
```diff
@@ -138,7 +138,7 @@
 
 /obj/structure/sign/kiddieplaque
 	name = "\improper AI developers plaque"
-	desc = "Next to the extremely long list of names and job titles, there is a drawing of a little child. The child appears to be retarded. Beneath the image, someone has scratched the word \"PACKETS\""
+	desc = "Next to the extremely long list of names and job titles, there is a drawing of a little child. The child appears to be developmentally disabled. Beneath the image, someone has scratched the word \"PACKETS\""
```
Zuhayr: Honestly this reads like a much classier/subtler/snarkier joke now.

PR #705, opened 2015-12-31
Minor edits, 12/31/2015

Mark9013100:
Adds the white cap to white wardrobes.
Adds black scrubs to the medical wardrobe.
Comments out the santa hats in custom loadout.
Minor change to the long forensics jacket names in custom loadout.
Removes the punctuation from the chef's classic apron.
Minor Z-1 area correction.
Re-adds maintenance access back to Medical Doctors, due to the map's size, and considering every other emergency role has it. They also had it prior to the introduction of the paramedic role.


code/modules/client/preferences_gear.dm
```diff
@@ -233,6 +233,7 @@ var/global/list/gear_datums = list()
 	cost = 1
 	slot = slot_head
 
+/*
 /datum/gear/santahat
```
Zuhayr: This may cause issues with runtimes upon trying to spawn if people have the item saved.

Mark9013100:
Right, noted. Would flat out deleting it work better?

Zuhayr:
No. It might be best to just increase the point cost and leave the item in.

Hubblenaut:
Most expensive santa hat in the world

Zuhayr:
Conflicts.

Mark9013100:
Considering I've no idea what's causing the conflict, I'll just close this PR and open another.

Hubblenaut:
It's the map file, just remerge it

PR #709, opened 2015-12-31
New PDA Sprites

SinTwo:
Adds new PDA sprites and a system to let you choose which one you want to use.
Apparently I also fixed a typo so.. Yay.
Also! If someone wants try and pretty up the little part in PDA.dm go ahead I'm not touching this anymore now that it works and doesn't runtime like crazy.


SinTwo:
Right, if there are no problems, going to go ahead and merge this in a few hours.

PR #711, opened 2016-01-01
May fix massive lag issues with autofire guns.

Zuhayr:
spawn() is very expensive and I have no idea why it was added, testing locally indicated that it was working a lot better than previously, but I am not sure if this will translate to an on-server improvement.


Zuhayr:
Doesn't seem to work.

Zuhayr:
Tested, looks like it works okay.

Issue #706, opened 2015-12-31
Viruses can not infect new monkey mobs.


Jakksergal:
As the title says. This makes curing viruses impossible without killing an infected human as you can't infect a monkey to create antigens.

Datraen:
See comments on #696.

EmperorJon:
Can you confirm that the issue here was no infection occurring when injecting infected blood?

Datraen:
Not really a duplicate bug report, but likely related to #696. Needs testing on new code.

Datraen:
Had it work recently, closing.

PR #722, opened 2016-01-03
Fixes #696 (IVs/Blood injection going to wrong place)

### LINKED ISSUE #696: IV Drips have no effect
When attempting to give someone blood via an IV drip, it will have no effect on the patient, it won't raise their blood level and admin intervention is required.
Zuhayr said this might be due to a container rewrite back on Bay.
Steps to reproduce:
1. Get someone with a low blood level.
2. Grab a IV drip and place a blood bag in it.
3. Attach the IV to the patient.

EmperorJon:
Fixes #696.

Blood was going into the bloodstream reagents of the mob, not actually vessel bloodstream as it were. A check was in place that mobs which required a heart did this, whereas the actual check should have been mobs that didn't require a heart. In other words, if you're some sort of species that doesn't have a heart, having blood put into you should put it in however you explain the "bloodstream" reagent container as opposed to actually increasing the mob's blood count.

I think this is all stuff relating around @Zuhayr's FBP changes so this may need to be looked at to work out if it's doing anything unusual (like if you try and hook a fully synth person to an IV, in which case the blood does end up in their reagents and can be syringed back out), but for the time being this fixes things like IV drips not working.

### DIFF
--- a/code/modules/organs/blood.dm
+++ b/code/modules/organs/blood.dm
@@ -216,7 +216,7 @@ var/const/BLOOD_VOLUME_SURVIVE = 122
 //Transfers blood from reagents to vessel, respecting blood types compatability.
 /mob/living/carbon/human/inject_blood(var/datum/reagent/blood/injected, var/amount)
 
-	if(should_have_organ(O_HEART))
+	if(!should_have_organ(O_HEART))
 		reagents.add_reagent("blood", amount, injected.data)
 		reagents.update_total()
 		return


Zuhayr:
Whoops, I shouldn't have assumed it was Kelenius' code.

EmperorJon:
Probably fixes #706.

PsiOmegaDelta:
Well, I'm not surprised you made it that way given the original line ```if(species.flags & NO_BLOOD)```.

PR #726, opened 2016-01-04
Adds Barista alt role to Bartender

Loganbacca:
Also increases bartender slots to 2.

There was some discussion in OOC about having a Barista role, so here it is.


EmperorJon:
This is something I was intending to do for a while, so this gets my vote, although we might want to think about different costumes and things. :D

Loganbacca:
If you've got any suggestions for costumes and the like I don't mind adding them to this PR.

PR #727, opened 2016-01-04
Edits/Changes, 1/3/2016

Mark9013100:
Adds maintenance access to Medical Doctors, due to the Map's size, and since they had it prior to the introduction of the paramedic role. - Revoked.
Adds a paramedic equipment locker, which comes with basic EMT gear. Removes the excess gear from the medical closet within the paramedic room.
Makes the santa hat unavailable.
Minor change to the name of the custom loadout forensics jackets.
Adds white caps to the white wardrobe.
Adds a black scrub and scrub caps to the medical wardrobe.
Corrects the name of the chef's classic apron.

Commit two:
Revokes maintenance access.
Fixes the station's cap sprite being off center.
Allows the security hailer to be worn on the ear slot.
Fixes the missing neck line on the atmos voidsuit.


ParadoxSpace:
-1 to maint access. That's why paramedics are there.

Mark9013100:
The reasoning why I've given Medical doctors maintenance access is due to them being an emergency role, that and considering you'll only have two paramedics, as compared to engineering and security where you usually have 3/4 people that are able to move through the colony relatively quickly.

Woodratt:
"Since they had it prior to the introduction of the paramedic role."

Don't think this is true. As far as I can recall they've never had maintenance access except for access to maint around the medbay. 

I'm somewhat against this merge cause as it stands the majority of the time medical doctors are already trying to do the paramedic's role. Not sure they need further reasoning and ability to do the role of the paramedic

Mark9013100:
Really? Could've sworn they used to. Still, I think it's reasonable for them to have maintenance access, but if it's unneeded, I'll just revert it then.

Zuhayr:
Please sort this out when you can.

Mark9013100:
Should be decided upon by the end of the week.

Mark9013100:
Since a decision never was reached upon, I revoked the maintenance access from medical doctors.

EmperorJon:
Conclusion reached in the meeting was that for now the maintenance access shouldn't be changed, and ideas such as merging Paramedic/EMT into Doctor are later concerns.
Concerns were also raised as to the usage of single commits to make so many changes (and many unrelated to each-other). More atomic commits and/or pull requests are preferred.

PR #729, opened 2016-01-04
Wheelchair fixes

Hubblenaut:
* Fixes buckled mobs being swapped with bumpers
 * Fixes wheelchairs being drivable without hands
 * Riding a wheelchair will prevent damage from unsplinted fractures
 * Moves the wheelchair overlay to its base file
 * Adds a few alpha-1 pixels to wheelchair and rollerbeds to make them easier clickable


Techhead0:
For cross-reference, this fixes https://github.com/Baystation12/Baystation12/issues/11784.

PR #739, opened 2016-01-08
Allows trays to drop their items, adds color to material items on trays

### LINKED ISSUE #557: Trays not working
The trays available in the kitchen allow for items to be picked up and carried with them as intended. However to remove items from the tray, the only way to currently do so is to smack it against something, scattering the tray's contents all over the place.

Datraen:
Trays once again drop their items properly.

Fixes #557

### DIFF
--- a/code/game/objects/items/weapons/trays.dm
+++ b/code/game/objects/items/weapons/trays.dm
@@ -192,24 +192,23 @@
 
 /obj/item/weapon/tray/dropped(mob/user)
 
-	var/mob/living/M
-	for(M in src.loc) //to handle hand switching
-		return
+	spawn(1)
+		if(istype(loc,/mob/living)) //to handle hand switching
+			return
 
-	var/foundtable = 0
-	for(var/obj/structure/table/T in loc)
-		foundtable = 1
-		break
+		var/foundtable = 0
+		for(var/obj/structure/table/T in src.loc)
+			foundtable = 1
+			break
 
-	overlays.Cut()
-
-	for(var/obj/item/I in carrying)
-		I.loc = loc
-		carrying.Remove(I)
-		if(!foundtable && isturf(loc))
-			// if no table, presume that the person just shittily dropped the tray on the ground and made a mess everywhere!
-			spawn()
-				for(var/i = 1, i <= rand(1,2), i++)
-					if(I)
-						step(I, pick(NORTH,SOUTH,EAST,WEST))
-						sleep(rand(2,4))
+		overlays.Cut()
+		for(var/obj/item/I in carrying)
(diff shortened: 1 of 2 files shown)


code/game/objects/items/weapons/trays.dm
```diff
@@ -192,24 +192,23 @@
...
-			spawn()
-				for(var/i = 1, i <= rand(1,2), i++)
-					if(I)
-						step(I, pick(NORTH,SOUTH,EAST,WEST))
-						sleep(rand(2,4))
+		overlays.Cut()
+		for(var/obj/item/I in carrying)
+			I.loc = loc
+			carrying.Remove(I)
+			if(!foundtable && isturf(loc))
```
Zuhayr: The entire foundtable thing could be replaced with
````
if(isturf(src.loc) && !(locate(/obj/structure/table) in src.loc))
````

Comment (Datraen) [inline code/game/objects/items/weapons/trays.dm]: I was focusing on having the code function as opposed to optimization. As it stood, it would call all of the locs prior to moving the item, and would always appear to be in someone's hands to this call in natural conditions.

Comment (Zuhayr) [inline code/game/objects/items/weapons/trays.dm]: I see, fair enough.

**Comment by Zuhayr** [comment]
:+1:

**Comment by Zuhayr** [comment]
I would like to add that this is really a hackfix and adding spawn() to fix timing issues should be avoided, but trays are overall really horrible code, so, whatever.

**Comment by Datraen** [comment]
I will admit this is a hackfix, but I have a feeling I would have to rewrite how dropped is called if I wanted to do it the way I wanted to.

File: code/game/objects/items/weapons/trays.dm [hunk 2]
```diff
@@ -192,24 +192,23 @@
...
-		if(!foundtable && isturf(loc))
-			// if no table, presume that the person just shittily dropped the tray on the ground and made a mess everywhere!
-			spawn()
-				for(var/i = 1, i <= rand(1,2), i++)
-					if(I)
-						step(I, pick(NORTH,SOUTH,EAST,WEST))
-						sleep(rand(2,4))
+		overlays.Cut()
+		for(var/obj/item/I in carrying)
+			I.loc = loc
```
PsiOmegaDelta: Recommend using ```I.forceMove(loc)``` instead to ensure light sources update properly among other things.

Datraen:
I'll do a proper rewrite when I'm not working (see: Monday afternoon, Tuesday.)

Datraen:
So, trying to avoid spawn() was not working out for me, so I set it to spawn() instead of spawn(1) and made some optimizations.

I also added colors to material based items that trays pick up.

PR #744, opened 2016-01-08
Fixes permanent blindness, deafness.

### LINKED ISSUE #742: FBPs have no way to fix their vision.
Optical sensors do not take damage or control sdisabilities, which causes FBPs to be permanently blind until admin intervention takes place.

Datraen:
Comments out the parts of code that held blindness and deafness constant, never allowing sdisabilities to update.

Fixes #742


code/modules/mob/living/carbon/human/life.dm
```diff
@@ -1034,10 +1034,10 @@
 			blinded =    1
 			eye_blurry = 1
 		else
-			//blindness
+			/*
```
EmperorJon: Why comment this stuff out?

PR #746, opened 2016-01-08
Gives telecrystals tech levels, restores PDA default to new PDA.

Datraen:
Does what it says on the title.

Was reminded to restore to new PDA by staff meeting.

It's been long enough for people to notice.


Zuhayr:
:+1:

Why the DNM?

Datraen:
DNM because tech levels are being discussed by staff.

Datraen:
Should be all good.

PR #748, opened 2016-01-09
Blindness fix.

Zuhayr:
Readds blindness removal to imidazoline and organ surgery. 
Stops imidazoline healing robotic eyes.
Renames the autopsy scanner to the biopsy scanner to make it clear to people that it can be used on living patients.
Fixes oversight with O_APPENDIX.

Alternative to #744.

### DIFF
--- a/code/__defines/mobs.dm
+++ b/code/__defines/mobs.dm
@@ -145,6 +145,7 @@
 #define O_BRAIN    "brain"
 #define O_LIVER    "liver"
 #define O_KIDNEYS  "kidneys"
+#define O_APPENDIX "appendix"
 #define O_PLASMA   "plasma vessel"
 #define O_HIVE     "hive node"
 #define O_NUTRIENT "nutrient vessel"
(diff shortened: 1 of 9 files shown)


Zuhayr:
Wasn't able to repro deafness, but I am 100% certain commenting out sdisabilities is not a proper fix.

PsiOmegaDelta:
Yeah, the proper fix is probably me going over my vision handling rewrite. I evidently missed details.

Datraen:
This one is better than mine by correcting the checks and targetting the correct life file, versus just ignoring sdisabilities during the check to handle healing.

Datraen:
So, testing results:
Blindness remains, eye_blind does not decrease.

Zuhayr:
I don't want to be rude but if your standard testing practices are using VV instead of player-facing mechanics then you are not going to get useful or indicative results. The whole point of this PR was to add BLIND removals to procs called when repair was conducted.

Datraen:
Worked under normal repair for synths, assuming will work for humans.

PR #769, opened 2016-01-12
Hydroponics tray power can be toggled

Hubblenaut:
* Hydroponics trays start offline by default.
 * A powerless tray will appear gray instead of blue.
 * A hydroponics tray draws 20 W when online.
 * Plantlife will not be simulated in powerless hydroponics trays.
 * Simple soil stays unchanged.


Zuhayr:
.. why does disabling the power stop plant life from growing? What's the purpose of this?

Hubblenaut:
If you want a fluff reason, my argument is that the growth of plants we see in the game as well as their quick death lies in how hydroponics trays work and speed up their life cycle. If disabled, the plant would grow in its natural speed.

Gameplay reason for this is to stop having all trays overgrown with dead plants when joining late as a gardener or simply not tending them for a while, inducing stress in a person and actually hindering them from roleplay.

My experience is that if you leave your trays for a chat with someone, you can completely start over when you come back.

ParadoxSpace:
That's pretty unrealistic and also unnecessary.

Hubblenaut:
How is it unrealistic? If it's unnecessary is not for a single person to say

Datraen:
That makes some logical sense.

Hubblenaut:
More fluff? Alright.

The reason the plants on the station grow so fast lies in the special fertilizer that is used. In soil, this fertilizer is always present. In hydroponics trays however, the fertilizer can be purposefully applied using an electric field to move the electrically charged fertilizer particles into the roots of the plants. No power, no fertilization.

Datraen:
I personally don't mind this getting added, has my approval.

Zuhayr:
I don't really like this. It means that soil and hydroponics trays are mechanically divergent for no real reason other than some minor convenience on the game mechanics side, and that lore is a really forced explanation for it. A preferable fix for the problems would be reducing the chance of an empty tray developing weeds without having had reagents or seeds planted previously.

Datraen:
If I recall correctly, the trays are actually parents of the soil plots, which would cause issues overall.

I'll rescind my approval because of this case.

Hubblenaut:
Trays have a 'mechanical' var which determines power, light, amount of information given and more.

Other for the power demand, which makes a lot of IC sense to me instead of plants naturally growing this quick, there are no other changes.

Making weeds less likely wouldn't achieve the same goal at all. Plants would still require most of a player's attention and quickly die if not given, versus being able to be toggled off for a break.

Neerti:
Sorry this took so long, but the consensus among staff while this was being discussed at the last meeting was that we don't want this feature, but would rather make weeds appear less often and plants to not die quite as fast instead.

See https://github.com/PolarisSS13/Polaris/issues/953 for what some of us might try out instead.

PR #770, opened 2016-01-13
Map Changes

Yoshax:
- Adds mirrors to the visitor rooms
- Adds disposal bins to the visitor rooms
- Adds drying racks to the garden and hydroponics


Zuhayr:
Conflicts.  May want to rebase rather than just keep pulling.

PR #772, opened 2016-01-13
Adds portable freezers to med lockers

Mark9013100:
Adds portable freezers to the following medical lockers:
medical doctor's locker
paramedic locker
CMO's locker


Yoshax:
Add them to cargo orderable if they aren't already?

Mark9013100:
Every time I tried to, when I tested it, the cargo shuttle would come empty. Better safe than sorry.

NortonDK:
What is a portable freezer and what use does it have?

Zuhayr:
:+1:

Mark9013100:
@NortonDK A portable freezer can be used to store organs and limbs on the go, without the fear of them rotting.

sabiram:
Also known as a 'cooler'

Mark9013100:
And also so you can store beer like the irresponsible sociopath you are.

PR #777, opened 2016-01-13
Adds Seromi.

Zuhayr:
Totally not Resomi (they are Resomi with different lore).


serenemelody:
I still liked Rigatoni.

PR #778, opened 2016-01-13
Fixes.

Zuhayr:


Zuhayr:
The hell is going on with Travis?

PsiOmegaDelta:
Travis' status reports that they're having issues with builds which require sudo (which Polaris still appear still use, Bay does not).

Datraen:
All looks good.

PR #787, opened 2016-01-15
Clarifies a few custom loadout items

Mark9013100:


code/modules/client/preferences_gear.dm
```diff
@@ -404,7 +404,7 @@ var/global/list/gear_datums = list()
 	cost = 1
 
 /datum/gear/harness
-	display_name = "gear harness"
+	display_name = "gear harness (FBP, Dionaea)"
```
Zuhayr: Should change it to 'full prosthetic' since FBP is not actually an official acronym or used anywhere in the codebase.

code/modules/client/preferences_gear.dm
```diff
@@ -1146,7 +1146,7 @@ var/global/list/gear_datums = list()
 	path = /obj/item/clothing/suit/storage/toggle/labcoat/yellow
 
 /datum/gear/labcoat/emt
-	display_name = "labcoat, emt"
```
Zuhayr: Sidenote, shouldn't EMT be all caps?

Mark9013100:
Noted, will change those in a few hours.

PR #790, opened 2016-01-16
Changes observe message to reflect correct respawn timer

Yoshax:


Datraen:
Ah, completely missed that one, whoops.

PR #797, opened 2016-01-18
Fixes #94

### LINKED ISSUE #94: Protolathe will not eject materials
What it says on the tin

Kelenius:
Protolathe and CI will now correctly eject materials.

### DIFF
--- a/code/modules/research/protolathe.dm
+++ b/code/modules/research/protolathe.dm
@@ -202,3 +202,16 @@
 			if(new_item.matter && new_item.matter.len > 0)
 				for(var/i in new_item.matter)
 					new_item.matter[i] = new_item.matter[i] * mat_efficiency
+
+/obj/machinery/r_n_d/protolathe/proc/eject(var/material, var/amount)
+	if(!(material in materials))
+		return
+	var/obj/item/stack/material/sheetType = getMaterialType(material)
+	var/perUnit = initial(sheetType.perunit)
+	var/eject = round(materials[material] / perUnit)
+	eject = amount == -1 ? eject : min(eject, amount)
+	if(eject < 1)
+		return
+	var/obj/item/stack/material/S = new sheetType(loc)
+	S.amount = eject
+	materials[material] -= eject * perUnit
--- a/code/modules/research/protolathe.dm
+++ b/code/modules/research/protolathe.dm
@@ -8,13 +8,14 @@
 	active_power_usage = 5000
 
 	var/max_material_storage = 100000
-	var/list/materials = list(DEFAULT_WALL_MATERIAL = 0, "glass" = 0, "gold" = 0, "silver" = 0, "phoron" = 0, "uranium" = 0, "diamond" = 0)
 
 	var/list/datum/design/queue = list()
 	var/progress = 0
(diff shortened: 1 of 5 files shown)


code/modules/research/protolathe.dm
```diff
@@ -202,3 +202,16 @@
 			if(new_item.matter && new_item.matter.len > 0)
 				for(var/i in new_item.matter)
 					new_item.matter[i] = new_item.matter[i] * mat_efficiency
+
+/obj/machinery/r_n_d/protolathe/proc/eject(var/material, var/amount)
```
Zuhayr: These procs are identical. Why not define them on r_n_d?

Kelenius: ``r_n_d`` doesn't have a materials var. Ideally there should be something like ``r_n_d/constructor`` where ``constructor`` would have all the common properties - or, better, unify CI, both lathes, and mechfab, but that's a pretty distant goal and the fix is needed. There are the same for CI and lathe.

Zuhayr: Just.. move the vars that need to be shared down to r_n_d? If they're identical in structure and purpose then I don't see an issue.

code/modules/research/rdconsole.dm
```diff
@@ -360,59 +360,10 @@ won't update every console in existence) but it's more of a hassle to do. Also,
...
-				res_amount = "uranium_amount"
-		if(ispath(type) && hasvar(linked_imprinter, res_amount))
-			var/obj/item/stack/material/sheet = new type(linked_imprinter.loc)
-			var/available_num_sheets = round(linked_imprinter.vars[res_amount]/sheet.perunit)
-			if(available_num_sheets>0)
-				sheet.amount = min(available_num_sheets, desired_num_sheets)
-				linked_imprinter.vars[res_amount] = max(0, (linked_imprinter.vars[res_amount] - sheet.amount * sheet.perunit))
-			else
-				qdel(sheet)
+		linked_imprinter.eject(href_list["imprinter_ejectsheet"], text2num(href_list["amount"]))
```
Zuhayr: I may just be being paranoid but isn't supply a raw href value to the ejection proc a potential exploit?

code/modules/research/protolathe.dm
Kelenius: I'd rather not refactor in a fix. Fix the issue now, refactor later. Like I said, there are plenty of places where they act in the same/similar way - ``process()``, ``dismantle()``, ``attackby()``, etc.

That being said, if you insist that it would be better, I'll change it.

code/modules/research/rdconsole.dm
Kelenius: Invalid material would quit the proc, invalid amount would not eject more than possible.

code/modules/research/protolathe.dm
Zuhayr: I'd rather not merge boilerplate that could be avoided by moving a single variable definition down one link in the object chain. It's barely a refactor.

Datraen:
Tested, works fine, the original interface is still used.

PR #798, opened 2016-01-18
Fixes for vendors

Kelenius:
-15 seconds from round start time and no double-defining.


code/game/machinery/vending.dm
```diff
@@ -19,9 +15,8 @@
 	src.product_path = path
 
 	if(!name)
-		var/atom/tmp = new path
+		var/atom/tmp = path
 		src.product_name = initial(tmp.name)
-		qdel(tmp)
```
GinjaNinja32: wtf even
initial() but creating an object anyway
what

GinjaNinja32:
:+1:

PR #808, opened 2016-01-19
Fixes #796

### LINKED ISSUE #796: Spiders cannot break through police tape
When there's an (inevitable) outbreak of spiders, and they become the size of a chair, they seem to be unable to break through police tape placed on a door.

I'm not sure if this applies to police tape placed in the usual way, or engineering tape. Though i'll probably test it later.

Yoshax:
As suggested by Psi, changes hostile mobs intent to harm.


PsiOmegaDelta:
You could've set it once on the base code/modules/mob/living/simple_animal/hostile type :P

Zuhayr:
Yep, what Psi said.

PR #810, opened 2016-01-19
Bay update.

Zuhayr:
Pulling from dev-freeze and master. Lots of conflicts.


Zuhayr:
This appears to work fine in local testing but I would suggest running it in its own branch on the main server for awhile to make sure nothing untoward has snuck in through the bloated merge.

Zuhayr:
@Neerti @SinTwo Can we test this on main at some point or no?

Zuhayr:
This seemed stable on main and nothing has come to light yet regarding feature loss etc. Going to unmark it and we can see what happens.

PR #811, opened 2016-01-20
Added functionality for two-handed gun use.

Zuhayr:
Holding a two-handed gun in one hand will check the other hand; if it is free, the icon will be set to a wielded icon and you will not suffer major accuracy penalties when firing the gun.

Code only; no guns actually use this functionality yet (due to a lack of icons for wielded guns).

http://i.imgur.com/mOmVW0X.png


code/modules/projectiles/gun.dm
```diff
@@ -161,6 +174,14 @@
 	user.setMoveCooldown(shoot_time) //no moving while shooting either
 	next_fire_time = world.time + shoot_time
 
+	var/held_acc_mod = 0
+	var/held_disp_mod = 0
+	if(requires_two_hands)
+		update_held_icon()
+		if((user.l_hand == src && user.r_hand) || (user.r_hand == src && user.l_hand))
```
GinjaNinja32: would `if(user.l_hand && user.r_hand)` not do the job? They must be holding it in *one* of their hands, so just check whether they're holding something in both or not?

PR #832, opened 2016-01-24
Removes Walking Mushrooms

EmperorJon:
Honestly, I thought these were gone a long time ago. 

Give me a compelling reason as to why we should keep them. All they do is walk is allow people to run around gibbering then explode, and you can't kill them because they're /sapient/! That makes you a /murderer/.


Datraen:
Honestly I'm sad they're gone, and they're as logical as the killer tomatoes.

They're fun little things, but these explode for some reason, which shouldn't be there.
Also, killing them because they're sapient stops people from killing animals?

PR #833, opened 2016-01-24
Removes Walking Mushroom seeds.

EmperorJon:
Removes mushroom people seeds too. When I mark something as Do Not Merge I mean "I'm not done yet" :p


EmperorJon:
See #832

PR #834, opened 2016-01-24
[DNM]Adds rolling up sleeves, eye swapping eyepatches

Datraen:
Icons courtesy of RedStryker, does what it says on the box.

Originally I was going to move things that weren't jumpsuits and similar clothing out, but I had a merge problem. Isn't a big deal.


Datraen:
Transcribed from now-deleted branch, since auto-merge was not possible. Errors ensued, were caught by Travis. Sorry about that.

PsiOmegaDelta:
It's minor, but I note that if I was to:
* Roll down the jumpsuit (or sleeves)
* Roll up the sleeves (or jumpsuit)
* Roll down the sleeves (or jumpsuit)

Then the rolled_down flag would end up de-synced with the actual state (or rolled_sleeves if you do it the other way around) and you'd have to roll down the jumpsuit twice for it to actually become rolled down.

GinjaNinja32:
Should likely do the `body_parts_covered` and `item_state_slots` updates in `update_clothing_icon()` to avoid that.

PsiOmegaDelta:
This still wouldn't set the ```rolled_down```/```rolled_sleeves``` vars properly. It might be better if the two procs shared a single var (```rolled_down_status``` or the like) and the value of that would then be used to decide the end-result of ```body_parts_covered```/```item_state_slots```.

It may also be a better idea to use something like ```body_parts_covered &= ~(UPPER_TORSO|ARMS)``` (or simply ```~ARMS``` for sleeves) in case there's ever a suit that doesn't normally provide lower torso/leg protection.

Datraen:
I'll look into properly processing it such that rolled_down and rolled_sleeves won't desync one another.

Also, isn't the &= operation ANDing the bitwise flags for body_parts_covered with the flags defined by UPPER_TORSO|ARMS ?

PsiOmegaDelta:
Well, let's assume you have a jumpsuit that for provides FEET protection (reinforced robes or somesuch), that protection would suddenly be lost if the player decides to roll up their sleeves. 
By inverting ARM or ARM|UPPER_TORSO (perhaps also hands) using the ~ operator you'll only remove as much as is strictly necessary.

Datraen:
That's true. I didn't consider the case of providing feet protection.

Datraen:
Forgot a suffix in the .dmi files, too.

PsiOmegaDelta:
Well, it was just an example more than anything else.

Datraen:
Redoing this.

Issue #766, opened 2016-01-12
disposals doesn't have airtight flaps


Dibasic:
not sure whether this is intended, but the lack of airtight flaps has already claimed the lives of three idiots i've seen so far. people can just walk straight into disposals and die

Zuhayr:
Not seeing a problem with this.

Neerti:
It causes the disposal room to overpressurise as more and more people flush things down, since the disposal bin actually takes air along with whatever gets flushed.

Issue #518, opened 2015-12-10
Missing emergency shutters II


sabiram:
Central Access door leading to civilian elevator thing(?)
Armory blast doors
Operating theatre 1 and 2, all 4 entrances
Research server room
Research storage room just north of the server room
Research restroom lavatory unit 1
Cargo emergency storage
Mailing room desk with sliding door
Escape pod 3, 4, 5, 6 (likely the rest)
All of the restrooms in the numbered suites (Visitor Room)
AI core maintenance hatch blast door
Large escape pod 1, 2
Medical restroom & lavatory 
Brig restroom & lavatory
Every cell in the brig including solitary confinement, every blast door leading to communal brig from cell (probably intentional?)
Unisex restroom lavatories in recreation area

Issue #404, opened 2015-12-03
AI core intercoms are inverted


GinjaNinja32:
![](http://nyx.gn32.uk/i/2015-12-03_10:06:07.png)

PR #841, opened 2016-01-24
Multiple map fixes

Yoshax:
Additionaly:

#404 is no longer an issue
#518 is no longer an issue
#766 likely isn't an issue, disposals is never open to, or connected to space


Mark9013100:
Does medical still get a CE's airlock?

Yoshax:
Take a look at medical maintenance, just below the substation. Woodrat added a reminder of my wonderful mapping mistakes.

PR #846, opened 2016-01-25
Fixes #843

### LINKED ISSUE #843: Communicator's hang up doesn't work
Title, this means the second tab's disconnect option has to be used.

Datraen:
Hang up button disconnects all connections to the user's communicator, while the disconnect buttons on manage connections disconnects individual communicators.

Also adds a period to the hung up statements.


EmperorJon:
Note that tags like "Fixes #843" won't work from titles.

PR #848, opened 2016-01-25
Removes special secret player requirements

Kelenius:
https://github.com/Baystation12/Baystation12/pull/11289


EmperorJon:
They may need alteration but we have them for a reason, and if we're to remove them that's a discussion that needs to be had amongst the highups.

Kelenius:
Possibilities: Extended

@EmperorJon Vote on forums or poking admins and asking them to give their input?

Kelenius:
Okay I've put the vote on the forum if that is of any help.

EmperorJon:
Discussed and was not wanted.

PR #875, opened 2016-01-29
Fixes eyecheck() not returning the proper value.

### LINKED ISSUE #835: Wearing thermal goggles with eye protection still damages eyes
To reproduce: get thermal goggles, doesn't matter which.
Grab a voidsuit helmet and welder.
Start welding, regardless of what you've done with the goggles, you'll still take eye damage from welding.

PsiOmegaDelta:
No longer should mob eyes' be burned when sufficiently protected. Fixes #835.
Also gives non-plan thermals an activation icon and updates a message with spans.


PsiOmegaDelta:
Actually, you don't have the Bay-vision rewrite. My mistake.

PR #878, opened 2016-01-29
Minor Z2 area edits

Mark9013100:
Adds two tables with tools in the hardsuit module room.
Adds a nanomed to the Centcom brig.
Places the webbings and emergency gear on the briefing table.


Zuhayr:
:+1:

PR #882, opened 2016-01-31
Adds a coverage check to xenoflora splattering, adjusts brainloss across the board.

Zuhayr:


Kelenius:
Why lower the brain damage?

zekesulastin:
Because brainloss didn't used to kill you; now it does.

Zuhayr:
It also kills at the 60 mark instead of capping at 200. Consensus was lowering it/removing its effects across the board.

code/game/machinery/cloning.dm
```diff
@@ -170,7 +169,7 @@
 			occupant.adjustCloneLoss(-2 * heal_rate)
 
 			//Premature clones may have brain damage.
-			occupant.adjustBrainLoss(-1 * heal_rate)
+			occupant.adjustBrainLoss(-(n_ceil(0.5*heal_rate)))
```
GinjaNinja32: this should not be `n_ceil`; for one, that's a useless proc, for another it's incorrect (it calculates `floor(x) + 1`, not `ceil(x)`; these are different for integer values)

`#define ceil(x) (-round(-(x)))` somewhere in helpers, and use `ceil()`.

Fox-McCloud:
>It also kills at the 60 mark instead of capping at 200

Could you point out where that's at? The brain item has 400 health.

Zuhayr:
Health value is irrelevant to organ survival/damage. Look at max_damage. It might even be 30.

Fox-McCloud:
For those of you who are interested/care: https://github.com/PolarisSS13/Polaris/blob/master/code/modules/organs/organ.dm#L127 governs how an organ dies.

The brain item has no max health specified, but the minimum broken damage is specified as 30 for the organ parent. https://github.com/PolarisSS13/Polaris/blob/master/code/modules/organs/organ.dm#L48 governs what max_damage is if one isn't specified (so 60 is correct here).

Wouldn't it make a bit more sense to just set the brain's max_damage to something like 100 as opposed to tweaking a whole bunch of brain damage values, especially considering there's quite a few more things that cause brain damage (ie: brain rot, the chaplain's Bible/setting your religion to something retarded)?

Zuhayr:
No. Arbitrarily increasing the amount of damage the brain can take relative to other organs is not more desirable than just adjusting brainloss. This PR was made after the staff chat agreed it was needed.

Bible/religion stupidity doesn't happen in Baycode, brainrot sets the value directly and has been adjusted by this PR.

PR #883, opened 2016-01-31
Removing Cryo bag damage

Telvayne:
This appears to cover everything?  It should only remove the damage-over-time effect of stasis bags.  Please review this code as I am not an expert.

Stasis bags are currently less than useless.  A simple shot of inapprovalin and a little hustle easily does more than they do.  As well, they cause genetic AND brain damage while in use!  This requires a trip to cryo to repair, and maybe more.  My suggestion is that we remove the damage so that stasis bags can be properly used without fear.

### DIFF
--- a/code/modules/mob/living/carbon/human/life.dm
+++ b/code/modules/mob/living/carbon/human/life.dm
@@ -86,7 +86,6 @@
 		if(!client)
 			species.handle_npc(src)
 
-	handle_stasis_bag()
 
 	if(!handle_some_updates())
 		return											//We go ahead and process them 5 times for HUD images and other stuff though.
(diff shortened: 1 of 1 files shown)


comma:
They were added back in the dark days of Internal Bleeders reign of fear, so this damage was a small tradeoff for ability to actually get guy to medbay without him dying on the way. As IB now, there's really so hurry to justify that, so seems reasonable to me. Actually I can't think of scenarios where it'd come in handy over inaprovaline / roller. Maybe severe poisonings?

Telvayne:
Right.  There won't be a lot of times when a bag will be the best option, but at least with the damage removed it becomes _an_ option.

zekesulastin:
The problem was they weren't really super useful even when IB had its reign of terror, what with being one-time-use bags that were a pain in the ass to replace that cannot be health analysed through and did not give any information at ALL about the patient inside.  Plus, they never did protect against space and such.

Oh well.

PsiOmegaDelta:
Statis/cryo bags is codewise a rather awful hack, where each relevant proc is responsible for checking and handling the stasis situation. Life() code would really be better off with it removed.

Kelenius:
Not to mention colon cancer.

Mark9013100:
Gotta worry about those colons!

Telvayne:
I wouldn't say no to just burning them to the ground and never mentioning them again.  I bet medical wouldn't even notice they were missing if you didn't tell them.

Mark9013100:
I've only seen a stasis bag used once on Polaris, and I'd use them myself if there wasn't the genetic/brain damage.

zaredman:
I've only used stasis bags to hold removed brains when JimBob McOHDEARGOD blew himself up.

Not going to lie? I don't think many people would miss them.

PR #888, opened 2016-02-02
Mulebots are now mobs

Kelenius:
obj/machinery/bot removed completely
Mulebot hacking removed, you can still disable their safety.
Mulebot PDA interface redone.


Telvayne:
Does this stop them from getting stuck in maintenance because they can't move past each other?

Kelenius:
One of two things will happen: if there's no other way to the target other than through the occupied tile, it will eventually give up and head home. If there is, it will go that way. So they still get stuck but will eventually get unstuck (the one closer to home will have a path to it) except in a rare situation where their homes are in entirely different places and they both block each other's only path to home. That requires the home to not be in cargo.

OhEmGee:
B-but what about mulebot TURBO SPEED and surfing them?

Kelenius:
You can still ride them, I'll look into the speed.

Kelenius:
They move at the speed of one tile per second and it can't be changed.

Zuhayr:
If you can't get these back to being feature complete with the previous MULEs then I don't think this is going to be merged.

Kelenius:
Instant five-tile bursts? Hacking that every quartermaster does at the start of the round if they ever plan to use the mules at all because there are literally no downsides to it? Is it really better than smooth, relatively fast movement?

Neerti:
Antags can't run people over with these anymore if I'm not mistaken.

Kelenius:
No, safeties and load restrictions are still disableable, you just don't need to hack it.

GinjaNinja32:
If you poke at BYOND's gliding movement vars and do some `sleep()` you should be able to do faster movement.

Kelenius:
That's not a problem. They are processed once per tick, and if I want to make them move multiple times per tick, I can just toss in some sleeps to make them step several times per process, with pauses. I just don't think they need faster movement via hacking. It's a straight upgrade that uses no resources, so there's no reason to not do it - so everyone who plans to use them will most likely do it. It also has a problem that mulebots with disabled safeties that move too fast leave people no time to react. One tile per second (two per tick) is reasonably fast for deliveries, and can easily knock someone down by surprise, but not a problem if you are actively avoiding them.

Kelenius:
Aright, so, any issues? Do you want me to put the hacking back?

Kelenius:
Beep boop.

Neerti:
Conflicts prevent merging currently.

Kelenius:
Resolved

PR #898, opened 2016-02-04
EMP tweaks/fixes

Neerti:
For borgs, EMP no longer results in a stun.  Instead, they are confused, and their HUD becomes static-y for a moment.  Their module items are also forced back into the borg, effectively disarming them.
EMP now drains powercells from current charge instead of max_charge, so two blasts should not empty the cell, for borgs.
Fixed bug where EMP was being applied twice to human mobs, including FBPs, resulting in them literally dying in one hit.
Synthetic organs take much less damage from EMP.
Synthetic external organs also take less damage, but more than their organs.
Bonus: Runtime fix for spawning in as a borg.


zaredman:
>Synthetic organs take much less damage from EMP.

'Much less' = ?

code/modules/organs/organ.dm
```diff
@@ -231,6 +231,7 @@ var/list/organ_cache = list()
 
 //Note: external organs have their own version of this proc
 /obj/item/organ/proc/take_damage(amount, var/silent=0)
+	world << "take_damage([amount], [silent]) called on [src.name]."
```
PsiOmegaDelta: World

code/modules/organs/organ_external.dm
```diff
@@ -213,6 +222,7 @@
 	return (vital || brute_dam + burn_dam + additional_damage < max_damage)
 
 /obj/item/organ/external/take_damage(brute, burn, sharp, edge, used_weapon = null, list/forbidden_limbs = list())
+	world << "take_damage([brute], [burn], [sharp], [edge], [used_weapon], [forbidden_limbs]) called on [src.name]."
```
PsiOmegaDelta: More of it.

Neerti:
Previously all synthetic limbs and all synthetic organs received 20 damage.  As FBP cells have a health of 35, they'd drop dead in two hits (and due to bugs they actually died on the first pulse).  Now organ damage and limb/external damage is seperate, organs taking 5 damage per pulse, and external things taking 8.  Remember that every single limb is being hit for that, and generally in two or three pulses, FBPs will be dropping their stuff and falling over frequently.

code/modules/organs/organ_external.dm
Neerti: I'm a dummy.

martin509:
How much of an inconvenience is, say, an EMP grenade for FBPs/borgs now? I remember being nearly killed (hardcritted I believe) by a tator with a single point-blank EMP grenade, so how much will one EMP grenade do now?

Neerti:
A point blank EMP grenade will still do roughly 88 damage in total, or 8 per external organ, which isn't too bad so long as they don't get hit by another EMP, as the second pulse tends to see bad effects manifesting.  Any energy weapons they have will also be empty, so EMP grenades and the ion rifle are still threats, but hopefully this makes them not be an instant win weapon.

Fox-McCloud:
Wouldn't this pretty much mean the only thing that can stun a borg now is a flash and flashbang?

PR #899, opened 2016-02-04
Allows suit cooling units to be attached to voidsuits.

Neerti:


code/modules/clothing/spacesuits/void/void.dm
```diff
@@ -241,5 +265,16 @@
 			W.forceMove(src)
 			tank = W
 		return
+	else if(istype(W,/obj/item/device/suit_cooling_unit))
+		if(cooler)
+			user << "\The [src] already has a suit cooling unit installed."
+		if(tank)
+			user << "\The [src]'s airtank is in the way.  Remove it first."
```
Techhead0: This should be `else if(tank)`, otherwise you'll be able to install a cooler when you already have one installed. Ditto above with tanks.

Telvayne:
I love you Neerti.

PR #907, opened 2016-02-06
Teshari tweaks.

Zuhayr:
Pardon the dumb branch name.


Minijar:
Maybe I'm missing something, but is Teshari a new race? Or just a renamed Resomi?

EmperorJon:
Their language looks a bit Schechi, but other than that, accepted. ;)

PR #909, opened 2016-02-06
Adds  small departmental posts

Minijar:
This adds small departmental posts for security,engineering and medical. In the form of two checkpoints, a monitoring station and a first aid station.


EmperorJon:
Hmm, something the staff will probably have to deliberate on but I actually quite like the idea of Security having little posts around.

Earthcrusher:
What EJ said. Adding a few details like this here and there might help break up the monotony in the hallways.

Spookerton:
I like this in concept, but I have problems with the inclusion of records consoles at the posts - I'm not particularly happy about the one in arrivals to begin with, as these places are usually empty as the grave and are effectively just used as either overflow access to security gear when there's none left in the brig, or for antags to get a quiet moment to futz with records with zero risk.

Minijar:
I could replace the cargo records console with the mining camera console.

Minijar:
Okay so I added two little additional rooms from @EmperorJon  advice. A small monitoring station for engineering and a little first aid station for medical.

Yoshax:
Seems really unecessary, imo. Security don't need to be up in everyone's business, which this will promote more, I believe. The other two just seem unnecessary as well, and I don't believe they'll be used at all.

Zuhayr:
I like it and I can see it being used a lot given how much Engineering likes to build things outside of its borders.

Jakksergal:
I like the idea of more outposts for departments considering the size of the station, but the Engineering post is very poorly placed. You might want to move it to the  location below and add some very very basic gear like a welding backpack and some metal/glass. https://gyazo.com/e73ba265b7b5a882b5cecd4bd461bb46

Minijar:
This is ready for merging if anyone if staff are happy to do so.

sabiram:
Maybe it's just the screenshot but the orange in the engineering one is more orange than the orange in the engineering wing I think

Minijar:
Seems some conflicts have magically appeared for this so I'll commit it again in a bit.

Telvayne:
I love it.  Small places for comfy RP, away from your department, but still 'on the clock'.

Minijar:
Reopening this to fix conflicts due to previous map edits

OhEmGee:
There should be a small cargo office too, since cargo is so far away from most of the other departments. It could have a chute for mailing and a supply request console.

PR #916, opened 2016-02-07
Switch places of No and Yes in give verb to Yes and No

Lorwp:
I fixed it boys (i hope).

### DIFF
--- a/code/modules/mob/living/carbon/give.dm
+++ b/code/modules/mob/living/carbon/give.dm
@@ -15,7 +15,7 @@
 		usr << "<span class='warning'>You don't have anything in your hands to give to \the [target].</span>"
 		return
 
-	if(alert(target,"[usr] wants to give you \a [I]. Will you accept it?",,"No","Yes") == "No")
+	if(alert(target,"[usr] wants to give you \a [I]. Will you accept it?",,"Yes,"No") == "No")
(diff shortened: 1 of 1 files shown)


Neerti:
:+1: && Travis

Lorwp:
Welp, let me try and fix it (again)

Zuhayr:
Why? It was changed so that you can't accidentally accept something.

Zuhayr:
Keep in mind that this includes things like grenades, things that people might want your prints on, contaminated shit...

Lorwp:
@Zuhayr Yeah, probably worth quickly polling the forums about this

Neerti:
The situations where you're being given a live grenade or other malicious items by baddies is very much in the minority compared to normal people giving normal things and having to do it twice or three times because the button positions were swapped.

Also, given time, people will accept being handed live grenades once they get used to moving their mouse to the other side, so changing the positions is really not that helpful overall imo.

Lorwp:
Yeah, i'm going to close  this. Unless we suddenly get outrage about people  not wanting to move their mouse, this probably will work out better anyway.

Zuhayr:
@Neerti the point is that if you are typing and hit space when the prompt pops up, it will default to 'Yes' before you even see the window, like what happened with positronic brains before they were changed in the same way.

PR #930, opened 2016-02-08
Some extra Teshari mechanics.

Zuhayr:
- Teshari get hungrier faster.
- Teshari have lower blood volume.
- Blood volume is now calculated per species and handled as a percentage.


Zuhayr:
This isn't working and I -fucking despise- debugging this kind of health code so fuuuuck it.

PR #931, opened 2016-02-08
Rubbers and beanbag rounds will now respect armor

Neerti:
Halloss bullets now will be reduced by melee armor.
Rubber round damage (brute) halved.


Zuhayr:
:+1:

PR #933, opened 2016-02-08
Small mob reagent changes.

Zuhayr:
- Small species now take smaller bites/gulps from food and drink.
- Small species are now effected by alchol and toxins twice as much.
- blood_volume is now a species-level var and Teshari have lowered blood volume.
- Hunger is a species var and Teshari get hungry faster.
- Small mobs have a reduced climb delay.
- Teshari gain more nutrition from meat.


html/changelogs/Zuhayr-tesharireagents.yml
```diff
@@ -0,0 +1,9 @@
+author: N3X15
```
Neerti: Think you forgot to change this.

Zuhayr: Cheers.

PR #934, opened 2016-02-08
Removes cloaking device

Kelenius:
Adminspawn-only device that is never used, but every time a human
updates icons, they check if they have one. Neat!


Zuhayr:
Better handled with invisibility or alpha now anyway. :+1:

PR #938, opened 2016-02-08
Makes stasis bags better

Neerti:
**``Life()`` no longer checks if the mob's inside a stasis bag and if so, sets the bag to used every life tick.  The bag itself does this now, and this allows for alternate sources of the stasis effect.**
Cryo (the log-out kind) applies the stasis effect.  Robot cryo, gateway, elevator, etc, do not apply this effect.
Stasis bags now have a tech level.
Stasis bags protect the occupant from hostile environments such as vacuum.  This is to attempt to fit the stasis bag in a specific niche, where an EMT may have to rescue someone who's in a breach or other horrible thing but can't bring them out in a timely manner, as pressure damage is quite high and crits you in under a minute.
Stasis bags can be examined to examine the occupant if you are next to them.
Stasis bags can be hit with a health analyzer to analyze the occupant.
Stasis bags can be picked up if unrolled, if they were not already used.
Examine tab ``description_info`` filled for both stasis bag types.

### DIFF
--- a/code/datums/supplypacks.dm
+++ b/code/datums/supplypacks.dm
@@ -456,7 +456,7 @@ var/list/all_supply_groups = list("Operations","Security","Hospitality","Enginee
 	group = "Medical"
 
 /datum/supply_packs/cryobag
-	name = "Statis bag crate"
+	name = "Stasis bag crate"
 	contains = list(/obj/item/bodybag/cryobag,
 				    /obj/item/bodybag/cryobag,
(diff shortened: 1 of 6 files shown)


EmperorJon:
Seems reasonable to me to see if they actually become useful.

Mark9013100:
Any chance you can fix the name for the stasis bag pack in cargo supplypacks?

Neerti:
Name should be fixed in the cargo supplypack now.

PR #970, opened 2016-02-12
Pull Down Gas Masks and Medical Masks

redstryker:
Allows you to pull down gas masks and medical masks.  Obviously you can pull them back up.  Tested, works fine.


code/modules/clothing/masks/gasmask.dm
```diff
@@ -27,6 +28,37 @@
 
 	return filtered
 
+/obj/item/clothing/mask/gas/attack_self()
+	toggle()
+
+/obj/item/clothing/mask/gas/proc/adjust_mask(mob/user)
+	if(user.canmove && !user.stat)
+		src.hanging = !src.hanging
```
Fox-McCloud: src. is implied; not necessary.

redstryker:
@Fox-McCloud , (sorry, I'm new to Byond code, and Git) what do I do then?

Fox-McCloud:
AhH! No problem. You'd just want to remove the src. is all so it would be:

`hanging = !hanging`

Zuhayr:
@Fox-McCloud you aren't a developer or a player on Polaris, please stop telling people what to do on our tracker.

@redstryker you do not need to remove src. There is nothing in the style guide about it and dev discussion has concluded that it isn't remotely important enough one way or the other to mandate.

tigercat2000:
@Zuhayr Just offering help, asshole.

tigercat2000:
@Zuhayr He offered advice, and using `src.` is bad. Fuck off.

Zuhayr:
I'm a maintainer on this repo and consistent unsolicited comments from people who aren't involved with the code are not particularly useful or appreciated.

Additionally, insults are unwarranted and childish. Please try to moderate yourself.

Spacemanspark:
@tigercat2000 To be fair, he does comment on a lot of things when not needed.

Spookerton:
@tigercat2000 treat the locals nicely, thanks. Critique is good, diatribe is bad. Respecting the difference is required, not optional.

Zuhayr:
@redstryker Would it be worth moving all of these boilerplate procs down to the general mask definition and allowing any mask with a flag set to be adjusted?

EmperorJon:
I had considered that myself. We'd rather have that than have it repeated in every type of mask that can be adjusted. The question indeed is do we want a flag for if it can be pulled down, or do we want to look through the remaining masks and implement the functionality for all of them? I'm not aware of exactly how many masks we have.

redstryker:
@EmperorJon and @Zuhayr , we have /many/ masks.  Given time, I could sprite them, but I really don't think that that's necessary.  I will try to figure out how to flag it or something.  Help would be appreciated, but I can talk to someone else about it if needed.

Datraen:
@redstryker Set it to check for a value, and make sure the general definition is set to make the check return. See https://github.com/PolarisSS13/Polaris/blob/c103bd1cf7a417141357de5ab8c63c5348cb49d4/code/modules/clothing/clothing.dm#L740 for more details.

EmperorJon:
Simply pull the shared code up to mask from gasmasks, breathmasks, etc.
Then have a check to only add the verb if a certain "can be pulled down" flag is true. Define that flag for masks with a default of 0 and set it to 1 for masks which have the required sprites and can bve pulled down.

Spookerton:
This works, but involves some code duplication and should probably see real gas masks refactored as children of a parent with pull_mask 0, since they are something of a minority.

PR #976, opened 2016-02-13
Fixes an inverted robot check.

Zuhayr:


Telvayne:
Like, a robot doing a handstand?

PR #977, opened 2016-02-13
Adds departmental posts [Second Edition]

Minijar:
This fixes the previous conflicts and is the same as the last pr.
#909


OhEmGee:
Reposting my idea, since I realized that I goofed with the PR being closed. There should be a small cargo office too, since cargo is so far away from most of the other departments. It could have a chute for mailing and a supply request console.

zaredman:
Staff chatting about this, @Minijar:
* Move the Engineering one further away from Engineering. Maybe by that cafe void.
* Medical one is liked.
* Additional security ones are sorta pointless as there is already a security post by Arrivals that's... Not really used. Additional security posts are not really liked by Staff/Management.

zaredman:
Additional chatter, and a direct quote from our forums:

* The security posts detract from security's ability to not sit alone in a box with only a radio for company, and including a full set of computers in each one is only asking for trouble. No security posts.
* The engineering post is like, ten seconds away from the /better/ set of consoles in engineering itself and is a waste of power.
* The MEDICAL post is nice, because of the station scale. Adding a second one somewhere around the holodeck would be nice.

So in short, scrap the engineering and security posts. The former's useless and the latter are, depending on the HoS of the round, either going to remain unoccupied and looted or used as punishment boxes.

[end quote]

* Change Engineering checkpoint to more of an 'Engineering emergency store-room.' The old boxstation map had a plethora of air pumps/emergency tools scattered around maint. NS does not. Having a place for that somewhere north-east of the station would be cool.

Minijar:
Update

OhEmGee:
Suggestion: Maybe the medposts could have the smaller nanomeds? Even the patient rooms have their own, so why not the posts?

Kelenius:
A /sleeper/? You know that it can basically save you from anything short of severe - and I mean SEVERE - organ damage?

Minijar:
Going to have to refork this and reopen and combine it with #1005 , whenever something else mapping wise is merged before this, it gets conflicts.

PR #981, opened 2016-02-13
Removes ACMEco cigarettes.

Mark9013100:


Minijar:
Fun police detected. ;)

Issue #925, opened 2016-02-07
Plant Analyzers don't report if a plant produces or consumes gases


RoseCG:
I have been noticing in Xenobotany I get plants that produce or consume gases, but nothing is reported on the plant analyser. Meaning that innocent looking exotic mushroom tray is now filling the lab with phoron.

I had heard that plant analysers reported on whatever a plant produced gases or not (without saying which kind of gas), so I'm not sure if this is a bug or not.

Datraen:
#983

PR #983, opened 2016-02-13
Fixes plant analyzer and hydroponics tray information

Ithalan:
Fixes plant analyzer not showing data about gasses consumed or exuded by plants
Fixes hydroponics trays not showing sensor information when empty

### DIFF
--- a/code/modules/hydroponics/trays/tray_tools.dm
+++ b/code/modules/hydroponics/trays/tray_tools.dm
@@ -96,7 +96,7 @@
 
 		dat += "<br>This sample contains: "
 		for(var/datum/reagent/R in grown_reagents.reagent_list)
-			dat += "<br>- [R.id], [grown_reagents.get_reagent_amount(R.id)] unit(s)"
+			dat += "<br>- [R.name], [grown_reagents.get_reagent_amount(R.id)] unit(s)"
 
 	dat += "<h2>Other Data</h2>"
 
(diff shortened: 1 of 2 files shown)


code/modules/hydroponics/trays/tray_tools.dm
```diff
@@ -197,18 +201,24 @@
...
 	if(grown_seed.get_trait(TRAIT_TELEPORTING))
 		dat += "<br>The fruit is temporal/spatially unstable."
-	
-	if(grown_seed.get_trait(TRAIT_EXUDE_GASSES))
-		dat += "<br>It will release gas into the environment."
-
-	if(grown_seed.get_trait(TRAIT_CONSUME_GASSES))
-		dat += "<br>It will remove gas from the environment."
-
-	if(grown_seed.get_trait(TRAIT_EXUDE_GASSES))
```
Zuhayr: I wonder why this was done twice?

Datraen: Because a commit didn't make it into the PR I did fixing that, I guess.

Datraen:
It was purposely excluded for Xenobotanists, who have access to gas scanners, to be able to experimentally figure out what the gases are, as a way to help discourage gardeners ordering exotic seeds.

Ithalan:
Couldn't that be accomplished much more effectively by just making the exotic seeds crate require xenoflora access to unlock?

Every other plant property can be identified either through the analyzer or visual inspection of the plant icon, so hiding the existence of the gas property specifically just seems like a pointless inconsistency waiting to fuck new players over (I certainly didn't find it enjoyable to be screamed at by others for creating overpressure the first time I planted some exotic seeds, when there was no way outside of codediving that I could have known that the gas property could exist.)

Datraen:
The plant analyzer was showing whether or not it exuded or consumed gas, if you scanned the seeds prior to planting them, you would see it. 

Yes, we could lock the crates to xenoflora access, but at the time that it was done, plants were very much the same, and it was the only variable that wasn't directly given to you by the scanner, save for the contents of the fruit.

Datraen:
For reference: #925 was a related bug report

PR #986, opened 2016-02-14
Restores ACMECo Cigs

ParadoxSpace:


Zuhayr:
No.

EmperorJon:
Ok so. Given you were complaining in OOC about just being told "no", here's a little explanation for you:

When the Head Developer merges a PR that removes something, trying to get it added back immediately after isn't going to get accepted.

PR #988, opened 2016-02-14
Cards Against The Galaxy.

Zuhayr:
Kinda-WIP while I wait to get more black and white card ideas, and run the lists past the adminbus. This could possibly be moved into config if it's a serious problem.

If you are unfamiliar with Cards Against Humanity, here is a helpful link: https://en.wikipedia.org/wiki/Cards_Against_Humanity


Zuhayr:
Mechanics are working just fine other than a runtime from card code that has been fixed.

code/modules/games/cards.dm
```diff
@@ -214,7 +224,7 @@
 	if(!discarding || !to_discard[discarding] || !usr || !src) return
 
 	var/datum/playingcard/card = to_discard[discarding]
-	qdel(to_discard)
+	to_discard.Cut()
```
Zuhayr: I am assuming that this list was nulled for a reason previously (the qdel) so I've done the Cut() here. No idea why it needs to be, though, considering this proc doesn't involve spawn or waitfor=0.

PsiOmegaDelta: Warning, tech talk ahead.
Lists should never be qdel()d, because the garbage collector can only handle subtypes of /datum which lists are not.
There are a few places where it happens due to my mass change of del() -> qdel(), and I should've checked for it.

Nulling a list is generally a bad idea unless BYOND has fixed a reference leakage bug. Cut() (and previously del()) is definitely preferable. It's just that some pieces of code check for null-lists to do initialization and similar sort of things.

Zuhayr: Roger.

Fox-McCloud: @PsiOmegaDelta 

Are you completely positive about that? I thought lists were finally moved to be a child of datum?

Spookerton:
Shiny. Admin :+1:
_Cards Against Humanity is [BY-NC-SA 2.0 CC](https://creativecommons.org/licenses/by-nc-sa/2.0/)._ according to [their page](http://cardsagainsthumanity.com/) So I imagine we should include a notice as code/modules/games/cah.license.txt or similar.

code/modules/games/cards.dm
PsiOmegaDelta:
```
world/New()
	..()
	var/list/a_list = list()
	world.log << istype(a_list, /datum)
```
Outputs 0

Fox-McCloud: Yeah; I just intentionally passed our `qdel` a list and it hard deleted it---noticed, then made the edit.

OhEmGee:
Finally.

Earthcrusher:
Literally best addition ever.

Zuhayr:
This is fine to merge if people are okay with the lists of cards. More can be added as people come up with them.

Datraen:
Include the license as a separate text file.

Zuhayr:
It's not literally Cards Against Humanity, it's a parody and doesn't include any rules or material actually taken from the game. Not sure why we would need to include a license file at all.

Datraen:
:+1:

PR #996, opened 2016-02-15
[WIP] Shuttletwo

Kelenius:
tl;dr: those are shuttles that can fly to custom coordinates. The part that moves the area is working, with some issues, and they can also dock/undock, but there's still a lot of work to do.


Minijar:
Is this similar to the old helm control console?

Kelenius:
>Is this similar to the old helm control console?

No idea, what did it do?

Minijar:
> No idea, what did it do?
Let you fly a ship around using coordinates, it also showed you your ships position on a map and you had to use little arrow buttons to jump tiles forward/backward etc.

Kelenius:
Yes and no. I want to remove pre-defined zones for the shuttles and replace them with saved coordinates, plus you could add your own - e.g. mercs will start with the same set of points they can move to, but they can also set their own. That would also allow you to have overlapping locations for merc and heist shuttle.

In a long-term, constructible shuttles.

Minijar:
Probably unfeasible, but what about crashing a shuttle into the colony using this system?

Kelenius:
Hmm... Well, what it does is determine the shape of the shuttle, then pick all turfs in the target area that would be covered by it, generate a new zone out of those turfs, move the shuttle there, then delete the old shuttle area. Doing it with turfs that are already filled with something would be no different. At the moment, however, it's set to only accept space turfs in a space area and abort the move if the target differs.

comma:
Can check for overlaps (solid turf here and at destination) and call ex_act(1) or something.

Kelenius:
It uses ``move_contents_to`` which AFAIK overwrites the old tiles.

Kelenius:
This is too much work to do at once, I'll need to clean up docking first. The transfer isn't the problem, the docking system is.

PR #998, opened 2016-02-16
Adds a variable to check for grenade loading, makes frags unloadable.

Datraen:
As the title reads, grenades now have a variable to check if they're loadable or not, and gives an error message when you're trying to load them into a grenade launcher.

This was requested by staff.


code/game/objects/items/weapons/grenades/explosive.dm
```diff
@@ -40,6 +40,7 @@
 
 	//The radius of the circle used to launch projectiles. Lower values mean less projectiles are used but if set too low gaps may appear in the spread pattern
 	var/spread_range = 7
+	loadable = null
```
Kelenius: ``null``? Why not ``0``?

EmperorJon:
Honestly, I'd prefer 0 myself rather than null, but I've hit the merge button. Dang.

PR #1000, opened 2016-02-16
Minor - Mouse players can now "rest" properly

RusskiSam:


Zuhayr:
Doesn't the root verb handle setting resting and printing out a message? Could you just call ..() and then update sprite?

PR #1001, opened 2016-02-17
Adds player panel, ckey info to Check Antagonists panel.

Datraen:
Administration stuff, but the panel after the PR:
 

Player Panel is directly linked, and the ckeys are visible at for any antagonist.

### DIFF
--- a/code/game/antagonist/antagonist_panel.dm
+++ b/code/game/antagonist/antagonist_panel.dm
@@ -27,13 +27,13 @@
 		var/mob/M = player.current
 		dat += "<tr>"
 		if(M)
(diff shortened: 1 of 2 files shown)


code/game/antagonist/antagonist_panel.dm
```diff
@@ -27,13 +27,13 @@
 		var/mob/M = player.current
 		dat += "<tr>"
 		if(M)
-			dat += "<td><a href='?src=\ref[src];adminplayeropts=\ref[M]'>[M.real_name]</a>"
+			dat += "<td><a href='?src=\ref[src];adminplayeropts=\ref[M]'>[M.real_name]/([player.key])</a>"
```
PsiOmegaDelta: I noted on Bay that this line was intended to give access to the player panel, the issue is that the href section is incorrect.

PR #1002, opened 2016-02-17
[DNM] [WIP] Xenobio rework

Datraen:
Adding a PR for tracking purposes.
This is incomplete, do not merge.


code/modules/mob/living/xenomorphic/life.dm
```diff
@@ -0,0 +1,94 @@
+
+/mob/living/xenomorphic/Life()
```
PsiOmegaDelta: If you're making a new mob/living subtype, without all the legacy of carbons, I strongly suggest that you override the procs called in /mob/living/life.dm instead of creating a monolithic Life() monster.

Kelenius:
Looking over it, seems that you might want to make them ``simple_animal`` subtype.

Datraen:
I considered just making them a simple mob subtype, but creating a new mob type is only a little more work.

Kelenius:
The point is that you won't need to copypaste things, and that any changes in the future won't need to be done twice. Still, your call - if you think they're too different, go for it.

Datraen:
I was considering taking most of simple_mob's life, copying wholesale and tweaking it. Something I'll decide soon, when my tests slow down.

PR #1014, opened 2016-02-20
Removed the Vox Pox.

Zuhayr:


EmperorJon:
I will support Zuh doing this.

Datraen:
As far as I'm aware, we're keeping them for external antagonists.

Anewbe:
I think the raiders at least should be left in the code, and if possible removed from the character select menu, but Pariahs can and should go, completely. Concept wise I really like the Vox, even if they've been played poorly in the past. Their weird biology, genuinely strange morals, and mysterious background make them one of the more interesting species I've seen in SS13.

Or at least comment it out, rather than a flat remove, just in case?

Zuhayr:
@Datraen asked the TS, people were fine with removing them. They have been played so badly for so long that they are not going to be recoverable, the limitations that drove their design process don't exist anymore, and there is nobody on staff to actually head the whitelist. Their lore is Byzantine, overengineered, and impossible for the average player to -play-, and it all just needs to go.

I would happily add a new antagonist species with a similarly weird perspective if you desperately want one, but Vox aren't currently used except once in a blue moon and removing them is not going to significantly negatively impact anyone.

@Anewbe Commenting things out is for codebases that don't have the entire history of the project backed up in source control.

Kelenius:
You forgot to remove their special chemical effects.

alabasterTempest:
I'm behind removing them. Heist gets run once in a blue moon if that, and most of the time it's either an all human crew, or a mixed crew of station-xeno and humans. Vox so rarely get even mentioned.

EmperorJon:
Regarding Ala's above point, we had the first Heist round I've seen in weeks earlier today. No Vox. It's also probably worth noting all the Vox-inclusive Heist rounds I can recall seemed to end with the Vox betraying and/or eating their human counterparts.

EmperorJon:
Just to confirm @Zuhayr, this is all good and ready to go? It's had the approval of our glorious leader Ala so I'm more than happy to merge it.

Zuhayr:
It's fine to merge, I'll do a commit to remove IS_VOX down the track.

zaredman:
@EmperorJon I'll do a forum thing today

PR #1021, opened 2016-02-20
Fixing the 'Succumb' verb for all species across the board

### LINKED ISSUE #1013: Succumb verb does not kill Teshari
What I did: Playing Teshari, floating through space with no suit, used succumb at critical and did not die for another ten minutes or so.

Guessing it has to do with the fact that I wasn't actually at 'normal' critical levels that a human mob would experience.

Telvayne:
This will need review, but may repair the succumb feature for all races, regardless of health levels.  I'm still learning this, and my syntax may be wrong.

Currently, succumb works off of the default health of 100.  This will adjust it to call the maxHealth variable of the mob and allow/deny succumb accordingly.  I think this is right(?)

This should fix #1013

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -149,9 +149,9 @@ default behaviour is:
 
 /mob/living/verb/succumb()
 	set hidden = 1
-	if ((src.health < 0 && src.health > -95.0))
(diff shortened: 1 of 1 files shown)


Zuhayr:
:+1:

PR #1027, opened 2016-02-20
Changes types of ghosts and eyes

Kelenius:
``/mob/dead/observer`` -> ``/mob/observer/dead``
``/mob/eye`` -> ``/mob/observer/eye``

Reason being that they are similar in that they both don't interact with the world in any way. Some procs were shared, some checks as well, and it overall makes more sense this way. Plus, there were no ``/mob/dead`` mobs.

### DIFF
--- a/code/modules/admin/verbs/debug.dm
+++ b/code/modules/admin/verbs/debug.dm
@@ -85,7 +85,7 @@
 	var/mob/choice = input("Choose a player to play the pAI", "Spawn pAI") in available
 	if(!choice)
 		return 0
-	if(!istype(choice, /mob/dead/observer))
+	if(!istype(choice, /mob/observer/dead))
 		var/confirm = input("[choice.key] isn't ghosting right now. Are you sure you want to yank them out of them out of their body and place them in this pAI?", "Spawn pAI Confirmation", "No") in list("Yes", "No")
 		if(confirm != "Yes")
 			return 0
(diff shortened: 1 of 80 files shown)


code/modules/admin/verbs/debug.dm
```diff
@@ -231,7 +231,7 @@
 	set name = "Del-All"
 
 	// to prevent REALLY stupid deletions
-	var/blocked = list(/obj, /mob, /mob/living, /mob/living/carbon, /mob/living/carbon/human, /mob/dead, /mob/dead/observer, /mob/living/silicon, /mob/living/silicon/robot, /mob/living/silicon/ai)
+	var/blocked = list(/obj, /mob, /mob/living, /mob/living/carbon, /mob/living/carbon/human, /mob/observer/dead, /mob/observer/dead, /mob/living/silicon, /mob/living/silicon/robot, /mob/living/silicon/ai)
```
comma: /mob/observer/dead twice

Zuhayr:
There's nothing particularly wrong with this but it does seem pretty pointless.

Kelenius:
I can imagine a lot of scenarios where you'd want to specifically avoid observers. Basically in any situation involving 'how do they interact with the world'. Teshari loneliness, for example, or give verb. Right now it doesn't give a lot of advantages, but can potentially save a lot of headache in the future. Clean code structure is important.

EmperorJon:
I think we need to decide if we're going to take this or not. I'm with Zuh here in that it looks like it's fine, and I can understand why it could under certain circumstances be useful, but I'm going to suggest @Neerti makes the call here.

Kelenius:
@Neerti

Kelenius:
Anyone?

Datraen:
I'm a little more in favor of Baystation12#12358 as opposed to this PR, as it still defines the mobs a level higher than /mob.

PsiOmegaDelta:
If the OP is anything to go by and a quick glance at the diff there is no real difference of mob definition depth between this PR and Bay's.

The main difference, besides the cleanup, is that were this PR has changed it from
```/mob/dead/observer``` -> ```/mob/observer/dead```
Bay instead went for
```/mob/dead/observer``` -> ```/mob/observer/ghost```

Datraen:
If I stop reading things quickly I'll notice that I'm no longer misreading. Looks okay, just need to take the time to read it over.

People were asking me about some class-related topics at school.

Ccomp5950:
You spent all that time fixing istype checks, but didn't change them to use the macros ( https://github.com/PolarisSS13/Polaris/pull/1027/files#diff-eb6c2c761051bf4fb9ae2aa4c36e6620R26 )

Jesus.

Issue #704, opened 2015-12-31
Welding Eye Trauma Inconsistency


Nalarac:
Fixing superficial dents on a prosthetic with a welder will yield no eye trauma. Robo-surgery with a welder does yield eye trauma, so this became apparent to me.

Zuhayr:
Misread reffed PR.

Nalarac:
Tested on: master - 2017-12-08 11757132352255fa88f0d06597edcb4efa816fa7 

This is no longer an issue.

Issue #768, opened 2016-01-12
Never-ending welder.


Cairo283:
Someone asked me to make a report about this, so I did: http://puu.sh/mtdcW/777e5e4c2d.png

This welder never ran out of fuel, no matter how many things it was used on (Blob blocks, doors, ciggs, etc). I'm not sure what caused it, but if I had to guess, I'd say it was when I used when *right* as it naturally drained from being on, but I really don't know.

PR #1028, opened 2016-02-20
Unlimited Welder Fix

Telvayne:
Fixes a bug where welders are infinite use.  With this, the Welder will not turn on unless it has at least 1u of welder fuel.  Previously, it worked on 'not zero', which allowed it to weld endlessly with fractions of 1u inside (as it was not Zero, but could not deduct 1 from it when used).

Corrects bug #768


Zuhayr:
:+1:

Telvayne:
I think I can fix #704 while i'm in here.  Don't merge this just yet.

PR #1031, opened 2016-02-20
pai bug fix + changes

### LINKED ISSUE #995: pAIs can open all airlocks, regardless of access.
As said in the title, as an unfolded pAI I could open all airlocks in medbay, including CMO's office.

Yoshax:
- Fixes #995
- Allows pAIs to have access added to them or removed, they can toggle the ability to do such with a verb

### DIFF
--- a/code/modules/mob/living/silicon/pai/pai.dm
+++ b/code/modules/mob/living/silicon/pai/pai.dm
@@ -416,7 +416,7 @@
 		if (idaccessible == 1)
 			switch(alert(user, "Do you wish to add access to [src] or remove access from [src]?",,"Add Access","Remove Access", "Cancel"))
 				if("Add Access")
-					idcard.access += ID.access
+					idcard.access |= ID.access
 					user << "<span class='notice'>You add the access from the [W] to [src].</span>"
 					return
 				if("Remove Access")
--- a/code/modules/mob/living/silicon/pai/pai.dm
+++ b/code/modules/mob/living/silicon/pai/pai.dm
@@ -8,6 +8,7 @@
 	mob_size = MOB_SMALL
 
 	idcard_type = /obj/item/weapon/card/id
+	var/idaccessible = 0
 
 	var/network = "SS13"
 	var/obj/machinery/camera/current = null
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/silicon/pai/pai.dm
```diff
@@ -406,3 +409,36 @@
...
+				user << "<span class='notice'>You add the access from the [W] to [src].</span>"
+				return
+
+			else
+				idcard.access = null
+				user << "<span class='notice'>You remove the access from [src].</span>"
+				return
+	else
+		user << "<span class='notice'>[src] is not accepting access modifcations at this time.</span>"
+		return
```
Techhead0: This `else` will fire if someone hits them with a non-ID item.

code/modules/mob/living/silicon/pai/pai.dm
```diff
@@ -406,3 +409,36 @@
 	grabber.update_inv_l_hand()
 	grabber.update_inv_r_hand()
 	return H
+
+/mob/living/silicon/pai/attackby(obj/item/weapon/W as obj, mob/user as mob)
+
+	if (istype(W, /obj/item/weapon/card/id) && idaccessible == 1)
```
Techhead0: Additionally, having `istype(W, /obj/item/weapon/card/id)` before `GetID()` prevents PDAs and wallets from working.

code/modules/mob/living/silicon/pai/pai.dm
```diff
@@ -406,3 +409,36 @@
...
+
+	if(idaccessible == 0)
+		idaccessible = 1
+		src << "<span class='notice'>You allow access modifications</span>"
+		return
+
+	else
+		idaccessible = 0
+		src << "<span class='notice'>You block access modfications</span>"
+		return
```
Techhead0: You could, if you wanted, condense these lines down to just:
```
idaccessible = !idaccessible
src << "<span class='notice'>You [idaccessible ? "allow" : "block"] access modifications.</span>"
```
The `return` is also unnecessary. And you dropped a period from the sentence.

Zuhayr:
>Allows pAIs to have access added to them or removed, they can toggle the ability to do such with a verb
Extremely leery of this. Can't really express why but it feels like way too much of a buff for a role that is explicitly not meant to operate without its master.

Yoshax:
I thought it would be nice, seeing as they only get what access is given to them, and can not have any access if they don't want any. I see most common usage being given their masters access, which would allow them to move around the workplace they're in, converse with other people. Example being a chemist with one, gives their access to their pAI, and whilst they're making something and are deeply busy or suchlike, the pAI can peruse around and chat with other doctors.

code/modules/mob/living/silicon/pai/pai.dm
```diff
@@ -406,3 +409,35 @@
...
 	grabber.update_inv_r_hand()
 	return H
+
+/mob/living/silicon/pai/attackby(obj/item/weapon/W as obj, mob/user as mob)
+	var/obj/item/weapon/card/id/ID = W.GetID()
+	if(ID)
+		if (idaccessible == 1)
+			switch(alert(user, "Do you wish to add access to [src] or remove access from [src]?",,"Add Access","Remove Access", "Cancel"))
+				if("Add Access")
+					idcard.access += ID.access
```
Zuhayr: Needs to be |= to avoid duplicate list entries.

Neerti:
This is DMN pending a community vote and a decision from Alabaster, due to a disagreement among staff.

Spookerton:
Relatively neutral feedback on the topic after we discussed it to death. Nothing particularly against, so admin :+1:

PR #1038, opened 2016-02-21
Adds maglights

Sabess:
Available in sec lockers and hacked autolathe.
Does less damage than sec batons and other weapons that are available in the hacked autolathe, but lights up.


Anewbe:
Had something similar on tg a while back, very much a fan... But you may want to tone down the damage, 13's pretty high, in my opinion.

Lorwp:
13 is a bit high, Maybe lower it to something like `8`. Otherwise this has my :+1:

Sabess:
Don't agree, it's all relative. It's only available to two departments. Both of these departments have items that do more damage.

Lorwp:
i guess so, though it does feel a bit weird to have a flashlight be a melee weapon.

Zuhayr:
13 is extremely excessive for a flashlight.

Sabess:
Disagree, but I'll tweak it down seeing as a fair few people want it that way.

Nienhaus:
I should really make a v3 Maglite. But it's still nice to see the sprite used.

Zuhayr:
Leery of approving this but unclear if that's just bias or not, paging @Neerti @EmperorJon @Datraen for final approval.

OhEmGee:
>13 is extremely excessive for a flashlight
have you guys SEEN a maglight IRL? The heavier ones could be used to bludgeon somone into submission, they're like the same as your run of the mill claw hammer in terms of bludgeoning

i'd say 10 (which is the same as the toolbox) would be a good median

Sabess:
It was changed to ten if you'll have another look at the code. :P

PR #1041, opened 2016-02-21
Resleeving.

Zuhayr:
Adds a use_cortical_stack config option. With this config option set:
- Prevents cloning from checking if a mob has a ckey before scanning it, and dumping a ckey from a scan into a freshly cloned body. ie. it disables old cloning mechanics.
- Cloning can produce mindless bodies from scans. They are produced as clone (n) in a similar manner to monkeys.
- A preferences option, defaulting to 'on', causes players to spawn with a neural lace head organ which, lorewise, periodically backs up an exact copy of their neural state and their 'echo', which is a handwaved lore excuse for 1:1 player:mob mechanics.
- This lace is required for 'cloning' in that it must be implanted into a living body to reinstantiate the player after death. This is done with normal organ transplanting surgery.
- The replacement will fail if the player is in another body or not connected to the game.
- Implanting into a body with no lace, but with a player, has a 30% chance of ghosting that player and overwriting their brain with the person being resleeved.

Todo:
- [ ] Get a better sprite for the lace.
- [x] Modify languages so they will transfer with resleeving.
- [ ] Possibly add morph rejection if someone is implanted into a body from a species other than their own, or a body that is entirely unlike their previous body.
- [ ] Add a machine or table that can efficiently remove a lace and swap it to a new body.
- [ ] Add a way to produce blank laces for implantation during gameplay and as a means of getting a usable mind-state from a corpse.

### DIFF
--- a/code/game/machinery/computer/cloning.dm
+++ b/code/game/machinery/computer/cloning.dm
@@ -56,7 +56,7 @@
 	var/num = 1
 	var/area/A = get_area(src)
 	for(var/obj/machinery/clonepod/P in A.get_contents())
-		if(!P.connected)
+		if(!P.connected || (P.connected == src))
 			pods += P
 			P.connected = src
 			P.name = "[initial(P.name)] #[num++]"
@@ -324,17 +324,24 @@
 					qdel(C)
 					menu = 1
 				else
-
-					var/mob/selected = find_dead_player("[C.ckey]")
-					selected << 'sound/machines/chime.ogg'	//probably not the best sound but I think it's reasonable
-					var/answer = alert(selected,"Do you want to return to life?","Cloning","Yes","No")
-					if(answer != "No" && pod.growclone(C))
+					var/cloning
+					if(config.use_cortical_stacks)
+						cloning = 1
+						pod.growclone(C)
+					else
+						var/mob/selected = find_dead_player("[C.ckey]")
+						selected << 'sound/machines/chime.ogg'	//probably not the best sound but I think it's reasonable
+						var/answer = alert(selected,"Do you want to return to life?","Cloning","Yes","No")
+						if(answer != "No" && pod.growclone(C))
+							cloning = 1
+					if(cloning)
 						temp = "Initiating cloning cycle..."
 						records.Remove(C)
 						qdel(C)
 						menu = 1
 					else
 						temp = "Initiating cloning cycle...<br>Error: Post-initialisation failed. Cloning cycle aborted."
+
 
 		else
 			temp = "Error: Data corruption."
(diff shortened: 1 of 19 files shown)


OhEmGee:
This is great. If I was a dev, I'd :+1: this very hard.

Hubblenaut:
About time cloning gets axed for a more lore-friendlier alternative.
I just hope that medical is not going to rip mindless bodies apart to use them for limb replacements ect.

nachomeep:
I think it's a good idea to be able to give limb replacements like that. Plenty of people already include it for their characters, and I don't see why it shouldn't be a thing. Especially if we get mindless clones.

Hubblenaut:
If growing limbs like that would be a lore thing I'd wonder why like half the people are amputees with prosthetics.

Kelenius:
Lasik is a thing, but you still see people with glasses. Gameplay does not represent lore 100%. Cloning does not take 5 minutes with no repercussions in the lore.

Zuhayr:
None of this discussion is relevant to this PR. If you want to discuss lore it do it on the forums.

Anewbe:
What would you want the new neural web/lace/cortical stack sprite to look like?
Will the lace be able to take damage (mechanically), and what effects if any will that have?
Assuming the quirks I saw in testing (removing the lace ripped off the head, and there wasn't a way to fully reattach the lace once it got transplanted) are intentional or dealt with, I'm so for this.

code/game/machinery/computer/cloning.dm
```diff
@@ -324,11 +324,17 @@
...
-					selected << 'sound/machines/chime.ogg'	//probably not the best sound but I think it's reasonable
-					var/answer = alert(selected,"Do you want to return to life?","Cloning","Yes","No")
-					if(answer != "No" && pod.growclone(C))
+					var/cloning
+					if(config.use_cortical_stacks)
+						cloning = 1
+						pod.growclone(C)
+					else
+						var/mob/selected = find_dead_player("[C.ckey]")
+						selected << 'sound/machines/chime.ogg'	//probably not the best sound but I think it's reasonable
```
Fox-McCloud: TG and ourselves have a specific sound for this called genetics.ogg. It's quite distinctive; may help notify players better and not cause confusion. https://raw.githubusercontent.com/ParadiseSS13/Paradise/master/sound/effects/genetics.ogg****
TG and ourselves have a specific sound for this called genetics.ogg. It's quite distinctive; may help notify players better and not cause confusion. https://raw.githubusercontent.com/ParadiseSS13/Paradise/master/sound/effects/genetics.ogg in case you're interested.

Fox-McCloud:
Alright, just wanted to make sure I'm understanding this properly.

Cloning will produce blank clones that look like the person in question--they don't have the original person's mind/ckey/etc.

When you transplant the cortical stack/neural net from the person who died, to the clone, then the clone has the mind/ckey/etc. of the original person, correct?

code/game/machinery/computer/cloning.dm
Zuhayr: I'm not touching anything that isn't explicitly required for the mechanical changes, sorry.

Zuhayr:
Correct.

Zuhayr:
Not going to have time/capacity to finish this. I will leave the branch up if anyone wants to adopt the project. It is largely feature complete other than the aspects noted in the OP.

dangranos:
I guess this won't be ported from bay either?

PR #1044, opened 2016-02-21
Adds welding goggles to autolathes.

Nalarac:
metal cost: 1500
glass cost: 1000
category: general


Zuhayr:
In the future please name your commits more informatively.
:+1:

PR #1052, opened 2016-02-21
Removes unnecessary code for mouse rest...

RusskiSam:
...Or how rookies don't know simple things.


RusskiSam:
Basicly, #1000 but without category set and message print.
Thanks @Zuhayr

Datraen:
Works either way, but ultimately both are valid.

This way is more ideal.

PsiOmegaDelta:
An even more ideal way is for the base lay_down() proc to call update_icon() which in turn sets the appropriate icon state for the situation.

PR #1059, opened 2016-02-22
Metabolic rate

Anewbe:
Makes metabolic_rate a species var, allowing species with fast or slow metabolisms


Zuhayr:
You are setting metabolic_rate on the reagents to some kind of global variable rather than to the species variable. You will need to cast the species of the person that the reagent is reacting in, get their metabolic rate, and apply it, rather than setting it all in this manner, every time metabolic_rate is going to be referenced. 

It's not a bad idea in principle but what you're doing here is not going to work and doing it properly is going to be very annoying to code.

Anewbe:
Didn't realize this was going to the master, thanks for the input.

PR #1066, opened 2016-02-23
Adds even more underwear

Yoshax:
Sprites courtesy of Serithi


Serithi:
Pics for those curious.
Front, back and side (slightly older version of it, but it's only off by a couple pixels) http://i.imgur.com/5bIQ0mY.png
Colors http://i.imgur.com/a6NA0xq.png

Hubblenaut:
..giant bald ladies.
(Also not in favor of that sprite Serithi)

sabiram:
That's a swimsuit

Zuhayr:
Why.

Neerti:
Not a fan of this myself, since it looks a bit too.. suggestive, let's say.

Datraen:
I feel the same way as Neerti on this one.

Serithi:
@Neerti You _have_ seen the "thong" underwear sprite we have, right :v. This technically covers stuff that thong _doesn't_.

@sabiram This is a type of halterneck bra. Felt like doing something a little more different than the generic designs we have right now. And bear in mind that swimsuits aren't exactly that much different to underwear anyway.

sabiram:
I'm fairly well educated when it comes to ladies' undergarments

I think this sprite would be better as a swimsuit, that's all~

Earthcrusher:
I like it. It's surprising the shape came out so well for how little space you had to work with.

zaredman:
#underweargate2560

Completely un-shockingly, I really couldn't care about what people wear. So I'm :+1:

Serithi:
@zaredman If it works for both, go ahead and pop it in swimwear too :v.

Spookerton:
Just adding another :+1: from the general staff side, since pixel nipple covers are scary.

Yoshax:
If anyone wants anything done or whatnot, it's basically Serithi's PR, I just made it because he can't into git.

PR #1075, opened 2016-02-24
[DNM] Revolution now requires 15 people to start

Mark9013100:
Makes rev a high pop gamemode, this should help with the ungodly lowpop rev rounds that get voted and turn into traitor because there's nothing to do whatsoever.


Kelenius:
Aka: turning rev from "very rare" to "never".

Mark9013100:
What a shame.

EmperorJon:
Discussed and was acceptable to headstaff.

PR #1080, opened 2016-02-24
Updates the NanoUI map

### LINKED ISSUE #1069: NanoUI map of the Colony is out of date
Title. If anyone can update the map thingamajig, that'd be great.

EmperorJon:
Fixes #1069


EmperorJon:
Note the generate map functionality is still not set up to be used for Polaris. It works at my end because I've done some magic in the past, but I haven't actually cleaned up and committed said changes. I probably should do.

PR #1083, opened 2016-02-25
There is now a ten-second delay for carding the AI

Kelenius:


Ithalan:
wouldn't the existence of this option just encourage AI core invaders to kill the AI outright instead of trying to card it? If you're trying to neutralize the AI core, it usually takes less time and effort to kill it (shoot it up from AI upload. Push a bomb inside, etc) than to secure the core chamber itself for the carding operation, and the possibility that it might not even be possible might just make them skip trying.

If a person is in a position to card the AI against its will, there's practically nothing it can do to stop that person from taking the additional step of killing it first anyway.

Kelenius:
>secure the core chamber itself for the carding operation

You mean run up to the core and card it before the turrets even pop up?

Spacemanspark:
Does Polaris have the ability to wrench a core up and drag the core away directly, out of curiosity?

Ithalan:
> You mean run up to the core and card it before the turrets even pop up?

If that's the issue, I would suggest making the act of carding the AI take 5-10 seconds instead. This would stop the hit and run, but leave carding as a non-lethal way to neutralize a hostile AI.

Kelenius:
There is a non-lethal way of neutralizing a hostile AI: AI upload console. It's unusable on malf (uncardeable currently) and traitor AI (can't be reset anyway, so being carded is as good as being dead).

zaredman:
>Does Polaris have the ability to wrench a core up and drag the core away directly, out of curiosity?

That's in pretty much every baycode fork yea.

Ccomp5950:
Todo:  Make death consensual.

Kelenius, I'm curious, what is this actually intended to solve?  (Trying to stop back and forth "well..." "well..." "oh yeah?....")

Mark9013100:
As it stands, right now you can bumrush the AI core with an intellicard and the AI won't be able to resist. :+1:

Kelenius:
Ccomp5950, intelicard use as a weapon, which is a one-click permanent inescapable unghostable imprisonment that also shows everyone your laws (and, consequently, antag status) and leaves you at full mercy of the capturer, even pAIs get more freedom than a carded AI. If the AI is being a baddie and reset is not an option then at least it should take some effort to force-card it. Shoot it or smack it with something heavy enough times and you can card it - it can be easily revived later - so the AI can ghost and isn't forced to follow you around without radio. The way the AI core is designed, you don't need to worry about turrets if you are fast enough, because the AI can't do anything to stop you from opening the windoor. Plus, as far as I know, there are plans to remove malf and merge some of it into traitor - consider this a first step.

Spookerton:
Preliminary chatter on this topic is against the idea of AI-chosen carding, but we do like the idea of disincentivizing the particular activity of dashing in and out to avoid the turrets. The favorite is currently adding a timed completion to carding, probably on the order of 8~10 seconds, which is plenty of time for turrets to utterly ruin someone.

alabasterTempest:
As I was out/missing when above-said convo was had, I second Sparky's alternative.

Minijar:
Is this by any chance related to the round the other day when this exact thing happened to you by any chance?

Neerti:
I strongly prefer Spookerton's solution to this issue, of adding a delay to carding.

Kelenius:
As you wisssssssh.

Spookerton:
:+1:

PR #1085, opened 2016-02-25
Re-adds four way manifold to port first aid hallway - Fixes #1081

### LINKED ISSUE #1081: Piping error in front of port first aid (near bridge/robotics)
The piping near the port first aid is incorrect and disconnects one of the scrubbers from the waste vacuum line.

A 4-way pipe manifold would be there to connect both the first aid closet and the scrubber next to the pipe. Instead, a 3-way manifold was put in the other way, and disconnected the scrubber therefore.

#### What actually happened
space station 13 2016-02-24 202544

#### Steps to reproduce
N/A: mapping issue.

#### Additional info:
Server revision: master - 2016-02-25
767a6e7c28ea01f05f7cb5b60858f9d64fdc22a0

Yoshax:


PsiOmegaDelta:
This is most optional, but by adding #1081 in a comment (not only the title/commit message) the linked issue will get a link back to the PR which also shows the current merge status.
The reference above has caused said link to be setup.

Yoshax:
Oh, coolio. I didn't realise referencing in a commit vs comment had different outcomes.

PR #1093, opened 2016-02-26
Adds hijabs

redstryker:
Adds hijabs in every base color as a loadout item.


OhEmGee:
Only if you add crucifixes and kipas as well :^)

Nah, I'm totally fine with this.

Minijar:
Kind of an odd addition but whatever.

OhEmGee:
>Kind of an odd addition but whatever.

it's the muslim agenda mini

Zuhayr:
How is it odd to add a headscarf/veil? You do realize that there are a Hell of a lot of Muslim folks in the world, yes?

redstryker:
@OhEmGee , I'm actually planning on adding Kippot and Turbans later on, so all in good time.

Yoshax:
You could look into a way of having the base sprites as greyscale, and then add a way for the player to select a color that is applied, similar to the way hair is done.

code/modules/client/preference_setup/loadout/loadout_head.dm
```diff
@@ -184,4 +184,88 @@
...
+	display_name = "Dark red hijab"
+	path = /obj/item/clothing/head/hijab/darkred
+
+/datum/gear/head/hijab/green
+	display_name = "Green hijab"
+	path = /obj/item/clothing/head/hijab/green
+
+/datum/gear/head/hijab/green
+	display_name = "Green hijab"
+	path = /obj/item/clothing/head/hijab/grey
```
PsiOmegaDelta: I believe the copy-paste demon been involved, as I have a feeling this grey hijab isn't intended to be under the green entry.

AltAccount111:
@redstryker Will crucifixes be added?

redstryker:
@AltAccount111 , Seeing that I am actually working on five different things right now, not in the near future, but perhaps.  Although, I'm hesitant to add crucifixes since there is no 'jewelry' system right now, and I think that making them basically ties would just be cheesy.  Plus, there are other 'Christian' items already in game.  Take the nun outfit, or all of the chaplain's clothes, for instance.  I'm just trying to add some diversity to the platform, and right now, adding crucifixes is the least of my worries.  I have plans to sprite/code other things before I do that, but dependent, I may get around to it.

PR #1095, opened 2016-02-27
Adds a timer to staffwho AFK notifications.

EmperorJon:
When using staffwho, AFK staff (currently apparently 5 minutes of inactivity) now report exactly how long they've been inactive for, such as follows. http://i.imgur.com/aUYkuwC.png
(Ignore the missing space before minutes, that was fixed)

This is so that when you use staffwho you can see which staff are more or less nominally AFK and which staff have stayed logged on to the server but actually have been AFK for literal hours.

### DIFF
--- a/code/game/verbs/who.dm
+++ b/code/game/verbs/who.dm
@@ -86,7 +86,10 @@
 					msg += " - Playing"
 
 				if(C.is_afk())
-					msg += " (AFK)"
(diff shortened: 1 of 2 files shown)


Minijar:
Is this viewable to staff only or players to?

EmperorJon:
This expands the pre-existing AFK notification for staff only.

PR #1096, opened 2016-02-28
Fixes ID computers.

Neerti:
Also fixes what appears to be a merge error in condiments, which made my compiler complain.
oh, someone also added FILE_DIR previously, as well.  Fixed that too.


EmperorJon:
(woopsy)

PR #1099, opened 2016-02-28
Z1/Z2 map fixes, 2/28/2016

Mark9013100:
Z-Level One:
Fixes a few decal errors in medical.
Moves a window over to the MULEBOT drop off.
Removes the airlock assembly in medical maintenance.
Adds a few more Orion Trail machines.

Z-Level Two:
Adds an Orion Trail machine.
Makes the ERT spawn a tiny bit more open.


Hubblenaut:
Why removing the airlock in medical maintenance? Thought it was a nice gimmick and reference.

Mark9013100:
It's resided in the glorious utopia of maintenance long enough, like all, it must leave sometime or another.

Kelenius:
Reference?

Telvayne:
Yoshax accidently put a copy of the CE's airlock in virology during a map update.  It went live and nobody noticed it for a while, due to 'haha virology'.  When we fixed the map, the airlock was just moved to maintenance as a nod to it.

Just a sort of joke, like engineering took it apart and stuffed it in maintenance,  in case they needed it later.

It's funny, and doesn't seem really out-of-place.

ParadoxSpace:
We could always not delete that door.

Mark9013100:
gib me money and I won't burn your sacred orphan-- airlock down

Telvayne:
I'm not that committed to polaris-centric meme doors.

PR #1107, opened 2016-03-01
Remaps xenoflora

Kelenius:
Before: http://i.imgur.com/CDrFdqQ.png
After: http://i.imgur.com/6pHF0wx.png + honey extractor next to machines at the bottom, tanks and crate pushed to the right
DNM pending feedback from http://ss13polaris.com/forums/viewtopic.php?f=18&t=667
Also fffffu conflicts

EDIT: http://i.imgur.com/LSA8bT3.png


Yoshax:
Are you planning to actually make it aesthetically pleasing like the old one?

Telvayne:
I'm a big fan of having two windoors as the only thing separating central primary from horrible phoron-generating madness plants.

Also, the smaller mechbay makes sense and I like it.  Mechbay feels far too big right now.

Kelenius:
Feedback received, additions made, merge conflict resolved.

Yoshax:
Still looks a bit plain, imo. Especially the mechbay, maybe add some windows from the robotics room to the mechbay.

Spookerton:
Too many trays, too big, too close to the primary corridor, not enough separation between xenoflora storage and xenoflora itself. If a xenoflorist wants to get their human interaction on, they can do it by using the multitude of different remote comms tools available, or just leave their lab for a while. This xenoflora change is negative.

The mechbay could stand to be a little smaller, though.

Kelenius:
Can we move this discussion to the forum, because this isn't the best place :<

Spookerton:
I'm mouth-piecing a selection of staff here so I mean sure, it can be a forum chat, but the general response to this has been pretty neutral-to-bad to begin with. As far as I can tell the primary goal for this is to attach xenoflora to the main corridor, and that's just not going to happen.

Kelenius:
Abandon the ship.

PR #1110, opened 2016-03-02
Equipment storage now requires Warden access

Mark9013100:
Pending staff discussion.

Screenshot: 

Also edits the vintage .45 pistol's name.

Properly tested, works fine.


zaredman:
Flag this as DMN pending staff talks.

Minijar:
Que the angry officers reporting warden when he refuses to hand over pistols.

Overall yes, I like it. If we do this, remove the rubber magazines from lockers since they serve no purpose when you can't get its gun.

nachomeep:
I don't think this is a good change. There's no way to take down an FBP if you remove the pistols.

Dibasic:
remove the detective gun too

comma:
Remove all fun

OhEmGee:
>I don't think this is a good change. There's no way to take down an FBP if you remove the pistols.

I concur.

Kelenius:
Sounds like a reason to nerf FBPs.

Telvayne:
FBPs should get flashed in the same way borgs do.  That'd sort things out right quickly.

zaredman:
FBPs can get tasered.

Hi, I have tasered them before.

This is not a popular pull by staff. Current vote is 8 no, 1 yes.

Neerti:
> I don't think this is a good change. There's no way to take down an FBP if you remove the pistols.

Could you elaborate on that? @nachomeep

Spacemanspark:
I'd like to say that I've very easily been tasered down while playing on Polaris, as an FBP.

Telvayne:
Right, they aren't immune to tasers.  I'm thinking of pepper spray that they're not immune to.

Neerti:
Vote was overwhelming towards 'no' for this.

PR #1117, opened 2016-03-04
Firing a weapon will now uncloak ninjas/changelings.

Neerti:


Neerti:
I swear I tested this fine, but I'm a tad busy on a different branch to fix this, I'll get to it later since ninja is so rare now.

PR #1127, opened 2016-03-06
Increases cell charge loss on mecha halloss hit

Hubblenaut:
Increases charge loss from (5 \* projectile_halloss) to (15 \* projectile_halloss).

Hitting a mecha ten times with an energy gun on stun would therefore result in a charge loss of 15\*40\*10=6000 points.
For Reference:
Normal Power Cell: 5,000
High-Capacity Power Cell: 10,000

Should make using tasers on mechs a way more viable option than it was before.

### DIFF
--- a/code/modules/reagents/Chemistry-Recipes.dm
+++ b/code/modules/reagents/Chemistry-Recipes.dm
@@ -1639,8 +1639,8 @@
 	name = "Pan-Galactic Gargle Blaster"
 	id = "gargleblaster"
 	result = "gargleblaster"
-	required_reagents = list("vodka" = 1, "gin" = 1, "whiskey" = 1, "cognac" = 1, "limejuice" = 1)
-	result_amount = 5
+	required_reagents = list("vodka" = 2, "gin" = 1, "whiskey" = 1, "cognac" = 1, "limejuice" = 1)
+	result_amount = 6
 
 /datum/chemical_reaction/brave_bull
 	name = "Brave Bull"


Mark9013100:
Face the wrath of the golden bolt.

Spookerton:
Using tasers on mechs should never BE a viable option. Close posthaste.

zaredman:
>Using tasers on mechs should never BE a viable option. Close posthaste.

Exactly this.

EmperorJon:
Now that I have been made aware of this, I very much do not like. Will note for review etc.

Hubblenaut:
This has been a thing for .. ages

Anewbe:
I actually wish Ions did something like this, instead of just nuking the mech,

Spookerton:
> This has been a thing for ages. Why has every mech to be shot to crap with lethals and ions..?
Because tazing mechs is fucking stupid.

Hubblenaut:
There should probably be another way to disable mechs without fully destroying them then. Maybe the ion rifle, as suggested above. It's definitely ridiculous that you need to go lethal and completely demolish a mech, even if it's just to get somebody out of a Ripley.

Anewbe:
Didn't the exosuit trackers/lock consoles disappear, too?

Dragor:
...So essentially you can take down a durand death mech with a few officers and tasers??

That's hilariously dumb. How do tasers even manage to drain the cell? How does this even make sense?

Hubblenaut:
Ask the person that initially coded it, I guess.

Kelenius:
On one hand, 15 shots is a lot. On the other hand, mech consumes power on its own, and tasers ignore armor.

Hubblenaut:
I'll close this for now, as it seems to go nowhere.

PR #1129, opened 2016-03-07
Cleans up and updates votes

Kelenius:
Absolute pathing, overall cleanup.
Removed two useless vars, simplified ``process()``.
Defines instead of strings.
Added "unvote" option that removes your vote.
Vote window now sized according to amount of options.
Added vote controller to debug controller list.

### DIFF
--- a/code/controllers/voting.dm
+++ b/code/controllers/voting.dm
@@ -1,9 +1,15 @@
 var/datum/controller/vote/vote = new()
 
-var/global/list/round_voters = list() //Keeps track of the individuals voting for a given round, for use in forcedrafting.
+var/global/list/round_voters = list() // Keeps track of the individuals voting for a given round, for use in forcedrafting.
 
-datum/controller/vote
-	var/initiator = null
+#define VOTE_RESTART "restart"
+#define VOTE_GAMEMODE "gamemode"
+#define VOTE_CREW_TRANSFER "crew_transfer"
+#define VOTE_ADD_ANTAGONIST "add_antagonist"
+#define VOTE_CUSTOM "custom"
+
+/datum/controller/vote
+	var/initiator = null // Key of the one who started the vote or "the server"
 	var/started_time = null
 	var/time_remaining = 0
 	var/mode = null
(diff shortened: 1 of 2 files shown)


RenScorch:
I have a protest of this, based on what I'm reading, you are hard coding config features into the code. Instead of simply having the config file handle it. Which completely rules the config file moot. 

I see no reason to add in Admin/mod flag checks when said config file is already suppose to handle it.

Kelenius:
@RenScorch point to a specific line, I don't have a clue what you're talking about.

RenScorch:
@Kelenius https://github.com/PolarisSS13/Polaris/pull/1129/files#diff-a0d72911aba915fa25b07cc5108d0700R208

code/controllers/voting.dm
```diff
@@ -11,385 +17,366 @@ datum/controller/vote
...
+				choices.Add(config.votable_modes)
+				for(var/F in choices)
+					var/datum/game_mode/M = gamemode_cache[F]
+					if(!M)
+						continue
+					gamemode_names[M.config_tag] = capitalize(M.name) //It's ugly to put this here but it works
+					additional_text.Add("<td align = 'center'>[M.required_players]</td>")
+				gamemode_names["secret"] = "Secret"
+			if(VOTE_CREW_TRANSFER)
+				if(!check_rights(R_ADMIN|R_MOD, 0)) // The gods care not for the affairs of the mortals
```
RenScorch: You should allow the config file to handle these sort of checks.

code/controllers/voting.dm
```diff
@@ -1,9 +1,15 @@
...
+var/global/list/round_voters = list() // Keeps track of the individuals voting for a given round, for use in forcedrafting.
 
-datum/controller/vote
-	var/initiator = null
+#define VOTE_RESTART "restart"
+#define VOTE_GAMEMODE "gamemode"
+#define VOTE_CREW_TRANSFER "crew_transfer"
+#define VOTE_ADD_ANTAGONIST "add_antagonist"
+#define VOTE_CUSTOM "custom"
+
```
RenScorch: What the point of the these defines, can you add commits on how they are used, and if able where they are being used, what are they replacing. That way later on down the road we know what the hell is actually going on.

Kelenius: They replace the strings. Previously, there were repeated uses of ``"restart"`` as a string. Which is prone to typos, because typing ``"rsetart"`` will not cause any compilation errors, but ``VOTE_RSETART`` will. This is a pretty common practice.

code/controllers/voting.dm
Kelenius: Why? How? It was already like that before.

code/controllers/voting.dm
```diff
@@ -11,385 +17,366 @@ datum/controller/vote
...
-					choices.Add(config.votable_modes)
-					for (var/F in choices)
-						var/datum/game_mode/M = gamemode_cache[F]
-						if(!M)
-							continue
-						gamemode_names[M.config_tag] = capitalize(M.name) //It's ugly to put this here but it works
-						additional_text.Add("<td align = 'center'>[M.required_players]</td>")
-					gamemode_names["secret"] = "Secret"
-				if("crew_transfer")
-					if(check_rights(R_ADMIN|R_MOD, 0))
```
Kelenius: Like, literally, I've changed  the check order to remove duplicated code but it's an exactly same check.

Zuhayr:
This all looks pretty much fine to me on a quick overview. 

@RenScorch defines are useful for repeated strings because they reduce the impact of typos. Additionally you mean comment not commit. A commit is a package of code changes. A comment is a comment.

code/controllers/voting.dm
RenScorch: My point was that add this in comments so that when someone breaks said file X amount of months down the road, and they are looking at the file, they can right off the bat tell what going on.

RenScorch:
@Zuhayr Why thank you Zuhayr. I know what a define is. And forgive me for writing late at night and performing a miss-spell. Second, the practice of refusing to comment of code so that later down the road we still have a blueprint of what that file is suppose to do, and what changes were made is irksome, it specially irksome when someone with a known history of pushing code that is not 100% bug-free(Granted bug-free code is pretty much a pipe dream in byond) and leaving behind no clues on what sort of changes is made, which can become extremely difficult for someone else to clean it, or repair it to the point of stability. 

Just because you know what XYZ means, and what files it connects to, does not mean everyone knows what it does. 

Anyways commenting is a basic practice that everyone should be using if you are working in a big community based coding projects.

Kelenius:
Comments are needed when you are explaining what a function does, what a variable is for, or when there's really complicated code section. This code is absolutely straightforward. There's nothing to comment on. There is no place where a comment would be needed because the code is too complicated to understand quickly. Using defines of strings is a **very** common practice in ss13.

Zuhayr:
@RenScorch spare me the sarcasm, you're acting like an authority in a PR that you self-demonstratively don't understand and I assumed it was out of ignorance. 

Current maintainers may disagree with me but none of your concerns are significant enough to warrant refusing to merge this PR. String defines are standard. [Excessive commenting indicates bad code](http://programmers.stackexchange.com/questions/1/comments-are-a-code-smell) and given that this codebases uses version control historical documentation is unnecessary. Anyone fucking with this system should be capable of doing a string search for a macro.

This is a very simple refactor. Pulling out ad hominem is a shitty thing to do and I really don't understand why you arbitrarily chose this PR, and these 'issues', to make a lot of noise about.

Kelenius:
Ping.

Kelenius:
Ping.

Kelenius:
Ping...

PR #1131, opened 2016-03-07
Fixes two bar recipes

Hubblenaut:
Changes iced tea and iced coffee recipes to make sense again and easily fill a 30 unit container.


Kelenius:
Mix ice and tea 1:2?..

Hubblenaut:
The other choice is 1:5

Kelenius:
I think that filling 1/6 of the glass with ice makes more sense than 1/3. But that's just my opinion.

Hubblenaut:
It both equally doesn't make much sense.
Yet another alternative: 1 part coffee, 1 part ice, 1 part cream

Spookerton:
:+1:

PR #1142, opened 2016-03-09
Retail Scanners, Price List Update and Interface Improvements

Hubblenaut:
- Fixed price path of Gargleblaster.
 - Added price for glass of water, booze-o-mat bottles and cigarette vendor items.
 - Purchasing a single item will not require a confirm swipe.
 - Amount of the same item type in transaction log is now tracked.
 - Current list of items in memory can be seen and edited on interface

image


Hubblenaut:
Do not merge yet.

Hubblenaut:
Bump.

PR #1150, opened 2016-03-10
Removes splash code

Kelenius:
Unused and not working properly.


Spookerton:
Has only self-references, touched only by maintenance changes in at least 4 years.
:+1: burn it all down.

PR #965, opened 2016-02-11
[NEEDS TESTING] Axes tcomms

Kelenius:
Replaces with a magic box that makes radios work when on. Works regardless of z-level.
Removes icon of the radio used because it's a hassle to maintain.
PDA spamfilter is now global for all messaging servers.

PR #1151, opened 2016-03-10
Clears up machine constructions

Kelenius:
Circuits contain paths instead of strings.
Circuit definitions moved to a single file. We don't need 100500 files with five lines in each.
Machines don't create parts inside them by default. Instead there are preset options that do; those are placed on the map. Previously, machine frame would cut the component list after creating a machine.
Telecomms are left untouched pending #965.
No functional changes to the game - pure cleanup.


Neerti:
Conflicts.

Kelenius:
Resolved.

Kelenius:
:<

Datraen:
Not really sure which did this in, but currently conflicts.

Kelenius:
Please don't make me fix map conflicts three times.

Kelenius:
Conflicts like hell, would be easier to remake.

PR #1158, opened 2016-03-11
Updates changelog

Kelenius:
I mean, SOMEONE has to


Neerti:
Someday I gotta look into switching to the autochangelog system bay and others are using nowdays.

PsiOmegaDelta:
The current auto-changelog system only generates the ymls. Someone still have to run the update changelog tool and PR the result.

PR #1160, opened 2016-03-11
Removes the need for slides in forensics

Mark9013100:
Port of: https://github.com/Baystation12/Baystation12/pull/12288

* Also changes the access requirement of x-ray gun crates to needing armory access, previously you only needed bridge access to pop open an x-ray crate.

* Removes a near-duplicate line in belts.


Spookerton:
Access rider is
*Makes the experimental energy weapon crate accessible to the warden + HoS instead of Command*
**/obj/structure/closet/crate/secure/weapon**
```
-	access = access_heads
+	access = access_armory
```

:+1:

PR #1162, opened 2016-03-11
Updates material costs

Kelenius:
The initial purpose was to fix odd costs of rglass, rods, and tiles in autolathe. It got out of hand.
DEFAULT_WALL_MATERIAL define replaced with MATERIAL_STEEL
Added MATERIAL_GLASS define.
Default sheet per unit amount is now 1000. Prices of everything adjusted. This will be changed later.
After this is merged, I'll clear up material use in research to draw numbers and names from relevant materials - right now they use their own procs and SHEET_MATERIAL_AMOUNT (which shouldn't exist in the first place).
The next thing that needs to be done now is design cleanup - they need to draw costs from the item. After than is done, there will be a final price adjustment for all items. Right now, let's just make it so all
prices make sense.
To clarify, you could make infinite material by making rods, tiles, or rglass, then feeding it to upgraded autolathe and taking them out as sheets.
Removed /obj/item/weapon/wire that served no purpose. It was in my way.


Spookerton:
So the summary of what this PR does is

**Deletes /obj/item/weapon/wire**
*which is vestigial, so this is fine*

**Renames DEFAULT_WALL_MATERIAL to MATERIAL_STEEL** 
*Makes sense, it's a badly named macro*

**Cuts the machine value of all material sheets in half**
*What, why*

**Adjusts the metal and glass costs of a non-exhaustive variety of machine-built objects while leaving all other material costs the same**
*Stahp*

...

TBQH I'd suggest just refactoring the macro and increasing the cost of the things you think are a problem *(and I haven't seen this particular one actually USED, just remarked on as technically possible - and known for years)* instead of making half-complete sweeping changes; and ideally in atomic pull requests, too.

Kelenius:
Oooooookay, fair enough. It kinda got out of control halfway. I was going to JUST fix the tiles and rods but it was hard to stop.

Kelenius:
As for "why": because 1000 is a nice, round number, and makes it easy to count how many sheets you need for something.

Spookerton:
I can totally understand that. You just changed that, /some/ of the metal and glass numbers (and not linearly where you did, even for things that didn't have any positive returns), and none of the other material numbers. Merging this would basically just be punishment for R&D, robotics, and mining. :(

PR #1166, opened 2016-03-11
Splits underwear into top underwear and bottom underwear

Yoshax:
-Splits underwear into top and bottom
-Removes gender split


Yoshax:
Should be good now, lest I've forgot anything.

Mark9013100:
Now I can cover my moobs, best PR of the year

Datraen:
:+1:

Dragor:
Yes.

Neerti:
Look a quick look at this, it appears you didn't change the underwear wardrobe or whatever it's called that you click on to change your socks/underwear/undershirt/etc.

Yoshax:
I did. See: d64bfb323516d8e3315ecd82e521200a905ff855

Neerti:
I'm blind, wooo.

An alternate to alert is to use input(), and supply a list of options, however this implementation (of the underwear wardrobe) should work just fine.

Datraen:
I'd honestly prefer another pick, rather than abusing the fact that you can close it.
It's more clear to the player when that is done.

Yoshax:
I think it works good as is, it remains basically the same as it is now, except it has top/bottom instead of male/female.

Datraen:
Apparently I misread something, never mind!
Back to :+1: from me.

PR #1169, opened 2016-03-12
Axes old wire, fixes costs in lathe

Kelenius:
#1162 but without insanity.


code/game/objects/items/stacks/rods.dm
```diff
@@ -9,7 +9,7 @@
 	throwforce = 15.0
 	throw_speed = 5
 	throw_range = 20
-	matter = list(DEFAULT_WALL_MATERIAL = 1875)
+	matter = list(DEFAULT_WALL_MATERIAL = SHEET_MATERIAL_AMOUNT / 2)
```
Zuhayr: I feel like fractional matter values may cause problems down the line in regards to lathe storage etc, might be worth doing a ceil() on this.

Kelenius: Does byond actually separate integers and doubles/floats? ``SHEET_MATERIAL_AMOUNT`` is 2000, so this should be 1000.

Zuhayr: It's a macro, so you should be planning around it potentially changing, either with comments on the macro or sanity stuff in the code. If someone changes it to an odd number for whatever reason it's going to end up a float and mess with the lathe.

BYOND assumes int until it needs to be a float I believe.

Kelenius: I'm planning to remove that macro.

comma: BYOND doesn't have int at all IIRC, all is float. That's why you can see 1.0 values in old vars, they come from decompiled code.

Zuhayr: I could have sworn that there are circumstances under which BYOND will do integer division. Maybe I am just confused.

Kelenius: It really doesn't matter. This macro is 2000. This code doesn't cause any issues currently. If someone is going to change the macro, it will be their responsibility to make sure it doesn't cause any issues.

Kelenius:
Ping.

Kelenius:
Ping...

Neerti:
64 bytes from Neerti: icmp_seq=1 ttl=2 m time=14 d

PR #1172, opened 2016-03-12
Adds Turbans and Kippot and a Hoodie (V2)

redstryker:
Adds the following to the loadout:

- Kippot (in all colors)

- Turbans (in all colors as well)

- Space Mountain Wind Hoodie

Also fixes some of the code with the Hijabs loadout. It is now cleaner. Instead of being listed as (for example) 'Black Hijab' on the loadout screen, it is listed as 'Hijab, black'. This allows it to be more orderly.

Kippa (in red):
http://puu.sh/mWkK6/541fe8a5f2.png
http://puu.sh/nCT6r/0e9f98eb20.png
http://puu.sh/nCT7W/2de785eea2.png
No front view since it cannot be seen from the 'north' position.

Turban (in red):
http://puu.sh/nCSOh/aa96aa764b.png
http://puu.sh/nCSS9/111bfeb905.png
http://puu.sh/nCSXQ/a1bf497524.png
http://puu.sh/nCT0G/7d9e33e420.png

Hoodie:
http://puu.sh/mm13w/e19ed2af1c.png (front's pretty plain excluding like, four pixels)


Hubblenaut:
Mhmm, I like these. If we ever have towels, putting them on the head slot should use this turban icon.

PR #1173, opened 2016-03-12
Traitor Borg lockdown, blowing-up fixed; can now self-hack.

Datraen:
Traitorborgs were supposed to not be lockable, but canmove was updated based on lockcharge. A separate variable was created so that borgs can appear to be locked down on the console.
Traitorborgs that were emagged were not supposed to be exploded when the button was pressed, changed to traitorborgs or emagged.

Traitorborgs can also now hack themselves.

### DIFF
--- a/code/game/machinery/computer/robot.dm
+++ b/code/game/machinery/computer/robot.dm
@@ -58,7 +58,7 @@
 			return
 
 		// Antagonistic cyborgs? Left here for downstream
-		if(target.mind.special_role || target.emagged)
+		if(target.mind && (target.mind.special_role || target.emagged))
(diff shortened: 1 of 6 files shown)


code/game/machinery/computer/robot.dm
```diff
@@ -58,7 +58,7 @@
 			return
 
 		// Antagonistic cyborgs? Left here for downstream
-		if(target.mind && target.mind.special_role && target.emagged)
+		if(target.mind.special_role || target.emagged)
```
Zuhayr: This will runtime on a mob with no mind.

PR #1174, opened 2016-03-13
Gives the mercs a nuke...

Mark9013100:
beer keg


Yoshax:
No flashing nuke lights :-1:

Mark9013100:
I'm a lazy asshole and that'd take effort

code/modules/reagents/reagent_dispenser.dm
```diff
@@ -217,6 +217,16 @@
...
 
+/obj/structure/reagent_dispensers/beerkeg/fakenuke
+	name = "nuclear beer keg"
+	desc = "A beer keg in the form of a nuclear bomb! An absolute blast at parties!"
+	icon = 'icons/obj/stationobjs.dmi'
+	icon_state = "nuclearbomb0"
+	amount_per_transfer_from_this = 10
+	New()
+		..()
+		reagents.add_reagent("beer",1000)
```
Kelenius: This will add beer twice. It's a subtype of regular keg, it inherits the procs.

Datraen:
What Kelenius said, + breaking current coding standards.

Beerkegs are older code, use relative pathing instead of absolute, which is the preferred method of pathing at the moment.
See https://github.com/PolarisSS13/Polaris/wiki/Code-Standards

Hubblenaut:
> use relative pathing instead of absolute
Don't you mean the other way around?

Datraen:
If you take that snippet, yes that is correct.
I'll write it out without the first comma:
Beerkegs are older code that use relative pathing instead of absolute pathing, which is the preferred method.

It probably reads better without the first comma, with less of a chance for miscommunication. We use absolute pathing, and not relative.

Neerti:
Pathing issue was fixed, so :+1:

PR #1183, opened 2016-03-14
Stunbatons now check to see if they're active while harmbaton.

### LINKED ISSUE #1177: Inactive Harmbatons do the same Hallos as Active ones.
It doesn't matter if your baton is on or off, powered or depowered, it will do the same hallos on harm intent.
The depowered baton to do no hallos damage.
#### What actually happened
It does 25 when powered, 25 when unpowered.
#### Steps to reproduce
Hit self with harmbaton.
#### Additional info:
 - **Server Revision**: Server revision: master - 2016-03-13
2aa4e7727c70cd03a50443d9e239613cc274447e

Datraen:
Fixes #1177

### DIFF
--- a/code/game/objects/items/weapons/stunbaton.dm
+++ b/code/game/objects/items/weapons/stunbaton.dm
@@ -122,8 +122,9 @@
 	if(user.a_intent == I_HURT)
 		if (!..())	//item/attack() does it's own messaging and logs
 			return 0	// item/attack() will return 1 if they hit, 0 if they missed.
-		agony *= 0.5	//whacking someone causes a much poorer contact than prodding them.
-		stun *= 0.5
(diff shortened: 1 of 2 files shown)


Kelenius:
...no

You literally did the oppoaite - now they will apply full pain in harm mode when off instead of half it did before.

Datraen:
I really shouldn't try to code in between classes at college.

mwerezak:
Eh, close. You just needed to do
```
if (!..())	//item/attack() does it's own messaging and logs
    return 0	// item/attack() will return 1 if they hit, 0 if they missed.
if(!status)
    return 1
agony *= 0.5	//whacking someone causes a much poorer contact than prodding them.
stun *= 0.5
```
instead.

PR #1188, opened 2016-03-15
A large amount of changeling changes.

### LINKED ISSUE #969: Changeling armblade and flavortext fuckery
Getting stunned or otherwise disarmed as a ling will instantly remove the armblade you just shelled 20 chems for, leaving lings to constantly reform it in combat.
The armblade does not seem deflect any projectile weapons. Not sure if this is an issue or if its just intended to work with lasers.

Changing identities while having snowflake text will retain the previous snowflake text, making lings a dead giveaway to the powergaymen in red suits.

Neerti:
Round-end report now displays a history of powers the changeling had bought for that round, including resets.
Renames purchasedpowers to purchased_powers because I like my underscores.
Changelings can always be absorbed by other changelings if the body is husked and wasn't already absorbed.
Reviving clears the husked status, fixing an exploit where a ling could be husked and alive, and be unabsorbable.
A stronger variant of the armblade is available if one also buys recursive enhancement (totaling 5 points).  The stronger variant has 30 armor pen, but no additional effects.
A stronger varient of the electric hand is available for those who buy recursive enhancement, which does more halloss and costs less chemicals to channel electricity, also costing 5 points.
Normal electric hand made a little stronger when using it by grabbing someone, since grabbing is harder.
Cryosting description now says it has a cooldown.
Delayed toxin sting also now says it has a cooldown.
Changes log produced when using finger-lockpick.
Endoarmor description made slightly more accurate, for those which may not have a base health of 100.
Enfeebling string now reduces a person's health by a percentage, so teshari are not instantly killed by a recursive enhanced version of it.  Enfeebling sting now has a 5 minute cooldown as well.
Replaces some sleep numbers with X SECONDS/SECOND to make it more clear how long it is.
Transforming will wipe your flavor text, instead of keeping your previous identity's flavor text.

Half-fixes #969

### DIFF
--- a/code/game/gamemodes/changeling/powers/absorb.dm
+++ b/code/game/gamemodes/changeling/powers/absorb.dm
@@ -1,10 +1,10 @@
 /datum/power/changeling/absorb_dna
 	name = "Absorb DNA"
-	desc = "Permits us to syphon the DNA from a human. They become one with us, and we become stronger."
+	desc = "Permits us to syphon the DNA from a human. They become one with us, and we become stronger if they were of our kind."
 	genomecost = 0
 	verbpath = /mob/proc/changeling_absorb_dna
 
-//Absorbs the victim's DNA making them uncloneable. Requires a strong grip on the victim.
+//Absorbs the victim's DNA. Requires a strong grip on the victim.
 //Doesn't cost anything as it's the most basic ability.
 /mob/proc/changeling_absorb_dna()
 	set category = "Changeling"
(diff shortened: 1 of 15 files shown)


Kelenius:
I don't see any issues with the code.

code/game/gamemodes/changeling/powers/absorb.dm
```diff
@@ -27,9 +27,10 @@
 		src << "<span class='warning'>We do not know how to parse this creature's DNA!</span>"
 		return
 
-	if(HUSK in T.mutations)
-		src << "<span class='warning'>This creature's DNA is ruined beyond useability!</span>"
-		return
+	if(HUSK in T.mutations) //Lings can always absorb other lings, unless someone beat them to it first.
+		if(!T.mind.changeling || T.mind.changeling && T.mind.changeling.geneticpoints < 0)
```
RemieRichards: the ```T.mind.changeling``` check isn't needed, as if it's missing the first if condition will fail.

comma: To clarify, first one isn't needed, just
```
if(T.mind.changeling && T.mind.changeling.geneticpoints < 0)
```
will do

Datraen:
Odds are loss of the unnecessary check wouldn't save much, but it should still be done.

PR #1191, opened 2016-03-15
Mob inventory cleanup

Kelenius:
Right. This does the following:

1. Moves inventory vars that were on ``/mob`` level to ``/mob/living``. This is to reduce the amount of clutter vars. If observers don't have inventory, they don't need inventory vars.
1. Moves some procs as well, leaving empty definitions on ``/mob`` level.
1. Where necessary, replaces uses of ``/mob`` with ``/mob/living`` to match the first change. This is mostly ``attackby``.
1. Adds more procs for inventory handling: ``/mob/living/proc/hands_are_full()`` returns 1 if both mob's hands are full, 0 otherwise; ``/mob/living/proc/item_is_in_hands(var/obj/item/I)`` returns 1 if ``I`` is in one of mob's hands; and ``/mob/living/proc/get_type_in_hands(var/T)`` returns the item of type ``T`` in the mob's hands, or ``null`` if one isn't found.
1. Uses said procs where possible to minimize the uses of ``l_hand`` and ``r_hand`` vars outside of mob code.

The goal here is to bring mob inventory closer to OOP it should be; right now the game assumes that playable mobs that have inventory have exactly two hands, one active and one inactive, one left and one right, and changing that would be pretty much impossible. I'm not going to try to tackle that in one PR, instead I'll make small, clear changes until all interaction with mob inventory is made via procs and variables are where they should be. Then we will have a much more flexible inventory system that would allow us to easily create mobs with non-standard inventories or hand configurations.

And eventually, cyborg hats.

### DIFF
--- a/code/ZAS/Fire.dm
+++ b/code/ZAS/Fire.dm
@@ -406,7 +406,7 @@ datum/gas_mixture/proc/check_recombustability(list/fuel_objs)
 	//Get heat transfer coefficients for clothing.
 
 	for(var/obj/item/clothing/C in src)
-		if(l_hand == C || r_hand == C)
+		if(item_is_in_hands(C))
 			continue
 
 		if( C.max_heat_protection_temperature >= last_temperature )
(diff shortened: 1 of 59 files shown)


Kelenius:
Ping.

Datraen:
:+1:

PR #1199, opened 2016-03-16
Changes alert level changing to no longer require ID swiping

### LINKED ISSUE #1187: Heads can't change alert level
Departmental Heads (ie not HoP/Captain) can't use the command console to change between green and blue alert. Confirmed with CE and HoS, may also affect AI.
Alert level would change to blue when the button was pressed on the console.
#### What actually happened
Error message came up, saying something about insufficient access; Alert stayed at green
#### Steps to reproduce
Go to the bridge command and communications console as a department head. Log in. Attempt to change the alert level.
#### Additional info:

Yoshax:
No longer need to swipe ID to change alert level, only have to be logged into the console with a head of staff or higher ID.

Puts it in line with other functions such as call shuttle.

Fixes #1187

### DIFF
--- a/code/game/machinery/computer/communications.dm
+++ b/code/game/machinery/computer/communications.dm
@@ -78,34 +78,23 @@
 			crew_announcement.announcer = ""
 
 		if("swipeidseclevel")
-			var/mob/M = usr
-			var/obj/item/weapon/card/id/I = M.get_active_hand()
-			if (istype(I, /obj/item/device/pda))
-				var/obj/item/device/pda/pda = I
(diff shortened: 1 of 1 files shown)


Yoshax:
Forgot a line.

Yoshax:
Fixed. Good now.

Ithalan:
Are AI also able to change the alert level?

Yoshax:
No, they can't. Outside of scope of this PR as that's a feature addition, this was a bugix and QOL change.

PR #1201, opened 2016-03-17
Doorcrushing now pushes people out of the door.

Neerti:
Being doorcrushed now pushes the mob being crushed to a random adjacent cardinal direction that is open.  Open is defined as the turf being nondense, and no dense objects existing on it.  This is done to prevent stunlocking by repeated doorcrushes.

### DIFF
--- a/code/game/machinery/doors/airlock.dm
+++ b/code/game/machinery/doors/airlock.dm
@@ -922,6 +922,20 @@ About the new airlock wires panel:
 	SetWeakened(5)
 	var/turf/T = get_turf(src)
 	T.add_blood(src)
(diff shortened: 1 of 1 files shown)


Kelenius:
Would it be possible to use something like ``if(Move(new_turf)``? This version won't work well with pass flags and glass panels on tile borders.

Neerti:
Wouldn't Move fail if they're unable to move due to stuns or other incapacitating conditions?

Kelenius:
That does not appear to be the case, though it does fail if the mob is buckled.

OhEmGee:
If doorcrushing is nerfed like this, the damage it does should be buffed. Maybe make it do

Neerti:
Probably won't have time for this, however Bay has an improved varient that we can port later, if anyone wishes to.

PR #1205, opened 2016-03-17
Adds cyro-sleepers to solitary, minor map fixes/edits

Mark9013100:
* Adds cyrogenic storage cells to the solitary cells on Z-1. Also gives the cells actual flooring.

* Adds a few more floor decals I forgot to add on Z-2, also adds a few tables to the Thunder-dome. And lino flooring to the tram lounge.

* Fixes a few /areas on Z-4.

* Fixes a few /areas on Z-5.

* Gives Brig Control standard security door access on Z-1.

* Gives armory lockers(excluding lethal ammo) standard security door access on Z-1.


Hubblenaut:
IMO putting cryo cells into solitary defeats the point of it being solitary

Mark9013100:
In that case, I might as well map in cyanide pills, solitary fucking sucks.

Hubblenaut:
IMO solitary should only be used to punish a prisoner that has offended regulations while already imprisoned, not for keeping somebody in there indefinitely.

They could make good use of timers next to the doors.

Kelenius:
>IMO solitary should only be used to punish a prisoner that has offended regulations while already imprisoned

Well it's not, what now?

Hubblenaut:
Putting anybody in there that is not basically a living weapon (changeling, wizard, ect) for an indefinite amount of time is poor security play and should never happen.
Still seems very much against the point of solitary to have a device in there that will just put you in cryogenic storage and make you avoid the punishment.

Datraen:
Pretty much what Hubble said, if someone tosses you into solitary when you're not already being in an issue while in the normal brig, ahelp and the staff is likely to help you out, or sort out any confusion.

Mark9013100:
> Putting anybody in there that is not basically a living weapon (changeling, wizard, ect) for an indefinite amount of time is poor security play and should never happen. If that happened to you, you should probably have ahelped.
> 

Solitary should never have happened but here we are.

nachomeep:
If you're in solitary, the round's basically over for you. If you don't cryo, your job slot's taken forever. Simple as that. We should have cryo in solitary for the purpose of opening slots.

Hubblenaut:
I put my two cents of an opinion in on this, I'm not here for an argument.

Looked over the map code, your new windows have one million health points. Why?

Mark9013100:
They do? I remember this happening with the equipment storage windows, but I think their health uses a different var. Still, it's possible these might be the old CentCom windows, I'll check tomorrow.

Mark9013100:
@Hubblenaut Are you sure windows use the health var? There's another var named maxhealth which is set to 40.

Neerti:
Windows get set to maxhealth upon New() anyways, so the huge health isn't actually there in-game.

Mark9013100:
Annnnnnnnd merge conflicts. Yay.

PR #1206, opened 2016-03-18
Adds wallets to loadout menu

Hubblenaut:
Spawns with cash ranging from 100 to 200 Thalers in a triangular distribution.
Cleans up code/modules/economy/cash.dm a tiny bit.


Kelenius:
Would it be better to just use ``spawn_money(rand(50, 100) + rand(50, 100), src)`` to spawn a single item?

PR #1207, opened 2016-03-18
Adds a check to the greytide virus to see if the APCs are operating.

### LINKED ISSUE #799: Brig Graytide Virus
Engineering receives a warning about the upcoming electric failures in the brig, however there is nothing they can do to prevent it.
The doors will magically open and the lights break even when there's no power.

Datraen:
Should fix #799

As a note: This checks whether or not the breaker's been turned on. If it's desired, it could be changed to whether or not it has a cell, and whether or not that cell has charge.

### DIFF
--- a/code/modules/events/prison_break.dm
+++ b/code/modules/events/prison_break.dm
@@ -61,9 +61,12 @@
 /datum/event/prison_break/tick()
 	if(activeFor == releaseWhen)
 		if(areas && areas.len > 0)
+			var/obj/machinery/power/apc/theAPC = null
 			for(var/area/A in areas)
-				for(var/obj/machinery/light/L in A)
-					L.flicker(10)
+				theAPC = A.get_apc()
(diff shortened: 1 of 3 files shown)


code/modules/events/prison_break.dm
```diff
@@ -61,9 +61,12 @@
 /datum/event/prison_break/tick()
 	if(activeFor == releaseWhen)
 		if(areas && areas.len > 0)
+			var/obj/machinery/power/apc/theAPC = null
 			for(var/area/A in areas)
-				for(var/obj/machinery/light/L in A)
-					L.flicker(10)
+				theAPC = A.get_apc()
+				if(theAPC.operating)	//If the apc's off, it's a little hard to overload the lights.
```
mwerezak: Wouldn't it be more appropriate to check the area's lighting channel instead?

mwerezak: Nevermind

PR #1220, opened 2016-03-21
Makes the hackersuit spaceproof

Yoshax:
The hackersuit is fairly useless as it is not spaceproof. This makes it spaceproof.

### DIFF
--- a/code/modules/clothing/spacesuits/rig/suits/light.dm
+++ b/code/modules/clothing/spacesuits/rig/suits/light.dm
@@ -38,9 +38,6 @@
 
 	req_access = list(access_syndicate)
 
-	airtight = 0
-	seal_delay = 5 //not being vaccum-proof has an upside I guess
-
 	helm_type = /obj/item/clothing/head/lightrig/hacker
 	chest_type = /obj/item/clothing/suit/lightrig/hacker
 	glove_type = /obj/item/clothing/gloves/lightrig/hacker
@@ -55,7 +52,6 @@
 		/obj/item/rig_module/vision,
 		)
 
-//The cybersuit is not space-proof. It does however, have good siemens_coefficient values
 /obj/item/clothing/head/lightrig/hacker
 	name = "HUD"
 	siemens_coefficient = 0.4


Woodratt:
I thought it was made non-spaceproof cause it doesn't look like a space worthy suit? Or am I wrong?

Mark9013100:
@Woodratt That was the originally reason I believe.

Yoshax:
It looks like less of a space suit, correct. However, the sprite is awesome, and the suit sees basically no use because it's not space-proof. Perhaps a little suspension of disbelief is required, but I believe it's a good chance.

mwerezak:
Why not find another way make the suit more appealing to players?

Yoshax:
Because it has the same usage, or is placed as such, as the other spacesuits. Nobody wants a rigsuit that isn't space-proof. It's a ninja-lite suit, and I just don't see how it can be useful without being spaceproof. I believe the only time it spawns is for heist anyway.

mwerezak:
Find new places to put it? Create some new and unique modules for it?

Yoshax:
Or just give it the functionality it should have?

mwerezak:
That's the point though. You're giving it functionality it should not have just because you can't think of a better solution.

Yoshax:
I'm giving it the functionality it should have, like it did in the first place.

mwerezak:
It only had that functionality because it was more-or-less accidentally inherited.

Yoshax:
And I believe, or think that I should have such functionality back. 

It's a suit designed for stealth, hacking and subterfuge. Only spawns during Heist, but the only way to get from the heisters ship to the station with the suit is by docking. That defeats the purpose of stealth. Docking your big ship to the station doesn't say stealth.

Kelenius:
I think you should take this discussion to forums.

mwerezak:
Not that there isn't another solution, like making the cybersuit available to other antags, but you can wear a space suit over the cybersuit you know. I've done it myself, just retract the head and chest pieces, put on your space suit, get aboard the station, then chuck your spacesuit in a locker somewhere and redeploy the cybersuit.

Mark9013100:
What's even the point of the Cybersuit? Why do the Heisters need it?

Yoshax:
It gives them a way to perform good stealth, subterfuge and other suchlike things.

Hubblenaut:
I got to agree that a rig spawned for heisters makes very little sense to not be spaceproof from a gameplay sense of view. With good enough modules, players will simply make it a habit of screwing them out of the hacker rig and putting them in another.

Though I can also understand why the sprite heavily suggests it shouldn't. The helmet is basically just an HUD tiara thing and doesn't seem to actually protect the head from anything at all.

The only viable solution I see here is to have this rig both return being spaceproof and with a sprite that properly suggests it.

Yoshax:
People already basically just rip everything out of the suit and put it into an industrial one. 

I like the sprite, it's different.

mwerezak:
I'd still suggest re-purposing it for different antags, but whatever.

mwerezak:
What if... the cybersuit allowed you to run faster? Make the suit pieces give a negative slowdown when powered.

PR #1221, opened 2016-03-22
Fixes Anomaly Analyzer reading the wrong variable for the artifact effect's range

Ithalan:
Anomaly analyzer erroneously read the wrong variable before for the effect's range (touch, aura or pulse).


Datraen:
:+1:

Issue #1219, opened 2016-03-21
Attack logs - Droppers always report as EMPTY


Mark9013100:
#### What you expected to happen
For the logs to say what chemical was injected from the dropper.
#### What actually happened
The chemical was reported as 'EMPTY.'
#### Steps to reproduce
Get a dropper.
Fill it with a chemical.
Use it on someone.
#### Additional info:
 - **Server Revision**: Found using the "Show Server Revision" verb under the OOC tab.
Server revision: master - 2016-03-21
996ce92f7ff8bceab4122366ba497b03fd3f3dac 
 - **Anything else you may wish to add** (Location if it's a mapping issue, etc)
Tested with water, and with whatever that eye medicine is.

Telvayne:
This looks like I could fix it pretty easily.  It seems that since it only holds 5u, and deploys that 5u on use, the dropper IS empty when the log is written.  If I just move the logging to occur prior to emptying the dropper, it should fix the problem.

Mark9013100:
Danke, closing.

PR #1223, opened 2016-03-22
Properly log dropper-related assault for admins

Telvayne:
Fixes bug #1219

Moving the logging from post-dropping (when the dropper is empty) to Pre-drop, post-dropable check (it should now be located at the point when a drop is going to occur, but just prior to actually doing so).

As always, please review this change to make sure I didn't break anything.  All I did was cut-paste the lines to a few lines above, but hey, never hurts to double-check!


Datraen:
:+1:

PR #1225, opened 2016-03-22
Changeling Blood Refilling

Telvayne:
The idea here is to generate blood along with the cellular matter we're already pulling out of nowhere.  This should fix the Teshari -> Human transformation that's currently leaving us with anemic changelings.

Do not merge until someone gives it a look, though.  Specifically, i'm not sure if my syntax is correct on calling fixblood()  (located in blood.dm, it says it's used to regenerate a living entity to 100% blood;  it's also called in an artifact code to do so, so I'm assuming it's coded right).

### DIFF
--- a/code/game/gamemodes/changeling/powers/transform.dm
+++ b/code/game/gamemodes/changeling/powers/transform.dm
@@ -39,6 +39,7 @@
 		H.b_type = "AB+" //For some reason we have two blood types on the mob.
 		for(var/flavor in H.flavor_texts) //Nulls out flavor text, so we don't keep our previous mob's flavor.
 			flavor = null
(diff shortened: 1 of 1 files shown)


Yoshax:
If my understanding is correct, fixblood has to be called on a human mob. And you can't call it on src.

If you do:
var/mob/living/carbon/human/H = src
H.fixblood()

It'll compile, but I dunno if it will work.

Yoshax:
Baystation12/Baystation#12473 may fix it, I'm not sure. Haven't had a proper look, just saw it a moment ago.

Telvayne:
Yeah, this looks like it's already fixed, or on the way to being fixed.  I'll close it, then.

PR #1230, opened 2016-03-23
Renames the security .45

Mark9013100:
I've been meaning to do this for weeks.


Datraen:
I could've sworn this was already done.

Mark9013100:
The old NT Mk.58 is named this. So that might be what you're thinking.

Zuhayr:
. at the start of names breaks VV.

PR #1231, opened 2016-03-23
Adds a new optional 'ambitions' system.

Neerti:
Antagonists can set custom objective-like goals for themselves, which everyone can see at the end of the round.  This otherwise has no impact on the round.
Pictures:
Picture of being turned into an antag
Picture of setting an ambition
Picture of round-end report

### DIFF
--- a/code/datums/mind.dm
+++ b/code/datums/mind.dm
@@ -61,9 +61,12 @@
 
 	//put this here for easier tracking ingame
 	var/datum/money_account/initial_account
-	
+
 	//used for antag tcrystal trading, more info in code\game\objects\items\telecrystals.dm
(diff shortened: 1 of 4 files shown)


Anewbe:
Seize the means of production! (+1)

Telvayne:
> Be ideal

oh you.

I really like this change.  I'd almost suggest it be /strongly encourage/ to fill out, and logged.  That way admins can give people a talkin' to when they were aspiring to 'murder everyone and eat them' or 'waaaahooo i'm crazy so i overloaded the SM'

Hubblenaut:
Do like this. Can admins see this in traitor panel?

Spacemanspark:
:+1:

PR #1233, opened 2016-03-23
Armblade Additions plus One Other Thing

Neerti:
Armblades can now force depowered or broken airlocks, firelocks, and blast doors, just like a crowbar.
Armblades can slash APCs to cut all the wires.  Fixing the damage is as simple as mending all the wires.
Cleans up the fireaxe wielding check in airlock.dm
Removes colon cancer in blast_door.dm
Allows supplying an argument to not shock the user when cutting wires directly in wires.dm
Not armblade related: Self Respiration is now a free passive ability instead of costing one genome.


Neerti:
Don't merge because the wire datum addition can be done better.

Neerti:
Can't be bothered to make the wire stuff better, so decided to remove the APC interaction.

PR #1235, opened 2016-03-23
Watercooler can be wrenched/unwrenched now

Yoshax:
Fixed capitalization in some cases

Ports Baystation12/Baystation12#12340


Telvayne:
I was waiting for this change.  Making those things immobile was annoying.

PR #1236, opened 2016-03-23
Light switches now glow

Yoshax:
- Adds a relatively weak light to light switches (slightly weaker than a light bulb), so it is not that hard to find them when the room is dark. Colour can be red or green depending on whether the light switch is turned on or off.
- Increases light switch power usage to 10 Watts. For comparsion, a light bulb has 30 so it feels apropriate.
- Light switches now run off LIGHT APC channel rather than EQUIP.
- Updated: Light switches now layer above darkness when powered. Light intensity slightly reduced.

Ports Baystation12/Baystation12#12209

### DIFF
--- a/code/game/machinery/lightswitch.dm
+++ b/code/game/machinery/lightswitch.dm
@@ -7,10 +7,12 @@
 	icon = 'icons/obj/power.dmi'
 	icon_state = "light1"
 	anchored = 1.0
+	use_power = 1
+	idle_power_usage = 20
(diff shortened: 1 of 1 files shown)


Datraen:
Lights currently use 2 power, due to the fact that there is a ghost load of 200W on each room.

Yoshax:
Will adjust appropriately.

Yoshax:
Adjusted. Same ratio. 1/3 cost of lightbulb.

PR #1243, opened 2016-03-24
Mob inventory cleanup, part 2

Kelenius:
Replaces three item removal procs with one
Adds a proc to delete an item on the mob


Hubblenaut:
Since you handle the removal of items as removal of equipped items, this makes it unable to drop items from your inventory that have canremove=0 set.

Kelenius:
Setting force to 1 will make it ignore canremove.

Kelenius:
Will resolve after approval.

Kelenius:
Why does a simple cleanup take two weeks to merge :V

Hubblenaut:
This will still handle hand- and pocket slots as equipment slots and make you unable to remove items from there that have 'canremove = 0' set.

Datraen:
The reason why this takes so long is because of the volume of files that have to be checked.
I've been personally too busy to take the time to review, I'll try to review this by Sunday.

Datraen:
Can't review, life got in the way, will be a week or two until I'm free enough to do this.

Datraen:
No glaring errors, :+1:

Kelenius:
Done.

PR #1252, opened 2016-03-26
Adds radiation sirens

Hubblenaut:
Adds red sirens that get activated upon being hit by radiation.
Mapped in key locations around the engine rooms.


Hubblenaut:
Do not merge yet.

Hubblenaut:
I currently open the polaris.dme file in notepad to get rid of the unnecessary DEFINES. If anyone knows a better workaround, please tell me.

code/game/machinery/siren.dm
```diff
@@ -0,0 +1,41 @@
+//////////////SIREN//////////////
+/obj/machinery/siren
+	name = "siren"
+	desc = "A small wall-mounted siren"
+	icon = 'icons/obj/stationobjs.dmi'
+	icon_state = "siren0"
+	light_color = "#a91515"
+	light_range = 2
+	light_power = 0
+	layer = 5
```
Kelenius: Better use a define here, I think.

Kelenius:
>I currently open the polaris.dme file in notepad to get rid of the unnecessary DEFINES. If anyone knows a better workaround, please tell me.

In DM, Build > Preferences for polaris > Automatically set FILE_DIR...

Hubblenaut:
Updated.

Kelenius:
Minor nitpick is that they won't react to radiation storms while they logically should. I'm not sure if there's a non-snowflakey method of doing that though.

mwerezak:
I dunno, a `/mob/living/rad_detector` that doesn't process, placed by the rad alarm on its tile and removed if the rad alarm is destroyed?

Kelenius:
>I dunno, a ``/mob/living/rad_detector`` that doesn't process, placed by the rad alarm on its tile and removed if the rad alarm is destroyed?

Eeeeeeeeh.

Maybe a global list of all objects that should react to radiation, that would include humans and sirens, and ``/atom/proc/radiation_act()``?

mwerezak:
That would require you to modify everywhere radiation is given to mobs, if that's what you want.

Datraen:
Cannot merge again, http://hastebin.com/awoxajisir.coffee to https://github.com/PolarisSS13/Polaris/blob/8f54e7778fd377296a9c895295d7977e6ce5a7eb/code/modules/events/radiation_storm.dm#L32 would work for doing radiation storms.

Hubblenaut:
I never iterated through world before, would have assumed it's quite resource-intensive. Will implement that, though.

Techhead0:
IIRC, there is a global `machines` list of all `obj/machinery` in order to speed up things like this.

Datraen:
machines should work.
I'm not sure what global lists exist, to be honest, and that was a code-wise simple solution.

Zuhayr:
Iterating over ```world``` is an absolutely terrible idea especially for something that might happen repeatedly in a round.

mwerezak:
Not to mention the fact that your implementation relies on putting
```
//set all sirens in range
for(var/obj/machinery/siren/S in range(src, round(sqrt(power / 2))))
		S.setAlarm()
```

Everywhere you irradiate nearby mobs, and won't work if someone ever forgets to do that.

Hubblenaut:
A global proc will be the more elegant solution and should only take up five lines. Will do when I have time.

Datraen:
Probably helps when I actually look for where the lists are. Thanks for the link, I quickly whipped something up when I had a break between classes.

I should stop doing that.

Kelenius:
I might be wrong, but IIRC, the fastest way is
```
for(var/A in myList)
	var/my/type/here/B = A
	B.doThing()
```
Assuming ``myList`` only contains my/type/here instances, and looping through world is slower, but still faster than looping through a list that contains other types and typechecking. So using world may actually be better than looping through machines here.

PsiOmegaDelta:
As I understand it

```
for(var/thing in a_list)
    var/some/type/ST = thing
```
The above is the fastest option when you know every entry of the list is of the desired type.

```
for(var/obj/O in world)
```
The above the fastest option when you specifically want base types (/area, /obj, /mob, etc.) because of some behind the scenes BYOND magic. That is, you don't gain anything by creating specific lists for them.

And finally
```
for(var/some/type/ST in a_list)
```
is the main option for when the list contains mixed types.

Neerti:
Anyways, this has lingered for a little over a month and doesn't seem to be going anywhere, so I'm gonna close it.  Such a shame to lose this to the abyss of time.

PR #1269, opened 2016-03-27
Balances Cult Hoods

Anewbe:
- Gives Cult Hoods the same armor values as the robes


Datraen:
Failed check is not as a result of Anewbe, it's a result of another PR not being run with the same tests.

PR #1287, opened 2016-03-29
Ports unit tests

Kelenius:
Port of https://github.com/Baystation12/Baystation12/pull/11808 and several others
Fixes three overlapping cables ([192 / 158 / 1], [214 / 161 / 1], [122 / 197 / 1])
Fixes drill and mech drill having the same ID.


Kelenius:
Yep, ready for the review.

code/unit_tests/map_tests.dm
```diff
@@ -0,0 +1,99 @@
...
+						/area/security/riot_control // This should probably be fixed,
+						)
+
+	var/list/exempt_from_apc = typesof(/area/construction,
+						/area/medical/genetics,
+						/area/mine,
+						/area/vacant/vacant_shop
+						)
+
+	for(var/area/A in world)
```
Datraen: I'd like to see this split up into different variables, to specifically state if the area is lacking an APC, scrubber or vent.

code/unit_tests/map_tests.dm
```diff
@@ -0,0 +1,99 @@
...
+
+			if(!A.air_vent_info.len && !(A.type in exempt_from_atmos))
+				log_unit_test("[bad_msg] lacks an Air vent.")
+				area_good = 0
+		
+			if(!area_good)
+				bad_areas.Add(A)
+
+	if(bad_areas.len)
+		fail("\[[bad_areas.len]/[area_test_count]\]Some areas lacked APCs, Air Scrubbers, or Air vents.")
```
Datraen: Stating the names of the areas wouldn't be bad, either.

Ccomp5950: that's handled with [bad_msg]

looks like this on Bay:  https://travis-ci.org/Baystation12/Baystation12/builds/118736597
(scroll down to the bottom and then up to the red text)

code/unit_tests/map_tests.dm
Ccomp5950: When I was writing this check on Bay I had the scrubbers and vents separated, the lists were identical so recombined it.  I've not looked over your map so unsure if that would be the case here.  If you are talking about the notification if one is lacking it on failure that already exists.  If you are talking about the enumerated exemptions I think it's redundant.

code/unit_tests/map_tests.dm
Datraen: I see. We could probably change bad_areas from a list to a number counter, then, considering it's utility.

Ccomp5950: Yeah i can see that, I think I made it a list so it could loop through the list at the end and report, I wound up instead reporting failures while in the main loop of the test.

Not sure why I left it like that, I'm pretty sure I had something else intended but I don't have my notes with me at work.

EmperorJon:
@Datraen So you're happy with it?

Datraen:
@EmperorJon @Kelenius I really would like color. It'd add to readability of the unit tests.

Kelenius:
[Okay](https://travis-ci.org/PolarisSS13/Polaris/builds/120566904)

PR #1288, opened 2016-03-29
Minor change to riot helm code

Mark9013100:
Original code was borrowed from cap flipping, now uses the same code in use by the Ushanka.


Datraen:
:+1:

PR #1292, opened 2016-03-29
Fixes Deadchat

Yoshax:


EmperorJon:
Unacceptable. Saltchat must remain muted forever. This is our best chance.

Yoshax:
It's clearly a feature :)

PsiOmegaDelta:
Technically the previous issue made deadchat work like a walke-talkie.
You could hear dsay as long as you kept it enabled, while you could only speak if you disabled it.

PR #1293, opened 2016-03-29
Adds progress bars to tasks that take time

Yoshax:
do_after()/do_mob() now comes with an optional progress bar, which is only seen by the client conducting the task.

Ports Baystation12/Baystation12#12291 &  Baystation12/Baystation12#12316


code/_helpers/mobs.dm
```diff
@@ -30,13 +30,10 @@ proc/random_hair_style(gender, species = "Human")
...
 	for(var/hairstyle in hair_styles_list)
 		var/datum/sprite_accessory/S = hair_styles_list[hairstyle]
-
-		if(gender != NEUTER && gender != PLURAL)
-			if(gender == MALE && S.gender == FEMALE)
-				continue
-			if(gender == FEMALE && S.gender == MALE)
-				continue
-
+		if(gender == MALE && S.gender == FEMALE)
```
EmperorJon: Not entirely sure what these changes are about

PR #1304, opened 2016-03-30
Command age limits

EmperorJon:
Adds the agreed-upon 25 year age limit for Head of Staff roles utilizng Zuhayr's refactored job age system that never actually got used!


Kelenius:
Config option, no?

EmperorJon:
That is not how the system was originally configured, to my knowledge.

PR #1310, opened 2016-03-31
Spikethrower Tweak

Anewbe:
Reduces spike generation rate by 50 percent.


Kelenius:
Seven commits to change one symbol, I think that's a record.

Zuhayr:
I don't understand how a commit history like this even happens if you have a nodding acquaintance with git. How are you authoring these commits?

Anewbe:
Woodrat wanted the regen rate lowered, rather than outright removing the weapon.

PR #1315, opened 2016-03-31
Updates Preference Permissions, Re-adds Verbs

Datraen:
Also ports Baystation12/Baystation12#12526

### DIFF
--- a/code/_helpers/logging.dm
+++ b/code/_helpers/logging.dm
@@ -31,7 +31,7 @@
 		diary << "\[[time_stamp()]]DEBUG: [text][log_end]"
 
 	for(var/client/C in admins)
-		if(C.is_preference_enabled(/datum/client_preference/admin/show_debug_logs))
+		if(C.is_preference_enabled(/datum/client_preference/debug/show_debug_logs))
 			C << "DEBUG: [text]"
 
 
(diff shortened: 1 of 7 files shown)


Yoshax:
Why re-add all the verbs?

Neerti:
Because removing them made a lot of people very unhappy.

Yoshax:
Yes, I'm incredibly dumb and didn't realise it was the admin verbs before commenting. Yay for assumptions.

Datraen:
I did re-add all of the verbs, because I got more than a couple "Where did the verbs go?"

Yoshax:
Change isn't in itself bad, people just need time to get used to change. I only really saw 1 person actually complain about the lack of verbs.

Datraen:
I started to feel that it was more of an inconvenience than an improvement when I was doing things on the server, as well.

Kelenius:
Verbs are bad :<

Datraen:
Verbs are nice. I'm not removing the alternatives, but being able to select a verb rather than constantly switching to global before joining or having to navigate in game is less convenient.

Nerezza:
My 2 cents is that it's strange to open a window to toggle OOC when it's drowning out everything else. For a user, what's been done is an inconvenience and not a convenience.

However, I do wanna note someone complained about the toggle prefix on a bunch of these verbs hijacking auto complete for toggle lock and toggle open or something.

Datraen:
For clarification: Bugfix tag is because the mods/developers didn't have access to attack/debug logs respectively, feature tag is the readdition of the verbs.

PR #1331, opened 2016-04-02
Lowers requirements for the new game modes.

Datraen:
Also adds them to the secret example list.


Yoshax:
That's the example config. The config used on server isn't stored in the repo.

Datraen:
I know. I said the secret example list.

Yoshax:
Ah. I misread.

PR #1347, opened 2016-04-04
Removes Head Gibbing

Neerti:
Adds variable to disallow gibbing of limbs, checked after the variable to disallow dismemberment.

The head can no longer be gibbed.  Severing the head via sharp weapons remains an option.

### DIFF
--- a/code/modules/organs/organ_external.dm
+++ b/code/modules/organs/organ_external.dm
@@ -56,6 +56,7 @@
 	var/disfigured = 0                 // Scarred/burned beyond recognition.
 	var/cannot_amputate                // Impossible to amputate.
 	var/cannot_break                   // Impossible to fracture.
+	var/cannot_gib                     // Impossible to gib, distinct from amputation.
 	var/joint = "joint"                // Descriptive string used in dislocation.
 	var/amputation_point               // Descriptive string used in amputation.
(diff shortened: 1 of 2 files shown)


Mark9013100:
What about destroying the head via lasers?

Neerti:
That is also covered by this.

Kelenius:
>hugbox

Hue.

Mark9013100:
hugbox medkits when

code/modules/organs/organ_external.dm
```diff
@@ -727,12 +728,16 @@ Note that amputating the affected organ does in fact remove the infection from t
 					"<span class='moderate'><b>Your [src.name] goes flying off!</b></span>",\
 					"<span class='danger'>You hear a terrible sound of [gore_sound].</span>")
 		if(DROPLIMB_BURN)
+			if(cannot_gib)
```
EmperorJon: Would it be better to do "if a and !b" rather than "if a, if b return" or is there something I'm missing here?

Zuhayr: You can't do composite conditionals in a switch.

Datraen:
:+1:

code/modules/organs/organ_external.dm
EmperorJon: Nice to know! Thanks.

PR #1362, opened 2016-04-05
Solars start wired

Kelenius:


Ithalan:
Did you also make the solar consoles and SMES units start out correctly set up like the one on the telecomm satellite is? If solar power is flowing from the start, it'd do a great deal to prevent the total blackout that happens if the round starts without an engineer

Kelenius:
They don't, because the station's powergrid is more complicated than the telecomms. You'd need to play with the SMES settings for a better result.

redstryker:
Meh, I don't like this, personally.  It gives the engineers something to do besides set up the engine, but that's just me. *shrug*

Nalarac:
As an engineer, I can say almost no one ever wants to set up the solar arrays.

Hubblenaut:
I'd rather having the SMES set along with this, so power is also sustained a while longer when no engineer is in the round.

Some charge on the SMES for buffer, then Input = max. possible input with our solar array setup (I believe 120 kW?), output roughly half of it. We can see where this goes.

Kelenius:
Outside of the scope.

Nerezza:
As an engineer I can't remember the last time solars were wired 'just because'. Nobody does it like that anymore, you have to have either a captain tell you or some power emergency for it to happen.

Furthermore, it's just strange that someone goes along and cuts all the solar wires between shifts. I can get that the engine would be offline while the operations deck is unmanned, but I don't get why someone would go out on the arrays, cut all the wire, and leave it laying by the arrays.

Telvayne:
I agree.  As Xica/Penance, I have never once wired solars for fun, because that stopped being fun a few weeks after I first played SS13.  It's dull, the airlocks are slow, there are no carp to make it !!FUN!!.  It's tedious clicking, with occasional frustration that you slightly misclicked by a pixel and have to cut the wire an redo it.

I absolutely, 100% support this PR.

Datraen:
:+1:

PR #1365, opened 2016-04-06
Adds Departmental Ponchos V2

redstryker:
Adds the following:
Medical Poncho
http://puu.sh/o6t7w/f6fa4486f9.png
http://puu.sh/o6t6P/70d09291e2.png
Has bio armor value of 10
Engineering Poncho
http://puu.sh/o6t31/d9e92a42c1.png
http://puu.sh/o6t1i/d685b637cb.png
Has rad armor value of 10
Science Poncho
http://puu.sh/o6t4Q/fafed2af43.png
http://puu.sh/o6t5L/1c0ccaa1d1.png
Has bomb armor of 10
Cargo Poncho
http://puu.sh/o6sYo/db6175d41c.png
http://puu.sh/o6sZH/00d5864fc4.png
Adds a melee armor value of 10 to the Security Poncho.
Adds all ponchos to the loadout. Makes them only available for their respective departments.


code/modules/client/preference_setup/loadout/loadout_suit.dm
```diff
@@ -125,6 +125,25 @@
...
+
+/datum/gear/suit/poncho/roles/engineering
+	display_name = "poncho, engineering"
+	path = /obj/item/clothing/suit/poncho/engineering
+	allowed_roles = list("Chief Engineer","Atmospheric Technician", "Station Engineer")
+
+/datum/gear/suit/poncho/roles/science
+	display_name = "poncho, science"
+	path = /obj/item/clothing/suit/poncho/science
+	allowed_roles = list("Research Director","Scientist", "Roboticist", "Xenobotanist")
```
PsiOmegaDelta: These ponchos are already automatically added as an option to the ```/datum/gear/suit/poncho``` above, due to being subtypes of /obj/item/clothing/suit/poncho.

redstryker:
Note:  I had forgotten to put the code in for the 'cargo' poncho in the loadout.  That's fixed now.
Side Note: I'm a dumbass.

code/modules/client/preference_setup/loadout/loadout_suit.dm
```diff
@@ -125,7 +125,31 @@
...
+
+/datum/gear/suit/poncho/roles/science
+	display_name = "poncho, science"
+	path = /obj/item/clothing/suit/poncho/roles/science
+	allowed_roles = list("Research Director","Scientist", "Roboticist", "Xenobotanist")
+
+/datum/gear/suit/poncho/roles/cargo
+	display_name = "poncho, cargo"
+	path = /obj/item/clothing/suit/poncho/roles/cargo
+	allowed_roles = list("Quartermaster","Cargo Technician")
```
PsiOmegaDelta: These gear datums are still not technically needed.
```/datum/gear/suit/poncho``` just above your additions will let players select from all subtypes of ```/obj/item/clothing/suit/poncho```, including ```/obj/item/clothing/suit/poncho/role```.

Comment (redstryker) [inline code/modules/client/preference_setup/loadout/loadout_suit.dm]: The issue is, I still want there to be allowed_roles.

Comment (PsiOmegaDelta) [inline code/modules/client/preference_setup/loadout/loadout_suit.dm]: Then this should do the trick. Change line 123 from
```for(var/poncho_style in typesof(/obj/item/clothing/suit/poncho))```
to 
```for(var/poncho_style in (typesof(/obj/item/clothing/suit/poncho)) - typesof(/obj/item/clothing/suit/poncho/role)))```

Comment (redstryker) [inline code/modules/client/preference_setup/loadout/loadout_suit.dm]: Thanks.

File: code/modules/client/preference_setup/loadout/loadout_suit.dm [hunk 3]
```diff
@@ -120,12 +120,36 @@
 /datum/gear/suit/poncho/New()
 	..()
 	var/list/ponchos = list()
-	for(var/poncho_style in typesof(/obj/item/clothing/suit/poncho))
+	for(var/poncho_style in (typesof(/obj/item/clothing/suit/poncho)) - typesof(/obj/item/clothing/suit/poncho/roles)))
```
mwerezak: Unbalanced parentheses.

PsiOmegaDelta: The ) before the -. My fault.

sabiram:
Why would NT develop armored poncho

Datraen:
I'm questioning the armor, too.

mwerezak:
To be quite honest a melee armour value of 10 is what I would expect from padded cloth... it's next to nothing.

mwerezak:
Keep in mind that the protective value of going from 0 to 10, is the same as the protective value of going from 40 to 42. Percentage based armour is not linear in its effect, and large changes at the low end of the scale really don't make much difference.

Kelenius:
10 is normal melee protection of security jumpsuits.

sabiram:
@mwerezak If that were the case then surely the others would be similarly armored. I don't think it's out of the question to ask that people _maybe_ wear a proper lab coat instead of relying on their poncho to protect themselves from infection. 

Style is one thing but automatically assigning the properties of real, time tested and true protective clothing to a cloak is a mite silly to me.

Yoshax:
Uhhh... That's not right.

If someone has 10 melee armor, and something does 10 damage. That reduces the damage by 10%, from 10 to 9.

If someone has 40 melee armor, and something does 10 damage. That reduces the damage by 40%, from 10 to 6. Therefore, if someone has 42 melee armor, and something does 10 damage, the damage is reduced by 42% from 10 to to 5.8. 

1 != 0.2

mwerezak:
@Yoshax have you thought about how the jump between 0 and 50 armor is different from the jump between 50 and 100 armor?

At 50 armor you can take twice as much damage than before. Someone has to deal 20 damage to you in order for you to take 10.

 At 100 armor you're invincible. Even if someone deals 90001 damage to you, still won't have taken 10.

"Damage dealt to the player" is a fine statistic for comparing weapons. For armor it's very misleading.

Let's go back to the specific armor numbers I choose though.

If someone has 10 armor, you need to give them 11 damage in order for them to take 10. Overall, they can take 10% more damage before kicking the bucket.

If someone has 40 armor, you need to give them 17 damage in order for them to take 10. Because of that armor, then can endure 66% more damage before they croak.

If someone has 42 armor, then they can endure 72% more damage. Okay, not quite a 10% increase, but blame me for eyeballing it. 43.5 is what I should have said.

Now all I've done here is use a different statistic to compare the armor values. So the question comes down to which one is more useful for comparing armor?

When a player compares two pieces of armor, which is more important to them: how much damage it prevents? Or how much longer they will live due to wearing it? (against a particular type of weapon).

If this doesn't make sense, don't worry about. I've given up on expecting anyone to understand this, it seems to require a shift in thinking.

Yoshax:
I suppose that makes sense, just, as you say, depends how you want to look at it. As a proper munchkin, I can see where you're coming from.

redstryker:
In my opinion, the low amount of armor that I gave them is almost negligible.  Why are people getting so up in arms about it?  The values are literally the same as the departmental jumpsuits for each of these respective departments. Plus, they are limited within departments.

I do not plan on changing it unless a dev tells me that I need to.  Even then, I don't want to remove them completely.  Otherwise, there is nothing besides the coloring distinguishing these ponchos from other departments, and /everyone/ will be wearing them, which is not the intention at all.  People outside of Sec already wear the Security poncho, and it irks me.

sabiram:
The concept of a colony station distributing woolen blankets with holes in them in departmental colors isn't so silly, even if the station is fully temperature controlled. The concept of the same woolen blankets with holes in them being made with lead plating or kevlar is silly. It's no less silly than the concept of them hiring a clown or a mime. It's silly. Sorry.

FenodyreeAv:
All jumpsuits have armour values. Ten is effectively nothing. It doesn't matter and everyone's overreacting.
The security poncho is not made of kevlar, it's just made of the kind of stuff military camo is made of, tough cloth.
Engineering poncho isn't lead lined, it's made of a space fabric that's slightly radiation retardant.

redstryker:
I may re-add them, I just need to speak to a dev about it.

PR #1366, opened 2016-04-06
Lunchboxes should now fit in backpacks

Anewbe:
Overrides the w_class from toolbox; makes lunchboxes small enough to fit in bags.


mwerezak:
`max_w_class` should be made 2, as well. I wonder if lunchboxes really should be a subtype of toolboxes after all.

Neerti:
I question the wisdom of making them inherit from toolboxes, but that can be fixed later if need be.

PR #1367, opened 2016-04-06
Ambition system tweaks

Neerti:
Ports Bay's port of this.  Did it manually because of ugly merge conflicts.
What Psi did:
Admins can now see and set ambitions in antag panel.
Makes is_special_character() a bit better.
What I did:
Setting ambitions, with the verb or with admin powers, is now logged.
Round-end text changed to be blue instead of bolded, for better readability.
Makes the ambition system known to the antag at the bottom of the chat log instead of before the 'you are a X' prompt.

### DIFF
--- a/code/datums/mind.dm
+++ b/code/datums/mind.dm
@@ -109,6 +109,8 @@
 			output += "<B>Objective #[obj_count]</B>: [objective.explanation_text]"
 			obj_count++
 
+	if(ambitions)
+		output += "<HR><B>Ambitions:</B> [ambitions]<br>"
 	recipient << browse(output,"window=memory")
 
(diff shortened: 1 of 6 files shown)


code/datums/mind.dm
```diff
@@ -109,6 +109,8 @@
 			output += "<B>Objective #[obj_count]</B>: [objective.explanation_text]"
 			obj_count++
 
+		if(ambitions)
+			output += "<HR><B>Ambitions:</B> [ambitions]<br>"
```
PsiOmegaDelta: On Bay this one is one (un)indentation further to the left.
As is, you'll now only see your ambitions if you also have objectives.

PR #1368, opened 2016-04-06
Ports Bay's new inventory system.

Neerti:
Manually ports the following,
https://github.com/Baystation12/Baystation12/pull/10095
https://github.com/Baystation12/Baystation12/pull/12000
https://github.com/Baystation12/Baystation12/pull/12199
Stuff I added,
Spacesuits, RIGs, PTR-7s and LWAPs are now w_class 5 and cannot go in your bag, ever.
Mandates boxes being capped to w_class 2 again.
Dufflebags now have approximately 28% more capacity than the normal bag, verses the previous ~58% increase.


code/game/objects/items/weapons/storage/boxes.dm
```diff
@@ -612,11 +612,10 @@
 	desc = "Eight wrappers of fun! Ages 8 and up. Not suitable for children."
 	icon = 'icons/obj/toy.dmi'
 	icon_state = "spbox"
-	storage_slots = 8
 	can_hold = list(/obj/item/toy/snappop)
 	New()
 		..()
-		for(var/i=1; i <= storage_slots; i++)
+		for(var/i=1; i <= 8; i++)
```
Kelenius: ``for(var/i = 1 to 8)`` is clearer and faster

code/game/objects/items/weapons/storage/fancy.dm
```diff
@@ -144,27 +142,45 @@
...
 
 /obj/item/weapon/storage/fancy/cigarettes/attack(mob/living/carbon/M as mob, mob/living/carbon/user as mob)
 	if(!istype(M, /mob))
 		return
 
-	if(M == user && user.zone_sel.selecting == O_MOUTH && contents.len > 0 && !user.wear_mask)
-		var/obj/item/clothing/mask/smokable/cigarette/W = new /obj/item/clothing/mask/smokable/cigarette(user)
-		reagents.trans_to_obj(W, (reagents.total_volume/contents.len))
-		user.equip_to_slot_if_possible(W, slot_wear_mask)
+	if(M == user && user.zone_sel.selecting == "mouth")
```
Kelenius: You've replaced a define with a string here.

code/game/objects/items/weapons/storage/fancy.dm
```diff
@@ -144,27 +142,45 @@
...
+		// Find ourselves a cig. Note that we could be full of lighters.
+		var/obj/item/clothing/mask/smokable/cigarette/cig = null
+		for(var/obj/item/clothing/mask/smokable/cigarette/C in contents)
```
Kelenius: This code could be replaced with ``var/obj/item/clothing/mask/smokable/cigarette/cig = locate() in src``

Yoshax:
Please no. It looks disgustingly ugly, and just isn't much fun.

code/game/objects/items/weapons/storage/boxes.dm
```diff
@@ -625,14 +624,13 @@
...
 	item_state = "zippo"
-	storage_slots = 10
 	w_class = 1
 	slot_flags = SLOT_BELT
 	can_hold = list(/obj/item/weapon/flame/match)
 
 	New()
 		..()
-		for(var/i=1; i <= storage_slots; i++)
+		for(var/i=1; i <= 10; i++)
```
Kelenius: and here

code/game/objects/items/weapons/storage/boxes.dm
```diff
@@ -650,7 +648,7 @@
 	icon_state = "syringe"
 	New()
 		..()
-		for (var/i; i < storage_slots; i++)
+		for (var/i; i <= 7; i++)
```
Kelenius: and here, I didn't want to be annoying and point at every one

Neerti:
How come it looks ugly?  I thought it looked nicer than the current slot-based system.  Can you explain why you feel it's not fun?

code/game/objects/items/weapons/storage/fancy.dm
```diff
@@ -68,14 +68,13 @@
...
 	item_state = "candlebox5"
-	storage_slots = 5
 	throwforce = 2
 	slot_flags = SLOT_BELT
 
 
 /obj/item/weapon/storage/fancy/candle_box/New()
 	..()
-	for(var/i=1; i <= storage_slots; i++)
+	for(var/i=1; i <= 5; i++)
```
Kelenius: here, too

Yoshax:
Take a look at the picture, items overlap the "boxes" their in, the stretched "boxes" look terrible, and the background doesn't resize to fit the number of "boxes" you have. In addition, I'm pretty sure for larger items, it's a nerf in total capacity.

See the screenshots from the PR you linked:
http://i.imgur.com/4e9NPlX.png
http://i.imgur.com/iyVc4pE.png
http://i.imgur.com/lGDwO8T.png

Neerti:
>items overlap the "boxes" their in

Unfortunately there's no way to fix this without retaining the box sizes used for communicating how much space you have left.

>the stretched "boxes" look terrible

I can't notice any degradation of quality on the stretched boxes on the image you showed me.

>I'm pretty sure for larger items, it's a nerf in total capacity.

For normal backpacks, they could always carry seven normal-sized items to my knowledge.  The new system still allows this.

Yoshax:
The boxes shouldn't be overly big or small, it looks bad. As does the unnecessary background when a slot isn't filled.

Other than being able to cary 400 pens, what does this actually add or improve?

Kelenius:
I agree, ID-filled box would probably look ugly.

Ithalan:
As a compromise for the HUD, how does auto-expanding rows of slots (like plant bags do, initially only showing one row, then adding more when rows are filled) sound? Current and max capacity could then be displayed with a thin bar underneath the bottom row (or maybe on the bag icon itself)

Hubblenaut:
The new inventory system was actually a driving point that made me stop playing on Bay. In my opinion, it looks ugly, the clipping icons would require a snowflakey fix (if there's an easy fix at all) and I don't see the reason for using it when there are no raised concerns with the current inventory system.

Beside that, Sin4 started working on an Inventory system despired from Sacred/Diablo, which I'd very much prefer over this one.

SparklySheep:
I actually like it a lot.

ParadoxSpace:
could we like, betatest this?

Nerezza:
I don't intend to make salt with this comment, but I've gotta say we're not trying to make a triple-A title here; looks don't really matter much as far as SS13 goes.

Also the stretch of the boxes for item icons signifies the amount of space the item takes up, which is why the plant bag thing wasn't used for this.

So far the only legit complaint I'm hearing is that it looks ugly. I agree, it looks ugly if you fill your backpack with 20 screwdrivers but WHEN do you do this? One proposed solution I've heard was to turn certain sprites 90 degrees, like screwdrivers, as the box stretch is done horizontally instead of vertically. Another possibility is to do the stretch vertically, but I think that would be even more ugly than it is right now.

As for things like ID card boxes, we could make form-fitted containers retain the current system as a box of ID cards - most likely only ever needing to hold ID cards - would not need a volume based inventory system. Same with a box of replacement lights, you don't need a volume system if you made that box form-fitted.

PR #1385, opened 2016-04-09
Names female shorts to have them show up in the loadout system.

Datraen:
Female shorts inherited their names from the male variants.

If someone wants to rename mens shorts to denote them later, feel free.


EmperorJon:
I believe it should be "women's", but that's just me being pedantic.

EmperorJon:
To be completely honest, the "female" clothing that was added is basically just one pixel row shorter for the most part. Example: image

We might as well just call them "short shorts". :P

Datraen:
Will adjust once I get out of work.

Issue #654, opened 2015-12-23
'off-hand' object drops to the ground then qdels when the holder is knocked down.


Zuhayr:
To reproduce:
- Hold a two-handed weapon in both hands.
- Target head.
- Click until head falls off or you are knocked prone.
- Observe off-hand object in the world for a bit before it del()s.

Zuhayr:
Legacy issue, code no longer exists.

PR #1387, opened 2016-04-09
Two-handed items drop check.

Datraen:
Adds a check to see if the item can be wielded when given or dropped when it is wielded.

This one was sitting around for a while.


EmperorJon:
Does this fix #654?

Neerti:
There is no off-hand obj anymore, ever since Zuh rewrote twohanding objects.

PR #1390, opened 2016-04-09
Shortens cryosleep timer

Anewbe:
Drops cryosleep timer to around 15, also affects robot storage timer

### DIFF
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -219,7 +219,7 @@
 	var/disallow_occupant_types = list()
 
 	var/mob/occupant = null       // Person waiting to be despawned.
-	var/time_till_despawn = 18000 // 30 minutes-ish safe period before being despawned.
+	var/time_till_despawn = 9000  // Down to 15 minutes //30 minutes-ish is too long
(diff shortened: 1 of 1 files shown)


Yoshax:
Isn't 15 minutes basically no change? Elevator is still immensely quicker, and more appealing. It just takes up a job slot for longer for no good reason.

Anewbe:
Admins said drop it to 15, I dropped it to 15.

EmperorJon:
I'd rather we just got rid of cryo. It doesn't really serve any purpose?

Yoshax:
It's nice fluffwise for some people.

Datraen:
It's also a nice escape pod, if you're in the area.

EmperorJon:
1. I don't really see why? If you really want to make it so you're cryogenically frozen between shifts or something, just say that happens on the lower level (which makes more sense, honestly).

2. And then put an escape pod in its place. :P

Kelenius:
I think this is better taken to forums.

Datraen:
So, nothing wrong with this, changing the cryo pod shuttle's outside of the scope of this PR. Merging.

PR #1392, opened 2016-04-09
1.5.0 Striped Socks

Woodratt:
- Added green, purple, blue, yellow, red, and orange striped socks. Thigh and knee high options.


EmperorJon:
Disgusting.

PR #1397, opened 2016-04-10
Adds full atmospherics reset admin verb

Yoshax:
- Adds "Fix Atmospheric Problems" verb, that is accessible to server admins.
- This verb will perform full reset of atmosphere. Unlike reboot-zas this fixes any possible griefs or bugs, by completely purging all existing gases and resetting them to roundstart values.

Ports Baystation12/Baystation12#12538

### DIFF
--- a/code/modules/admin/verbs/grief_fixers.dm
+++ b/code/modules/admin/verbs/grief_fixers.dm
@@ -1,0 +1,54 @@
+/client/proc/fixatmos()
+	set category = "Admin"
+	set name = "Fix Atmospherics Grief"
+
+	if(!check_rights(R_ADMIN|R_DEBUG)) return
+
+
+	if(alert("WARNING: Executing this command will perform a full reset of atmosphere. All pipelines will lose any gas that may be in them, and all zones will be reset to contain air mix as on roundstart. The supermatter engine will also be stopped (to prevent overheat due to removal of coolant). Do not use unless the station is suffering serious atmospheric issues due to grief or bug.", "Full Atmosphere Reboot", "No", "Yes") == "No")
+		return
+	feedback_add_details("admin_verb","FA")
+
+	log_and_message_admins("Full atmosphere reset initiated by [usr].")
+	world << "<span class = 'danger'>Initiating restart of atmosphere. The server may lag a bit.</span>"
+	sleep(10)
+	var/current_time = world.timeofday
+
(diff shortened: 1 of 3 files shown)


code/modules/admin/verbs/grief_fixers.dm
```diff
@@ -0,0 +1,54 @@
...
+	if(!check_rights(R_ADMIN|R_DEBUG)) return
+
+
+	if(alert("WARNING: Executing this command will perform a full reset of atmosphere. All pipelines will lose any gas that may be in them, and all zones will be reset to contain air mix as on roundstart. The supermatter engine will also be stopped (to prevent overheat due to removal of coolant). Do not use unless the station is suffering serious atmospheric issues due to grief or bug.", "Full Atmosphere Reboot", "No", "Yes") == "No")
+		return
+	feedback_add_details("admin_verb","FA")
+
+	log_and_message_admins("Full atmosphere reset initiated by [usr].")
+	world << "<span class = 'danger'>Initiating restart of atmosphere. The server may lag a bit.</span>"
+	sleep(10)
```
Kelenius: Why?

code/modules/admin/verbs/grief_fixers.dm
```diff
@@ -0,0 +1,54 @@
...
+	var/current_time = world.timeofday
+
+	// Depower the supermatter, as it would quickly blow up once we remove all gases from the pipes.
+	for(var/obj/machinery/power/supermatter/S in machines)
+		S.power = 0
+	usr << "\[1/5\] - Supermatter depowered"
```
Kelenius: Pretty sure you can send to ``src`` here.

code/modules/admin/verbs/grief_fixers.dm
Neerti: I don't see a problem here, can you elaborate, assuming you mean the sleep() call?

Kelenius: Yeah, it doesn't seem necessary.

Datraen: I question it, as well.

atlantiscze: As original author, i added it there because the message has sometimes appeared after the lag spike, not before

Datraen: Ah, that makes sense, then.

Datraen:
:+1:

PR #1398, opened 2016-04-11
Slowdown Fix

Neerti:
Slowdown is now correctly applied for objects intended to have slowdown, for example Dufflebags and Arm Guards, as many more slots are checked for slowdown instead of just the suit and shoe slots.  Slots checked are now suit, uniform, back, gloves, head, suit storage, and the left and right hands.  Shoes still get checked later on.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_movement.dm
+++ b/code/modules/mob/living/carbon/human/human_movement.dm
@@ -22,8 +22,15 @@
 	var/hungry = (500 - nutrition)/5 // So overeat would be 100 and default level would be 80
 	if (hungry >= 70) tally += hungry/50
 
-	if(wear_suit)
-		tally += wear_suit.slowdown
+	// Loop through some slots, and add up their slowdowns.  Shoes are handled below, unfortunately.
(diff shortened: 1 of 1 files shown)


Kelenius:
Does that mean you can get a speed boost by carrying extra boots in your hands?

Neerti:
Yes, fixing that unfortunate mistake now.

Yoshax:
If you're doing speed, then it might be worth removing and correcting shoes speeding you up.

Techhead0:
@Yoshax All you need to do is change the SHOES_SLOWDOWN define to 0 instead of -1. Add 1 to the base speed to compensate, if you wish.

Nerezza:
I prefer the idea of removing speed boosts from most, if not all, equipment. The only semi-reasonable explanation I can come up with for the boost is that the tiles we have on the station have poor traction for anything that isn't rubber, and that's a stretch of the imagination.

Yoshax:
Base move speed is defined in config. Better Neerti does it because he can change the config.

PR #1403, opened 2016-04-11
Lockdown fix

EmperorJon:
Borgs now show the correct operational status/lock-unlock button in the UI.
Autotraitor borgs cannot be locked down.

### DIFF
--- a/code/game/machinery/computer/robot.dm
+++ b/code/game/machinery/computer/robot.dm
@@ -91,7 +91,7 @@
 		if(!target || !istype(target))
 			return
 
-		var/istraitor = (target.mind.special_role == "Traitor")
+		var/istraitor = target.mind.special_role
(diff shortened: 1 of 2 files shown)


EmperorJon:
Aaand part 2 fixed.

EmperorJon:
What this doesn't do is resolve how Traitor and Autotraitor are still two different antag types as understood in many places in the code. For example, they both have the same "special_role" flag, and both show up in the traitor panel as "Traitor" (the one right at the bottom of the list is actually Autotraitor). If you were to assign both (somehow) then remove one, it would think they were still the other, but they'd have no special role.

This raises a point. There may be other checks in the code somewhere for if someone is a traitor that won't be flagged for autotraitors. This probably needs investigating at some point?

Datraen:
:+1:

PR #1407, opened 2016-04-12
Resist tweaks.

EmperorJon:
Ports https://github.com/Baystation12/Baystation12/pull/12360

Fixes not being able to resist cuffs.


Datraen:
Travis likes, I don't see any problems.

PR #1410, opened 2016-04-13
Cleans up and updates xenoarch

Kelenius:
See changelog
Replaced a lot of macros with spans.
Defines used instead of magic numbers.
Excavation amount is now in cm, not percent, who ever thought this is a good idea (wait wait don't tell me).

Aaaaaaaaaaaaaaaaaaaaaaaa I made it though.


EmperorJon:
Mmmm giant tasty code review time.

Ithalan:
I'm not quite sold on the anomaly locater also detecting normal archeological finds being a good change, at least if it still only show the distance to the nearest find. Anomalists have nothing to do until an anomalous artifact is actually brought into the lab (or more often, they manage to find one themselves), and I think making them harder to pick out from among the regular finds is the last thing that is needed.

Kelenius:
It shows both.

code/modules/xenoarcheology/effect.dm
```diff
@@ -12,20 +10,11 @@
...
 
-//0 = Unknown / none detectable
-//1 = Concentrated energy
-//2 = Intermittent psionic wavefront
-//3 = Electromagnetic energy
-//4 = Particle field
-//5 = Organically reactive exotic particles
-//6 = Interdimensional/bluespace? phasing
-//7 = Atomic synthesis
-
```
RenScorch: Did you move these comments or did you just delete them. If you deleted them can you put them back. If you move them, can explain where you moved them?  If they are not required anymore, can explain why they are not required anymore?

Kelenius: They're replaced by defines.

EmperorJon:
Something conflicted somewhere. :|

Kelenius:
Well. Like usual, will fix after approval.

Kelenius:
Aaaaaaaaaaaaaaaaaaaaany time now.

Kelenius:
:<

Zuhayr:
This is tragic.

Kelenius:
I've lost interest, someone else can pick this up.

PR #1412, opened 2016-04-14
Lockers can not trap live hostile mobs

Hubblenaut:
As it says on the tin. Live hostile mobs will not simply let themselves be caught in a locker.

### DIFF
--- a/code/game/objects/structures/crates_lockers/closets.dm
+++ b/code/game/objects/structures/crates_lockers/closets.dm
@@ -142,6 +142,8 @@
 /obj/structure/closet/proc/store_mobs(var/stored_units)
 	var/added_units = 0
 	for(var/mob/living/M in src.loc)
+		if(M.stat && istype(M, /mob/living/simple_animal/hostile))
+			continue
 		if(M.buckled || M.pinned.len)
 			continue
 		if(stored_units + added_units + M.mob_size > storage_capacity)


Kelenius:
Everything can be locked into them, except for the few arbitary mobs that for some reason can not.

OhEmGee:
...Why?

Hubblenaut:
Because those are hostile and shouldn't simply be stopped by closing a locker on them.
Alternatively, we can have them reopen the locker in their next tick, but that could be avoided by repeatedly closing it until locking or welding it shut.

Kelenius:
Locking humans in the lockers: fine.
Locking constructs in the lockers: fine.
Locking cyborgs in the lockers: fine.
Locking goats in the lockers: physically impossible unless the goat is dead.

Hubblenaut:
If you want to script elaborate behavior for mobs to avoid being locked into a locker, feel free.
But quit those passive aggressive remarks. Either give actual constructive criticism and possible solutions for the problem at hand or remain quiet.

Datraen:
I'd prefer simple animals as a whole having a chance to try to break out of a locker in their life code, failing if the locker's welded. Simply being unable to trap a giant spider or goat in a locker is rather... odd.

ParadoxSpace:
It should be odd to be able to trap a giant spider in a locker, that's stupid.

Kelenius:
I agree that it's silly, but this is bad implementation.

Hubblenaut:
Will close this quick bandaid PR then. I do not have much time on my hands right now, so don't expect any more elaborate fix for this by me soon.

PR #1415, opened 2016-04-15
Reinforced Walls

Hubblenaut:
- Lowers amount needed for building a reinforced wall from a girder from
four to two sheets.
 - Gets rid of rod products from dismantling a reinforced wall.
 - Adds one more extra step for undoing the deconstruction.

If I'm not mistaken, building reinforced walls used to only take two plasteel sheets per wall. With 10 sheets per stack of plasteel spawned in engineering, this will now make it possible to build five walls per stack and 15 walls in total with their given materials instead of half of it.

### DIFF
--- a/code/game/objects/structures/girders.dm
+++ b/code/game/objects/structures/girders.dm
@@ -143,7 +143,8 @@
 
 	user << "<span class='notice'>You begin adding the plating...</span>"
 
-	if(!do_after(user,40) || !S.use(2))
+	var/amount_to_use = reinf_material ? 1 : 2
+	if(!do_after(user,40) || !S.use(amount_to_use))
 		return 1 //once we've gotten this far don't call parent attackby()
 
 	if(anchored)
@@ -167,7 +168,7 @@
 		user << "<span class='notice'>\The [src] is already reinforced.</span>"
 		return 0
 
-	if(S.get_amount() < 2)
+	if(S.get_amount() < 1)
 		user << "<span class='notice'>There isn't enough material here to reinforce the girder.</span>"
 		return 0
 
@@ -177,7 +178,7 @@
 		return 0
 
 	user << "<span class='notice'>Now reinforcing...</span>"
-	if (!do_after(user,40) || !S.use(2))
+	if (!do_after(user,40) || !S.use(1))
 		return 1 //don't call parent attackby() past this point
 	user << "<span class='notice'>You added reinforcement!</span>"
 
--- a/code/game/turfs/simulated/wall_attacks.dm
+++ b/code/game/turfs/simulated/wall_attacks.dm
@@ -217,8 +217,7 @@
 				if (istype(W, /obj/item/weapon/wirecutters))
 					playsound(src, 'sound/items/Wirecutter.ogg', 100, 1)
 					construction_stage = 5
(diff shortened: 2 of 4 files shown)


Datraen:
I'd think a welding tool would make more sense for mending a grille, rather than wire cutters.

Hubblenaut:
Considered that, though I fear that would confuse some.

Datraen:
It'd have to be updated on the wiki prior to updating live, but otherwise makes sense. People would be confused from the lack of rods, though.

Hubblenaut:
If I had enough time and motivation I'd completely redo how reinforced walls are assembled and dismantled, because there are quite some things wrong with them.
And yes, wiki will be updated. I was going to do it after merge/confirmation.

Datraen:
I was just making sure. While odds are this won't be too prevalent, I'd like to avoid any issues like those we've had in the past.

Nalarac:
Wouldn't this make building a reinforced wall cost the same amount of resources as a regular wall?

Hubblenaut:
It would, yes, but that's how it always was.
An alternative though would be to spawn more plasteel for engineering, as it's definitely not enough.

Nalarac:
I understand where you're coming from, but you can reinforce walls with things other than plasteel now as well, so you could make a steel reinforced steel wall for the same cost as a normal steel wall. Unless, of course, you're only changing the value for using plasteel.

mwerezak:
I believe standard rwalls are plasteel-reinforced steel as well, so the cost is actually 2 plasteel sheets and 2 steel ones (correct me if I'm wrong). If you use 4 plasteel sheets you make a wall that has even more health than a standard r-wall.

Hubblenaut:
No, Nalarac is right on this, and I considered what could be done to reason the differences between normal and reinforced walls.
In the end, you can currently also use 6 sheets of metal for a reinforced wall that may not be as strong as plasteel, but takes the same lengthy time to take down.

Datraen:
Is this still being debated?

Nerezza:
I can confirm that current r-walls are finished with plasteel, but I don't know if they're steel reinforced girders underneath.

A compromise currently popping up in engineering is to do Girders>Steel>Plasteel so 4 sheets of steel and the 2 plasteel you get from deconstructing the wall. These are indistinguishable from the original r-walls, and frankly I've begun to wonder if reinforcing girders with something other than steel should be a thing.

SinTwo:
Actually, you can make walls and r-walls out of any building material. You make girders with steel sheets. The apply your materials. For example, you can make a glass wall reinforced with wood.

Kelenius:
Compromise: replace station r-walls with steel-reinforced and keep the recipes as they are. Normal walls will cost 4 steel, rwalls will cost 6.

Nerezza:
Yeah, I'm aware that the r-walls can be built from 2 steel and 2 and 2 (how does this grammar?) of any other material. I'm just saying, once the outer plasteel is on the exterior the steel reinforced girders and the plasteel reinforced girders are both completely indistinguishable visually, and I doubt there's much difference in health between them too.

I'm also worried that steel r-walls would be much weaker than reinforced r-walls. In particular, the engine core's r-walls and the toxins mixing r-walls both need to be strong to prevent incidents.

PR #1376, opened 2016-04-07
Ghosts can now follow any /atom/movable object.

Yoshax:
Ports Baystation12/Baystation12#12168

PR #1420, opened 2016-04-16
Revert "Ghosts can now follow any /atom/movable object."

Yoshax:
Reverts PolarisSS13/Polaris#1376

Fixes#1419

I'm probably not going to fix it, or figure out how to, for reasons, and I don't see anyone else doing so soon, and it's kinda decent thing to have.


Yoshax:
Or not and it magically works now.

PR #1438, opened 2016-04-18
Adds three new mixed modes.

Datraen:
Mercenaries & Renegades:
Requirement: _16_
Secret Requirement: 15
Odds are it won't hit secret requirement if it's going to be a secret round.

Mercenaries & Wizard:
Requirement: **15**  I'm blind and look at the wrong files.

Traitors & Renegades:
Requirement: 10
Secret Requirement: 11
Not autotraitor.


zekesulastin:
Uhh, isn't the regular mercenaries mode minimum of fifteen people or did that get changed?  I would think that the player requirements would be /more/ onerous if you are adding more powerful external threats.

Mark9013100:
**GOD SAVE US ALL.**

Ithalan:
Idea for future additions: Mercenaries & Cultists

Datraen:
Zeke: For renegades, I'm not anticipating antag status as a reason TO murderbone, but as more of a defensive thing. If it's entirely seen as an issue as renegades go around murderboning, I'll be sure to bump that back up to the 20 I had it originally. (AKA never played unless people really really want it.)

Wizard could do neat wizardy things. I don't know how it'll play out, and these are just requirements for the purpose of playtesting and seeing how things go. It could get bumped up easily.

In response to Mercenaries & Cultists:
I hate current Cult with a burning passion. If someone else wants to do it, so be it. The framework is there and easy to get the game mode implemented.

EmperorJon:
I don't like the idea of Merc + Renegades only being 12 voted. I've got a feeling that this will mean people will vote for it over Merc, because it'll basically be Merc with EVEN MORE ACTION as the renegades screw the Mercs plans over by being armed. That plus when Merc fails people might go for the easier option. I'm just dreading seeing 12 people games where we have like 6 Mercs, 3 Renegades, and 3 people who aren't antags.

I do support Renegades as a concept, however, as long as we make it clear that your survival object is just that - survival - and that doing something like pre-emptively hacking the AI isn't survival, it's asking to be caught. Renegades as an "illegal" antagonist way of causing more friction with other antagonists, without directly increasing the amount of antagonist stuff occurring, gets my thumb-up.

EmperorJon:
We can balance the numbers later if need be. I'm going to merge this.

PR #1445, opened 2016-04-18
Adds Vey-Med prosthetic limbs.

Anewbe:
Adds human-like Vey-Med prosthetics.


martin509:
Don't we already have Zeng-Hu to fill that niche? Or is this more human-like than ZH?

Anewbe:
It is more human than ZH, yes.

Datraen:
Pictures?

Hubblenaut:
Zeng-Hu doesn't look much human-like to me at all? It's a mix of black amd yellow.

EmperorJon:
Definitely going to want pictures here.

Anewbe:
veymed

EmperorJon:
I like them, although I feel they somewhat invalidate the existence of the ZH ones?

Anewbe:
I would have made these the default ZH sprites, but I decided (with some prompting) that it was probably better to keep something similar to the old sprites, which I did, the results of which have already been merged.

PR #1450, opened 2016-04-19
Ports attack call sequence cleanup from Bay

mwerezak:
Porting this because I noticed Polaris is missing the attack code cleanup from last summer and it's only going to get more difficult to port with time.

Port of https://github.com/Baystation12/Baystation12/pull/10287

Attempts to clean up the item-attacking-mob call sequence cleaner by reorganizing each call to have a clearly defined purpose.

The call sequence now looks like this:

- `mob/ClickOn()` calls the item's resolve_attackby() proc. *Unchanged.*
- `item/resolve_attackby()` calls the target atom's attackby() proc. *Unchanged.*
- `mob/living/attackby()` after checking for surgery, calls the item's attack() proc.
- `item/attack()` is what you override to define what an item does when it is used on a mob. *By default, it assumes the item is a weapon, and does what every weapon should do: generates attack logs before calling the mob's attacked_with_item() proc.*
- `mob/attacked_with_item()` defines what happens when a mob is attacked with a weapon.
*Not just any time an item is used on a mob, but specifically when it is used as a weapon. This is largely where combat code begins.*
For humans, this does all the standard attack roll and checking what body part gets hit and that stuff. If the attack is a successful hit, it calls apply_hit_effect().
- `item/apply_hit_effect()` defines what happens when a weapon successfully makes contact. **This is something really basic yet was sorely missing from the codebase, and is what got me started on this cleanup.** *For example, stun batons now just override this instead of re-duplicating the attack roll and hit/miss messages, since that's now done by attacked_with_item()!*
While this can be overriden to do whatever you want, "standard" weapons (or weapons that act like standard weapons in addition to doing other stuff) should call the target mob's hit_with_weapon() proc to ensure that all regular weapons have the same code for applying damage and armour, bloodying the weapon and making blood splatters, embedding, all that stuff, without having to duplicate everything.
- `mob/living/hit_with_weapon()` defines how "standard" weapons are handled when they successfully hit a mob. This is intended to be more of a helper proc that is called by apply_hit_effect() overrides when they want to re-use the default weapon behaviours. I guess it would benefit from being split up further but that's a job for another time.

This mainly cleans up the call sequence for mob/living/carbon/human. Other mobs have been updated so that they will work, but a serious attempt to clean up every mob was is included in this PR. That said, the bulk of attack code deals with humans, for obvious reasons.


code/modules/mob/living/carbon/carbon_defense.dm
```diff
@@ -0,0 +1,98 @@
...
+		weapon_sharp = 0
+		weapon_edge = 0
+
+	apply_damage(effective_force, I.damtype, hit_zone, blocked, sharp=weapon_sharp, edge=weapon_edge, used_weapon=I)
+
+	//Melee weapon embedded object code.
+	if (I && I.damtype == BRUTE && !I.anchored && !is_robot_module(I))
+		var/damage = effective_force
+		if (blocked)
+			damage /= blocked+1
```
mwerezak: This needs to be updated.

code/modules/mob/living/living_defense.dm
```diff
@@ -144,6 +144,42 @@
...
+
+	if(I.damtype == BRUTE && prob(33)) // Added blood for whacking non-humans too
+		var/turf/simulated/location = get_turf(src)
+		if(istype(location)) location.add_blood_floor(src)
+
+	return blocked
+
+//returns 0 if the effects failed to apply for some reason, 1 otherwise.
+/mob/living/proc/standard_weapon_hit_effects(obj/item/I, mob/living/user, var/effective_force, var/blocked, var/hit_zone)
+	if(!effective_force || blocked >= 2)
```
mwerezak: This too.

code/modules/mob/living/carbon/human/species/xenomorphs/alien_facehugger.dm
```diff
@@ -0,0 +1,232 @@
+//This file was auto-corrected by findeclaration.exe on 25.5.2012 20:42:32
```
mwerezak: Got carried in with the port, will remove.

Hubblenaut:
Good thing! Does this also happen to fix the bullet hit messages showing on seemingly random occasions?

mwerezak:
I don't know, I haven't experienced that issue myself. Is it a Polaris specific bug?

mwerezak:
With any luck, it should be good to go.

mwerezak:
Well, I am dumb

Neerti:
Is this ready for merging?

mwerezak:
It could use some more testing

Yoshax:
Merge it and test it live :)

Datraen:
Needs some updating with the removal of the removeItem and similar procs.

mwerezak:
The ironic thing is that those were what caused most of the cherry-pick conflicts in the first place.

mwerezak:
Sorry guys, I don't really have much desire to continue working on this, I'd rather spend the time on other things. You guys have the diff, and you have these commits so if anyone else around here thinks this is worth having you should be most of the way there to finishing it.

For those who don't already know how to fetch a pull's commits, it's just `git fetch polaris pull/1450/head`

PR #1453, opened 2016-04-20
Round start timer now remains at 0 until the round actually starts.

Yoshax:
The transfer vote should thus now happen after 3 hours of gameplay, instead of after 3 hours of uptime.

Ports Baystation12/Baystation12#12691


EmperorJon:
Saw this on Bay the other day. Was going to port. You beat me. :+1:

PR #1455, opened 2016-04-20
Gender tweaks and adjustments

Yoshax:
Ports:
Baystation12/Baystation12#12669
Baystation12/Baystation12#12702

Splits gender into biological gender and gender identity.
- allows species' gender selection to be limited
- biological gender determines sprite where appropriate
- both are shown in medical records
- gender identity determines pronouns upon examine and suchlike to allow for androgynous/agender/other things

### DIFF
--- a/code/_helpers/global_lists.dm
+++ b/code/_helpers/global_lists.dm
@@ -23,6 +23,9 @@ var/global/list/joblist = list()					//list of all jobstypes, minus borg and AI
 
 var/global/list/turfs = list()						//list of all turfs
 
+#define all_genders_define_list list(MALE,FEMALE,PLURAL,NEUTER)
+#define all_genders_text_list list("Male","Female","Plural","Neuter")
+
 //Languages/species/whitelist.
 var/global/list/all_species[0]
 var/global/list/all_languages[0]
--- a/code/_helpers/text.dm
+++ b/code/_helpers/text.dm
@@ -323,3 +323,5 @@ proc/TextPreview(var/string,var/len=40)
 			if(48 to 57)			//Numbers
 				return 1
 	return 0
+
+#define gender2text(gender) capitalize(gender)
--- a/code/datums/datacore.dm
+++ b/code/datums/datacore.dm
@@ -1,3 +1,4 @@
+
 /hook/startup/proc/createDatacore()
 	data_core = new /datum/datacore()
 	return 1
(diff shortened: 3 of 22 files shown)


nachomeep:
Best pull request in the past months.

Kelenius:
>No apache helicopter option

EmperorJon:
Please, the "I identify as an attack helicopter" meme got old several years ago.

Yoshax:
Can we keep the stupid jokes and comments out of the discussion? They're not helpful.

PsiOmegaDelta:
This is technically an improvement over the current Polaris situation, were the main gender var can be set to male/female/neuter/plural for any and all species during character setup, and were the neuter/plural options tend to cause interesting effects for human-like (not humanoid) mobs as some sections of the code checks if the mob is male and otherwise assumes female while other locations switch it around.

There is certainly a compromise baked in here, allowing people to select their preferred gender identity and having it show up when examined, however, if you really do want to identify as an Apache helicopter that is now possible without breaking BYOND.

RoseCG:
If this works, would it be possible to do the same thing for hair colors?

This was something discussed on Bay, and I know some characters only have their hair dyed instead of genetically modified. Once they're cloned, they come to life with their original hair color.

Yoshax:
Probably wouldn't be too difficult. But I'm not sure where hair color is used or checked or anything of the sort. But likely simple.

Datraen:
:+1:

Neerti:
One possible issue I can imagine with this is that for certain FBP synthetics, their 'sex' would be locked to MALE or FEMALE, which might be odd for some people.

PsiOmegaDelta:
That's a tricky one, as I assume FBPs are not their own species template anymore but simply a full body prosthesis?

Neerti:
Yes, as FBP is an anagram of that.  Unfortunately I can't think of a nice way to 'test' for being an FBP for a character that exists as a preference setup instead of a mob instance.

PsiOmegaDelta:
I have an idea, that should do the trick, just need some spare time.

Yoshax:
To do:
- Change appearance changer to check for and allow FBPs to select genders other than male and female.

EmperorJon:
DNM pending (to do)?

Yoshax:
To do is done.

Datraen:
I didn't notice any glaring at a quick glance, someone else might want to look at it.

Zuhayr:
@Neert belatedly - check if the chest is set to prosthetic in the preferences robot limb data.

nachomeep:
Do neuters use the male chest, female face still?

Yoshax:
I didn't change any of the sprites, or where they're used.

PR #1459, opened 2016-04-21
Breacher Helmet Force

Anewbe:
Gives Breacher helmets force 5, makes their headbutts more dangerous


Nerezza:
They still have that horn right?

EmperorJon:
Yes. What this does is make a headbutt with a breacher helmet do more force, as headbutts apparently check headwear force which none have!

mwerezak:
I wonder if the horn should be `sharp`, as well...

PR #1470, opened 2016-04-22
Changes to SMES Interface

Hubblenaut:
Image

 - Transition to using kilowatts and kilowatt hours
 - Adds an input load bar with value
 - SMES charge rate and coil capacities slightly adjusted for using
   kilowatt units

### DIFF
--- a/code/modules/power/smes.dm
+++ b/code/modules/power/smes.dm
@@ -1,7 +1,7 @@
 // the SMES
 // stores power
 
-#define SMESRATE 0.05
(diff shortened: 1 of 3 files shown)


Datraen:
Provided it handles decimals, which you said it did in dsay, :+1:.

Hubblenaut:
To clarify, nothing has been changed to custom input other than a division by 1000. Float values are possible, but we're rounding at one digit behind the dot.

Datraen:
Sounds good.

mwerezak:
You may want to grab https://github.com/Baystation12/Baystation12/pull/12914

PR #1476, opened 2016-04-22
Adds All of the Winter Coats to the Loadout.

redstryker:
Does as the title says.  Nothing more.  Nothing less.  I _would_ have made it more organized with the new system, but I don't know how to do that/if it works if you want there to be job restrictions for the items.


Hubblenaut:
New system is self-organizing IMO.
One question:
Why is the engineering coat available for atmospheric technicians, but the atmospheric one not available for engineers?

Datraen:
We have to add a system that does job restrictions at some point, if it's not done by the time Xenobio2 is done, I'll take a crack at it.

Zuhayr:
The current loadout system does do job restrictions.

Datraen:
@Zuhayr: The current "issue" is that the implementation that allows for listed selections doesn't handle job restrictions on a item-by-item basis, it's all or nothing.

PR #1478, opened 2016-04-22
Makes opening/closing pizza boxes easier

Nerezza:
Alt-click now opens/closes pizza boxes like a laptop.

The current method is:
1: Pull pizza out of box
2: Pick box up in other hand
3: Put pizza back in box
4: Use pizza box
5: Optional, put box back on table

I've made sure this only works if the user is living and ajacent to the boxes, and alt-click on a stack of pizza boxes is ignored.


Datraen:
:+1:

PR #1480, opened 2016-04-22
Makes inventory slots forward clicks to items in them.

Yoshax:
In effect that means you don't have to pixelhunt item in the slot, just click the slot. Smoking cigs is easy mode now.


EmperorJon:
I cannot approve of this enough.

PsiOmegaDelta:
Note that this doesn't help with container/storage slots, only the main inventory ones.

PR #1481, opened 2016-04-22
Hardsuits are now only checked for access when not sealed

Yoshax:
Access is checked when you are trying to seal the suit, you cannot seal it without the proper ID. If the suit is not sealed you require ID to use any part of it.

If the suit is sealed, you no longer need an ID to use it, as it was checked upon sealing.


Anewbe:
So in lazyman's terms, you only need the ID to get the hardsuit active, after that you're good to go?

Yoshax:
Essentially yes.

Datraen:
:+1:

PR #1482, opened 2016-04-22
Fixes Taj/Unathi being unable to eat mice.

Anewbe:
Taj/Unathi will now be able to eat mice and lizards again.

Adds two mob_helper checks for mob size, istiny and ismini.

EDIT: Teshari can also eat mice and lizards again/now.

### DIFF
--- a/code/modules/mob/mob_grab_specials.dm
+++ b/code/modules/mob/mob_grab_specials.dm
@@ -140,14 +140,14 @@
 
 /obj/item/weapon/grab/proc/devour(mob/target, mob/user)
 	var/can_eat
-	if((FAT in user.mutations) && issmall(target))
+	if((FAT in user.mutations) && ismini(target))
 		can_eat = 1
 	else
 		var/mob/living/carbon/human/H = user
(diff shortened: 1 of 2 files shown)


Nerezza:
What about Teshari?

Anewbe:
They're not edible, but yes, they can eat mice and lizards too. Editing title.

Yoshax:
If all these species can eat mice, why can't humans?

PJB3005:
Because humans have some level of dignity `:^)`

Nerezza:
I'd imagine it's because humans aren't so beastly as the beast races so they can't eat them raw. They can still take a kitchen knife to the mice, butcher them, cook that, and eat that if they want... Actually they could eat the butchered meat too I guess?

Honestly, I didn't know Tajaran and Unathi could eat mice also. I only knew Teshari could since they're pretty much the beastiest of beasties on the station, literally speaking of course.

sabiram:
A human is fully capable of eating a mouse whole

Zuhayr:
It was added for gameplay variance.

PR #1483, opened 2016-04-22
Moves antag text to antag levels

EmperorJon:
The text which says "YOU ARE AN ANTAGONIST be an opposing force" etc. etc.

Changes renegade text from default to make it clear they're minor antagonists meant to resist the major ones. Changes ERT text to make it clear they're not antagonists (not that anyone ever comments on it, but that is what it says)...

### DIFF
--- a/code/game/antagonist/antagonist_create.dm
+++ b/code/game/antagonist/antagonist_create.dm
@@ -99,11 +99,14 @@
 		player.current << "<span class='notice'>[leader_welcome_text]</span>"
 	else
 		player.current << "<span class='notice'>[welcome_text]</span>"
+	if (config.objectives_disabled)
+		player.current << "<span class='notice'>[antag_text]</span>"
 
 	if((flags & ANTAG_HAS_NUKE) && !spawned_nuke)
 		create_nuke()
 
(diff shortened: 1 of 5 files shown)


code/game/antagonist/antagonist_create.dm
```diff
@@ -99,11 +99,14 @@
...
 	else
 		player.current << "<span class='notice'>[welcome_text]</span>"
+	if (config.objectives_disabled)
+		player.current << "<span class='notice'>[antag_text]</span>"
 
 	if((flags & ANTAG_HAS_NUKE) && !spawned_nuke)
 		create_nuke()
 
-	show_objectives(player)
+	if (!config.objectives_disabled)
```
Zuhayr: The purpose of not checking config here was for people given objectives by admin via the interface, since making it depend on the config resulted in being unable to give people objectives without enabling it for all subsequent antagonists.

EmperorJon: Not sure I get what you're saying. Wouldn't this only mean that an antagonist wouldn't see their objectives if the config was disabled if they got objectives _before_ they became an antag?

PR #1486, opened 2016-04-24
Merge Map Changes

ursamedium:
GIT HATES MAP AS TEXT OH GOD


Zuhayr:
For reference, you need to run 1prepare_map before editing the map, and 2clean_map after, to avoid this immense diff. The scripts are in tools/mapmerge.

ursamedium:
I AM SO EMBARRASSED THAT I MADE THIS PR TO THE ACTUAL POLARIS MASTER WHILE I WAS FUMBLING AROUND gosh that's what I get for trying to do things at way tired I am so sorry
Thank you though! That will help a lot ahaha! I'm not experienced at ALL with mapping so .. yeah.

Zuhayr:
Happens all the time on Bay, don't worry about it too much.

PR #1490, opened 2016-04-25
Removes move delay after clicking

Kelenius:
http://ss13polaris.com/forums/viewtopic.php?f=18&t=700 feedback is positive


Datraen:
I'd honestly prefer a poll.

Kelenius:
12/2 good enough?

PR #1491, opened 2016-04-25
Morgue Trays interface with occupant minds

Hubblenaut:
- Morgue Trays will show a green blinking light when its occupant can be cloned.
 - A ghost returning to a body in a morgue tray will alert doctors with HUDs.


Datraen:
Why was this not merged yet.

PR #1495, opened 2016-04-26
Buildables v1.2

SinTwo:
Removes text2path, adds constructability to a few things, adds a locker_painter.dm, and several other fixes.


Kelenius:
I think it's better to do something like #1151 before adding more buildables.

SinTwo:
...I did? Unless you want to point out a more specific part that I didn't do.
This is mostly a fix update hence the v1.2 instead of v2.

Kelenius:
Oh. Huh.

code/game/machinery/mass_driver.dm
```diff
@@ -9,12 +9,40 @@
...
 	active_power_usage = 50
+	circuit = /obj/item/weapon/circuitboard/mass_driver
 
 	var/power = 1.0
 	var/code = 1.0
 	var/id = 1.0
 	var/drive_range = 50 //this is mostly irrelevant since current mass drivers throw into space, but you could make a lower-range mass driver for interstation transport or something I guess.
 
+/obj/machinery/mass_driver/map/New()
+	circuit = new circuit(src)
```
PsiOmegaDelta: Missing call to ..()

PR #1500, opened 2016-04-27
Adds some dresses

Yoshax:
Pic:
https://i.gyazo.com/d28a6892f7f99515d3702e6a73f59bab.png

From left-to-right:
Long-wide dress
Plain dress
Long dress
White wedding dress
Red belt with dress


EmperorJon:
Not really a fan of the colours by more clothing = almost always good.

PR #1509, opened 2016-04-28
Santizes ambitions before they're set to the mind variable.

Datraen:
Just a simple swap of lines.


Neerti:
Datraen proves I'm a dummy.

Datraen:
Like I mentioned in TS to him just now, it's the kind of mistake I'm prone to make.

PR #1516, opened 2016-04-29
Map QOL changes

### LINKED ISSUE #1514: ORs for medical are missing sinks.
see title
For the rooms to have sinks
#### What actually happened
they don't
#### Steps to reproduce
move to OR 1 and 2 and see they don't have sinks.

Yoshax:
- Adds more sinks to medbay
- Adds crowbars to spawn locations
- Ensures flavour airlocks & rooms are not open to space
- Adds some vending machines to the arrivals shuttle
- Adds a plotted plant or two around the station

Fixes #1514


Datraen:
Why are we adding crowbars to spawn locations?

Yoshax:
Because people join when there's no power and are indefinitely trapped in a room. I saw a round yesterday where for about a half hour there was a good four or five people trapped in spawn locations because no power and no crowbars.

mwerezak:
Adding a couple of emergency toolboxes in some lockers instead of just straight crowbars might be more thematic.

Yoshax:
I guess, it was more a clarity thing. Not everyone realises there's crowbars potentially in those lockers or stuff.

Datraen:
Lockers should contain equipment that is useful. I do prefer that solution personally.

Yoshax:
It makes a bit more sense, but as I said, many people do not realise there's even a possibility for that, nor will they.

mwerezak:
At what point do you stop making concessions for players that can't be bothered to look around for solutions when faced with a problem?

Yoshax:
Because it's not clear at all? There's no indication or anything of the sort there would be crowbars in the lockers that are known by almost everyone to hold internals.

Datraen:
Alternative: Create emergency lockers, add those.

mwerezak:
Incidentally, O2 lockers have always had a chance of spawning with crowbars.

But if you really want to be obvious about it then you can probably reuse the red fire equipment locker sprite for the emergency lockers.

Yoshax:
Yeh, I'm aware of that, the problem is that it isn't clear or obvious to anyone that they have the possibility, or anything of the sort to have crowbars.

sabiram:
It's not clear/obvious that the box in your backpack has emergency oxygen gear, either. If the wiki didn't have formula for medication or the steps required for building things, it would all be privileged information.

If they have to open an oxygen locker, they might find a crowbar, and now they know that crowbars can be in there sometimes. I think it's more interesting to have The Player explore a little. Not that this is a big deal, anyway

PR #1517, opened 2016-04-29
Buildables v1.2

SinTwo:
Removes text2path, adds constructability to a few things, adds a locker_painter.dm, and several other fixes.

- Adds a locker painter
- Frames built are directional
- Sleepers, body scanners, mass drivers, and supply request consoles can be built
- Parts are once more up-gradable


Datraen:
Why was #1495 closed and the branch deleted, if you were going to just open another PR with the same commits?

EmperorJon:
@Neerti What's our process for merging stuff like this, do we squash commits or no?

Neerti:
Up to you, really.  In this case it might be nice to squash.

PR #1518, opened 2016-04-29
Adds tear gas grenades and small changes

Yoshax:
- Adds tear gas grenades
    - Adds a box to the armory
    - Adds a box to ert area
    - Adds them to the antag uplink
- Changes smoke grenade box to having 7 smoke grenades
    - Removes single stacked smoke grenades from merc base and ert area and replaced them with box/es

Tear gas grenades are slow acting, and have a long delay after exposure/inhalation before the full effect is in place. They are however incredibly crippling and powerful if exposed to for a long period of time. In addition, they also have a large area of effect.


Yoshax:
DNM whilst I port the thing that makes them hilariously not OP.

Yoshax:
Or not, because it refuses to work properly and I'm not entirely sure how it works.

PR #1519, opened 2016-04-29
Smoke grenade changes and adjustment

Yoshax:
- Adjusts smoke grenade code
- Smoke grenades can now change their color on-the-fly with a multiool. High-tech stuff.
- Adds a smoke_strength variable to smoke grenades
    - Smoke strength controls how much smoke is produced
    - Future smoke grenades can be made to have more or less smoke

### DIFF
--- a/code/game/objects/items/weapons/grenades/smokebomb.dm
+++ b/code/game/objects/items/weapons/grenades/smokebomb.dm
@@ -38,4 +38,6 @@
 
 /obj/item/weapon/grenade/smokebomb/attackby(obj/item/I as obj, mob/user as mob)
 	if(istype(I,/obj/item/device/multitool))
-		smoke_color = input(user, "Choose a color for the smoke:") as color|null
+		var/new_smoke_color = input(user, "Choose a color for the smoke:", "Smoke Color", smoke_color) as color|null
+		if(new_smoke_color)
+			smoke_color = new_smoke_color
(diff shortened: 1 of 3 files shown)


code/game/objects/items/weapons/grenades/smokebomb.dm
```diff
@@ -37,3 +35,7 @@
 	sleep(80)
 	qdel(src)
 	return
+
+/obj/item/weapon/grenade/smokebomb/attackby(obj/item/I as obj, mob/user as mob)
+	if(istype(I,/obj/item/device/multitool))
+		smoke_color = input(user, "Choose a color for the smoke:") as color|null
```
PsiOmegaDelta: Just checking, is it intentional that when the user clicks cancel during color selection that the set color is reset?

Yoshax: No, not particularly. If the user cancels the color selection should remain as whatever the smoke_color var holds.

PsiOmegaDelta: You want to check that the return value from input isn't null before assigning it to smoke_color then.
As a QoL thing you can supply smoke_color as the 4th argument to input and it'll be pre-marked during selection (the 3rd argument is the title).

Yoshax:
To do:
- Make smoke code work

DNM whilst To Do.

Yoshax:
To do is done.

PR #1520, opened 2016-04-29
Allows Robotics To Print Drone Brains

ParadoxSpace:
it's about time

### DIFF
--- a/code/modules/research/designs.dm
+++ b/code/modules/research/designs.dm
@@ -656,7 +656,17 @@ other types of metals and chemistry for reagents).
 	build_path = /obj/item/device/mmi/digital/posibrain
 	category = "Misc"
 	sort_string = "VACAB"
-
+	
+/datum/design/item/mmi/digital/robot
+	name = "Robot Intelligence Circuit"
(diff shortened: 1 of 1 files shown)


Yoshax:
Does that work? It seems to me from a cursory glance that the brain will never get a client.

ParadoxSpace:
nh you're right
IIRC drone brains couldn't be manually activated the last time I checked??

Yoshax:
You could make it work.

Zuhayr:
What is the point of positronic brains if the software brains, which were explicitly intended to remove the 'pick it up and put it in a new chassis' aspect of robots, are just going to become posibrains mk2? I don't get why people have been steadily adding more functionality to them when they were intended to be a one-shot variant for the robot equivalent of playing a DNC crewmember.

Kelenius:
>Available both in R&D and robotics
>5 minutes of research
>Only metal and glass

Are you fucking high

ParadoxSpace:
considering that a software brain is less advanced than a pAI, and a pAI only requires metal and glass

PR #1526, opened 2016-05-01
Cleans up black jumpskirt icon and genericises description

sabiram:
The old north facing icon had a leftover seam on the skirt portion from the 'female jumpsuit' icon I modified. The seam has been removed. 

The old description made reference to a specific size of clothing, as it was originally going to be a custom personal item for a particularly lithe individual. It's readily available for anyone, so it's not particularly fitting (:>) to describe it otherwise.


sabiram:
I think I may have potentially fixed it

sabiram:
hello

PR #1530, opened 2016-05-01
Adds a check for can_speak to prevent null language procs being called.

Datraen:
Prevents runtimes* from procs on a null.

### DIFF
--- a/code/modules/mob/language/language.dm
+++ b/code/modules/mob/language/language.dm
@@ -158,7 +158,12 @@
 
 // Can we speak this language, as opposed to just understanding it?
 /mob/proc/can_speak(datum/language/speaking)
-	return (speaking.can_speak_special(src) && (universal_speak || (speaking && speaking.flags & INNATE) || speaking in src.languages))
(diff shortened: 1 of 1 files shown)


code/modules/mob/language/language.dm
```diff
@@ -158,7 +158,12 @@
 
 // Can we speak this language, as opposed to just understanding it?
 /mob/proc/can_speak(datum/language/speaking)
-	return (speaking.can_speak_special(src) && (universal_speak || (speaking && speaking.flags & INNATE) || speaking in src.languages))
+//Prevents someone from speaking a null language.
+	if(speaking)
+		return (speaking.can_speak_special(src) && (universal_speak || (speaking && speaking.flags & INNATE) || speaking in src.languages))
```
Zuhayr: I know it's not part of this PR, but you're touching the line anyway.

```(speaking && speaking.flags & INNATE)``` should really be ````(speaking && (speaking.flags & INNATE))```` to avoid && being evaluated before &.

PR #1535, opened 2016-05-02
Fixes two very common runtimes

Neerti:
Fixes some null.Topic() runtimes caused by AI hotkeys.
removeItem() now checks if the item actually exists before modifying variables.


Anewbe:
Fixit fixit fixit fixit!

Datraen:
:+1: waiting on Travis.

PR #1536, opened 2016-05-02
Remaps robotics and mechbay slightly, other adjusts and adjusts cell charge rate

Yoshax:
Picture: https://i.gyazo.com/3acd21782407ccb86af0c056d4132aec.png

- Moves stuff into neat closets
- Adds 2nd cell charger
- Extends lab by 2 tiles length
- Adds medical records console with robotics access
- Adds autolathe
- Puts anesthetic into a cupboard
- Buffs starting steel and glasl amount
- Makes cell chargers use more power to up charge speed


sabiram:
Was it such a pain to go and ask the medbay for a printout?

Yoshax:
It was requested by some folks, so I obliged. It makes sense to me as well.

sabiram:
Why does a mechanic need access to medical records, anyway?

Mark9013100:
I'd rather robotics have hard back copies of the med records. I don't trust them with editable records.

Neerti:
Why do you not trust them with that?

Yoshax:
They're professionals likely with degrees, entrusted just as much as medical to handle live patients. There's no reason they shouldn't have them. They're doctors as much as MDs are in their own way. They deserve their own console to be able to access records for their patients. If they mess with records then they get fired and punished by staff.

Searif:
I am the one who asked for med records, I could honestly care less if its physical or a computer. Not so sure why there is distrust though, how are roboticists less trustworthy then nurses.

Also, as for why mechanics as you put it would need medical records, mostly to check a patients on death instructions as I have had issues with medical just not telling me, other reasons are mostly just RP, checking for allergies and psychological issues.

Mark9013100:
>Why do you not trust them with that? 
Because I can see antag roboticists having a field day setting everyone's records to cyborgify on death. I wouldn't mind if they had read-only versions of the records, ie, the ones in the medical PDA cartridges.

sabiram:
'Roboticists' are mechanics, it's not a dirty word. From a roleplay perspective, I just don't like the idea of having 'my' medical records freely available to someone who isn't a qualified medical doctor with no oversight, no logging of who is looking at them and when, etc.

Searif:
Your mechanic knows brain surgery, how to make and install prosthetic limbs, make AIs? That's impressive.

Sass aside, roboticists deal with a bit more then fixing up cars(mechs) and the like, a portion of their work revolves around repairing prosthesis, which are technically medical devices, on living people that may or may not have medical needs that need to be known about.

And regardless of whether or not they should be performing cyborgifications, they are who medical goes to for it and ultimately their original purpose in SS13.

Yoshax:
Mechanic? Are you shitting me? These so-called mechanics are the ones who repair all the synthetic beings, AIs, Robots, FBPs. They're just as intelligent as doctors, and just as capable in their own way.

sabiram:
It's your own negative interpretation of the word, not mine. I don't appreciate how you're talking to me here, so I'll see myself out.

Serithi:
@sabiram Prosthetics are usually wired up to the nervous and muscular systems at the very least so they can be used in the first place. You think a simple mechanic's going to pull that off? Probably a lot of actual MD's wouldn't know how to do that. It's not just Mechanic Joe slapping on a robotic arm and it magically works by itself, these have to be carefully integrated with the biological body, Roboticists are specifically trained for the job which is _why_ Medical defers to Robotics for prosthetic stuff.

So them being qualified to at least have _access_ to medical records is a no-brainer. Your statement about "oversight" and such can just as easily apply to the MD's, what's stopping _them_ from editing records? An MD can just set people to be borged, kill them, and put the Roboticist to work. Literally the only difference from @Mark9013100 's point is that the MD's not the one then borging people, but it's still the same otherwise. So it's entirely a moot point.

Hubblenaut:
Could you replace the tinted windows with normal ones?
A single tinted window does not work in the way BYOND has its vision algorithm laid out, it's being counted as full block.

Neerti:
Med records aside, on the picture, it looks like the shutters are misaligned, on the bottom-right of the picture.

Mark9013100:
I still think the best solution would be giving the roboticists MED-U cartridges.

PR #1538, opened 2016-05-03
Allows for uniforms to hide shoes with HIDESHOES

Yoshax:


Jakksergal:
Finally I can sprite a proper onezie.

PR #1539, opened 2016-05-03
Adds ability for players to ignore other players.

Neerti:
Players can now choose to not receive messages in OOC and deadchat from specific players, with the Ignore verb, or in the global preferences tab.
Ignores can be reverted with the Unignore verb, or with the global preferences tab as well.
This has no effect on IC chat. LOOC is also exempted from this.


Zuhayr:
This has been needed for a very long time.

code/game/verbs/ignore.dm
```diff
@@ -0,0 +1,42 @@
...
+		prefs.ignored_players -= key_to_unignore
+		prefs.save_preferences()
+		usr << "<span class='notice'>Reverted ignore on <b>[key_to_unignore]</b>.</span>"
+
+/mob/proc/is_key_ignored(var/key_to_check)
+	if(client)
+		return client.is_key_ignored(key_to_check)
+	return 0
+
+/client/proc/is_key_ignored(var/key_to_check)
```
Zuhayr: Should probably not be able to put admins on ignore for simple administrative reasons.

Neerti: the dsay verb doesn't get blocked with this if I did it correctly.

Datraen: Admins still shouldn't be ignorable in OOC, and normal deadchat isn't far from that.

Datraen: As an example of this case being important: An admin tells people to knock off the chat in OOC. One of the people chatting is unable to hear it because they've ignored the admin.
That person gets spoken to by the admin, and then you have to check their ignore list, and hope that they didn't change it from lack of logging.

Neerti:
Admins now shouldn't be blocked by this.

PR #1543, opened 2016-05-04
Manifestcheck

silveryferret:
Adds a simple "Be Cultist" check before spawning in a manifest ghost.


silveryferret:
shit. I totally forgot to rename this. Woops.

Datraen:
Well, this'll definitely be squashed when it's merged.

PR #1544, opened 2016-05-04
Reverts the second mob clean up because of incomplete implementation.

Datraen:
This caused an error that would then cause machinery to hang on the server, and was reproduced successfully.

I don't like hanging machinery.

The reason why everything isn't forceMove is because this was a revert, with forceMove being implemented where conflicts existed and manually setting the loc was previously in place.
Reverts #1243, #1493.

### DIFF
--- a/code/datums/mind.dm
+++ b/code/datums/mind.dm
@@ -391,7 +391,7 @@
 		switch(href_list["common"])
 			if("undress")
 				for(var/obj/item/W in current)
-					current.removeItem(W, force = 1)
+					current.drop_from_inventory(W)
 			if("takeuplink")
 				take_uplink()
 				memory = null//Remove any memory they may have had.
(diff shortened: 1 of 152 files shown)


mwerezak:
I'm curious about the steps taken to reproduce the machinery hang.

Kelenius:
Same.

Datraen:
The steps to reproduce:
1. Pick up laundry basket
2. Drop laundry basket
3. Be concerned about laundry basket disappearing (enters the mob in an unusual way, not in any inventory slots)
4. Go to cryo after being worried about disappearing laundry baskets.

We tried to resolve the laundry basket issue directly, but it was not cooperating when we were adjusting the code last night.

The reason why the steps weren't posted was merely to help prevent griefers who might see this and utilize it before it was merged.

Something that was already alarming was the missed implementation of the new removal code for suit coolers and oxygen tanks in voidsuits, considering that other parts of the dropped proc were changed to use the new removal implementation.

PR #1547, opened 2016-05-04
Ninja - Fixes and Changes

Hubblenaut:
- Improves Emergency Leap to make it a lot less likely to fail.
 - Ninja Rig now spawns with a hyper capacity cell. This is both to make the rig seem more high-tech and prevent players from wasting time trying to get a better cell. (effectively 3x charge now)
 - Teleport cost increased from 500 to 2000 to both accomodate for the increase of charge available and to make it be considered a little less of a way of moving around to avoid too much jackety sax.
 - Charging the suit using the d-sink will now roughly take 40 seconds (from empty to full) regardless of cell size.
 - D-sink will report consumed power in cell units.
 - Fixes failed teleport attempts drawing power.
 - Fixes the suit not being able to charge up to 100%.


Datraen:
:+1: waiting on Travis

PR #1548, opened 2016-05-04
Change chemmaster custom preset to 15 and adds options for 30 and 60 units

Dibasic:
60 is more commonly used for filling a whole bottle. The preset was 30 back when bottles only held 30.

### DIFF
--- a/code/modules/reagents/Chemistry-Machinery.dm
+++ b/code/modules/reagents/Chemistry-Machinery.dm
@@ -21,7 +21,7 @@
 	var/obj/item/weapon/storage/pill_bottle/loaded_pill_bottle = null
 	var/mode = 0
 	var/condi = 0
-	var/useramount = 30 // Last used amount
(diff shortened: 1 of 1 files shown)


Datraen:
I still use 30 at times for things, but it's not that important and it's easily changed.

Dibasic:
Better way would just be to add 30 and 60 buttons since they are commonly used

Datraen:
So why not do that?

Dibasic:
I will do that, should I make a separate PR or add a commit to this one?

Datraen:
Might as well add it to this one, as this would be a moot point otherwise. We can squash commits on Github now, instead of having to have someone do it on their own client.

Dibasic:
did that. added options for 30 and 60 units. set the default to 15 because that's a reasonable value I think.

Datraen:
:+1:

PR #1551, opened 2016-05-05
Makes slime people more interesting and thematic.

Zuhayr:
I had a bit of a fey mood but I'm running out of energy and I won't quite get everything finished. Here goes:

- Breaks species.dm up a little.
- Converts numerous species appearance variables procs.
- Adds a 'shapeshifter' species class which implements a 'wrapper' system for the above appearance getters.
- Adds an argument to several getters to allow for by-mob distinctions.
- Refluffs slime people as 'prometheans'. This doesn't make a huge amount of sense now but it at least feels more interesting. See 'future plans'.
- When irradiated, slime people will glow in the dark.
- Allows slime people to change their bodytype, gender and hair on the fly (using the shapeshifter wrapper class). Valid bodytypes are Unathi, Tajaran, Skrell, Teshari, Monkey and Human.

**TODO**
- [x] Add shock to glomp.
- [x] Test species_restricted etc.

**Things I would have liked to implement but didn't**
- Immune to brute damage and oxyloss.
- Submerging a slime core in water and nutriment will gradually rebuild a body for it. It is otherwise incompatible with other brain stuff like MMIs.
- Has several toggles for various features (having the ability to talk vocally, having hands, not leaving a slime trail) which cost metabolism to use.
- Toxins increase metabolic rate/food consumption.
- Burn damage directly translates to food consumption.
- Starvation directly translates into cloneloss.
- Physical food metabolism rate is quartered.
- Drunk food metabolism rate is quadrupled.

**Future plans:**
- Slime toxin should turn a slime into a promethean and grab a player from ghosts, and then set the colour based on the colour of slime that it originated from (using slime_set_colour()). They should also get some powers from this relating to the colour of slime.
- The nonsolid var on limbs is a terrible hack, should be fixed down the track.
- Golems should be changed into a more complex form of slime, requiring components from adamantine slimes to build. Something like a regular promethean in an adamantine shell, with more Hebrew golem concepts like a directive engraved into their forehead or some other form of engagement other than 'lol ur master is X'.


Zuhayr:
I haven't modified the atmos variables because a. I don't really know how they should differ from humans and b. I can't test Polaris without downgrading to BYOND 509. I'm going to call this 'finished' in the sense that I've tested the mechanics and they all seem to work. It's not like slime people are common on-server in any case.

Zuhayr:
This isn't really a WIP, I'm not going to be changing it again unless there are problems.

PR #1552, opened 2016-05-05
Removes circuit boards from relevant checks for microwaves, cooking.

Datraen:
Also some spacing for ease of reading.
Tested for all operations of microwaves.


PsiOmegaDelta:
Perhaps https://github.com/Baystation12/Baystation12/pull/12579 is something to consider.
A little more centralized if nothing else.

PR #1560, opened 2016-05-05
Adds the ability to feed telecrystals to emags for more uses

Yoshax:
- Adds half default uses per telecrystal use, which is 5.
- Gives you more per telecrystal than buying another emag, to balance utility of having separate emags and spreading them out across different people.


Datraen:
Why not have a specific var for the amount of uses to gain per telecrystal, which is currently 5, rather than having the default_usages/2 every time?

Yoshax:
I feel it's a little clearer in the future if the uses ever gets changed to 20, or something like that. The person can just change both 10s to a 20 rather than forgetting to change the gain. It helps the math remain always the same.

PR #1567, opened 2016-05-06
Makes smoke grenades last longer and adjusts checks

Yoshax:
- Makes smoke linger for longer
- Makes the amount of smoke slightly more/heavier
- Changes boxes of smokes from 5 to 7 grenades
- Changes mapped in single smoke grenades to boxes
- Also removes old smoke grenade code that damages blob because it's dumb


Datraen:
Blob replacement when

PR #1568, opened 2016-05-06
Adds the ability for vending machines to vend 2 items with a 1% chance

Yoshax:
Also  changes it to de-increment the amount in the vending machine after vending the item.

### DIFF
--- a/code/game/machinery/vending.dm
+++ b/code/game/machinery/vending.dm
@@ -518,14 +518,12 @@
 	if (src.icon_vend) //Show the vending animation if needed
 		flick(src.icon_vend,src)
 	spawn(src.vend_delay)
+		new R.product_path(get_turf(src))
+		R.amount--
 		if(prob(1))
-			for(var/I = 1 to 2)
-				new R.product_path(get_turf(src))
-				R.amount--
-			src.visible_message("<span class='notice'>\The [src] clunks as it vends an additional item.</span>")
-		else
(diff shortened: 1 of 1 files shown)


Kelenius:
```
var/spawnItems = 1 + rand(1)
for(var/i = 1 to spawnItems)
~~~magic~~~
```
Less copypasted code.

Zuhayr:
.. why not just
````
	new R.product_path(get_turf(src))
	R.amount--
	if(prob(1))
		new R.product_path(get_turf(src))
		R.amount--
		src.visible_message("<span class='notice'>\The [src] clunks as it vends an additional item.</span>")
````
For loops are not particularly cheap in DM afaik.

**Comment by Datraen** [comment]
:+1:

File: code/game/machinery/vending.dm [hunk 1]
```diff
@@ -509,8 +509,6 @@
 			qdel(coin)
 			categories &= ~CAT_COIN
 
-	R.amount--
```
Zuhayr: May be worth flagging the machine as busy and stopping it accepting additional orders during the vending spawn() since it might be possible to cause it to vend below zero by spamming the interface during the delay. Either that or just put the first amount-- here and check if it's above zero before vending the second item.

mwerezak:
Aside: I see you guys are still using vendors that allow you to add half-eaten snacks and get full ones back out.

PR #1575, opened 2016-05-06
----- HOLD MAP CHANGES ----- NO MAP CHANGES ABOVE THIS LINE -----

EmperorJon:
A large map change will be incoming. Due to map merging issues, we're not going to accept map changes for some time.


Yoshax:
What kinda change?

EmperorJon:
A reconfiguration of the civilian sector.

Yoshax:
Are currently open map PRs going to be merged first, or are they going to have to wait until this is done?

And does it only affect map1?
And what about any bugs that around found on the map, are they just not allowed to be fixed?

EmperorJon:
We are going to attempt to process map changes that are currently open first.
It should only by the main z-level, yes.
As for bugs, we don't know of any large outstanding map issues and hopefully no large issues will found. If they are they'll be dealt with as we see fit.

Yoshax:
How long is 'some time'?

PR #1577, opened 2016-05-06
Changes all references to 'Mars University' from 'Mars University' to 'Mojave University'

redstryker:
Does exactly as the title says.


redstryker:
As a note for this change: http://ss13polaris.com/blog/mars-university-name-change/

PR #1578, opened 2016-05-06
Skirt addition and loadout refactor

SinTwo:
Adds skirts by JakkSergal, cleans up skirts, simplify skirts on loadout.

mob


Datraen:
1) Pictures
2) Maybe it'd be better to keep the job-specific skirts out of the general skirt, maybe consolidate job clothing into selections for job checks?

PR #1581, opened 2016-05-07
Adds PDA Multicaster

Neerti:
Adds a new machine to telecomms, which contains special PDAs corresponding to each department on the station.  Sending a PDA message to one of these PDAs will duplicate it to everyone in the department the PDA is assigned to.  It determines what department someone is in by cartridge type.
Adds cartridges for several jobs that were missing them, for the purposes of being recognized as being in a department.  The cartridges otherwise do nothing.
New circuitboard to build the new pda multicaster.  It can also be deconstructed with the standard machinery deconstruction steps of screwdriver > crowbar.


Neerti:
Finally got Travis to stop hating me.  Issue was that my comments weren't ended with a ,

PR #1589, opened 2016-05-08
Lowers Human Minimum Age To 16

ParadoxSpace:
Exactly what it says on the tin.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/station.dm
+++ b/code/modules/mob/living/carbon/human/species/station/station.dm
@@ -11,7 +11,7 @@
 	num_alternate_languages = 2
 	secondary_langs = list("Sol Common")
 	name_language = null // Use the first-name last-name generator rather than a language scrambler
-	min_age = 17
+	min_age = 16
 	max_age = 110
 
 	spawn_flags = CAN_JOIN


OhEmGee:
we really should have a public vote about this

SpadesNeil:
Gotta ask. Why would a 16-year-old be on a highly dangerous orbital research station? Not exactly your mom-and-pop pizza parlor. I mean, IRL, there's laws against these kind of work conditions, but I know that we're in like the 26th century on the outer fringes of human inhabited space and NanoTrasen probably doesn't care but...

... I don't know, this seems like a very bad idea that only serves to allow 16-year-old players to have self-insert characters of an accurate age.

Zuhayr:
>Why would a 16-year-old be on a highly dangerous orbital research station?
Polaris is set on a colony.

SpadesNeil:
Sorry, I'm an idiot. Disregard my opinion.

SpadesNeil:
Wait, I thought it was an asteroid. Therefore, in orbit. Or am I really that out of the loop? If so, disregard my opinion.

ParadoxSpace:
yeah see i figured that 16 was the age where like
'hello congrats you can finally visit the work level'

SpadesNeil:
>yeah see i figured that 16 was the age where like
>'hello congrats you can finally visit the work level

Hm, I suppose that makes sense. Very well.

Verarch:
I'd be for it as long as there's restrictions on jobs for 16 year olds. I doesn't make sense for a 16yo to be a skilled medical doctor, or a security officer trusted with potentially lethal weaponry if they haven't even graduated school.

SpadesNeil:
I'd agree, but I also take it to an extreme when I say 16-year-olds should be hard coded to be restricted to assistant only.

EmperorJon:
I have word that this isn't going to be happening.

PR #1593, opened 2016-05-08
New Player/Character Setup (Fake)NanoUI

Arokha:
Used the fake NanoUI 'browser' datum to render the pages for the New Player panel and the Character Setup screen.

While I consider Polaris' character setup screen to be mechanically better than others, it wasn't prettier. Well. It was a bit. But now it's more pretty-er. Did this since we're switching to your codebase, so, this is the first upstream PR.

@Datraen mentioned this might be desired here.

Fake Nano Setup

### DIFF
--- a/code/modules/client/preference_setup/general/01_basic.dm
+++ b/code/modules/client/preference_setup/general/01_basic.dm
@@ -43,8 +43,8 @@ datum/preferences/proc/set_biological_gender(var/gender)
 /datum/category_item/player_setup_item/general/basic/content()
 	. = "<b>Name:</b> "
 	. += "<a href='?src=\ref[src];rename=1'><b>[pref.real_name]</b></a><br>"
-	. += "(<a href='?src=\ref[src];random_name=1'>Random Name</A>) "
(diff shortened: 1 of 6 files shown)


Datraen:
:+1: It's better than white background and black text.

EmperorJon:
It certainly is an improvement, that's for sure.

Spacemanspark:
When you first log into the server, and click the character setup button, does it take three minutes to open?
Paradise had and fixed that issue for awhile.

Arokha:
After the server is done starting (as in, done initializing geometry and ZAS) the button works just as fast as the black/white character setup screen. I don’t notice any difference in speed. This isn’t really even NanoUI, just the fake CSS styles that make it look like it, so if they had some shortfall in NanoUI, this would be unlikely to have the same issue.

Arokha:
Also I should mention the colored text for the toggle between "Join as assistant if pref unavailable/Return to lobby if pref unavailable/Join as random if pref unavailable" on the occupations screen is now just white text for all three choices. Sorry. There's just no colors that look good on NanoUI buttons.

Spacemanspark:
Gotcha, thanks.

PR #1594, opened 2016-05-08
Fixes broken logs

Kelenius:


Kelenius:
Travis is wrong.

Kelenius:
That should do it.

code/modules/mob/new_player/new_player.dm
```diff
@@ -501,7 +501,7 @@
 /mob/new_player/hear_say(var/message, var/verb = "says", var/datum/language/language = null, var/alt_name = "",var/italics = 0, var/mob/speaker = null)
 	return
 
-/mob/new_player/hear_radio(var/message, var/verb="says", var/datum/language/language=null, var/part_a, var/part_b, var/mob/speaker = null, var/hard_to_hear = 0)
+/mob/new_player/hear_radio()
```
mwerezak: Are you sure this won't runtime if calls are made to any new_players with optional arguments set?

Kelenius: Yes, as long as there are no named arguments.

mwerezak: > if calls are made to any new_players with optional arguments set?

Kelenius: ``hear_radio(thing, differentThing, anotherThing, oneMoreThing)``
is fine
``hear_radio(thing = someThing)``
is not

The code currently only uses calls without named arguments. The first form. So yes, it won't runtime.

mwerezak: Alright. I guess it won't be much of a problem to just re-add all of these again when someone inevitably uses some named arguments for this 7-argument proc. Hopefully.

Kelenius: If someone adds code that uses named arguments with this proc, it will be their problem to make sure THEIR code does not runtime.

Datraen: I honestly see no real reason to remove it right now, instead of just changing the mob to part_c.

Kelenius:
Do I need to start posting WGW here or what?

PR #1607, opened 2016-05-10
Hacktool

Hubblenaut:
Ports hacktool from Baystation12 and adds it to uplink.


code/game/objects/items/devices/hacktool.dm
```diff
@@ -0,0 +1,100 @@
...
+
+	var/in_hack_mode = 0
+	var/list/known_targets
+	var/list/supported_types
+	var/datum/topic_state/default/must_hack/hack_state
+
+/obj/item/device/multitool/hacktool/New()
+	..()
+	known_targets = list()
+	max_known_targets = 5 + rand(1,3)
```
Kelenius: ``rand(6, 8)``

Kelenius:
>costs two emags or six toolboxes
>only works on doors
>takes time to hack
>does the same as a toolbelt, yellow gloves, and some preparation
>toolbox that costs 1 tc contains gloves and all toolbelt items

Hubblenaut:
You're implying everyone would ICly know how to use a toolbox and forgetting that the hacktool is disguised as a multitool and inheriting its functionality.

code/game/objects/items/devices/hacktool.dm
Hubblenaut: And would you elaborate why?

PsiOmegaDelta:
I should quickly note that there is (currently) no remote connection, you have to be adjacent to the door.
What it does do with stored airlocks is allow you to access the AI door controls without spending the time to hack.

code/game/objects/items/devices/hacktool.dm
Zuhayr: It's literally an identical value, and removes the overhead of an addition operation.

PsiOmegaDelta: I'd be concerned if the overhead of an addition operation in a New() proc had any effect in the short or long term.
The purpose the statement was written as it was, was to clarify that 5 was the intended base amount and the rand() is just a bonus.

Kelenius: It's one less operation and it makes the code clearer. It won't have a noticeable effect on the speed of the game, but it's just. There's no advantage to writing it like that.

Kelenius:
Just because it's complete, unchanged port doesn't mean it needs to be accepted without discussion. It would cost HALF of your crystals just so you can open doors.

PsiOmegaDelta:
Ah, you're used the Bay-cost. We more than doubled the telecrystal amount standard traitors begin with with various price adjustments to go with it, meaning it's indeed more than likely costing much more than it should on Polaris.

Kelenius:
So it costs about the same as emag? Yeah, that makes more sense.

code/game/objects/items/devices/hacktool.dm
Hubblenaut: The advantage is to make it clear what is wanted. It's quite silly to have a discussion about this.

Hubblenaut:
Will edit it to use half.

PR #1616, opened 2016-05-11
AI Language tweaks

Anewbe:
Adds Siik'Maas and Schechi to AI language list


Neerti:
I'm not sure if the Taj language you're adding is still used in lore, but I could be wrong.

Anewbe:
It's still in the code in some form or other, is why I added it.

Zuhayr:
You should be using macros for this. See https://github.com/PolarisSS13/Polaris/blob/4456f77d751d3f19bb67c04421a8a602191fad9c/code/__defines/species_languages.dm

Additionally Siik'Maas is the wrong language. Taj speak Tajr.

Nerezza:
It's true, however our lore page still claims Siik'Maas is the main language and Siik'Tajr is the special kitty sign language. Idk why we didn't make the spoken language, Siik'Maas, into the language used in Polaris.

PR #1628, opened 2016-05-12
Prevent Unexpected SQL Errors

Arokha:
Even when you have SQL disabled, it prints errors to the console about not being able to connect.

99% of people starting a Polaris server are devs who won't have SQL set up.

Also updates the example config to be less misleading, as the SQL option is 1 or 0, not comment/uncomment. The old default was 'enabled' due to the configuration.dm having a default of 1, so that is still the same. Devs will need to set it to 0 to squelch the message.

If you want, change it to 0 in the example config file, but I didn't want to mess with your defaults.

Another from downstream.

### DIFF
--- a/code/world.dm
+++ b/code/world.dm
@@ -558,8 +558,9 @@ var/failed_db_connections = 0
 var/failed_old_db_connections = 0
 
 /hook/startup/proc/connectDB()
-	if(!config.sql_enabled) return 1
-	if(!setup_database_connection())
+	if(!config.sql_enabled)
+		world.log << "SQL connection disabled in config."
+	else if(!setup_database_connection())
 		world.log << "Your server failed to establish a connection with the feedback database."
 	else
 		world.log << "Feedback database connection established."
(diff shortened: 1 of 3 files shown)


Spookerton:
To expand on the root of this, we do hope at some point to tackle rewriting the old storage stuff to make use of byond's native sqlite capabilities, which should result in it becoming a non-issue as we will simply generate a new database file with correct tables if one isn't available.

As is, though, I'm hesitant to :+1: a silent exit for a config switch that affects server behavior and the assumptions of a number of other options and game features fairly significantly. It might be better to include an info line like `SQL connections disabled in config, skipping connection setup` - any message about database availability is preferable to silently failing in the case of incorrect config options, which you correctly point out are parsed poorly in some cases and ship with a bad template file; something I imagine has been inherited several generations of forks over by this point :b

Arokha:
Using sqllite will be nice. If we work on that I'll port what we do, if you want it.

Meanwhile, I've updated it to not be silent, and just made the configuration.dm setting more in-line with everything else. No other 1/0 options in the config file, so now it's just comment/uncomment (default uncommented to keep the same default that's been inherited, like you said, by about everyone).

code/world.dm
```diff
@@ -600,6 +602,8 @@ proc/establish_db_connection()
 
 
 /hook/startup/proc/connectOldDB()
+	if(!config.sql_enabled)
+		world.log << "SQL connection disabled in config."
 	if(!setup_old_database_connection())
```
Spookerton: That thur's going to need to be elif.

Spookerton:
**Once the line note I popped in there is dealt with**, this looks fine to me. :+1:

Arokha:
Corrected. I had added it to the message above but not that one.

PR #1631, opened 2016-05-12
Changelings will now store and use the flavor text of their victims

Yoshax:
When a victim is absorbed, their flavour text will be stored, and then applied when the changeling transforms into that person.

In addition, the gender identity will now also correctly be stored and applied.


Datraen:
:+1:

PR #1632, opened 2016-05-12
[DNM] Adds fingerless variants of gloves

Mark9013100:
* Adds fingerless variants of all colored gloves. Available in custom loadout.

* Moves white gloves to the normal color gloves.dm.

* Spaces a line in misc gloves.dm.

Needs to be retested.


Yoshax:
You should be using the selection for that. Normal gloves really should as well, but nobody's done it yet. From a cursory examination it doesn't appear that they're wearable by species who can't wear gloves, which they probably should be, considering they're fingerless.

Mark9013100:
Crap, forgot about species.

Anewbe:
The species part is easy, just change the path of whatever your "default" fingerless glove is so it's **clothing/gloves/fingerless** then put **species_restricted = null** in it... Other than latex, black, and yellow, pretty much all of the regular gloves have the same stats, so you'll only have to put in code bits under those few that are different.
Fingerprints I dunno, but I can find out if you want.

Mark9013100:
Probably going to go with Anewbe's solution for now. I'll work on this some more later.

Datraen:
We can do **species_restricted = list()** instead of null, to just avoid any kind of runtimes that may or may not occur.

Mark9013100:
Grand, thanks Datraen.

Mark9013100:
Just need to get the fingerprints part done and this'll be (hopefully)clear for merging.

Yoshax:
You might still want to use the geartweaks to put the gloves into a selection, where you pick your selection from a pop-up box like the skirts. It's a little neater code-wise, but would require a bit more work.

code/modules/clothing/gloves/color.dm
```diff
@@ -84,3 +84,92 @@
...
+	item_state = "lightbrowngloves_alt"
+
+/obj/item/clothing/gloves/fingerless/brown
+	name = "fingerless brown gloves"
+	desc = "A pair of gloves, they don't look special in any way."
+	icon_state = "brown_alt"
+	item_state = "browngloves_alt"
+
+/obj/item/clothing/gloves/fingerless/black
+	desc = "These work gloves are thick and fire-resistant."
```
Iamgoofball: why would they still be fire resistant after being made fingerless

Mark9013100: Good point. Probably also going to knock them down a point in custom loadout.

Mark9013100:
Won't be able to finish this for awhile, closing.

PR #1633, opened 2016-05-12
Gives Diona pay the same as Taj and Unathi

Yoshax:
I was playing a Diona chef, and I couldn't even buy ramen to make hell ramen. What is this.

### DIFF
--- a/code/modules/economy/economy_misc.dm
+++ b/code/modules/economy/economy_misc.dm
@@ -48,11 +48,12 @@
 
 
 /var/list/economic_species_modifier = list(
-												/datum/species/human	= 10,
-												/datum/species/skrell	= 12,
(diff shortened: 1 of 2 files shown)


ParadoxSpace:
why do dionaea need as much money

Yoshax:
Because slavery is bad. And because none of those numbers actually matter.

ParadoxSpace:
it's not slavery my man
i'm pretty sure dionaea lorewise don't need to be paid all that much from a corporate standpoint

Yoshax:
The numbers are worthless lorewise. They don't mean anything, neither does the amount of money anyone starts with. Just having no money is limiting in-game when you want to do certain things.

PR #1641, opened 2016-05-13
Fixes not being able to stop wounds from bleeding

Yoshax:
Ports Harpy's elegant solution instead of the hackfix I used.
Ports: https://github.com/Baystation12/Baystation12/commit/b83b977ee4d26de30e0715a529387464feb7f7fb


Datraen:
:+1:

PR #1645, opened 2016-05-13
Uplink changes and additions

Yoshax:
- Uplink categories are now split into their separate files
- Mercs now get an uplink each

Adds a new feature, this is a weighted random discount on uplink items for those with an uplink. This is weighted in a way that there is a high chance of a low discount, and a low chance of a high discount. This picks a random item every 15 minutes, and offers a discount on that item to the person. This is random per uplink. An item will always cost a minimum of 1.


Yoshax:
To do is done. Ready for review, I believe.

PR #1653, opened 2016-05-14
Fixes Loyalty Implants

Arokha:
Makes loyalty implants correctly 'implant' on spawn for jobs that get that, if the server config is enabled. Was completely broken for jobs that would have this, as the datum specified that the job datum itself would be passed for implanting.

Why would you define a proc on /mob/living/carbon/human for implanting something then ask to be passed your own src? Also why would you try to pass /job/datums as something to implant? Silly.

FYI this has been broken forever in Bay. At least a year. It was broken when we ported last June to the Bay at the time.

### DIFF
--- a/code/datums/mind.dm
+++ b/code/datums/mind.dm
@@ -344,7 +344,7 @@
 				log_admin("[key_name_admin(usr)] has de-loyalty implanted [current].")
 			if("add")
 				H << "<span class='danger'><font size =3>You somehow have become the recepient of a loyalty transplant, and it just activated!</font></span>"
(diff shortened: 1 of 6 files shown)


Searif:
To be fair it was broken on bay and polaris simply because we never used them ourselves.

Arokha:
Fair enough. But. It's tiiiime. Been broken for a year and it was a simple fix.

And we do use them on ours, so thought we'd share.

Searif:
Yeah I didn't have an issue with it just thought I'd let you know why its been broken.

PR #1658, opened 2016-05-15
Fix the letter ÿ in nano-ui displays

Leshana:
* Particularly noticable on the cartridge button of the PDA, any time we transmit a name containing \improper or \proper using nanoui it shows up as ÿ
* This is becuase BYOND internally handles these escapes by simply using character 0xFF etc.
* We'd better strip them out before sending them as JSON data so they aren't rendered as ÿ when treated as utf-8 encoding.

### DIFF
--- a/code/_helpers/text.dm
+++ b/code/_helpers/text.dm
@@ -324,4 +324,11 @@ proc/TextPreview(var/string,var/len=40)
 				return 1
 	return 0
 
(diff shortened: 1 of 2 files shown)


Spookerton:
latin-1 is a dirty word. :c

Arokha:
This would be nice

PR #1664, opened 2016-05-16
Changes Death Alarm

silveryferret:
Changes the Death Alarms from announcing to general comms, to just
Security and Medical. Because sometimes hearing your best friend has
died over Common channel is not the best scenario to hear that someone's
died.

Used the captain's headset to keep things slightly cleaner, and it gets
deleted after as far as I can tell.


Datraen:
Honestly, I like this. Doesn't make sense to announce it to an entire crew, where it may cause panic.
:+1:

Mark9013100:
Can it also be announced over the response team radio as well?

silveryferret:
Pretty easily done, yeah. But generally speaking if you've got response team channel you've probably got all the channels anyway

Arokha:
This is neat, makes sense. No sense in generating gossip on common since I've never seen a DA fire and then not a whole wall of peanut-gallery people start blabbing about it.

N8-Toe:
I think it should be over  command radio as well

PR #1674, opened 2016-05-16
Revert to old_x and old_y after attack animations

Arokha:
Does not use initial(pixel_x) and initial(pixel_y) for 'finishing' attack animations, so that mobs can have old_x and old_y updated to reflect a natural pixel offset (e.g. for mobs that are longer/taller than 32x32 and are nudged to fit into a tile).

For example, we have a 32x64 cyborg module which sets pixel_x, pixel_y, old_x, and old_y when selected, however attacking with it resets it to 0 as this doesn't respect old_x and old_y like all other mob anims do (see floating, etc).

Also why define this on atom and then literally never use it on anything other than a mob, in the... mob/animations.dm file?

_Sorry to keep making QOL PRs. I promise I'll send something cool eventually._
_Edit: Actually your Teshari benefit from my code to sprite holders months ago so I did, kinda._

### DIFF
--- a/code/modules/mob/animations.dm
+++ b/code/modules/mob/animations.dm
@@ -140,11 +140,11 @@ note dizziness decrements automatically in the mob's Life() proc.
 	//reset the pixel offsets to zero
 	is_floating = 0
 
-/atom/movable/proc/do_attack_animation(atom/A)
+/mob/proc/do_attack_animation(mob/M)
 
 	var/pixel_x_diff = 0
 	var/pixel_y_diff = 0
-	var/direction = get_dir(src, A)
+	var/direction = get_dir(src, M)
 	switch(direction)
 		if(NORTH)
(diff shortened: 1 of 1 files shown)


Mechoid:
Out of curiosity regarding this as a non-dev, does this mean if someone edits a mobs variable offset to move a sprite, when it attacks, will it revert back to before the variable edit, or back to where the admin set it?

Arokha:
That is both the current behavior and the behavior after this, as you've written your question. Just editing pixel_x and pixel_y will make the mob go back to their original offset after playing an animation.

**BUT** if said person varedits the mob's old_x and old_y variables as well, it will prevent that.

Basically all the other animations like floating (in zero-g) returned to 'old_x' and 'old_y' when done floating, but the attack animations did not.

Mechoid:
Ah, I see. Good to know!

PR #1678, opened 2016-05-17
Disables alternative title validation to allow for admins to give people/themselves retainable non-standard titles

Yoshax:
@Spookerton said he would at least like to have the ability


Spookerton:
:+1:

PR #1687, opened 2016-05-18
Fixes beating people up during surgery

### LINKED ISSUE #1684: Surgery on help intent hits people + other issues
If you're doing surgery, and you attempt to use an item that isn't appropriate, such as a hemostat when you can't use it, you hit the person when you're on help intent.
To hit the person.
#### What actually happened
You hit the person.
#### Steps to reproduce
Do surgery and attempt to use the wrong tool. See that you hit the person.
#### Additional info:
- The return here is likely the cause: https://github.com/PolarisSS13/Polaris/blob/8f54e7778fd377296a9c895295d7977e6ce5a7eb/code/modules/surgery/surgery.dm#L113

Yoshax:
Fixes #1684 

My description in the issue was wrong, it's missing the return 1 that was causing issues. This will however make it so you can't use items on someone who on the table while on help intent, such as health analysers or such. Such is life.

### DIFF
--- a/code/modules/surgery/surgery.dm
+++ b/code/modules/surgery/surgery.dm
@@ -110,6 +110,7 @@ proc/do_surgery(mob/living/carbon/M, mob/living/user, obj/item/tool)
 
 	if (user.a_intent == I_HELP)
 		user << "<span class='warning'>You can't see any useful way to use [tool] on [M].</span>"
+		return 1
 	return 0
 
 proc/sort_surgeries()


EmperorJon:
Not sure I'm happy with items not being usable on people on tables, especially if that includes things like injecting.

Yoshax:
So, uh... I was going by the words of the PR that introduced it, but it appears to be wrong.

You can still use syringes, and other items, it just produces a:
`You can't see any useful way to use Syringe (inaprovaline) on Lennox Hegarty.`

So! All is good.

EmperorJon:
That message has always bugged me. At some point we might want to alter it to clarify it's specifically surgical usages.

Yoshax:
Yeh, would be a good idea. Or better the whole thing so it doesn't appear when you're trying to inject people or such.

Changing the message to something like, "You can't see any useful way to perform with with \the [src] on [affected]."

But that's maybe for another day.

Issue #1683, opened 2016-05-18
C4 is not guaranteed to destroy a wall


Yoshax:
#### Brief description of the issue
C4 is not guaranteed to destroy a wall
#### What you expected to happen
For C4 to always destroy the wall and girder it's planted it.
#### What actually happened
The C4 can explode and only destroy the covering of the wall, and not the girder.
#### Steps to reproduce
Plant C4 on the wall, wait and possibly see if only the material covering is destroyed.

Arokha:
Fixed by #1688, can be closed.

PR #1688, opened 2016-05-18
Makes plastique better

Arokha:
It destroys the wall it's placed on, rather than slightly somewhat dismantling it a little and requiring you to have a toolbelt to finish the job.

@Yoshax @Yoshax @Yoshax 8D


Yoshax:
Fixes #1683

PR #1690, opened 2016-05-19
Fixes runtimes with suits

Arokha:
Naked and trying to wear a suit, it will attempt to access w_uniform.inv_flags which will be null and runtime. Things like putting on a rig and trying to deploy them would do this.

Moves shoe redrawing to uniform and suit update_inv code in update_icons, rather than in inventory.dm. update_icons already had code for this as the top line in update_inv_shoes checked if it should be hidden.


Datraen:
:+1:

PR #1694, opened 2016-05-19
Buffs forensics investigation equipment crate

Mark9013100:
* Removes the "Auxiliary forensic tools" crate.

* Adds a Crime Scene Investigation kit to the 'Forensics investigation equipment' crate.

* Increases the price of the 'Forensics investigation equipment' crate to 40 points, from 35.


Spookerton:
:+1:

PR #1696, opened 2016-05-19
Cable coil adjustments and changes

Yoshax:
- Adjusts and adds more colors of cables
- Removes limitation of being unable to join different colors of cable
- Allows merging of different kinds of cable
- Adjusts the color changing proc to be nicer as per Baystation12/Baystation#11645
  - Does not add the fancy color changing via multitool due to lack of prerequisite


Datraen:
:+1:

PR #1700, opened 2016-05-19
[DNM] Communicator tweak

Arokha:
Communicators ain't be wavin' 'n shit. Use 'hearers' instead of 'viewers'.

Makes it so communicators can be heard while inside other ~~people~~ objects. So if you're hiding inside a ~~tajaran~~ locker you can hear the communicator in the same room as you.

:wink: **COUGH** :sweat_smile:


Arokha:
Going to make it show the face of the last person to talk on the communicator because Datraen says they do video.

Arokha:
Nope. Going to do this better.

PR #1711, opened 2016-05-21
Fixes never being able to light molotovs and makes throwing them less rubbish

Yoshax:
Did you even know you could make molotovs?


Yoshax:
So yeh, apparently it does work without the src and just wasn't for some random season when I was testing it. But like, it doesn't hard and I kinda can't be bothered removing it. But the chance table change is valid and actually does something.

PR #1716, opened 2016-05-21
Warden Jacket protects legs

Anewbe:
HoS coat does it, and the Warden coats reach low enough to have this make sense.


Spookerton:
Fair enough. :+1:

PR #1719, opened 2016-05-22
Pathing Tweaks

Anewbe:
Toeless jackboots are now jackboots/toeless instead of jackboots/unathi
Breachers are now rig/breacher instead of rig/unathi


code/modules/clothing/spacesuits/rig/suits/alien.dm
```diff
@@ -1,4 +1,4 @@
-/obj/item/weapon/rig/unathi
+/obj/item/weapon/rig/breacher
 	name = "NT breacher chassis control module"
```
Zuhayr: Worth making this \improper while you're modifying the file, maybe?

Anewbe:
Good catch, thank you.

Spookerton:
Includes mapping changes re: #1575 
Will need updating before merge when 1575 is resolved. Otherwise looks fine.

PR #1721, opened 2016-05-22
Tweaks whisper to work for held mobs

Arokha:
Like Teshari, when scooped. This is how Say already works, I am not sure why Whisper was written differently. Probably by a different person or at a different time.

Also removed a question-comment, because "yes".

### DIFF
--- a/code/modules/mob/living/carbon/human/whisper.dm
+++ b/code/modules/mob/living/carbon/human/whisper.dm
@@ -109,24 +109,18 @@
 			message = replacetext(message, "u", "µ")
 			message = replacetext(message, "b", "ß")
 
-	var/list/listening = get_mobs_or_objects_in_view(message_range, src) //VOREStation Edit - GMOOIV proc gets people in preds/objects(holders)
+	var/list/listening = get_mobs_or_objects_in_view(message_range, src)
 	listening |= src
 
 	//ghosts
-	for (var/mob/M in dead_mob_list)	//does this include players who joined as observers as well?
+	for (var/mob/M in dead_mob_list)
 		if (!(M.client))
 			continue
 		if(M.stat == DEAD && M.is_preference_enabled(/datum/client_preference/ghost_ears))
 			listening |= M
-	/* VOREStation Removal - Unnecessary with GMOOIV proc above as it does this.
-	//Pass whispers on to anything inside the immediate listeners.
-	for(var/mob/L in listening)
-		for(var/mob/C in L.contents)
-			if(istype(C,/mob/living))
-				listening += C
-	*/
+
 	//pass on the message to objects that can hear us.
-	for (var/obj/O in listening) //VOREStation Edit - Already got a list above.
+	for (var/obj/O in listening)
 		spawn (0)
 			if (O)
 				O.hear_talk(src, message, verb, speaking)
--- a/code/modules/mob/living/carbon/human/whisper.dm
+++ b/code/modules/mob/living/carbon/human/whisper.dm
@@ -109,7 +109,7 @@
 			message = replacetext(message, "u", "µ")
 			message = replacetext(message, "b", "ß")
 
-	var/list/listening = hearers(message_range, src)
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/whisper.dm
```diff
@@ -109,24 +109,18 @@
...
 
 	//ghosts
-	for (var/mob/M in dead_mob_list)	//does this include players who joined as observers as well?
+	for (var/mob/M in dead_mob_list)
 		if (!(M.client))
 			continue
 		if(M.stat == DEAD && M.is_preference_enabled(/datum/client_preference/ghost_ears))
 			listening |= M
 
-	//Pass whispers on to anything inside the immediate listeners.
```
Arokha: This was removed because it was redundant with using GMOOIV proc, as that proc also returns objects (unless you ask it not to).

code/modules/mob/living/carbon/human/whisper.dm
```diff
@@ -109,24 +109,18 @@
...
-	for(var/mob/L in listening)
-		for(var/mob/C in L.contents)
-			if(istype(C,/mob/living))
-				listening += C
-
 	//pass on the message to objects that can hear us.
-	for (var/obj/O in view(message_range, src))
```
Arokha: Another unnecessary call to view, as we've already gotten a list of objects in listeners, as GMOOIV returns objects and mobs both.

code/modules/mob/living/carbon/human/whisper.dm
```diff
@@ -109,24 +109,18 @@
...
-	for (var/mob/M in dead_mob_list)	//does this include players who joined as observers as well?
+	for (var/mob/M in dead_mob_list)
 		if (!(M.client))
 			continue
 		if(M.stat == DEAD && M.is_preference_enabled(/datum/client_preference/ghost_ears))
 			listening |= M
 
-	//Pass whispers on to anything inside the immediate listeners.
-	for(var/mob/L in listening)
-		for(var/mob/C in L.contents)
```
Arokha: This was removed because it was redundant with using GMOOIV proc, as that proc also returns objects in objects (unless you ask it not to).

code/modules/mob/living/carbon/human/whisper.dm
```diff
@@ -109,24 +109,18 @@
 			message = replacetext(message, "u", "�")
 			message = replacetext(message, "b", "�")
 
-	var/list/listening = hearers(message_range, src)
+	var/list/listening = get_mobs_or_objects_in_view(message_range, src)
```
mwerezak: Lovely.

mwerezak: You do realize though that this performs a recursive search into the contents tree of every mob or object in view though, right?

Arokha: Yes, that is how Say currently works. I realize it's expensive, but without digging, there's not really a particular way to do it since scooped Teshari are mobs in an object in another object or a mob.

Considering that this his how Say already works, and Say iterates over a much larger range, I doubt you would see any performance impact moving Whisper to the same thing.

Ccomp5950: That's how the "broken" version of say works which for some reason Polaris is still using.  And yes the performance impact is drastic.  Recursing through every mob and every object looking for other objects and mobs to recurse in is much slower then just checking turfs and the player list.

https://github.com/Baystation12/Baystation12/blob/dev/code/modules/mob/living/say.dm#L245-L265

nachomeep:
whisper for synths when

code/modules/mob/living/carbon/human/whisper.dm
Arokha: I'm cloning Bay, so give me a minute to read through it in a less ugly interface than GitHub. I may be back with questions. Or if it's super-awesome I may just close this and redo it with your suggestion.

Arokha:
Somebody shoot me because I'm rewriting say.

PR #1722, opened 2016-05-22
Thunderdome and Heavy armor actually have armor values now.

SpadesNeil:
- Thunderdome armor has armor comparable to "heavy armor vest" used by mercs.
- Heavy vest is super robust against melee and projectile damage, but totally useless against energy damage. Also it makes you slow as shit, so, it's balanced because you're a stupidly easy target.

Why did I give Thunderdome armor any armor? It helps prolong fights, making them more exciting. Thunderdome events are fun, both PVP and PVE, and I've had a lot of luck in the roleplay aftermath of such events. I don't know if Polaris will ever use this, but I don't see the harm in adding it in the event you ever find a use.


SpadesNeil:
Whoops, I did this wrong. Disregard.

PJB3005:
You should make a new branch with git got every change you make to prevent this. Ask in IRC for help.

PR #1724, opened 2016-05-22
Removes 2 spells from wiz until we get NewWiz (*COUGH NEERTI COUGH*)

Anewbe:
Removes Horsehead and Mindswap from Wizard spellbooks


Mark9013100:
horsehead did nothing wrong

PR #1726, opened 2016-05-23
Adds a framework to allow guns to be locked to dna

Yoshax:
This adds a framework that allows guns to be locked to dna.

- The DNA is stored as a list, and as such, can have a number of allowed users.
- The first person to donate DNA is the primary controller. The primary controller controls whether or not the gun will allow more users to give their DNA.
- If the primary controller removes their DNA, the next person to add their DNA is the new primary controller.
- DNA locking has 2 security levels. At the first level, the gun merely refuses to be shot. At the second level, the gun will self-destruct causing a minor explosion after 10 seconds.

This is not applied to any existing guns. This only adds the ability to create future guns with this ability.


Anewbe:
I see many issues arising from this.

1. What happens if the first person to DNA lock a gun doesn't allow more locks, and goes SSD or dies as a DNC? That gun's now bricked.

2. What happens if the first person to DNA lock a gun doesn't allow more locks, and goes horrifically insane/traitory/whatever? That gun's now bricked.

3. I really hate the idea of a self-destruct. What happens if person A (DNA locker) sets the thing to go bang, then goes SSD, dies as DNC, or antags? That gun's a time bomb. Also the whole concept of a firearm self-destruct sounds like a cheesy spy movie gimmick to me... A zap or something, maybe, but the gun going boom? Just silly and annoying.


Also why DNA lock, out of curiosity, and not ID? ID lock would negate 2/3 of my questions flat out.

Yoshax:
DNA lock is more future fun and allows it to be locked to singular users rather than departments etc.

1. Send it back to the factory to get factory reset.
2. Send it back to the factory to get factory reset.
3. It explodes because I wanted it to explode.

Emag functionality to negate locks or suchlike can be added. I just haven't.

code/modules/projectiles/gun.dm
```diff
@@ -114,6 +125,19 @@
...
 
 	var/mob/living/M = user
+	if(dna_lock && stored_dna)
+		if(!authorized_user(user))
+			if(safety_level == 0)
+				M << "<span class='danger'>\The [src] buzzes in dissapoint and displays an invalid DNA symbol.</span>"
+				return 0
+			if(safety_level == 1)
+				M << "<span class='danger'>\The [src] hisses in dissapointment.</span>"
+				for(var/mob/O in hearers(src, null))
```
Zuhayr: Should be visible_message().

code/modules/projectiles/gun.dm
```diff
@@ -426,4 +450,76 @@
...
 /obj/item/weapon/gun/attack_self(mob/user)
-	switch_firemodes(user)
\ No newline at end of file
+	switch_firemodes(user)
+
+/obj/item/weapon/gun/proc/get_dna(mob/user)
+	var/mob/living/M = user
+	if(!controller_lock)
+
+		if(stored_dna && !M.dna in stored_dna)
```
Zuhayr: Should be !(M.dna in stored_dna).

Anewbe:
I don't want to send my ion rifles back to the factory because some Warden DNA locked them both, then went home. That's annoying.

code/modules/projectiles/gun.dm
```diff
@@ -426,4 +450,76 @@
...
+			controller_lock = 1
+			M << "<span class='notice'>\The [src] beeps and displays a locked symbol, informing you it will no longer allow DNA samples.</span>"
+		else
+			controller_lock = 0
+			M << "<span class='notice'>\The [src] beeps and displays an unlocked symbol, informing you it will now allow DNA samples.</span>"
+	else
+		M << "<span class='warning'>\The [src] buzzes and displays an invalid user symbol.</span>"
+
+/obj/item/weapon/gun/verb/allow_dna()
+	set name = "Toggle DNA samples allowance"
```
Zuhayr: Inconsistent capitalization.

Zuhayr:
Seems odd to have all this stuff at /gun level when it isn't used and logically would not be applicable to guns like crossbows or pneumo cannons.

Mark9013100:
Only if I can get my guns laundered by Drebin 983.

Yoshax:
If the warden dna locks all the guns then you've got worse problems with the warden than not being able to use all the guns.

Fox-McCloud:
May want to look at TG's firing pins, as it's a more robust and modular system.

SpadesNeil:
_Something something angry political post about smart guns._

No I'm kidding. This is neat. Though I worry it may take some risk away which I think should stay in the game. I don't think it's a good idea to be able to lock guns because it's bad for gameplay. What if I steal a gun? What if I need a downed officer's gun? Mostly, what's stopping one person from locking all of the guns like a dick?

I'd rather not see this feature. It removes an element of risk that should stay in the game.

Yoshax:
It's not applied to any guns, and I don't intend to go applying it to every gun. It's just something I made because I was bored.

It's not going to be applied to tasers, ion rifles, etc unless you go do it.

SpadesNeil:
I'm a stickler so I'd rather not see the feature _at all_ but I understand if it _is_ added, it's limited to certain guns. I am okay with that if it happens.

code/modules/projectiles/gun.dm
```diff
@@ -426,4 +452,78 @@
...
+			stored_dna -= user.dna
+			M << "<span class='notice'>\The [src] beeps and clears the DNA it has stored.</span>"
+			if(M.dna == controller_dna)
+				controller_dna = null
+				M << "<span class='notice'>\The [src] beeps and removes you as the primary controller.</span>"
+				if(controller_lock)
+					controller_lock = 0
+			return 1
+	else
+		M << "<span class='warning'>\The [src] buzzes and displays a locked symbol. It is now allowing DNA modifcation at this time.</span>"
```
Datraen: Is this supposed to be a not?

PR #1731, opened 2016-05-24
a few things

ursamedium:
aaaagh i'm tired


ursamedium:
I'm tired and I pushed this to polaris I'm so dumb.

PR #1732, opened 2016-05-24
Fixes two issues

Arokha:
Listed in the commits.

### DIFF
--- a/code/modules/mob/living/carbon/human/life.dm
+++ b/code/modules/mob/living/carbon/human/life.dm
@@ -154,21 +154,6 @@
 	if(stat != CONSCIOUS) //Let's not worry about tourettes if you're not conscious.
 		return
 
-	//Vision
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/human/life.dm
```diff
@@ -970,19 +955,18 @@
 		if(species.vision_organ)
 			vision = internal_organs_by_name[species.vision_organ]
 
-		if(!vision) // Presumably if a species has no vision organs, they see via some other means.
+		if(!species.vision_organ) // Presumably if a species has no vision organs, they see via some other means.
 			eye_blind =  0
 			blinded =    0
```
Arokha: Line 973 didn't work because vision would ALSO be null if they were missing eyes, if it was unable to pick from the internal_organs_by_name list. So it was treating that case as if the species had no eyes.

PR #1734, opened 2016-05-24
Removes unused xeno stuff + a bug fix

Yoshax:
Will most likely fix changelings having a xeno icon randomly sometimes on medical huds.


Yoshax:
Uhh wot.

PR #1736, opened 2016-05-25
Adds towels

Yoshax:
Can be worn on:
- Belt
- Outerwear
- Head

Added to loadout.

Sprites courtesy of Pawn.


Datraen:
Just in time for the best day!

PR #1741, opened 2016-05-26
Say/Whisper Rewrite

Arokha:
This rewrites say/whisper and removes the code duplication between them. They had inconsistent behaviors between them, and this makes them the same, because they are now one proc that accepts a whispering param (I could make this a whisper_range parameter if someone wants? but maybe that's excessive).

So I'm not expecting this to just automatically be perfect, and I'm ready for the wave of critique about it. Or questions on why I did something. Or pointing out some obvious mistake.

---
Test string: "Hello, I'm just testing things now. This is a decent sized message, though."

Testing (200 say, 200 whisper, 2 people 1 tile away, 1 person 3 tiles away, 1 person 7 tiles away)
New Version (s+w): 0.502 | 4.015 | 4.027 | 400 calls

Old Version (say): 0.051 | 3.123 | 3.123 | 200 calls
Old Version (wsp): 0.018 | 1.402 | 1.402 | 200 calls

Difference: +0.433 | -0.51 | -0.498 | X calls

Makes sense that 'self cpu', the first column is more since it's more stuff in one proc than before. The rest is faster, though.

---

This cherrypicks a few commits from Bay, then builds on them. So if you pull stuff from Bay it should be able to construct a path to get here from ancestors.

Saycode is much simpler and has more informative comments throughout. So hopefully the next generation can do good things with it.

The programatic differences are probably the most important: Say/Whisper now use a new dview-based listing proc to find mobs, and will include mobs in objects and such, like scooped Teshari, who can now whisper in/out of their objects. This also allows Cyborgs to whisper, and for people to whisper into radios with the correct verbs, etc. 

I'm not 100% sure... could you whisper in languages before? You can now, anyway. I dunno if you could before or not. 

It didn't appear to have the spam protection in 'say', only 'whisper'? Maybe it was hidden somewhere else.

Slightly unrelated, also ports a rewrite of get_turf to use locs[1], which is faster, and also ports the fixes to the speech problems such that you don't whisper with a weird blank verb anymore. No more `Person  softly, "Thingthingthing"`


Datraen:
Makes sense logically and helped tested, so :+1: merging now.

Issue #1606, opened 2016-05-10
Cyborg Recharger deconstruction bug


Edwardc120:
#### Additional info:
 was told it was a mapping bug that the chargers wouldn't recharge

Nalarac:
Exosuit Fabricators seems to have this same problem along with being unable to be upgraded with the Rapid Part Exchange device.

SinTwo:
Will test Exosuit Fabricators but I could not reproduce this with Cyborg Rechargers.

SinTwo:
Seems only Exosuit Fabricators have this problem. Problem has been solved, just need to clear somethings first.

PR #1746, opened 2016-05-26
Fixes Exosuit Fabricator Deconstruction and makes Jukeboxes constructable

Leshana:
* Fixes Issue #1606 - You need to have the circuit be an actual instance, not a path.
* Makes Jukeboxes constructable
  * Adds circuit board.
  * Adds design to print board on autolathe.
  * Makes jukeboxes deconstructable back to frame.

### DIFF
--- a/code/modules/research/designs.dm
+++ b/code/modules/research/designs.dm
@@ -756,6 +756,13 @@ CIRCUITS BELOW
 	build_path = /obj/item/weapon/circuitboard/arcade/orion_trail
 	sort_string = "MAAAA"
 
+/datum/design/circuit/oriontrail
+	name = "jukebox"
+	id = "jukebox"
+	req_tech = list(TECH_MAGNET = 2, TECH_DATA = 1)
+	build_path = /obj/item/weapon/circuitboard/jukebox
+	sort_string = "MAAAB"
+
 /datum/design/circuit/seccamera
 	name = "security camera monitor"
 	id = "seccamera"
--- a/code/modules/research/designs.dm
+++ b/code/modules/research/designs.dm
@@ -756,7 +756,7 @@ CIRCUITS BELOW
 	build_path = /obj/item/weapon/circuitboard/arcade/orion_trail
 	sort_string = "MAAAA"
 
-/datum/design/circuit/oriontrail
+/datum/design/circuit/jukebox
 	name = "jukebox"
 	id = "jukebox"
 	req_tech = list(TECH_MAGNET = 2, TECH_DATA = 1)
--- a/code/game/machinery/frame.dm
+++ b/code/game/machinery/frame.dm
@@ -218,7 +218,10 @@
 				if(component_check)
 					playsound(src.loc, 'sound/items/Screwdriver.ogg', 50, 1)
 					var/obj/machinery/new_machine = new src.circuit.build_path(src.loc, src.dir)
+					// Handle machines that have allocated default parts in thier constructor.
 					if(new_machine.component_parts)
+						for(var/CP in new_machine.component_parts)
+							qdel(CP)
 						new_machine.component_parts.Cut()
 					else
 						new_machine.component_parts = list()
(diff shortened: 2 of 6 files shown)


code/modules/research/designs.dm
```diff
@@ -756,6 +756,13 @@ CIRCUITS BELOW
 	build_path = /obj/item/weapon/circuitboard/arcade/orion_trail
 	sort_string = "MAAAA"
 
+/datum/design/circuit/oriontrail
```
Yoshax: Duplicate definition. Should be `datum/design/circuit/jukebox` or something along those lines.

Yoshax:
You'll likely want to follow the format of making a `map/new()` for the machinery that's buildable. Ensures they correctly spawn with their parts, and there's no issues with the buildable ones.

code/modules/research/designs.dm
Leshana: Oops, you are absolutely correct about this, copy/paste error.

Yoshax:
Still needs the machine given a new `map/new()`
Otherwise the ones you build will have a whole other duplicate set of parts.

Furthermore, on that note, by doing so you also need to change the mapped in jukebox whiiiiiich can't be done because of #1575

SinTwo:
Pretty much what Yoshax said here. On the note of the mech fabricators, I have a fix ready for them but require the map freeze to be lifted before being able to apply and merge it.

Leshana:
Actually @Yoshax machines you build will not contain duplicate parts even with the current code (I have confirmed this via testing). Please see: https://github.com/PolarisSS13/Polaris/blob/master/code/game/machinery/frame.dm#L222 for the line that removes any parts from a newly built machine (that were added by its constructor) _prior_ to adding the parts from the frame.
Based on my reading, I believe the map/new() design is an anti-pattern that should be avoided, because:

1. The machine code is already explicitly designed to handle machines that self-populate their component parts in their constructor.
2. The only benefit it provides is a very slight performance optimization, preventing the component parts from being allocated and then gc'd in the case of in-game building (a rare event).
3. It makes it more likely that mappers will accidentally place the wrong type, resulting in machines *without* any components.
4. It makes it more difficult to subclass machines, requiring every subclass to define its own map/New(), duplicating code.

If I have misunderstood the operation of the code please let me know.

Yoshax:
I'll do some testing on the matter and get back to you. I'm just going by what I've been told.

Leshana:
Okay I did a bit more digging!  In my earlier testing I:
1. Constructed jukebox from frame & circuit
2. Deconstructed it.
3. Verified I got back exactly the same parts I started with.
4. Found the line of code responsible for this ( https://github.com/PolarisSS13/Polaris/blob/master/code/game/machinery/frame.dm#L222 )

This was all correct, but a closer inspection shows that doing it this way *does* duplicate the parts in the machine's _contents_ but does *not* duplicate it in the component_parts list.  Thus you don't get them back when you deconstruct it.     Some might argue that this is "close enough" but eh, we can do better.

However this is only true for machines that _contain_ their component parts  (See the circuit.contain_parts variable).   Machines for which circuit.contain_parts = 0  do not contain their parts, and thus  Line 222 would remove the only reference to them, allowing them to be gc'd. 

Give the earlier points I mentioned regarding the map/New() pattern, I propose either
A. Changing the machine to be circuit.contain_parts = 0,  or
B. Changing frame.dm to properly handle the circuit.contain_parts = 1 part by explicitly deleting the parts instead of simply Cut()'ing the component_parts list.

Obviously Option A has less potential side effects.  I'll double check it.

Leshana:
However, Option B has the advantage of simultaneously fixing this problem for **all** machines everywhere, now and forever, without the inconvenience of having to do map/New()
I've made this change and confirmed no duplicates exist even in the contents of the machine.  Also checked a photocopier and photocopier/map for good measure.

Datraen:
Technically no longer a map issue, so...

PR #1747, opened 2016-05-26
Allow building lights on windows.

Leshana:
I noticed that the current station map has many locations where lights and such are mounted on windows.  But you cannot actually construct lights this way in-game.   In the interest of making stuff doable in-game, this enables placing of frames on anchored windows.
example of lights on windows


Datraen:
:+1:

PR #1764, opened 2016-05-27
Adds preference to mute remote LOOC chatter heard by admins.

Yoshax:
* Lets people remain on standby for adminhelps without having all that spam scrolling stuff off the screen if they are not taking on the role of policing LOOC.


Anewbe:
:+1:

PR #1775, opened 2016-05-28
Ports Models for Cybernetic Limb Manufacturers + More Monitor-head Wallpapers

KasparoVy:
As requested!

Ports prosthetic head manufacturer models from Para.
Head and Hesphiastos - Frontier monitor sprites by **Dumbdumn5**. They're the reason why the original PR exists.

_Manufacturer selection for anything but the head right now._
headmanufacturerselect

_Manufacturer selection for the head. 17 entries in total._
headmanufacturerselect

_Preview of Hesphiastos - Frontier head in character selection._
altheadpreview

_And in-game._
altheadpreview2

Furthermore, the monitor-head mask thing has been changed such that it offers possible monitors to select based on the allowed monitor styles as defined on a by-manufacturer basis in robolimbs.dm. There is a list of standard monitor styles defined that the Morpheus and all other heads suffixed with 'mtr.' use. The Hesphiastos alt1. definition specifies a list of monitor styles that head will be able to use. If the user with the synthetic head does not have a definition of monitor styles they can use (meaning, not monitor-heads) they are seen as incompatible with the 'monitor' upgrade, and attempting to wear it will tell the user such a thing.

_Preview of the 'monitor-upgrade' mask on the ground._
It starts with the same icon as the Morpheus default head south-facing sprite (in hands as well) until it is equipped, then it defaults to the 'blank' or 'off' screen for whatever monitor type is detected.

ASIDE: I apologize for the odd darkness overlay. For whatever reason, the 'blind' overlay appears whenever I spawn in as a mob/living/carbon/human/whatever (whether I spawn in normally or posess a mob I created in the game panel) and I can't seem to get rid of it. Happened before I made any changes.
maskpreview

_Mask menu for a Hesphiastos alt. head and the selected screen in-game._
maskmenu1maskonbody1

_Mask menu for a standard monitor head and the selected screen in-game._
maskmenu2maskonbody2

---------------------------------------------------

**Please refer to this page for more pictures of and information about the feature: https://github.com/ParadiseSS13/Paradise/pull/4262**

**Things from that page that will not apply here: Antennae, Optics, changes to the Zeng-hu head sprite and bugfixes (as those bugs aren't present here).**


code/modules/clothing/masks/monitor.dm
```diff
@@ -40,8 +56,17 @@
 	..()
 	var/mob/living/carbon/human/H = loc
 	if(istype(H) && H.wear_mask == src)
+		var/obj/item/organ/external/E = H.organs_by_name[BP_HEAD]
+		var/datum/robolimb/robohead = all_robolimbs[E.model]
 		canremove = 0
-		H << "<span class='notice'>\The [src] connects to your display output.</span>"
+		if(robohead.is_monitor)
+			if(robohead.is_monitor == 1)
```
Zuhayr: Uugghhhhhh. :( :(

Zuhayr:
Couple of issues I have with this code that I would like to note (in before 'piss off you aren't a dev anymore').

- The subtype code and additional menu dialogue seems completely unnecessary, and it makes the code a fair bit harder to follow. Defining the datums is sufficient for making the sprites available and doesn't add more complexity to the interface and the code itself.

- The hardcoded monitor stuff is not ideal - why do the lists need to be instantiated on every mask item? Can't they go on the head itself via the manufacturer datum? Could the hesphastios monitor stuff be its own subtype of the monitor item instead?

I don't mean to be overly critical but this PR feels very untidy.

code/modules/clothing/masks/monitor.dm
```diff
@@ -9,29 +9,45 @@
...
-		"heart" =    "ipc_heart",
-		"monoeye" =  "ipc_monoeye",
-		"nature" =   "ipc_nature",
-		"orange" =   "ipc_orange",
-		"purple" =   "ipc_purple",
-		"shower" =   "ipc_shower",
-		"static" =   "ipc_static",
-		"yellow" =   "ipc_yellow"
+
+	var/list/standard_monitors = list(
```
Zuhayr: The ````global```` was there to ensure that this list is not represented multiple times in memory and should probably be put back.

KasparoVy:
No worries, I appreciate the criticism

The two things you mentioned were integral to the PR when I first wrote it and I just hadn't put much thought into gutting it for this codebase, my mistake

Already working on the second point, but the first point I need a tiny bit of clarification on.

You'd rather just one list with all possible models listed at once, not a menu listing just manufacturers and another with just models for the selected manufacturer?

code/modules/clothing/masks/monitor.dm
KasparoVy: Thanks, I was unsure of its importance

Spacemanspark:
:+1:

Zuhayr:
Personally - I want to stress that because I am not a player let alone staff on Polaris and probably shouldn't be commenting at all - I think that if you can just add the alt manufacturers as their own datums without losing functionality, you should.

Datraen:
@Zuhayr brings up good points. Unless it is felt that the list is sufficiently large, it shouldn't need condensing.

KasparoVy:
I just figured it made sense to have, but I've no problem gutting it

ParadoxSpace:
pull Polaris again
Aronai's limb changes and the map PR have likely caused conflicts

Hubblenaut:
Could we perhaps not call them 'alt' and instead give them a manufacturer tag?

KasparoVy:
Bad timing, I'm unable to do anything until I get back from a wedding in a few hours. I'll sort it then. My apologies.

I'll try to think of some appropriate names as well, but they were designed to be models offered by the same manufacturer.

Yoshax:
No worries, there's no rush.

KasparoVy:
The names of the alt heads have been changed as requested. I modified the descriptions to be less copypaste. Menu picture updated.

If anyone has better ideas for the names, please comment

ParadoxSpace:
Sounds good to me

PR #1778, opened 2016-05-28
Fix in-hand sprites for suits

Leshana:
**Symptom**: Suits did not appear on mob sprite when held in the hand.
**Cause**: The item_state variables were referring to non-existent icon states (XXX_voidsuit) in lefthand.dmi and righthand.dmi, where the actual icon states were XXX_hardsuit.
**Solution**: Rename the icon states in the dmi files to be what is actually used:
_Changed_:
eng_hardsuit -> eng_voidsuit
mining_hardsuit -> mining_voidsuit
syndie_hardsuit -> syndie_voidsuit
wiz_hardsuit -> wiz_voidsuit
medical_hardsuit -> medical_voidsuit
sec_hardsuit -> sec_voidsuit
atmos_hardsuit -> atmos_voidsuit


Datraen:
I'd rather this be handled after Xenobio2 gets merged, as this will create conflicts with that.

Leshana:
The same change has been done again after the Xenobio2 merge.

PR #1779, opened 2016-05-28
Ports Bay's Robolimb Changes

Arokha:
Applies to FBPs and such. Changes how robolimb repairs are handled (not how to do them ingame, but rather just the mechanisms behind the scenes).

They also fixed two bugs I fixed earlier, but it'd be better if we took Bay's changes so that they can merge later if you pull into those same sections. I also fixed some runtimes with bruise packs and the like and notified the user if they were trying to heal a non-existent limb.

Yes a lot of files changed, as they removed 'ORGAN_ROBOTIC' flag and made it part of an enumeration instead. This is important for later when I do 'lifelike' FBPs!

The removed surgery_limbattachment was an unused file. Just removing it as it's not relevant.


Yoshax:
You re-added `code/modules/mob/living/carbon/human/species/xenomorphs/alien_powers.dm`
We removed Xenos.

PR #1784, opened 2016-05-28
Fixes invisible gibs

Yoshax:
Due to gib4 not existig.


EmperorJon:
Looks like it's not existed for years, and by coincidence we find out on a round where there are bugs galore and we're convinced something just broke it! Wahey.

Yoshax:
Praise bugs.

Datraen:
Praise Bugs!

PR #1785, opened 2016-05-28
Ports attack call sequence cleanup from Bay

Yoshax:
Revives #1450 

Porting this because I noticed Polaris is missing the attack code cleanup from last summer and it's only going to get more difficult to port with time.

Port of https://github.com/Baystation12/Baystation12/pull/10287

Attempts to clean up the item-attacking-mob call sequence cleaner by reorganizing each call to have a clearly defined purpose.

The call sequence now looks like this:

- `mob/ClickOn()` calls the item's resolve_attackby() proc. *Unchanged.*
- `item/resolve_attackby()` calls the target atom's attackby() proc. *Unchanged.*
- `mob/living/attackby()` after checking for surgery, calls the item's attack() proc.
- `item/attack()` is what you override to define what an item does when it is used on a mob. *By default, it assumes the item is a weapon, and does what every weapon should do: generates attack logs before calling the mob's attacked_with_item() proc.*
- `mob/attacked_with_item()` defines what happens when a mob is attacked with a weapon.
*Not just any time an item is used on a mob, but specifically when it is used as a weapon. This is largely where combat code begins.*
For humans, this does all the standard attack roll and checking what body part gets hit and that stuff. If the attack is a successful hit, it calls apply_hit_effect().
- `item/apply_hit_effect()` defines what happens when a weapon successfully makes contact. **This is something really basic yet was sorely missing from the codebase, and is what got me started on this cleanup.** *For example, stun batons now just override this instead of re-duplicating the attack roll and hit/miss messages, since that's now done by attacked_with_item()!*
While this can be overriden to do whatever you want, "standard" weapons (or weapons that act like standard weapons in addition to doing other stuff) should call the target mob's hit_with_weapon() proc to ensure that all regular weapons have the same code for applying damage and armour, bloodying the weapon and making blood splatters, embedding, all that stuff, without having to duplicate everything.
- `mob/living/hit_with_weapon()` defines how "standard" weapons are handled when they successfully hit a mob. This is intended to be more of a helper proc that is called by apply_hit_effect() overrides when they want to re-use the default weapon behaviours. I guess it would benefit from being split up further but that's a job for another time.

This mainly cleans up the call sequence for mob/living/carbon/human. Other mobs have been updated so that they will work, but a serious attempt to clean up every mob was is included in this PR. That said, the bulk of attack code deals with humans, for obvious reasons.

DNM while I ensure nothing is broken and there's nothing else to do.


Yoshax:
So, it kinda works. Except it doesn't with the new Xenobio procs and I've no idea how to make them work.

code/game/objects/items/weapons/swords_axes_etc.dm
```diff
@@ -1,18 +1,10 @@
...
  * Contains:
- *		Banhammer
  *		Sword
  *		Classic Baton
  */
 
-/*
- * Banhammer
- */
-/obj/item/weapon/banhammer/attack(mob/M as mob, mob/user as mob)
```
Datraen: RIP Banhammer

code/game/objects/items/weapons/weaponry.dm
```diff
@@ -68,27 +56,6 @@
 		user << "<span class='notice'>You hit the floor with the [src].</span>"
 		call(/obj/effect/rune/proc/revealrunes)(src)
 
-/obj/item/weapon/sord
```
Datraen: RIP SORD

code/modules/hydroponics/trays/tray_reagents.dm
```diff
@@ -48,20 +48,6 @@
 	toxicity = 8
 	pest_kill_str = 7
 
-/obj/item/weapon/material/minihoe // -- Numbers
```
Datraen: Duplicate removed, for those looking.

Datraen:
:+1:

EmperorJon:
:+1:

PR #1787, opened 2016-05-29
Manual port of BS12 setup changes

EmperorJon:
Manually ports 
https://github.com/Baystation12/Baystation12/pull/13042
and
https://github.com/Baystation12/Baystation12/pull/13086
as close as currently can be done. This had to be done manually because of the number of differences between the codebases that are becoming apparent. In order to resolve this and some of the changes that couldn't be brought across entirely, I've got a list of other things I want to retroactively port across over the next day or so, namely some inventory.dm changes, and underwear-as-datums which complicated this in a few places. A few little bits and bobs also had to be ported from previous PRs to get this all working, so I really need to clean up after.

This adds the new character preview stuff by @PsiOmegaDelta which ensures the setup screen only updates the preview image when things actually change, completely overhauls the way the preview image is actually created in a nice manner, and also adds some nice functionality whereby you can hide equipment/clothes on the preview image.

I've tested all the obvious things and it all seems to be a-ok on my end!


EmperorJon:
image

Here's an example (with a random character) of the new preview stuff that Psi did. That's pretty much the only front-end difference the user's going to really see, but it's damn useful!

EmperorJon:
https://github.com/Baystation12/Baystation12/pull/8973
This is something which got a bit intertwined which we may want to port.

EmperorJon:
Oh! I almost forgot. This also makes loadout items appear on your character preview.

image

PR #1788, opened 2016-05-29
Xenobio2 tweaks

Datraen:
Will create conflicts with https://github.com/PolarisSS13/Polaris/pull/1785, that should probably be merged before this is.


Datraen:
Huh, it didn't create conflicts.

PR #1794, opened 2016-05-29
Underwear Datums

EmperorJon:
Manually ports https://github.com/Baystation12/Baystation12/pull/12400. I'm finding it a lot easier to copy-paste across what's needed and alter on the fly than use a merge tool...


EmperorJon:
One more step to bringing our preferences/setup stuff in line with Bay.

PR #1796, opened 2016-05-29
The Terminator Update

Arokha:
This updates synthetic limbs to make more sense. It tweaks what you can 'see' when examining people. If someone has a robotic leg but is wearing pants, how would you know that? If someone has a burn on their arm, but their jumpsuit sleeves down, how would you know that? If someone has a replacement Vey-Med arm, how would you know it's robotic? It also treats examining FBPs more 'realistically'. If they are covered except for their head, it doesn't matter if their whole body is Bishop robotic. If their head is Vey-Med and that's all you can see, they just look human to you.

Remember, if you want to know where you're hurt, just helpclick yourself and you'll know where you feel pain.

So FBP manufacturers can have a 'lifelike' var set. Vey-Med has this. This makes the limbs not show obviously non-organic damage (dents) until they are more damaged and start showing wires/metal. Attempts to treat these limbs with medical stuff results in a different message. Manufacturers can also set individual blood colors. Vey-Med blood is now white, ala Bishop from Aliens.

isSynthetic proc = Is the mob actually synthetic, as in, mechanically for breathing/tox purposes?
looksSynthetic proc = Does the mob display outward signs of being synthetic? Based on head and torso and what's revealed.

Other fixes:

- Ports Ccomp's GMAOIVF improvements.
- You can no longer attach limbs to non-existent parents. You can't give somone a foot on a leg that doesn't exist.
- You can't attach fleshy limbs to robotic ones. BODIES DO NOT WORK THAT WAY.
- 'Synthetic' var on humans points to manufacturer if you need to grab it quickly. isSynthetic returns this as well.
- Robolimb count (and thus overheating speed) updates whenever your limbs change.
- Lifelike FBPs do not show a 'system offline glyph'.
- isSynthetic and looksSynthetic moved to human_helpers becasue they were defined on human in mob_helpers
- Nanopaste correctly repairs limbs using the new procs (both burn and brute, making it an expensive welder+wire)

### DIFF
--- a/code/__defines/damage_organs.dm
+++ b/code/__defines/damage_organs.dm
@@ -44,10 +44,11 @@
 #define DROPLIMB_BURN 2
 
 // Damage above this value must be repaired with surgery.
-#define ROBOLIMB_SELF_REPAIR_CAP 30
+#define ROBOLIMB_REPAIR_CAP 30
 
-#define ORGAN_ASSISTED 1
-#define ORGAN_ROBOT    2
+#define ORGAN_ASSISTED 1 // Like pacemakers, not robotic
+#define ORGAN_ROBOT    2 // Fully robotic, no organic parts
+#define ORGAN_LIFELIKE 3 // Robotic, made to appear organic
 
 //Germs and infections.
 #define GERM_LEVEL_AMBIENT  110 // Maximum germ level you can reach by standing still.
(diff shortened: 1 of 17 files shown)


Datraen:
:+1:

nachomeep:
Zeng-Hu still has a system offline glyph? That feels a bit odd.

Arokha:
Basically everyone has an 'offline light' if they appear to be synthetic. I assume if you're vey-med head and zeng-hu body, the light isn't visible on the head. Must be on the chest or something, as that's when you 'appear synthetic', when your chest is exposed in that case.

PR #1799, opened 2016-05-29
Prevents the evaluation of gravity in null areas.

### LINKED ISSUE #1795: Runtimes from FBPs in character setup

proc name: has gravity (/proc/has_gravity)
  source file: areas.dm,314
  usr: (src)
  src: null
  call stack:
has gravity(Summer Hardie (/mob/living/carbon/brain), the man-machine interface (Sum... (/obj/item/device/mmi))
Summer Hardie (/mob/living/carbon/brain): mob has gravity(null)
Summer Hardie (/mob/living/carbon/brain): Life()
Summer Hardie (/mob/living/carbon/brain): Life()
/datum/updateQueueWorker (/datum/updateQueueWorker): doWork()
/datum/updateQueueWorker (/datum/updateQueueWorker): doWork()
runtime error: undefined proc or verb /mob/living/carbon/human/dummy/mannequin/has gravity().

Not that.

#### What actually happened
Yeah.

#### Steps to reproduce
Start a round, then turn a character slot into an FBP and watch.

#### Additional info:
 - **Anything else you may wish to add** Looks like setup mannequins are being added to mob/living mob lists if they're FBPs, despite the fact they're removed when they're first created. I suspect there's something weird going on with creating a posibrain/MMI adding them to these lists. This needs to be hunted down and fixed because if this goes live it may cause big problems.

Datraen:
Catches get_turf(AT) returning null before continuing.

Partially fixes* #1795.

### DIFF
--- a/code/game/area/areas.dm
+++ b/code/game/area/areas.dm
@@ -310,6 +310,7 @@ var/list/mob/living/forced_ambiance_list = new
 /proc/has_gravity(atom/AT, turf/T)
 	if(!T)
 		T = get_turf(AT)
+		if(!T) return 1	//Still no turf, give up. Used for null space entities that shouldn't be bouncing up and down.
 	var/area/A = get_area(T)
 	if(A && A.has_gravity())
 		return 1


EmperorJon:
I personally wouldn't consider this more than a possibly dangerous hack-fix. Mannequin mobs will still be present in the mobs/living mobs list, which could have unintended consequences.

Datraen:
I forgot the living mobs issue.

Datraen:
Runtime persisted, didn't let run long enough.

PR #1800, opened 2016-05-29
[WIP] Gun Stuff

SpadesNeil:
Let's try this one more time without screwing it up.

Chances are I've probably still screwed it up. If so, I'm just going to close this and try again.


SpadesNeil:
Nope. Did this wrong. Contains all this crap I don't want.

PR #1801, opened 2016-05-29
PUNCH SPIDERS, GET MONEY

Anewbe:
Adds stronger hostile mobs for shenaningans
Switches the attack text on spiders to "punches", from "pokes"


Hubblenaut:
There doesn't seem to be any visual difference between all those - intended?

Anewbe:
I debated about giving them different sprites, but anyone who can spawn these is also able to modify the sprites... Plus it'll keep you on your toes.

PR #1807, opened 2016-05-30
Adds working double beds.

AtomicWorm:
Adds double beds along with appropriate bedshets to the code. They are not currently buildable.

Mob sprites get shifted up by 13 pixels when a mob is buckled to the bed, to simulate using the left side.

Example:
http://i.imgur.com/0XE2oaU.png


SpadesNeil:
brb porting this to Vorestation because admit it we need it more than you do. :)

PR #1812, opened 2016-05-30
Underwear + setup merges

EmperorJon:
Manually merges https://github.com/Baystation12/Baystation12/pull/12595 and https://github.com/Baystation12/Baystation12/pull/12461.

There's another thing I want to fix in the setup that I've seen, and with that I'll put a mass changelog.


Yoshax:
Please re-add the Lacy Bra, Violet D:

EmperorJon:
Renames the one with the red stripe to alt, stripe and changes the previous violet sprite to be alt, making it colour-able instead of violet. Feel free to take a look at it.

PR #1820, opened 2016-05-30
Adds more ties to loadout, strangulation not included

Mark9013100:
Adds the blue tie with clip, long red tie, yellow tie, and navy blue tie to custom loadout.


EmperorJon:
So these already have sprites, they just weren't available?

Yoshax:
Correcto. They were added by the port from Bay that split several suits in jacket, coat + under.

EmperorJon:
Excellente.

Issue #1789, opened 2016-05-29
Auxillary tool storage has the wrong access and improper pipes.


Jakksergal:
#### Brief description of the issue
Auxiliary tool storage has the wrong access set. It appears to have maintenance and some other access that's not meant to be on the station.
#### What you expected to happen
The door to just be maintenance access.
#### What actually happened
It was inaccessible
#### Steps to reproduce
- Be any job with maintenance access
- Touch door
- Get denied

#### Additional info:
 - **Server Revision**: master - 2016-05-28 01b3182004b7098f57baa249511fcb26946cb23b 
 - **Anything else you may wish to add** 
 - Located next to the coffee shop.
 - Pipes under door do not have hidden set to one and are on top of the floor tiles.

Neerti:
Should note I was able to enter as a security officer, which is odd.

Yoshax:
The front door has access = 12. Which is maintenance.

Woodratt:
aux tool storage has always had maintenance as its access. Security officers, paramedics, cargo techs all have access. The pipes being visible however is a bug, which I am sorting.

Issue #1277, opened 2016-03-28
Plastic flaps in disposals access are not airtight.


PapaDrow:
#### Brief description of the issue

Plastic flaps on disposals line (specifically, the area with the mass driver) are not of the 'airtight' variety.

#### What you expected to happen

For the conveyor line either to have no easy access (ie an airlock) or to have airtight flaps at the end. Or at least a warning sign.

#### What actually happened

Whenever the chute to the mass driver is activated and the mass driver itself is fired, there appears to be leakage of pressure in the section of disposals with the conveyor. If anyone opens the airlock onto the conveyor belt, it tends to depressurize the whole room for a moment and causes injury.

#### Steps to reproduce

Go to the disposals line via maint, activate both the conveyor (when there's something on it so the chute at the end activates) and then use the mass driver.

#### Additional info:
 - master - 2016-03-27 ddf4cceb963cf24e2f36f502b4c9c3d2eac651d0

Nerezza:
Just to describe it in another way. The chute in disposals is pressurizing off ambient air, and it quickly sucks up the air inside the conveyor portion.

I'm not sure if it needs air to actually operate, if not then airtight flaps would work like Drow says. If it does need air, then that section of disposals needs a vent.

Yoshax:
Duplicate of #766

Issue #817, opened 2016-01-21
[Map] Medical doesn't have a point for MULE bots


Mark9013100:
Nitpick, but medical doesn't have a drop off point for MULE bots.

NortonDK:
So is command

Atermonera:
This has been resolved

PR #1833, opened 2016-05-31
Fixes bugs with current map

Datraen:
Fixes:
#1789
#1277
Comment in #817, since the original was already fixed.


Datraen:
In addition to fixing #1277, an air alarm was placed inside the mass driver disposals section.

PR #1836, opened 2016-06-01
Fixes box of EMP grenades only ever spawning 1 grenade

Yoshax:


Datraen:
It... doesn't look properly indented.

Neerti:
It looks fine to me, Github might be just misleading you due to how wide their tabs are on the preview.

PR #1842, opened 2016-06-01
Fixes being able to buckle more than one mob to bucklable things

Yoshax:
Required 2 checks to be neat and because buckle code is shit.

### DIFF
--- a/code/game/objects/buckling.dm
+++ b/code/game/objects/buckling.dm
@@ -5,6 +5,7 @@
 	var/buckle_lying = -1 //bed-like behavior, forces mob.lying = buckle_lying if != -1
 	var/buckle_require_restraints = 0 //require people to be handcuffed before being able to buckle. eg: pipes
 	var/mob/living/buckled_mob = null
+	var/has_buckled
 
 /obj/attack_hand(mob/living/user)
 	. = ..()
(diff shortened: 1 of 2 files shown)


code/game/objects/buckling.dm
```diff
@@ -5,6 +5,7 @@
 	var/buckle_lying = -1 //bed-like behavior, forces mob.lying = buckle_lying if != -1
 	var/buckle_require_restraints = 0 //require people to be handcuffed before being able to buckle. eg: pipes
 	var/mob/living/buckled_mob = null
+	var/has_buckled
```
Techhead0: Instead of adding another var, why not just check `buckled_mob`?

Yoshax: To remove any possible fuckery where a mob is removed from the buckled_mob but is still technically buckled. If that was the case, then my added var wouldn't have been changed, and would still recognise there's a mob buckled, which there is.

Techhead0: Instead, you leave the possibility for fuckery when the new var isn't updated properly.

*shrugs*

Zuhayr: If someone is unbuckled but the buckled_mob var isn't updated then that particular thing is bugged. Adding a new var in case the other var breaks is very strange practice.

PR #1793, opened 2016-05-29
Bay ports, fixes and stuff

Yoshax:
Putting it up so folks can see and work on it from here. Still got a bunch more stuff to grab.

Ports Baystation12/Baystation12#11281
Ports Baystation12/Baystation12#11988
Ports Baystation12/Baystation12#11507
Ports Baystation12/Baystation12#12683
Ports Baystation12/Baystation12#11431

PR #1845, opened 2016-06-01
Scheduler now uses btime, and other improvements

Yoshax:
Port of Baystation12/Baystation12#11364

This is in preperation of porting other things, and was one of the reasons why I had so much difficulty with #1793 

This seems functional alone, but only includes the original PR by Psi. Other modifications and additions will come in additional PRs along with their accompanying things.


Yoshax:
There's a diff difference between this and Psi's PR because of Github just figuring stuff out differently.

PsiOmegaDelta:
And here I'm about to remove the use of btime as 510 has been officially released :P

Yoshax:
AAAAAAAAAAAAAAH. PSI. WHY YOU DO THIS. HOW AM I EVER MEANT TO PORT ALL YOUR STUFF.

But really. This'll do for now, until I can port all the rest of the stuff, and Psi's done whatever he's doing then I can port that.

Datraen:
For right now, let's use btime until we're 510 ready.

PR #1859, opened 2016-06-02
Brings SmartFridges under the vending machine datum system.

Datraen:
The items remain the specific item that you start with, rather than having the opportunity to duplicate another item.

It was reported to me that Xenobotanical samples could be converted if you put them in a fridge and then take them back out, this will fix that.

EDIT: Seed storage checks against the actual traits of the seed as well as contains a list of actual objects, no change needed on the machinery.

### DIFF
--- a/code/game/machinery/kitchen/smartfridge.dm
+++ b/code/game/machinery/kitchen/smartfridge.dm
@@ -1,3 +1,63 @@
+/*
+*	Datum that holds the instances and information about the items in the smartfridge.
+*/
+/datum/data/stored_item
+		var/item_name = "name"	//Name of the item(s) displayed
+		var/item_path = null	
+		var/amount = 0
+		var/list/instances		//What items are actually stored
+		var/fridge				//The attatched fridge
+		
+/datum/data/stored_item/New(var/fridge, var/path, var/name = null, var/amount = 0)
+	..()
+
+	src.item_path = path
+
+	if(!name)
+		var/atom/tmp = path
+		src.item_name = initial(tmp.name)
+	else
+		src.item_name = name
+		
+	src.amount  = amount
+	src.fridge = fridge
+
+/datum/data/stored_item/Destroy()
+	fridge = null
+	if(instances)
+		for(var/product in instances)
+			qdel(product)
+		instances.Cut()
+	. = ..()
+	
+/datum/data/stored_item/proc/get_amount()
+	return instances ? instances.len : amount
+	
+/datum/data/stored_item/proc/get_product(var/product_location)
+	if(!get_amount() || !product_location)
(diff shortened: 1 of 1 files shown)


code/game/machinery/kitchen/smartfridge.dm
```diff
@@ -15,7 +75,8 @@
 	var/icon_on = "smartfridge"
 	var/icon_off = "smartfridge-off"
 	var/icon_panel = "smartfridge-panel"
-	var/item_quants = list()
+	var/list/item_records = list()
+	var/datum/data/stored_item/currently_vending = null	//What we're putting out of the machine.
```
PsiOmegaDelta: Not properly qdel()d in Destroy().

code/game/machinery/kitchen/smartfridge.dm
```diff
@@ -162,20 +226,19 @@
...
+			if(S.dry) continue
+			if(S.dried_type == S.type)
+				S.dry = 1
+				S.name = "dried [S.name]"
+				S.color = "#AAAAAA"
+				I.get_product(loc)
+			else
+				var/D = S.dried_type
+				new D(loc)
+				qdel(S)
```
PsiOmegaDelta: Leaves null entries in I.instances.

code/game/machinery/kitchen/smartfridge.dm
```diff
@@ -162,20 +226,19 @@
...
-			qdel(S)
-		return
+	for(var/datum/data/stored_item/I in item_records)
+		for(var/obj/item/weapon/reagent_containers/food/snacks/S in I.instances)
+			if(S.dry) continue
+			if(S.dried_type == S.type)
+				S.dry = 1
+				S.name = "dried [S.name]"
+				S.color = "#AAAAAA"
+				I.get_product(loc)
```
PsiOmegaDelta: Removes the first (last) product from I, not necessarily this specific S instance.

code/game/machinery/kitchen/smartfridge.dm
```diff
@@ -222,35 +285,38 @@
...
+				plants_loaded++
+				var/hasRecord = FALSE	//Check to see if this passes or not.
+				for(var/datum/data/stored_item/I in item_records)
+					if(istype(G, I.item_path))
+						stock(G, I)
+						hasRecord = TRUE
+				if(!hasRecord)
+					var/datum/data/stored_item/item = new/datum/data/stored_item(src, G.type)
+					stock(G, item)
+					item_records.Add(item)
```
PsiOmegaDelta: Very similar to the insertion done above, easily made into a proc to share the code.

PR #1863, opened 2016-06-02
Add a unit test which verifies research designs' materials lists

Leshana:
* Adds a unit test which verifies all research designs' materials lists
* Fixed a bad material discovered by the new unit test!
  * "mutagen" is a reagent, not a material; it goes in the chemicals list.


PJB3005:
A suggestion: make the materials used into defines, if you typo it'll compile error.

PR #1866, opened 2016-06-03
Deadbay: Act One, Scene One: Overdoses cause Liver Damage

Anewbe:
As an effort to make medical more "exciting", I've been tasked with making it more dangerous. Enjoy.

* Overdoses now cause organ damage _and_ toxins damage. Watch your volumes!
* Chems now have an overdose type. Currently only Hyperzine affects an organ other than your liver. More will be added as deemed appropriate.
* Changes the way livers work, having over 30 TOX damage without treating it will cause your liver to take damage, which will in turn cause you to take more TOX damage.
* Chloral Hydrate should be dangerous and/or lethal again, use with caution.
* Dead internal organs will now be rejected by your body, with all the fun that entails.
* Peridaxon will now stave off organ rejection. Transplants should eventually start being rejected again, but revived organs should not.


code/modules/reagents/Chemistry-Reagents.dm
```diff
@@ -86,9 +89,29 @@
...
+				H.damage += 1
+				M << "<span class='danger'>You feel a stabbing pain in your chest!</span>"
+		else
+			M.adjustToxLoss(2)
+		return
+	else
+		if (istype(M, /mob/living/carbon/human))
+			var/mob/living/carbon/human/C = M
+			var/obj/item/organ/internal/liver/L = C.internal_organs_by_name[O_LIVER]
+			L.damage += rand(3,6)
```
Yoshax: I would personally prefer a solid constant number. Not a fan of RNG.

Anewbe: Constant number kills way too hard with heart damage, plus this makes it feel like the heart is tearing itself apart, instead of slowly rotting away.
As for the liver bit, I couldn't figure out a good number that'd make OD scary, but not too scary, and the 3-6 felt right when I was testing it.

SpadesNeil:
Really not a fan of more internal organ damage. Medicine for fixing organ damage is difficult to make IIRC, and surgery will quickly become tedious at this rate.

Anewbe:
Medicine to fix internal organ damage isn't all that hard to make. Yes, it's the most complicated chemical, but most chemists will make at least 2 bottles. I've seen Chemists complain about how easy medical is because of the chems, so there's most certainly an issue.

SpadesNeil:
You make a fair counter-argument.

code/modules/organs/organ_internal.dm
```diff
@@ -190,15 +190,29 @@
...
 			else
 				var/obj/item/organ/internal/O = pick(owner.internal_organs)
 				if(O)
 					O.damage += 0.2  * PROCESS_ACCURACY
+		//So much !!FUN!!, a broken liver won't be able to handle smaller amounts of toxins either
+		if(owner.getToxLoss() >= 10 && !owner.reagents.has_reagent("anti_toxin"))
+			if (src.damage > min_broken_damage)
+				var/obj/item/organ/internal/O = pick(owner.internal_organs)
+				if(O)
+					O.damage += 0.2 * PROCESS_ACCURACY
```
Hubblenaut: This block for damaging the liver will get called in addition to the above one if both the requirements are met, which doesn't seem intended. It's also unnecessary copy-paste of checks that will under circumstances (toxloss 30+) be called twice.
There's surely a nicer solution to this.

code/modules/organs/organ_internal.dm
```diff
@@ -190,15 +190,29 @@
...
+
+
+//	Broken liver means your blood turns to poison
+		if(is_bruised())
+			owner.adjustToxLoss(0.3 * PROCESS_ACCURACY)
+		else if(is_broken())
+			owner.adjustToxLoss(0.8 * PROCESS_ACCURACY)
```
Hubblenaut: Death spiral - damaged liver producing toxins that in turn damage the liver further. The logic here doesn't make too much sense to me.

code/modules/reagents/Chemistry-Reagents.dm
```diff
@@ -55,7 +56,9 @@
 		return
 	if(!affects_dead && M.stat == DEAD)
 		return
-	if(overdose && (volume > overdose) && (location != CHEM_TOUCH))
+	if(overdose && (volume >= overdose) && (location != CHEM_TOUCH))
+//		M.overdosing += volume * 0.2
+//	if(M.overdosing > 0)
```
Hubblenaut: What's with this commented code you add here?

code/modules/reagents/Chemistry-Reagents.dm
```diff
@@ -86,9 +89,29 @@
 /datum/reagent/proc/affect_touch(var/mob/living/carbon/M, var/alien, var/removed)
 	return
 
-/datum/reagent/proc/overdose(var/mob/living/carbon/M, var/alien) // Overdose effect. Doesn't happen instantly.
-	M.adjustToxLoss(REM)
-	return
+/datum/reagent/proc/overdose(var/mob/living/carbon/M, var/alien)
+//	M.overdosing -= REM
+	if(od_type == "heart")
+		if (istype(M, /mob/living/carbon/human))
```
Hubblenaut: ishuman()

code/modules/reagents/Chemistry-Reagents.dm
```diff
@@ -86,9 +89,29 @@
 /datum/reagent/proc/affect_touch(var/mob/living/carbon/M, var/alien, var/removed)
 	return
 
-/datum/reagent/proc/overdose(var/mob/living/carbon/M, var/alien) // Overdose effect. Doesn't happen instantly.
-	M.adjustToxLoss(REM)
-	return
+/datum/reagent/proc/overdose(var/mob/living/carbon/M, var/alien)
+//	M.overdosing -= REM
+	if(od_type == "heart")
```
Hubblenaut: Usage of flags would be faster here. Also cleaner to make use of OOP and have this block be handled by the organs themselves, the implementation here seems quite snowflakey.

code/modules/organs/organ_internal.dm
```diff
@@ -190,15 +190,29 @@
...
 			if (src.damage < min_broken_damage)
-				src.damage += 0.2 * PROCESS_ACCURACY
+				src.damage += 0.3 * PROCESS_ACCURACY
 			//Damaged one shares the fun
 			else
 				var/obj/item/organ/internal/O = pick(owner.internal_organs)
 				if(O)
 					O.damage += 0.2  * PROCESS_ACCURACY
+		//So much !!FUN!!, a broken liver won't be able to handle smaller amounts of toxins either
+		if(owner.getToxLoss() >= 10 && !owner.reagents.has_reagent("anti_toxin"))
```
Iamgoofball: defines

also once again anti-toxin making ODs obsolete

code/modules/organs/organ_internal.dm
```diff
@@ -190,15 +190,29 @@
 	if(owner.life_tick % PROCESS_ACCURACY == 0)
 
 		//High toxins levels are dangerous
-		if(owner.getToxLoss() >= 60 && !owner.reagents.has_reagent("anti_toxin"))
+		if(owner.getToxLoss() >= 30 && !owner.reagents.has_reagent("anti_toxin"))
```
Iamgoofball: defines

also once again anti-toxin making ODs obsolete

code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine.dm
```diff
@@ -310,6 +311,7 @@
 	description = "Hyperzine is a highly effective, long lasting, muscle stimulant."
 	reagent_state = LIQUID
 	color = "#FF3300"
+	od_type = "heart"
```
Iamgoofball: magic strings

code/modules/reagents/Chemistry-Reagents.dm
```diff
@@ -86,9 +89,29 @@
...
+/datum/reagent/proc/overdose(var/mob/living/carbon/M, var/alien)
+//	M.overdosing -= REM
+	if(od_type == "heart")
+		if (istype(M, /mob/living/carbon/human))
+			var/mob/living/carbon/human/C = M
+			var/obj/item/organ/internal/heart/H = C.internal_organs_by_name[O_HEART]
+			playsound(M.loc, 'sound/effects/singlebeat.ogg', 5, -1)	//Don't know how to keep this mob specific
+			C.pulse += 2
+			if (prob (5))
+				H.damage += 1
```
Hubblenaut: Don't access damage directly, use take_damage() that includes necessary checks.

code/modules/reagents/Chemistry-Reagents.dm
Hubblenaut: Also here, take_damage()

code/modules/reagents/Chemistry-Reagents.dm
```diff
@@ -86,9 +89,29 @@
...
-	M.adjustToxLoss(REM)
-	return
+/datum/reagent/proc/overdose(var/mob/living/carbon/M, var/alien)
+//	M.overdosing -= REM
+	if(od_type == "heart")
+		if (istype(M, /mob/living/carbon/human))
+			var/mob/living/carbon/human/C = M
+			var/obj/item/organ/internal/heart/H = C.internal_organs_by_name[O_HEART]
+			playsound(M.loc, 'sound/effects/singlebeat.ogg', 5, -1)	//Don't know how to keep this mob specific
+			C.pulse += 2
```
Hubblenaut: Pulse is updates in handle_pulse() and shouldn't be accessed outside of that proc.

Hubblenaut:
Ugh, having looked this over a bit, it's not a nice implementation for what you intend to do and already has some glaring mistakes in it. This should probably be changed quite some first.

Techhead0:
AFAIK, there currently isn't any downsides to missing a liver, while having one can cause problems with no upsides. This PR just makes the situation worse.

code/modules/reagents/Chemistry-Reagents.dm
Anewbe: That's for a later process, assuming I can make it work.

code/modules/organs/organ_internal.dm
Anewbe: That's intended. If the liver's badly damaged, having a lesser amount of toxins will cause a random organ to take damage, and having greater than 30 toxins will cause it to pick twice.

Iamgoofball:
:-1:

continues the trend of "anti-toxin is magic cure-all"

if you want to make ODs deadly take the goon/tg approach to ODs where each chem that can OD has its own effects

code/modules/organs/organ_internal.dm
Anewbe: Death spiral is intended. TOX has no consequences right now, this creates some.

code/modules/reagents/Chemistry-Reagents.dm
Anewbe: Can you elaborate on that second sentence?

Hubblenaut: You're adding 2 to the pulse every call of overdose(), which makes no sense simply for pulse using flags. Depending on where overdose() is called in the code, it will either break the value or be overwritten by handle_pulse().

Anewbe:
Again, this is step one of many. I was told to keep this PR simple by staff, so I did.

Admittedly there are some bits that can be neater codewise, but this PR isn't going in today, tomorrow, or any time this week, if it goes through. I can clean it up before then, and almost certainly will. I put this up when I did to let people see and discuss it, and so that the staff will have reason to look into it at the next staff meeting.

Missing organs doing nothing is somewhere in the list of things to be added. So is more varied overdoses, and possibly moving overdose() to something in the mob, rather than having it in the reagent

The anti-tox thing I will think about, though. Good catch.

code/modules/organs/organ_internal.dm
Hubblenaut: That seems quite arbitrary to me, and the checks here are still being called twice, while not necessary.

code/modules/reagents/Chemistry-Reagents.dm
Anewbe: Ah, thank you.

code/modules/organs/organ_internal.dm
mwerezak: I don't really see any issue with that. You should be using take_damage() and is_broken() though.

code/modules/organs/organ_internal.dm
mwerezak: It's dynamic

code/modules/reagents/Chemistry-Reagents.dm
mwerezak: Yeah, you would be much better off with `organ/process_overdose()` or something like that instead of this.

code/modules/organs/organ_internal.dm
mwerezak: Yeah, I'd just leave anti_toxin to deal with the toxin and let players deal with the organ damage.

Iamgoofball:
>Missing organs doing nothing is somewhere in the list of things to be added. So is more varied overdoses, and possibly moving overdose() to something in the mob, rather than having it in the reagent

oh fuck no that's snowflake as hell

do it on the reagent

Anewbe:
Yeah, but if it works it opens up a lot of possibilities for it, which is nice.

mwerezak:
What exactly would be the benefit of handling overdose on the mob side? It seems to me like most of the variation would be based on the reagent.

Anewbe:
It would resolve some interesting bugs I ran into while trying to make the overdose linger, rather than being done the moment the chem left the blood.

Iamgoofball:
That's overcomplication for the sake of overcomplication. Just make it end.

Anewbe:
If it lingered, it would be a much milder set of damage, because right now a non-massive overdose does essentially nothing.

Techhead0:
If you want a lingering overdose, it seems the thing to do would be make it generate another reagent that lasts a while and does light tox damage.

Anewbe:
Either way, that whole idea is for another PR.

Arokha:
Um, I thought overdoses already did this. I guess I'm remembering wrong? @_@

Yoshax:
So, what is going to be done with this? Do we have a consensus? Is it good to merge?

Anewbe:
Closing this for now.

PR #1872, opened 2016-06-04
Tweaks FM-2t to possibly be usable

Anewbe:
* Renames FM-2t to burst laser, FM-2t remains in description
* Drops burst laser to weak beams, adds more shots to compensate
* Fixes the single shot mode being unusable

### DIFF
--- a/code/modules/projectiles/guns/energy/nuclear.dm
+++ b/code/modules/projectiles/guns/energy/nuclear.dm
@@ -23,12 +23,12 @@
 
 
 /obj/item/weapon/gun/energy/gun/burst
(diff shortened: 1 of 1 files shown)


mwerezak:
I find it curious that the burst has inaccuracy and dispersion steps. Does the laser have recoil somehow?

comma:
Bright flashes scare people and make them wave gun around crazily!
Obviously recoil is from tiny explosion at the muzzle when air rushes into ionized pathway or some technobabble like that.

EmperorJon:
It gets hot and you have to wave it around screaming. *nod*

PR #1878, opened 2016-06-05
Ammo-type fix

Anewbe:
Corrects the calibers used by our old firearms, so they do the damage they're supposed to after #1877


Yoshax:
The bullpup loads 5.56, did you take that into account?

code/modules/projectiles/ammunition/boxes.dm
```diff
@@ -270,7 +270,7 @@
 	caliber = "a556"
 	matter = list(DEFAULT_WALL_MATERIAL = 1800)
 	ammo_type = /obj/item/ammo_casing/a556
-	max_ammo = 10
+	max_ammo = 30
```
Yoshax: Why is this 30 when the previous number was 20?

code/modules/projectiles/ammunition/boxes.dm
```diff
@@ -377,7 +377,7 @@
 	caliber = "a762"
 	matter = list(DEFAULT_WALL_MATERIAL = 4000)
 	ammo_type = /obj/item/ammo_casing/a762
-	max_ammo = 20
+	max_ammo = 10
```
SpadesNeil: There already exists a 10 round magazine for 762 rounds. This change is redundant.

code/modules/projectiles/ammunition/boxes.dm
```diff
@@ -286,7 +286,7 @@
 
 /obj/item/ammo_magazine/a556m
 	name = "20rnd magazine (5.56mm)"
-	icon_state = "5.56mid"
+	icon_state = "5.56"
```
SpadesNeil: This breaks the icon entirely.

code/modules/projectiles/guns/projectile/automatic.dm
```diff
@@ -51,17 +51,17 @@
...
 	force = 10
-	caliber = "a762"
+	caliber = "a556"
 	origin_tech = list(TECH_COMBAT = 6, TECH_MATERIAL = 1, TECH_ILLEGAL = 4)
 	slot_flags = SLOT_BACK
 	load_method = MAGAZINE
-	magazine_type = /obj/item/ammo_magazine/c762
-	allowed_magazines = list(/obj/item/ammo_magazine/c762, /obj/item/ammo_magazine/SVD)
+	magazine_type = /obj/item/ammo_magazine/a556m
+	allowed_magazines = list(/obj/item/ammo_magazine/a556, /obj/item/ammo_magazine/a556m) //, /obj/item/ammo_magazine/SVD)	//wrong caliber
```
SpadesNeil: If you're going to change this, you need to change the code below, or the code breaks.

SpadesNeil:
This is kind of a mess. I'd suggest just reverting the change to the damage I did and just say screw it. This breaks more things than it fixes.

code/modules/projectiles/ammunition/boxes.dm
SpadesNeil: We already have a 20 round magazine of the 5.56 ammunition as it is. If you're going to change either of them change the bigger one.

Or make the Z8 compatible with the a556m magazines, which is probably the better route.

code/modules/projectiles/guns/projectile/automatic.dm
```diff
@@ -305,7 +306,7 @@
 	icon_state = "p90smg"
 	item_state = "p90"
 	w_class = 3
-	caliber = "5mm"
+	caliber = "9mm"
```
SpadesNeil: Okay that was my bad. I swear I fixed this.

code/modules/projectiles/guns/projectile/automatic.dm
```diff
@@ -104,19 +105,19 @@
...
+	caliber = "a762"
 	origin_tech = list(TECH_COMBAT = 8, TECH_MATERIAL = 3)
-	ammo_type = "/obj/item/ammo_casing/a556" // Is this really needed anymore?
+	ammo_type = "/obj/item/ammo_casing/a762" // Is this really needed anymore?
 	fire_sound = 'sound/weapons/Gunshot.ogg'
 	slot_flags = SLOT_BACK
 	load_method = MAGAZINE
-	magazine_type = /obj/item/ammo_magazine/a556
-	allowed_magazines = list(/obj/item/ammo_magazine/a556)
+	magazine_type = /obj/item/ammo_magazine/c762
```
SpadesNeil: This will make the Z8 spawn with 20 rounds of powerful 762. Is this intentional?

SpadesNeil:
Another suggestion could be add the new carbine 5.56 rifle to Cargo's inventory as I originally intended instead of all of this. That would give a viable counter to the STS without ruining game balance because the Z8 is all station security can obtain.

That, and undo the changes I made to the damages of the bullets. Or, just make them exactly the same.

code/modules/projectiles/guns/projectile/automatic.dm
```diff
@@ -73,10 +73,11 @@
 
 /obj/item/weapon/gun/projectile/automatic/sts35/update_icon(var/ignore_inhands)
 	..()
-	if(istype(ammo_magazine,/obj/item/ammo_magazine/SVD))
```
SpadesNeil: This should have been changed rather than commented out.

code/modules/projectiles/guns/projectile/automatic.dm
Anewbe: Icons shouldn't be broken, I played around with them, and renamed the sprites to fit the appropriate sprite/caliber. Worked when I was testing.

code/modules/projectiles/guns/projectile/automatic.dm
Anewbe: No, this will make the Z8 spawn with 10 rounds of 7.62, as per the change to ammo_magazine/c762.

code/modules/projectiles/ammunition/boxes.dm
Anewbe: Whatever the STS used to use was my intention for this, if it was 20, then the bullpup is (as far as I can tell) pretty much just an STS resprite. If it was originally 30. I'm not sure what exactly you meant there, Yosh.

code/modules/projectiles/ammunition/boxes.dm
Anewbe: The 10 round mag isn't labeled as a 7.62 mag, it's neither coded nor mapped in anywhere that I know of, and I'm pretty sure the Z8 Bulldog uses a 10 round magazine, so I made the 7.62 mag that actually shows up have the proper number.

Anewbe:
I did take that into account, Yosh, from what I understand it's supposed to be a less-powerful stationside version of the STS.

From what I saw when I was actually testing this, it actual puts a lot of the stuff back into order.

I adjusted the sprites to match what I made the various magazines look for, so the icons should be fine.

Yoshax:
I'm honestly just somewhat confused by all the new guns and the proposed changes here. It seems like the end result is different than what it should be.


Right now two guns load 762, the STS, and the Saw. There are 3 'magazines' for 762. The 50 round Saw, 20 round mag and 10 round mag. These do 35 damage.

After this, two guns load 762, the Bulldog, and the Saw. There are 3 'magazines' still. The 50 round Saw, 10 round mag and 10 round mag. These do 35 damage.


Currently two guns load 556, the Bulldog and the Bullpup. There are 2 magazines. A 10 round mag and a 20 round mag. These do 25 damage.

After this, two guns load 556, the STS and the Bullpup. There are 2 magazines. A 30 round mag, and a 20 round mag. These do 25 damage.

Perhaps you see my confusion here. Perhaps not. 

However, you've left a lot undone. Currently the ERT are provided with Bulldogs which load 5.56, with 5.56 magazines. You've made no map changes. They will have incompatible magazines. Same with merc base, they have STSs which load 7.62, with 7.62 magazines. They will have incompatible magazines.

All in all, maybe I'm stupid, maybe not, but I'm confused.

Anewbe:
Map changes will happen when I find those maps and change them, didn't think of that until about an hour ago.

BEFORE #1877 , the 762 rounds (SAW and STS) did 25 damage, and the 556 rounds (Bulldog, of the old guns) did 35. That made the STS work as an assault rifle, and the Bulldog work as more of a Marksman rifle. Currently, the STS is obscenely damaging, and the Bulldog is far less useful than it should be. This fixes that, bringing the guns back to their old values and roles.

I don't know what the bullpup's intended role is, precisely, I _thought_ it was supposed to be a less effective version of the STS for station use, which would imply use of the 556 round. Giving it the current 762 would again make the Bulldog less than useful, and probably overshadow the STS.

Headstaff told me not to worry about the SAW, but it will be even more damaging if it ever gets accessed.

Yoshax:
My confusion lies within the other stuff being affected.

The Bulldog should be more damaging than the STS, I don't care what they load, I don't care if it's puppies and pineapples. As for the Bullpup, it, in my opinion, seems to have no purpose. It can't be more damaging than the Bullbog, and making it the same as the STS makes it an STS. So... I'm whatever. It doesn't actually used anywhere afaik.

After the map changes, this can be merged, or something, I don't know. The rest of the stuff seems to be disrupted by this, but I dunno.

code/modules/projectiles/ammunition/boxes.dm
SpadesNeil: Scroll down a few lines, and you'll see it. Rename it if you like.

SpadesNeil:
You just changed the magazine capacities but didn't change anything else relating to these weapons.

Now the bulldog holds 30 bullets, and is only compatible with the magazine holding 30 bullets. Meanwhile he changes the STS to hold only 10 bullets. Not only do the Bulldog and STS switch roles, but the Bulldog is now an STS on steroids while the STS becomes a nerfed bulldog.

You've also not even touched any of the files relating to traitor menus or autolathe recipes so now everything here is labeled wrongly. A traitor will spawn an STS rifle and a 20 round magazine to go with it only to find he gets a 10 round magazine for a gun that isn't compatible with that ammo.

Pardon my french but this is a clusterfuck. It'd be really better if we just undid what I changed with damage values and go back to the way things were. Screw realism. I seriously regret changing these damage values. Just change the damage values back and leave everything else alone. This is a really round-about way of fixing the problem and doesn't actually fix it.

code/modules/projectiles/guns/projectile/automatic.dm
SpadesNeil: But we still have two types of magazines that fit into the gun, neither of which are the correct icon for 556 mags.

SpadesNeil:
>BEFORE #1877 , the 762 rounds (SAW and STS) did 25 damage, and the 556 rounds (Bulldog, of the old guns) did 35. That made the STS work as an assault rifle, and the Bulldog work as more of a Marksman rifle. Currently, the STS is obscenely damaging, and the Bulldog is far less useful than it should be. This fixes that, bringing the guns back to their old values and roles.

Let's just revert that then, please. You're **absolutely right** that I screwed that up, but I'd feel so much better if we just switched the values back around rather than all of this.

code/game/machinery/autolathe_datums.dm
```diff
@@ -490,37 +490,37 @@
 
 /////// 5.56mm
 /datum/autolathe/recipe/rifle_556
-	name = "10rnd rifle magazine (5.56mm)"
+	name = "30rnd rifle magazine (5.56mm)"
 	path = /obj/item/ammo_magazine/a556/empty
```
SpadesNeil: I uh, kinda fucked this up. ;_; That shouldn't be /empty.

code/modules/projectiles/ammunition/boxes.dm
Datraen: The bulldog has an underbarrel grenade launcher and fires a different round. It's not just an STS resprite.

SpadesNeil:
The sprites look offset.

code/modules/projectiles/ammunition/boxes.dm
Anewbe: Not the Z8 Bulldog, the new Bullpup (automatic/carbine).

code/modules/projectiles/guns/projectile/automatic.dm
Anewbe: The correct icon for 5.56 mags is the icon that used to be 7.62, because of the arbitrary caliber size of the guns. The magazine sprites were made for those specific guns, and labeled with the old caliber. From my own testing of this PR, the magazines used match both the damage and icon expected for the guns in question.

This's keeps all the sprites and damage in order, as it was before any of the recent gun changes. I checked, manually, that the sprites work.

Anewbe:
I cannot seem to say enough times that I've _tested_ the sprites, capacities, and damages for everything I've changed here, so that they match the values, icons, and capabilities that they had before any of the related PRs went through.

Datraen:
I'm not a fan of swapping the rounds for the bulldog and the STS.

SpadesNeil:
> I'm not a fan of swapping the rounds for the bulldog and the STS.

I agree.

> In what manner, Datraen: Sprites, damage, name, or other?

Sprites, name, behavior of the guns, what the guns are based on... I mean the STS is meant to be a space Kalashnikov. Kalashnikovs most certainly do not use 5.56 rounds. The thought of an STS with a STANAG mag jammed into it is just ugly to me.

PR #1879, opened 2016-06-05
Reverts ammo change, fixes P90, changes SVD mag name

SpadesNeil:
Reverts a change I made in https://github.com/PolarisSS13/Polaris/pull/1877 to the 5.56 and 7.62 ammunition types because this has been regarded as a poor decision by me for balance reasons because at the time of the change being decided I was assuming the carbine rifle would be in cargo, and it was not included in cargo.

I'm not about to argue better ways to fix the balance right now because it's more complicated than just changing some damage values, so let's just undo the damage I did. Hah. That was a pun.

Also corrects the P90's ammunition type, as was addressed in
https://github.com/PolarisSS13/Polaris/pull/1878

Also corrects a mistake I left behind when undoing my autolathe changes from the "Guns Guns Guns" update after a staff meeting said not to change that stuff.

Also changes the SVD magazine's name, because it's just a 10 round 762 magazine now, and it's compatible with the STS.

tl;dr this sets the Z8 Bulldog and STS damage values back to what they were before without changing anything else. The guns go back to being exactly what they were before.

### DIFF
--- a/code/datums/uplink/ammunition.dm
+++ b/code/datums/uplink/ammunition.dm
@@ -60,14 +60,14 @@
 	name = "20rnd Rifle Magazine (7.62mm AP)"
 	path = /obj/item/ammo_magazine/c762/ap
 
-/datum/uplink_item/item/ammo/SVD
+/datum/uplink_item/item/ammo/s762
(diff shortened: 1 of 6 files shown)


Datraen:
:+1:

PR #1881, opened 2016-06-06
Prosthetics don't require research

Anewbe:
Having them tied to research made them unfortunately difficult to actually create.


Yoshax:
Not removing it from the cell and eyes too, aren't those required?

Anewbe:
I didn't think there was a req on them, I will correct that later, good catch.

Searif:
Eyes are, cell is not

Anewbe:
According to the code it's both.

PR #1883, opened 2016-06-06
Ports GinjaNinja32's drinks rewrite from Bay

Yoshax:
Ports Baystation12/Baystation12#13131

To-do
- [x] Find out why coffee cups turn into square glasses when they have reagents and fix it
- [x] Convert the protein shakers/port Ginja's thing for them
- [x] Listen to the tears of the damned because no more snowflake icons
- [x] Cry because coffee cups dont work


SpadesNeil:
Coffee cups won't work!? Cannot merge. Broken update. I hate it. :U PR closed.

(I'm kidding. This should be neat once it's fixed.)

EmperorJon:
I'd much rather have different glass types and general fun like this than special snowflake icons for each drink.

SpadesNeil:
Yea the icons changing was always immersion breaking for me.

Hubblenaut:
Will this still work with prices for retail scanners?

Datraen:
:+1:

PR #1902, opened 2016-06-09
Fixes #1830

### LINKED ISSUE #1830: Vermin infestation event spawns mobs in areas that don't exist
This area does not exist on our map!
capture

Spiders!

#### What actually happened
No spiders!

#### Steps to reproduce
Trigger vermin infestation until getting an invalid area.

#### Additional info:
 - **Server Revision**: https://github.com/PolarisSS13/Polaris/commit/dbcf3b18c1f392a7913fb2349f59108eb8797287
 - **Anything else you may wish to add**: Is actually really easy to fix. I can fix it tomorrow.

SpadesNeil:
Although me being me, I want to take it a step further.

This purely just fixes issue #1830 but wouldn't it be so much nicer if we could have more areas, more vermin, more things that can be spawned? 

Adding more areas is simple.

As for more mobs, how about space carp? Those are Fun™.

Someone who's a dev, gimmie some feedback before you merge this if you want more stuff added. Right now I'm just fixing the broken area.

### DIFF
--- a/code/modules/events/infestation.dm
+++ b/code/modules/events/infestation.dm
@@ -102,7 +102,7 @@
 #undef LOC_HYDRO
 #undef LOC_VAULT
 #undef LOC_TECH
-#undef LOC_TACTICAL
+#undef LOC_GARDEN
 
 #undef VERM_MICE
 #undef VERM_LIZARDS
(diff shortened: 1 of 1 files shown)


code/modules/events/infestation.dm
```diff
@@ -6,7 +6,7 @@
 #define LOC_VAULT 5
 #define LOC_CONSTR 6
 #define LOC_TECH 7
-#define LOC_TACTICAL 8
+#define LOC_GARDEN 8
```
SpadesNeil: I wonder; why bother even using these defines? Just for dev QoL? Can't you accomplish the same feat with a simple integer and no define attached to it?

PR #1898, opened 2016-06-08
Adds IDs and Wallets

Arokha:
Adds variants of ID cards for each department and head, and adds several new colors of wallet. The wallets self-color the little ID in the corner based on the ID inserted so they always show the right picture of the ID, without needing to sprite each ID. Some are snowflaked, notably Gold and Centcom, but the rest are generated on the fly based on colors on the ID.

Per the forum thread here:
http://ss13polaris.com/forums/viewtopic.php?f=18&t=1004&p=9572

new IDs

(White wallet not depicted)

PR #1904, opened 2016-06-09
IDs and Wallets

Arokha:
H'okay. Mk2. Let's try this again.

Makes new wallets and IDs. One ID for each department, one with gold highlights for heads. HOP gets a silver one, captain gets the classic gold.

Wallets display a little correctly-colored ID in the corner, assuming the ID sprite exists in wallet.dmi. The sprites for all the department, silver, gold, centcom, antag, etc are pre-made for you. There are several 'extra' ones if someone wants to use cyan, pink, green, etc.

The "Polychromic Wallet" is one that can be recolored. You can pick whatever color you'd like. This is 2560, after all. You can still use the classic orange if you want. If the polychromic one is EMP'd, it flickers weird colors, then returns to normal after a bit.

Sorry that this image is blurry. I have to play spess upscaled because I have a 3200x1800 monitor.


Datraen:
Seeing overwhelming positive response from http://ss13polaris.com/forums/viewtopic.php?f=18&t=1004, merging.

For reference, #1898 was the old PR.

PR #1907, opened 2016-06-10
SmartFridge Code Simplification

Datraen:
Based off of Baystation12/Baystation12#13339 and comments on #1859 

Datums are unified, stocking has a general proc for SmartFridges.

Should resolve any outstanding bugs that may be missed, such as loading named items like fruits or bottles by hand instead of using a plantbag in the case of fruits.

### DIFF
--- a/code/game/machinery/kitchen/smartfridge.dm
+++ b/code/game/machinery/kitchen/smartfridge.dm
@@ -162,7 +162,12 @@
 		icon_state = icon_off
 	else
 		icon_state = icon_on
-	if(contents.len)
+	var/hasItems
+	for(var/datum/stored_item/I in item_records)
+		if(I.get_amount())
+			hasItems = 1
+			break
+	if(hasItems)
 		overlays += "drying_rack_filled"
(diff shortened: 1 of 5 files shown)


code/game/machinery/kitchen/smartfridge.dm
```diff
@@ -226,17 +168,17 @@
...
-	for(var/datum/data/stored_item/I in item_records)
+	for(var/datum/stored_item/I in item_records)
 		for(var/obj/item/weapon/reagent_containers/food/snacks/S in I.instances)
 			if(S.dry) continue
 			if(S.dried_type == S.type)
 				S.dry = 1
 				S.name = "dried [S.name]"
 				S.color = "#AAAAAA"
-				I.get_product(loc)
+				stock(S)
```
PsiOmegaDelta: S is still part of I.instances in both this and the else case below.
In the first case it means that one can magically dispense the object twice (it'll even teleport across the universe to end up at the machine), in the second it means null entries will end up in the I.instances list.

PR #1909, opened 2016-06-10
Heavy weapons changes

Yoshax:
This PR proposes changes to the Laser Cannon & PTR Sniper.

PTR:
- More recoil
- Increased scoped accuracy
- Lowered unscoped accuracy

Laser Cannon:
- Is now singleshot
- Does 90 damage, with 100 armor penetration
- Self-recharges which takes 60 seconds
- Is slightly more accurate

Mounted laser cannon retains the current values.

Possible PTR Changes:
- Higher damage

Possibly changes to cannon:
- Higher recharge time
- Remove self recharge

My goals here are to:
- Make the PTR more reliable.
- Make the Cannon into more of a Laser RPG.

### DIFF
--- a/code/modules/projectiles/guns/energy/laser.dm
+++ b/code/modules/projectiles/guns/energy/laser.dm
@@ -68,13 +68,15 @@ obj/item/weapon/gun/energy/retro
 	fire_sound = 'sound/weapons/lasercannonfire.ogg'
 	origin_tech = list(TECH_COMBAT = 4, TECH_MATERIAL = 3, TECH_POWER = 3)
 	slot_flags = SLOT_BELT|SLOT_BACK
-	projectile_type = /obj/item/projectile/beam/heavylaser
-	charge_cost = 400
-	max_shots = 6
-	fire_delay = 20
+	projectile_type = /obj/item/projectile/beam/heavylaser/cannon
+	max_shots = 1
+	fire_delay = 30
 //	requires_two_hands = 1
 	one_handed_penalty = 6 // The thing's heavy and huge.
-	accuracy = 2
+	accuracy = 3
+	self_recharge = 1
+	recharge_time = 60
+
 
 /obj/item/weapon/gun/energy/lasercannon/mounted
 	name = "mounted laser cannon"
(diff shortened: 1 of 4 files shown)


SpadesNeil:
All of this sounds pretty awesome actually.

I absolutely advise removal of self-recharge from laser cannon. I didn't know that was a thing. It wasn't always a thing.

Mark9013100:
With this change, the laser cannon will be able to destroy limbs in one shot.

SpadesNeil:
https://github.com/goonstation/goonstation-2016/blob/master/sound/weapons/railgun.ogg

New laser sound maybe?

SpadesNeil:
I still disagree with the laser cannon being self-recharging. I think it should be required you have to recharge it normally, but it has such a big battery it takes upward of 60 seconds to finish.

Datraen:
It'd be worth either toning it down, or remove the self-charging feature.

Yoshax:
Toning down the damage puts it back where it is, which is outclassed by every other beam weapon in the game. 

The fact of the matter is, Laser Cannon aren't exactly common. It isn't exactly common for the circumstances for them to exist to happen. If you want the self-charging removed, I guess I can. But I foresee it putting it in a lower place than it is even now.

SpadesNeil:
Removing the self-recharge in exchange for a weapon that can do massive amounts of damage and completely ignores armor is fine by me. That's a guaranteed 90 damage when you land a hit, plus better accuracy than before. You're gonna fuck the dude up. You can finish him off with a crowbar for all I care while he's screaming in agony on the ground with a smoldering hole in his chest.

Yoshax:
Removed the self-recharge because folks don't like it. But I'm all but 100% sure it's going to make the weapon utterly and entirely unused even more so than it already is.

Datraen:
:+1:

zekesulastin:
So, given that pretty much the only source of laser cannon currently in the game is science's protolathe, what's the real benefit of spending diamond on a one-shot weapon when you can use that same diamond with silver to make crates of SMGs (which do a better damage type anyways except for the AP) or just uranium to make adveguns?

I get that it's cool to be able to delay someone wearing the mercenary rig long enough to jab themselves with combat drugs, but.

Yoshax:
The point was to have it as an auto-recharging support weapon. You nail someone for 90 damage, then tag out to your melee or side arm. The point of it was/is that it does the most damage per shot, and in one shot.

Neerti:
The laser cannon is in the uplink as well, @zekesulastin

zekesulastin:
@Neerti: that's almost worse, I think - you can get all kinds of crazy crap with that :P

SpadesNeil:
Honestly the SMG costs are stupid but that's a problem for another day.

Yoshax:
This has been open for a long time.

@Neerti, @EmperorJon, @Datraen.

PR #1910, opened 2016-06-10
Xenobio2 Update #7

Datraen:
- Slime cores would sometimes linger in the slime replicator, allowing for duplication from the same core. The replication and extraction machines null core and occupant variables to avoid this.
- Datums are cleaned up on Destroy() to avoid further issues.
- Slimes being hostile is now more apparent. Before they would be hostile, but remain on the neutral faction, causing no damage to those around them under the same faction.
- Simple mobs out of the hostile subtype would follow their targets after death, this is resolved.
- Additional step added to the mutation proc for a 40% chance of mutation toxin or an extra chemical if the toxin is already present.

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/hostile.dm
+++ b/code/modules/mob/living/simple_animal/hostile/hostile.dm
@@ -5,11 +5,4 @@
 	destroy_surroundings = 1
 	a_intent = I_HURT
 	hostile = 1
-
-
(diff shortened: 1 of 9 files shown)


Yoshax:
Fix any open issues?

Datraen:
None on the tracker, but things I've noticed/had reported to me.

Yoshax:
Toot. Toot.

PR #1913, opened 2016-06-10
Makes Diona Understand All

ParadoxSpace:
All languages, that is. Nymphs can do it, why not the adults? They would still only be able to physically speak one alternate language however.


Spookerton:
No, diona cannot magically understand all languages.
Nymphs haven't had love in a long, long time, and applying the mechanical traits of a vestigial shiny to a proper species is silly.

PR #1920, opened 2016-06-11
Removes activation sounds from glasses that have them

Yoshax:
It's a horrifying noise and sounds like a bloody dentist's drill or something.

### DIFF
--- a/code/modules/clothing/glasses/glasses.dm
+++ b/code/modules/clothing/glasses/glasses.dm
@@ -46,8 +46,6 @@ BLIND     // can't see anything
 			active = 1
 			icon_state = initial(icon_state)
 			user.update_inv_glasses()
-			if(activation_sound)
-				usr << activation_sound
 			usr << "You activate the optical matrix on the [src]."
 		user.update_action_buttons()
 


SpadesNeil:
But I like it. D: It sounds like a set of night vision goggles when you turn them on.

https://www.youtube.com/watch?v=TC0KTDMPx5E

Yoshax:
It's far more shrill, and high-pitched than that.

SpadesNeil:
Is it? Maybe I didn't notice it. Well, let's replace the awful sound, then. Not remove it! :( I like the sound...

Yoshax:
Finding sounds is hard.

SpadesNeil:
Gimmie a minute! (And as usual by 'a minute' I mean _'probably like an hour just hang on I'm doing a thing'_)

SpadesNeil:
Initial research into this states that the sound is really just a fantasy thing and RL goggles don't make that sound. Reality is it's not realistic.

But I think the sound is still cool so I demand we keep it! :U

At least if we don't, I won't feel so bad.

Spookerton:
http://www.freesound.org/people/reg7783/sounds/321800/
While real nogs don't **usually** make noise *(splinter cell is PROBABLY the trope origin, and it's basically just an audio cue for the player)*, capacitor whine can still be a thing. Defibs are one of the better examples of the same sound that's true(r) to life.

Hubblenaut:
And here I thought you mean drinking glasses and got all confused.

Yoshax:
@Neerti, @EmperorJon. @Datraen.

This has been idle for some tmie.

Datraen:
People suggested alternative sounds, which I would prefer to just removing it.

While it's not something that's realistic, it's something that's fun.

Yoshax:
Three sounds have been proposed. One is from Splinter Cell. The other two are just as bad as what we have now. It isn't fun, it's irritating.

Datraen:
We could tone down the sound in the second link that @SpadesNeil posted, http://www.freesound.org/people/Syna-Max/sounds/60345/ , and I think that would be a better route.

Yoshax:
I dislike it, it's ridiculously high-pitched and irritating.

Spookerton:
Perhaps we could just get a http://freesound.org/search/?q=switch sound. :star2:

Hubblenaut:
We definitely need that switch sound for lighting switches, levers ect.

Zuhayr:
Clearly toggling your glasses should trigger Shia Labeouf shouting JUST DO IT.

Serithi:
I'm going with Zuhayr's suggestion myself. :v

SpadesNeil:
Agreed. *nodnod*

PR #1923, opened 2016-06-11
Adjusts cargo categories

Yoshax:
Open to comments/suggestions/etc.


code/datums/supplypacks/robotics.dm
```diff
@@ -0,0 +1,103 @@
...
+	/obj/item/weapon/disk/limb/wardtakahashi,
+	/obj/item/weapon/disk/limb/xion,
+	/obj/item/weapon/disk/limb/zenghu,
+			)
+	cost = 40
+	containertype = /obj/structure/closet/crate/secure/gear
+	containername = "Robolimb blueprints (adv)"
+	access = access_robotics
+
+/datum/supply_packs/robotics/mecha_ripley
```
SpadesNeil: I don't really know why we still have these. These were only ever put in Cargo to begin with because a long time ago the circuit printer wasn't a thing. I'd remove these, honestly.

code/datums/supplypacks/robotics.dm
```diff
@@ -0,0 +1,103 @@
...
+			/obj/item/weapon/book/manual/ripley_build_and_repair,
+			/obj/item/weapon/circuitboard/mecha/ripley/main,
+			/obj/item/weapon/circuitboard/mecha/ripley/peripherals
+			)
+	cost = 30
+	containertype = /obj/structure/closet/crate/secure
+	containername = "APLU \"Ripley\" Circuit Crate"
+	access = access_robotics
+
+/datum/supply_packs/robotics/mecha_odysseus
```
SpadesNeil: I don't really know why we still have these. These were only ever put in Cargo to begin with because a long time ago the circuit printer wasn't a thing. I'd remove these, honestly.

SpadesNeil:
Weapons and Ammo should probably be under the same category for ease of access. If you're buying a box of guns, you're going to want to buy ammo in the same menu IMO.

Atmos looks good.

Costumes and clothes should definitely be its own category. That looks good.

Engineering looks good.

Hospitality looks good. Yea laser tag doesn't belong there. Also remind me to give you guys some banners later so you can play Capture The Flag. :)

Hydroponics and the livestock in it should be two separate categories in my opinion. Just, Livestock should be its own section completely separate from Hydroponics. Livestock aren't grown in trays, and more than just the kitchen uses livestock.

I'd say "materials" should be "building materials" but that's just me nitpicking. The category on its own looks good.

Medical looks good.

Misc looks good. Lotta removals.

Operations category is gone and that's good. Made no goddamn sense.

I had some criticisms of Robotics but otherwise looks good.

Science could probably use MORE things, I just don't know what.

Supplies looks good.

"Voidsuits" I'd rename to "EVA Equipment" and include oxygen tanks in it, and _maybe_ some jetpacks--at least CO2 (but, I also don't often play traitor rounds where people steal jetpacks, so, up to you). Otherwise looks good.

Like I said, weaponry and ammo should be merged. Otherwise good.

code/datums/supplypacks/robotics.dm
Datraen: Requires research, sometimes you don't have R&D and you want/need to build a mech.

SpadesNeil: Well if that's the case, who is going to build the parts to the mech if you don't have anyone to build the circuits? Robotics has a circuit printer, right?

I assume you keep Research and Robotics separated, so if it's research levels you're worried about, I'd just suggest lowering the research requirements for the APLU and Ody if you want the circuits to still be accessible. It takes very little research to make them, though I can see how this would still be a blocker if there simply _is no scientist_ to research them.

Yoshax:
This is done, maybe. IDK.

Datraen:
:+1:

PR #1928, opened 2016-06-12
Move preferences copy_to() new player mobs into the preference_setup datums.

Leshana:
* The /datum/category_item/player_setup_item datums did a good job of organizing the code for loading/saving/editing preferences data, but all of the code that applies preferences to newly created player mobs was still in a single function.
* This change adds a new proc to player_setup_item datums:  copy_to_mob()  which is called from the traditional copy_to() proc, allowing each preferences datum to apply its own character data to the mob.
* This allowes new preferences to easily compartmentalize their new code.
* I also moved all the code for existing preferences from copy_to()  into the copy_to_mob() on their respective preferences datums.

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -251,118 +251,15 @@ datum/preferences
 /datum/preferences/proc/copy_to(mob/living/carbon/human/character, icon_updates = 1)
 	// Sanitizing rather than saving as someone might still be editing when copy_to occurs.
 	player_setup.sanitize_setup()
+
+	// This needs to happen before anything else becuase it sets some variables.
 	character.set_species(species)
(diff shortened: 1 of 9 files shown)


code/modules/client/preferences.dm
```diff
@@ -251,118 +251,15 @@ datum/preferences
 /datum/preferences/proc/copy_to(mob/living/carbon/human/character, icon_updates = 1)
 	// Sanitizing rather than saving as someone might still be editing when copy_to occurs.
 	player_setup.sanitize_setup()
```
PsiOmegaDelta: Noting that you're making a sanitation call here, and then again for each preference datum in copy_to_mob.

Leshana:
That is true.  Its already been sanitized by then, so we don't need to do it twice.

PR #1929, opened 2016-06-12
Fixes #1905

### LINKED ISSUE #1905: FBP Cutting Limbs Off and Items
so like
i'll try my best to explain this like
if you cut off a robot's limbs, and he has items on those limbs, he later cannot interact with said items and you cannot remove them, they do not show up
to be able to whip off my glasses/boots/display monitor
#### What actually happened
I can't
#### Steps to reproduce
give robot glasses, cut robot head off, put robot head back on, watch as glasses are nullspace
#### Additional info:

Arokha:
*screams forever* :crying_cat_face: 

This one almost got me. Organs were defined in two places, probably because of a merge error, and /obj/item/organ/external/head/removed() was defined in THREE places. Pls. Why you do this. PLSSSS.

The actual problem was that calling u_equip is not the thing to call to unequip things. Don't call it. Use drop_from_inventory.


SpadesNeil:
That reminds me. I wonder if the Captain's space suit is still defined in two places because Baystation never seems to fix that.

PR #1931, opened 2016-06-13
Adds actual video calls to communicators

Arokha:
'chu heard me. S'what it does. When you're on a call, there's a "Start Video" button next to each person. When you click it, you actually look through their thing, like a ghost might if they ghostcalled. You have to stay within 1 tile of a video showing communicator to see it.

Multiple people can share one communicator video stream by setting it on a table or whatever, and examining it. There's a message like "It appears to be showing a video: [view]" and you can click view and you can all look at the video. COLLABORATION. SYNERGY. OTHER BUZZWORDS.

I also added the feature of DECLINING voice requests from people. No reason they should sit there forever if you don't wanna talk to them.

Also communicators show up on a special camera list on security consoles, but the same EPv2 network visibility setting turns this off as well if you wanna be all hidey. This does mean that people with the visiblity on serve as sort of roaming AI cameras for the AI as well. So the AI can watch you repair the outside of the station or whatever if you want.

Edit: Oh right, and I added an iconstate for communicators showing video, which is like, a little blue screen with scanlines. So you can know when a good time to walk up and examine them would be.


Arokha:
For fluff context, they were always video calls. I asssummmeeee they are like Skype video calls if you do multi-way video and it has video thumbnails? So, 'Start Video' must just like... enlarge one person's video on the call! Even not doing video still shows you emotes if you're within a few tiles of the screen, I presume because you can see the little video thumbnails.

EmperorJon:
I should probably have asked for a changelog. Oh well. :\

Arokha:
@EmperorJon Sorry. I've never done one. You'd have to teaaaach me. :O

PR #1932, opened 2016-06-13
Fitness Vendor Changes + Fixes

Yoshax:
See commit messages.

### DIFF
--- a/code/game/objects/items/weapons/towels.dm
+++ b/code/game/objects/items/weapons/towels.dm
@@ -12,3 +12,7 @@
 /obj/item/weapon/towel/attack_self(mob/living/user as mob)
 	user.visible_message(text("<span class='notice'>[] uses [] to towel themselves off.</span>", user, src))
 	playsound(user, 'sound/weapons/towelwipe.ogg', 25, 1)
+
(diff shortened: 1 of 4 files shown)


code/game/objects/items/weapons/towels.dm
```diff
@@ -12,3 +12,7 @@
 /obj/item/weapon/towel/attack_self(mob/living/user as mob)
 	user.visible_message(text("<span class='notice'>[] uses [] to towel themselves off.</span>", user, src))
 	playsound(user, 'sound/weapons/towelwipe.ogg', 25, 1)
+
+/obj/item/weapon/towel/random/New()
+	..()
+	color = "#"+get_random_colour()
```
Techhead0: `get_random_colour()` doesn't include the `#`?  That's dumb.

Arokha: I generally agree. I was halfway through writing a proc that did before I realized GRC even exists and with heavy sigh decided to use GRC because _conformity_

Techhead0: I did a search, there are not many uses of it. Would probably not be hard to change them.

PR #1934, opened 2016-06-14
Loyalty implant for HoP

SpadesNeil:
I'm pretty sure this used to be the case and I don't know why it was changed.

Why would you NOT give a loyalty implant to the guy who can give himself (and others) all access, is second in command to the Captain, and has access to everyone's bank accounts?

HoS, HoP, and Captain all have the power to individually single handedly doom the entire station if they decide they want to be jerks at work that day. If NanoTrasen is going to buy loyalty implants for them, they should have implants for all three of them or none at all.

### DIFF
--- a/code/game/jobs/job/captain.dm
+++ b/code/game/jobs/job/captain.dm
@@ -30,13 +30,13 @@ var/datum/announcement/minor/captain_announcement = new(do_newscast = 1)
 			if(2) H.equip_to_slot_or_del(new /obj/item/weapon/storage/backpack/captain(H), slot_back)
 			if(3) H.equip_to_slot_or_del(new /obj/item/weapon/storage/backpack/satchel/cap(H), slot_back)
 			if(4) H.equip_to_slot_or_del(new /obj/item/weapon/storage/backpack/satchel(H), slot_back)
-		H.equip_to_slot_or_del(new /obj/item/clothing/under/rank/captain(H), slot_w_uniform) 
+		H.equip_to_slot_or_del(new /obj/item/clothing/under/rank/captain(H), slot_w_uniform)
 		if(H.age>49)
-			// Since we can have something other than the default uniform at this  
+			// Since we can have something other than the default uniform at this
 			// point, check if we can actually attach the medal
 			var/obj/item/clothing/uniform = H.w_uniform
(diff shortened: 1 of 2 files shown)


Serithi:
Because we're not grimderp?

SpadesNeil:
Grimderp? So, loyalty implanting the guy with access to all of the station's weapons, and implanting the guy with access to the whole station, but we're not implanting the guy who can give himself access to the whole station just because he feels like it?

Respectfully, that doesn't sound like a sensible and consistent way of doing things. Considering this is how it is on other servers I've played, this seems like an oversight.

mwerezak:
I think you missed the point.

SpadesNeil:
Then let's remove them entirely instead if that's the opinion of them.

I like the idea of loyalty implants but that's just me.

Yoshax:
AFAIK, Loyalty implants aren't used at all.

zekesulastin:
@SpadesNeil It actually is a config option in game_options.txt; the implant_loyalty function checks that option.  USE_LOYALTY_IMPLANTS is commented by default.

SpadesNeil:
Well then I'm going to re-open this after all. :U

Serithi:
I'm still failing to see how this would be legally accepted in the slightest.

zekesulastin:
@Serithi It's not on for Polaris or by default in Polaris code: https://github.com/PolarisSS13/Polaris/blob/master/config/example/game_options.txt#L66 - Spades is just trying to fix an inconsistency for when that config option is enabled.

I'm not really for or against it, as it's mainly a balance discussion that does not affect Polaris since we do not have those implants available.

Datraen:
There's no real reason not to, but also no real reason for us to.

I'm good with the change going through for downstreams, as it's not overridden in implementation here.

:+1:

EmperorJon:
If downstream maintenance of loyalty implant mechanics is wanted, then this makes sense, yes.

Spookerton:
There's no real value in this change for Polaris, but it does represent correctness for a feature that we have retained, if only because it can be disabled. Sure why not. :+1:

Yoshax:
I'll probably not run the changelog for this when I do so, cause it'll no doubt confuse people. But all is well.

PR #1936, opened 2016-06-14
Allows cards to go on the ear slot

Mark9013100:
Allows ID cards, emags, and disks to go on the ear slot.
Port of: https://github.com/Baystation12/Baystation12/pull/13382


redstryker:
I thought you meant playing cards and got really excited. :(  But support anyway.

Hubblenaut:
This doesn't actually do as it says. Be aware that the slot_flags field gets overwritten by several card subclasses.

PR #1944, opened 2016-06-15
Makes snowy overlays have directions in the sprites instead of different obj's

Yoshax:


Spacemanspark:
This is rather...
Snowflake.
Ayyyyyyyyyyyy

PR #1947, opened 2016-06-16
Adds mob sprites for the headsets

Mark9013100:
Implements on-mob sprites for the normal headsets.

Normal:

Security:

Medical:

Engineering:

Service:

Command/Science:

Cargo:

Also renames the ERT headset and gives it the unused centcom sprite:

*Also comments out the unused medical research and roboticist headsets.


Yoshax:
I dunno if I'm the only person, but I kinda like them not having sprites. Gives a nice clean look ;-;

Also you probably shouldn't that out, it's not causing any problems, and can be spawned by admins or such for stuffs. Is no real need to comment it out.

Mark9013100:
I'll un-comment it out then.

Datraen:
Is this above or below the hair layer?

Mark9013100:
Goes above.

Anewbe:
If only there was a way to have it both visible and invisible... 
Because I like this in theory, but in practice I don't want that sticking up from my head.

mwerezak:
What happened to the original headset mob sprites? Those were pretty subtle, I liked them.

Mark9013100:
It was inconsistent, as only the default headset had a sprite while the rest didn't(excluding the bowman headsets).

mwerezak:
Just a matter of adding some colours right?

Datraen:
Yeah, I figured it was above based on how headtail decorations are.

I'm against this entirely. It will be an issue with Skrell headtails.

nachomeep:
I'd love this if it were under the hair layer. Also, I just want consistency between the default existing and these not.

PR #1951, opened 2016-06-17
Makes cult sword an actual sword

Yoshax:


Datraen:
:+1: Thought someone would done this by now.

PR #1953, opened 2016-06-18
Adds NT brand prosthetics

Anewbe:
Adds in some spiffy new NanoTrasen brand prosthetics. You can show up with 'em or you can build 'em (don't need a disk, either).

Moves the Morpheus torso back to the old ways, where the limbs actually match the body. The current Morpheus torso is the base for the NT set.


Datraen:
Pictures of these changes/additions?

Anewbe:
nanotrasen


Morpheus is on top, NanoTrasen is on the bottom.

Datraen:
I don't see a problem with them as they are currently presented, so :+1:.

Spacemanspark:
Reminds me of Worker Joes

PR #1955, opened 2016-06-18
510 Improved Overlays

OrbisAnima:
Fixes auto blindness and no click on 510 servers, lessens the lag of
mesons and material scanners, and I suspect the general speed as well.

I only made it work for Polaris (and by extention, to our server, Eros),
and adapted it a bit. ( did readded the Allies cocktail too)

Ported from:

https://github.com/Baystation12/Baystation12/commit/7c2cc890abf840b1a9967bc26c36581301e7229c
https://github.com/Baystation12/Baystation12/pull/13430


PsiOmegaDelta:
You'll also want https://github.com/Baystation12/Baystation12/pull/13430.

OrbisAnima:
@Datraen, I also forgot to mention that this will also re-add the Allies cocktail into the game

OrbisAnima:
@PsiOmegaDelta  thanks, I will check it out and add it!

PsiOmegaDelta:
Actually, just went through the diffs. The linked PR is also included, simply not mentioned.

OrbisAnima:
Yeah, I started this one on Eros, and then we though it would be better to make it upstream. I did have a bit of a problem with the add click catcher, but it seems the system took the commented version. Let me test it again with them enabled and see what happens

OrbisAnima:
sadly it throws an error
 runtime error: wrong type of value for list
proc name: add click catcher (/mob/proc/add_click_catcher)
  source file: hud.dm,387, and I think it's mostly related to the 
client.screen += client.void
wich happens to other instances of screen += void if not removed, any ideas on how to solve it, anyone?

code/modules/client/client procs.dm
```diff
@@ -157,7 +157,7 @@
 	nanomanager.send_resources(src)
 
 	if(!void)
-		void = new()
+		void = void.MakeGreed()
```
PsiOmegaDelta: You've replaced the line ```void = new()``` while it still needs to be there, as otherwise you're calling null.MakeGreed().

OrbisAnima:
I see, between everything I forgot to check what MakeGreed did, thanks Psi!

OrbisAnima:
There we go, tested on local, looking good

code/modules/client/client procs.dm
```diff
@@ -157,7 +157,7 @@
 	nanomanager.send_resources(src)
 
 	if(!void)
-		void = void.MakeGreed()
+		void = new()
```
PsiOmegaDelta: You need both lines :p
```
	void = new()
	void.MakeGreed()
```

OrbisAnima:
Ammended :P

OrbisAnima:
@Yoshax we seem to be good!

Yoshax:
Toot, toot. 510 train.

Datraen:
Question I would have asked if I wasn't otherwise occupied was whether or not this would cause compiling issues on 509, but since it's merged, it's a moot point.

OrbisAnima:
Sadly I didn't tested it on 509, wich in restrospective would had been a good idea...

PsiOmegaDelta:
Bay is running on 509 and things seem to be running, superficially at least, just fine.

PR #1973, opened 2016-06-25
Makes Buildmode Great Again

Neerti:
Adds the following;
Control-click to copy object type to adv. buildmode
Ability to instantly build a rectanglar room with custom walls and floor by clicking two points.
Ability to create two ladders, linked to each other, and have correct icons, by clicking two points.
Ability to instantly move one object into another object by left/right clicking.
Ability to make anything glow, with custom range, intensity, or color, or to reset glowing.
Partial pathnames for all buildmode path inputs.
Adds spans.


EmperorJon:
We're going to build a wall.
And it'll be really easy, thanks to this update!

PR #1985, opened 2016-06-27
Removes the password to delist the server

Yoshax:
We really need a better way of doing this.

### DIFF
--- a/code/hub.dm
+++ b/code/hub.dm
@@ -1,8 +1,8 @@
 /world
 
 	hub = "Exadv1.spacestation13"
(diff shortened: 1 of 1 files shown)


Spookerton:
We're going to see if -invisible works first.

Yoshax:
PR's here if you need it.

Anewbe:
:+1: :+1:

Spookerton:
Invisible does it, but we should probably do this anyway so that our lewd downstreams don't accidentally cry tears of blood if they have odd startup parameters.

PR #1987, opened 2016-06-27
Increases max record and book length to higher levels

Yoshax:
Long live the snowflakes.

### DIFF
--- a/code/__defines/misc.dm
+++ b/code/__defines/misc.dm
@@ -91,8 +91,9 @@
 
 // Setting this much higher than 1024 could allow spammers to DOS the server easily.
 #define MAX_MESSAGE_LEN       1024
-#define MAX_PAPER_MESSAGE_LEN 24576
+#define MAX_PAPER_MESSAGE_LEN 6144
 #define MAX_BOOK_MESSAGE_LEN  24576
+#define MAX_RECORD_LENGTH	  24576
 #define MAX_LNAME_LEN         64
 #define MAX_NAME_LEN          52
 
--- a/code/__defines/misc.dm
+++ b/code/__defines/misc.dm
@@ -91,8 +91,8 @@
 
 // Setting this much higher than 1024 could allow spammers to DOS the server easily.
 #define MAX_MESSAGE_LEN       1024
-#define MAX_PAPER_MESSAGE_LEN 6144
-#define MAX_BOOK_MESSAGE_LEN  12288
+#define MAX_PAPER_MESSAGE_LEN 24576
+#define MAX_BOOK_MESSAGE_LEN  24576
 #define MAX_LNAME_LEN         64
 #define MAX_NAME_LEN          52
 
--- a/code/modules/admin/verbs/custom_event.dm
+++ b/code/modules/admin/verbs/custom_event.dm
@@ -7,7 +7,7 @@
 		src << "Only administrators may use this command."
 		return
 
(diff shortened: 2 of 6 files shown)


SpadesNeil:
Oh so that's where this is handled. I was once annoyed by my books getting truncated by turning paper into a book.

Datraen:
If you need more than 12k characters in a book, you probably shouldn't be making it into a book in SS13.

:-1:

Spookerton:
What is the value of 24KiB papers

Techhead0:
I was curious and/or bored, so I did the math. Maxing out the character count on letter-size paper (8.5 × 11 in or 215.9 × 279.4 mm) would require a ~4 point font size. That's letters with the height of 1/18 of an inch or 1.4 mm.

Yoshax:
It was done for records mostly, to allow people who want to write more and have more detail to do so. I had some complaints directed my way that people did not have enough space to write all that wish to, which is somewhat problematic. In addition, I have had complaints that people can't make long enough books. So this is the product of such.

Yoshax:
If preferred, I can make records use book length and return paper length to it's current value. Which allows for longer records and books while retaining the current paper. Although I'm not sure how that will play with printing off records that are longer than the paper length.

Datraen:
Perhaps splitting paper and record lengths may be a better alternative.

Yoshax:
It'll perhaps pose problems when printing them off and suchlike, I'm not certain.

Datraen:
Researching paper code suggest that it checks against a variable called "free_space", which inherits the max paper size when cleared and initialized, and so it should not prevent larger records from existing.

Yoshax:
Converted records to use their own MAX_RECORD_LENGTH and converts most uses of MAX_BOOK_MESSAGE_LEN to use  MAX_PAPER_MESSAGE_LEN and leaves Books as the only thing to use MAX_BOOK_MESSAGE_LEN.
 
Returns MAX_PAPER_MESSAGE_LEN to the current live values, while setting MAX_RECORD_LENGTH to my new proposed value. In addition, MAX_BOOK_MESSAGE_LEN retains its new proposed value as it is only used for books. And saying they shouldn't be writing long books is a terrible argument @Datraen. Allow people to write longer books so our library actually has life.

Datraen:
@Yoshax , having a library full of free-use works doesn't constitute breathing life into it, and if it's such a long user-created book, it should also be archived elsewhere, instead of just a series of books on the server, in the event of a failure.

Yoshax:
It's better to be able to put your book into one book, rather than having to split it into volumes.

SpadesNeil:
I kind of agree with Yoshax here. Is there any _technical_ reason why we shouldn't do this?

Datraen:
Technically, no. Otherwise, it's a mechanical stop for people that want to try to grief the server with overly-large records/pages/etc to take up space and processing slots.

SpadesNeil:
That's kind of a really elaborate way to grief.

Yoshax:
If people are spending their time to grief the server by making books,  then I don't think there's much to worry about. There's far worse things you could be doing.

Datraen:
I'll concede to book sizes. Pages are much easier to create in large quantities very quickly.
For reference: The current restrictions on pages would result in a max-length page taking up ~6KB of memory, and a max-length book taking up ~12KB of memory, in readable characters alone.

Yoshax:
Paper length is unchanged.

Datraen:
I'm just stating the current settings. With the changes proposed in the PR, max-length books, records will be 24KB.

PR #1991, opened 2016-06-27
Clicking a stack no longer transfers one sheet at a time.

aztectornado:
Know how when you click a sheet stack (eg metal) in your other hand, you'd pull one sheet off, and then clicking again would add one extra sheet, so every time you needed only a small amount, you'd have to sit there spamclicking against the server ticks?

No more!

This adds a simple little popup so that when you click a sheet stack in your other hand, you'll instead get asked how many sheets you want to split off from the main stack, skipping tedious spamclicking! Since you now no longer need to transfer them one sheet at a time, this also makes it so clicking a sheet in your offhand with a sheet in your primary hand will merge the two stacks together, just as if you clicked the sheets while they were lying on the ground. 

This is useful for times when you maybe don't want to take /all/ of the metal from engineering storage, or if you want to divide up the spoils of mining between robotics and RnD to some specific ratio. The input defaults to 1 when first opened though, so if you're in a hurry and only want one diamond to throw into the deconstruction machine, you can just click or hit enter after it opens.

### DIFF
--- a/code/game/objects/items/stacks/stack.dm
+++ b/code/game/objects/items/stacks/stack.dm
@@ -298,14 +298,16 @@
 
 /obj/item/stack/attack_hand(mob/user as mob)
 	if (user.get_inactive_hand() == src)
-		var/obj/item/stack/F = src.split(1)
-		if (F)
-			user.put_in_hands(F)
-			src.add_fingerprint(user)
-			F.add_fingerprint(user)
-			spawn(0)
-				if (src && usr.machine==src)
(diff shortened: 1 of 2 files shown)


Yoshax:
Can you add a changelog, please?

aztectornado:
Done. Sorry I forgot that. Places I used to code for almost never did changelogs. x_x

Belsima:
Should remove the 'needs changelog' tag.

PR #1992, opened 2016-06-27
Adjusts and attempts to normalize cargo costs

Yoshax:
See diff for full list of changes.

- Many things are less expensive and no longer have ridiculous prices.
- Uniforms/Clothing crates have been normalized to 10 points across the board
- Job supply crates are 20
- Head job supply cost 50

Future plans for upcoming PRs:
- Adds a new categories for job supply and clothing
- Ensure every job has a crate
- Add more crates in general

Potential ideas for the future:
- Possibly include a special requests option that contacts the admins.


Datraen:
:+1:

PR #1984, opened 2016-06-26
Taste and Chemistry rebrand

aztectornado:

PR #1995, opened 2016-06-27
Add Taste

aztectornado:
(Dammit git, why and HOW did you make another new commit for something already done?! And why do you say /I/ did it?)

Whatever. I tried my best to get that commit to go away, but git rebase keeps picking up another one every time I get it to drop the one it has. Best I could do is pick the shortest one available so the conflict it causes is easy to fix. x_x

Anyway, this adds in Bay's new taste system, just like in #1984, but WOW am I glad I redid it from the top. Somehow all the changes that they made to Snacks.dm didn't get added or even merged properly, which is a shame since they seem to have cleaned up how nutriment gets added, and made it so snacks and foods have unique flavors, too.

This time since I wasn't half asleep, I think I managed to catch all the issues from before, and then a few more. Lemme know if you find any more though, because I'm still human!

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Toxins.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Toxins.dm
@@ -1,8 +1,8 @@
 /* Toxins, poisons, venoms */
 
 /datum/reagent/toxin
-	name = "bitterness"
-	id = "bitterness"
+	name = "Toxin"
+	id = "toxin"
 	description = "A toxic chemical."
 	taste_description = "bitterness"
 	taste_mult = 1.2
--- a/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Toxins.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Toxins.dm
@@ -1,10 +1,10 @@
 /* Toxins, poisons, venoms */
 
 /datum/reagent/toxin
-	name = "Toxin"
-	id = "toxin"
+	name = "bitterness"
+	id = "bitterness"
 	description = "A toxic chemical."
-	taste_description = "toxin"
+	taste_description = "bitterness"
 	taste_mult = 1.2
 	reagent_state = LIQUID
 	color = "#CF3600"
@@ -286,7 +286,7 @@
 	name = "Soporific"
 	id = "stoxin"
 	description = "An effective hypnotic used to treat insomnia."
-	taste_description = "toxin"
+	taste_description = "bitterness"
 	reagent_state = LIQUID
 	color = "#009CA8"
 	metabolism = REM * 0.5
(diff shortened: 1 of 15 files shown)


code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Toxins.dm
```diff
@@ -1,9 +1,11 @@
 /* Toxins, poisons, venoms */
 
 /datum/reagent/toxin
-	name = "Toxin"
-	id = "toxin"
+	name = "bitterness"
```
Yoshax: Why does this need renamed?

aztectornado: ....Fuck, thank you for pointing this out. Last time I managed to catch that. (This actually breaks a few things)

aztectornado:
Dangit, quit breaking this! ;A;

Spookerton:
Generalized :-1: on the basis of text-based description of flavor being mediocre at best and a bad joke at worst. Does not add to user experience, do not want.

Neerti:
How does it not add to the experience, when the alternative is literally nothing?

aztectornado:
Have you /read/ the descriptions in this game lately? They're all mediorce at best, jokes at worst. 

...Also, ^what Neerti said. It might not be perfect, but it adds something.

EmperorJon:
Descriptions aren't good in a lot of places, yes. That doesn't mean we should make it worse by giving the user "It tastes like a hull breach in your mouth" every time they drink space-up.

Spookerton:
@Neerti Nothing is often better than something when we're talking in terms of a collaborative roleplaying experience. Adding specific definitions of flavor to things reduces the flexibility available to the players in describing them themselves, and when they're as weak as most of these are ...

The code isn't bad. Just the feature.

Yoshax:
I'm a fan of the system after testing with it, but the descriptions are... A tad horrifying. I'm not sure what would be better, it's hard to describe what something tastes like.

PsiOmegaDelta:
Have everything taste of chicken?

ParadoxSpace:
There is a good cooldown for taste messages, i.e. not every sip of space up will tell you what it's like.

Hubblenaut:
Spookerton, while not saying this is the best way to implement it, having a description of how something tastes like is a missing element in the game.
As it is right now, a character can be given a flask and can't know the difference of whether they just drunk water or vodka.

I wish these taste descriptions didn't make use of a double space and had punctuation though.

Iamgoofball:
did you address any of the comments from the previous PR?

Neerti:
It's been conflicted for quite some time, and it doesn't seem any work has been done in a few months, however the public does want something like this to be made, so I'm going to close this for now, and in the future, someone (me if nobody else decided to do it) will write a variant of this with different descriptions, as that seemed to be the major point of contention.  It's quite unfortunate that this had sat here for months, and for that I'm sorry.

Techhead0:
I guess it's true, then.
Polaris players still have no taste.

PR #2008, opened 2016-06-30
Allows Circuit-Brains (Digital) to be chosen as an FBP brain-type.

Mechoid:
I've opened this so it is available for staff discussion, and not gathering dust sitting in my repo-folder.

This will allow FBP's to choose to have a circuit-based brain in character set-up, similarly to MMI or Positronic brains, through the selection of the 'Digital' option in brain selection.

### DIFF
--- a/code/modules/organs/organ.dm
+++ b/code/modules/organs/organ.dm
@@ -259,6 +259,10 @@ var/list/organ_cache = list()
 	min_bruised_damage = 15
 	min_broken_damage = 35
 
+/obj/item/organ/proc/digitize() //Used to make circuit organs like robobrains
+	robotize()
+	src.status &= "digital"
+
 /obj/item/organ/emp_act(severity)
 	if(!(robotic >= ORGAN_ROBOT))
 		return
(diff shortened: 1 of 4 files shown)


ParadoxSpace:
I'm not sure how to properly express my love for you right now.

Spookerton:
:+1:

code/modules/organs/organ.dm
```diff
@@ -259,6 +259,10 @@ var/list/organ_cache = list()
 	min_bruised_damage = 15
 	min_broken_damage = 35
 
+/obj/item/organ/proc/digitize() //Used to make circuit organs like robobrains
+	robotize()
+	src.status &= "digital"
```
Zuhayr: What in the heck is this supposed to do? Status is a bitfield.

Mechoid: I believe I may have mixed it up with a different 'status', then. Testing without that now.

Zuhayr: &= is also not what you want, that's a bitwise operation. If you were trying to flip the digital status bit you would need |= and if you were just straight up adding it you would need +=.

Mechoid: Ah, alright. Thanks for helping sort that out.

PR #2015, opened 2016-07-01
Dry-dock phase for Technomancer

Neerti:
As time goes on and on, it's becoming increasingly harder to avoid merge conflicts due to the gamemode's code branching out.  Thus, I've figured that adding this in now would make it easier to develop for.  It would also make it much easier to run tests by putting it on an actual server, yet having full control of when the tests should start and end.

Technomancer and Wizard are not votable at this time, as the former replaces the latter, and the former is not polished enough for a release at this moment.  Some things to note about the code;

* This regurgitates a new spell hud system, called the ability master, taken from the spell master.  Spell master remains where it is right now due to constructs depending on it, however consider the spell master to be depreciated, the ability master instead replacing it.  The reason for replacement was to make a hud system that was more generic than the spell master system, which worked for only wizard spells.  Ability master works can contain any 'type' of ability, both object and verb based.  Changeling and Technomancer abilities are currently included.  One could expand the system to include things like Ninja modules, cult abilities, etc.
* Old spell system is also staying because constructs (I'm slightly mad at them now).
* Being covered in water gives a 1.5 siemens_coefficent malus.
* Fire stacks can now go below 0, as was previously intended based on the minimum stack define.
* Support for icy floors is available.
* Accuracy bonus/penalty variable is available.
* Reflected projectiles can no longer be reflected, to avoid infinite reflection loops.
* Probably a lot more stuff I'm forgetting.


Neerti:
Also Travis will probably fail due to the fact that my comments are not formatted in a specific way.

Yoshax:
*stares at line changes and shudders*

code/game/gamemodes/technomancer/devices/shield_armor.dm
```diff
@@ -0,0 +1,84 @@
...
+	var/datum/effect/effect/system/spark_spread/spark_system = null
+	var/block_percentage = 75
+
+/obj/item/clothing/suit/armor/shield/New()
+	..()
+	spark_system = PoolOrNew(/datum/effect/effect/system/spark_spread)
+	spark_system.set_up(5, 0, src)
+
+/obj/item/clothing/suit/armor/shield/Destroy()
+	spark_system = null
```
PsiOmegaDelta: You still qdel() pooled objects, otherwise they'll never return to the pool.

Yoshax:
Conflicts present. Cannot merge.

In addition, you noted on TS this is finished just not polished. Can we not go live with it as being votable? Will probably aid in finding all the bugs and such easier.

Neerti:
I mentioned that it wouldn't be votable when this is merged.

Yoshax:
Indeed, as noted in the original post. I am asking why we cannot put it live as a votable gamemode.

Neerti:
I'd prefer to be able to do controlled tests in a real environment with a lot of players, so I can tweak and balance everything, and be able to see how everything goes personally, verses vague feedback given by people when I'm not there.

A lot of the art is also not finished, and it'd look pretty bad for the buttons to all be blank.

PR #2020, opened 2016-07-01
Makes flashes confuse and blind the person instead of stunning

Yoshax:
We have removed most of the previously existing instant stuns but kept flash as such. In its current state you can flash someone and pin them onto the floor, and there is basically nothing they can do. There is no reason for it, and it should go.

Flash will now:
Conufse for 5 seconds
Blind for 5 seconds
Blurry for 10 seconds

### DIFF
--- a/code/game/objects/items/devices/flash.dm
+++ b/code/game/objects/items/devices/flash.dm
@@ -74,13 +74,17 @@
 		if(M.stat!=DEAD)
 			var/safety = M:eyecheck()
 			if(safety <= 0)
-				var/flash_strength = 10
+				var/flash_strength = 5
 				if(ishuman(M))
 					var/mob/living/carbon/human/H = M
 					flash_strength *= H.species.flash_mod
-				if(flash_strength > 0)
-					M.Weaken(flash_strength)
-					M.flash_eyes()
+
+					if(flash_strength > 0)
+						H.confused = max(H.confused, flash_strength)
+						H.eye_blind = max(H.eye_blind, flash_strength)
+						H.eye_blurry = max(H.eye_blurry, flash_strength + 5)
+						H.flash_eyes()
+
 			else
 				flashfail = 1
 
--- a/html/changelogs/Yoshax - flashnerf.yml
+++ b/html/changelogs/Yoshax - flashnerf.yml
@@ -1,0 +1,36 @@
+################################
+# Example Changelog File
+#
+# Note: This file, and files beginning with ".", and files that don't end in ".yml" will not be read. If you change this file, you will look really dumb.
+#
+# Your changelog will be merged with a master changelog. (New stuff added only, and only on the date entry for the day it was merged.)
+# When it is, any changes listed below will disappear.
+#
+# Valid Prefixes: 
+#   bugfix
+#   wip (For works in progress)
+#   tweak
+#   soundadd
+#   sounddel
+#   rscadd (general adding of nice things)
+#   rscdel (general deleting of nice things)
+#   imageadd
+#   imagedel
+#   maptweak
+#   spellcheck (typo fixes)
+#   experiment
+#################################
+
+# Your name.  
+author: Yoshax
+
+# Optional: Remove this file after generating master changelog.  Useful for PR changelogs that won't get used again.
+delete-after: True
+
+# Any changes you've made.  See valid prefix list above.
+# INDENT WITH TWO SPACES.  NOT TABS.  SPACES.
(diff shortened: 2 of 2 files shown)


Datraen:
Makes insta-stun the way of the dodo, :+1:

Spookerton:
I'd be totally willing to give this a try.

I'm not sure if I agree with totally nixing stuns in general, because it moves away from being able to rely on anything but batons and tasers to actually **stop** someone; a move like this means that flashes in particular no longer prevent someone from running - it just reduces their ability to run effectively - and doesn't stop them from attacking effectively at all, which is probably half of what flashes are used for preventing in close quarters.

It's not a problem I think is easily fixed though, and pepper spray can still fill in the role. Just less effectively. Whiiiich I suppose could be a positive.

Yoshax:
Instant stuns are never going to be a productive part of the game. They are anti-rp, and anti-fun. The only one that remotely makes sense, and could remain is the parapen because of it being a traitor item and costing TC.

Spacemanspark:
Never liked instastun mechanics myself.
We have these on Paradise (Minus the blurring part) for our flashes. It doesn't render them unusable, but it makes them more of a side tool rather than an instastun device that fits into your pocket.

Spookerton:
You're not wrong that instant stuns are a mechanical win button, but that doesn't mean that there's an alternative available that won't aggravate people even more. Removing the option to use something like a flash in its current format means that people will just jump straight to whacking and zapping. Regardless, though, I said: totally willing to give this a try.

Spacemanspark:
One thing I will ask; does this have the AoE flash to it? Where you click a flash in hand and it flashes anyone in a 3x3 area around the user? The AoE flash is also weaker.

EmperorJon:
Doesn't confusion etc. etc. have no impact on attack accuracy right now? I think if this change went in you'd want to alter that, because right now you can leave people blundering around running into things and hurting themselves whilst also accurately ripping the shit out of you with bursts of assault rifle fire.

Yoshax:
If someone wants to go refactor how mobs are affected by various status effects, they're free to.

zekesulastin:
It seems like that would be a pretty important prerequisite for this, then.

Anewbe:
I don't think this affects mounted or portable flashes, is that intended?

Hubblenaut:
Confusion lowers accuracy with weapons, but otherwise has no effect on combat to my knowledge.

Yoshax:
Intended @Anewbe

Anewbe:
I'd be willing to give it a try, then.

FenodyreeAv:
You can still resist while flashed.
It's getting harder and harder to restrain someone safely.
They only have to get lucky once, because they'll kill you. An officer has to be lucky every time.
Very big -1, confused status doesn't do shit.

Hubblenaut:
Can port Baystation#13381 and add a modifier for being confused.

Yoshax:
Both have been @Hubblenaut https://github.com/PolarisSS13/Polaris/pull/1956

SparklySheep:
I was always wary about doing this because of the hard counters to flashes (sunglasses, welding helmets, etc.) but since most of those have been removed from both the map and from the loadout, I'd approve of this. Instant stuns are not a fun game mechanic especially with the reduced speed on this server compared to others.

I give my cautious :+1:

Anewbe:
Still willing to at least give this a try.

kimiza:
I would say this is a bad idea. As is, the flash has an incredibly common hard-counter. A welding helmet, goggles or glasses(which are fairly easy to obtain) all utterly negate the flash, and further nerfing it will probably relegate it to the stage where it is completely unused.

Yoshax:
@Neerti @EmperorJon 

This is a thing. This thing requires attention.

EmperorJon:
So now the accuracy penalties are through we're totally fine for this, right? It still has Requires Staff Input but idk if there was anyone who said no? I'm certainly happy to give it a try.

Anewbe:
#2134 was merged for this, yes. Still willing to give this a try.

Neerti:
We can revert if it turns out that the change goes poorly.

EmperorJon:
Well merge it then! :P

Woodratt:
As someone who has been on both ends of this. It does not work.

People can still disarm spam, people can use baseball bats. People can still melee at close range. 

Make them do halloss or something, cause if you're close enough to use a flash then you're going to get punched, pushed down and dead.

Rand0mname:
So basically as a detective I now have to shoot every threat I run into? I can live with that.

PR #2023, opened 2016-07-02
Allows hardsuits to carry backpacks in their suit storage

Yoshax:


Anewbe:
:+1:

Spookerton:
:100: :+1:

Neerti:
I'm going to take the controversial stand and say that the lack of a convenient inventory slot should remain, as RIGs are really strong as it is.  Getting rid of this limitation tips the balance in a way that makes the choice of using a RIG from a hard choice between having easy inventory or having strong EVA armor with extra toys attached, to being a no-brainer.
Personally a :-1: from me.

EmperorJon:
A massive :+1: :+1: :+1: :+1: :+1: for me combined with a :-1: for balance concerns so abstaining.

Datraen:
:-1: On concerns of balance.

Having storage on a RIG would be nice, but there would have to be sacrifices if they're going to have suits with tons of toys built into them + extra storage that made voidsuits preferable.

ursamedium:
I like this, it's always a tossup on 'do I want the big toys, or do I want a bag', but balance.

Might I offer that, if a whole backpack is determined to be too much, instead perhaps the ability to hold a smaller general 'suit storage compartment' that's roughly box sized or so?

SpadesNeil:
>Might I offer that, if a whole backpack is determined to be too much, instead perhaps the ability to hold a smaller general 'suit storage compartment' that's roughly box sized or so?

Actually, yea. Why don't we do that? A cargo compartment specifically designed for hardsuits, but has a fraction of the capacity of backpacks.

Someone already changed hardsuits in Vorestation to allow backpacks because it was too useful for us to care about balance, but that's an even better solution that doesn't break balance.

Anewbe:
I'm working on a set of smaller backpacks that'd probably work for this purpose, if it is decided that Hardsuits need more storage (The CE one certainly could). They've got a max_storage size of 15 and a max_weight of 3, subject to change.

Spookerton:
I dunno. I've occasionally allowed some rig chests to hold capacity-edited ERT packs (because they look neat), with enough capacity under the old system for 3~4 objects of the same size as backpacks (large?) which have 7.

Either way this turns out, I'm all for storage for rigs. I find it a little ridiculous that you can't just hang a satchel strap off your shoulder/pointy bits, no matter what kind of bulky suit you're wearing, but if we care that much about balance *(although I'm interested in circumstance, since I can only see it being something to discuss for the mercenary and ninja suits, really)* a smaller container specific to rigs would be totally fine too.

Yoshax:
The problem is @Spookerton, that was designed for old balance and storage system, where all items took up the same space and the larger items would never fit into bags. These days you can put your roller bed or laser cannon into your bag. It takes up a lot of space, but it fits into your bag. It no longer requires a whole suit storage space because we reworked storage.

Yoshax:
@Neerti @EmperorJon 

This is a thing. This thing requires attention.

EmperorJon:
Definitely not something I'm comfortable with making a decision on. Can we get Neerti and top level staff to decide?

SpadesNeil:
My own opinion: I say there should be specialized backpacks with less capacity explicitly for this purpose.

Neerti:
I've already stated my opinion above.

PR #2028, opened 2016-07-03
Hardsuit gauntlets cover gloves

Anewbe:
Adds a new glove type called gauntlets, that cover gloves. All RIG gloves are now gauntlets.

### DIFF
--- a/code/game/objects/items/weapons/handcuffs.dm
+++ b/code/game/objects/items/weapons/handcuffs.dm
@@ -60,7 +60,7 @@
 		user << "<span class='danger'>\The [H] needs at least two wrists before you can cuff them together!</span>"
 		return 0
 
-	if(istype(H.gloves,/obj/item/clothing/gloves/rig) && !elastic) // Can't cuff someone who's in a deployed hardsuit.
+	if(istype(H.gloves,/obj/item/clothing/gloves/gauntlets/rig) && !elastic) // Can't cuff someone who's in a deployed hardsuit.
 		user << "<span class='danger'>\The [src] won't fit around \the [H.gloves]!</span>"
 		return 0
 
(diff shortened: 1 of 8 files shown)


Datraen:
:+1:

Hubblenaut:
Wearing them over insulated gloves will not actually protect the user from electricity, which is kind of counter-intuitive.

FenodyreeAv:
Would be neat if arm guards were considered gauntlets.

Anewbe:
That's intended Hubble, staff didn't want it to work that way for balance reasons. Plus, most if not all RIGs have a siemens of 0.2 or higher, so you're already shock resistant. The RIGs that are most likely to play around with zappy stuff either already have or will soon have insulated gauntlets anyway.
Also, if it's a bigass piece of metal and stuff that covers a whole glove, it's likely to make contact with skin, or some other part of the suit that's not necessarily insulated, so whatever.

Feno: that's the exact reason I added the base gauntlet item, rather than just adding the stuff to glove/rig/, but people had reservations about making the armguards work that way. If they feel the need to explain further, they can and probably will here. Plus the base gauntlet has some odd behaviors that the hardsuit version does not, and I cannot for the life of me figure it out right now, so it's probably best that the armguards remain as is for the time being.

Spookerton:
:+1:

Anewbe:
Tag this with a DNM for now please, found a critical issue in testing that wasn't happening yesterday.

Anewbe:
Alright, only bug I have now is as follows:
Using the equip hotkey (e) with gauntlets, while not wearing gloves, causes the gauntlets to enter the back slot, instead of the hands. This does not happen with magboots.

Anewbe:
Ready to go as is, base item doesn't actually exist outside of adminnery.

Screemonster:
On the insulated glove thing - fun fact: people working on high-voltage lines wear conductive suits specifically so the current passes through the suit instead of the person wearing it.
This is, of course, assuming that rigsuits are actually metal and not some space-age polymer or something.

PR #2029, opened 2016-07-03
Adds JackeryFox's Industrial Welder sprites

Anewbe:


Hubblenaut:
Looks a bit weird, to be honest

Yoshax:
Seems kinda odd it goes normal sprite -> this sprite -> normal sprites, for the three levels of welder.

Anewbe:
The only hugetank welder I can think of is in Robotics, and I think that one deserves a sprite of its own, because it holds 4 times as much fuel as the base. I was considering adding the new one to that as well, but I thought it still didn't look quite right.

ursamedium:
ind ind2 indh indh1
Since there wasn't a sprite preview in the pr

SinTwo:
I'm am quite opposed to this new sprite; it looks like there's a growth or something on the handle.

SpadesNeil:
It looks fugly. :-1:

Anewbe:
225px-neutral_president
Really fine with either sprite, seems to be mostly negative reactions.

JackeryFox:
Aye don't worry, I suck at everything anyways.

mwerezak:
The sprite would be improved IMO by reducing the amount of clutter decoration on the welder. More detail is not always better, especially when you're restricted to 32 x 32.

ursamedium:
It's a little murky, color-wise, imo, too. The yellow and red kind of bleed into each other.

JackeryFox:
Eh, just delete it. It's all a waste of time, dunno why I even bothered with it.

Anewbe:
That's not true, Fox, it looks pretty cool on its own, or I wouldn't have made this PR in the first place.

Verarch:
I honestly prefer this sprite. Maybe could use a bit work, but IMO, it's good.

Datraen:
Putting my two cents in: it's a good attempt, not a bad sprite, just not sure how much of an improvement it really is.

PR #2033, opened 2016-07-04
Removes "is" from pose

Anewbe:
Removes the word "is" from the Set Pose text.

I have no feelings regarding this, just happened to be easy to do.


Spookerton:
Makes poses less stilted. Sure! :+1:

Anewbe:
225px-neutral_president

PR #2037, opened 2016-07-04
Adds the ability to lower a sterile mask.

redstryker:
Allows people to lower their sterile masks below their chin.

Preview:
http://puu.sh/pQgEb/62efe03451.png
http://puu.sh/pQgG6/1e716ac643.png


redstryker:
Closing so that I can have this change merged with another change to the masks/miscellaneous.dm

PR #2039, opened 2016-07-04
Water-Coolers, Floor Lamps, and Benches

SinTwo:
Adds benches, floor lamps, and upgrades the Water-Coolers so they now can have a cup dispenser that holds up to 10 paper cups.

bench1

lamp

watercooler


code/modules/reagents/reagent_dispenser.dm
```diff
@@ -203,13 +203,24 @@
...
 		..()
 		reagents.add_reagent("water",120)
-	else
-		icon_state = "water_cooler_0"
+	update_icon()
+
+/obj/structure/reagent_dispensers/water_cooler/examine(mob/user)
+	..()
+	if(cupholder)
+		user << "\blue There are [cups] cups in the cup dispenser."
```
Yoshax: You should be using span classes.

EmperorJon:
Can we also get a changelog considering it adds NEW THINGS.

Yoshax:
Since they are children of tables, what happens when you build one next to a table?

Hubblenaut:
I thought of adding benches once, but any character positioned on a turf with a vertically laid out bench will just look like they are STANDING on it, opposed to sitting...

SinTwo:
@Yoshax Nothing! They build normally and do not try to connect.
@Hubblenaut Indeed it does; however, the same is also true for stools, no? (albeit to a lesser degree, but still!)
@EmperorJon Dammit, I knew I was forgetting something.

PR #2041, opened 2016-07-05
Modernizes [Re]Spawn Character

Arokha:
This updates the verb Respawn Character (changing it into 'Spawn Character') and makes it make sense in the modern era. The verb was from 2011 so it needed some love.

This is the verb that spawns a client's loaded character, either for replacing an accidentally deleted character (or otherwise somehow removed), silently spawning someone's loaded character, or spawning someone's already in-play character for... you know... star trek mirror universe events.

Your options when spawning are:
- Which client to spawn the loaded character of
- If you want to spawn them in arrivals or under you
- If you want to equip them as their previous job (or assistant, otherwise)
- If you want to announce their arrival
- If they are new, if you want to inject medical/sec/general records

### DIFF
--- a/code/modules/admin/verbs/randomverbs.dm
+++ b/code/modules/admin/verbs/randomverbs.dm
@@ -356,15 +356,15 @@ Traitors and the like can also be revived with the previous role mostly intact.
 		return
 
 	//I frontload all the questions so we don't have a half-done process while you're reading.
-	var/client/picked_client = input(src, "Please specify which client's character to spawn.", "Client", "") in clients
+	var/client/picked_client = input(src, "Please specify which client's character to spawn.", "Client", "") as null|anything in clients
 	if(!picked_client)
 		return
 
-	var/location = input(src,"Please specify where to spawn them.", "Location", "Right Here") in list("Right Here","Arrivals","Cancel")
-	if(!location || location == "Cancel")
+	var/location = alert(src,"Please specify where to spawn them.", "Location", "Right Here", "Arrivals", "Cancel")
+	if(!location)
 		return
 
-	var/announce = alert(src,"Do an announcement as if they had just arrived?", "Announce", "Yes", "No", "Cancel")
+	var/announce = alert(src,"Announce as if they had just arrived?", "Announce", "Yes", "No", "Cancel")
 	if(announce == "Cancel")
 		return
 	else if(announce == "Yes") //Too bad buttons can't just have 1/0 values and different display strings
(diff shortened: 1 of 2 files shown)


Screemonster:
For the latter use case, be sure to go to a mirror after spawning and give them a goatee.

SpadesNeil:
Oh! This is neat.

code/modules/admin/verbs/randomverbs.dm
```diff
@@ -349,111 +349,117 @@ Traitors and the like can also be revived with the previous role mostly intact.
...
-			break
+	//I frontload all the questions so we don't have a half-done process while you're reading.
+	var/client/picked_client = input(src, "Please specify which client's character to spawn.", "Client", "") in clients
+	if(!picked_client)
+		return
 
-	if(!G_found)//If a ghost was not found.
-		usr << "<font color='red'>There is no active key like that in the game or the person is not currently a ghost.</font>"
+	var/location = input(src,"Please specify where to spawn them.", "Location", "Right Here") in list("Right Here","Arrivals","Cancel")
+	if(!location || location == "Cancel")
```
PsiOmegaDelta: Why not utilize the spawn location list?

code/modules/admin/verbs/randomverbs.dm
```diff
@@ -349,111 +349,117 @@ Traitors and the like can also be revived with the previous role mostly intact.
...
+	if(announce)
+		AnnounceArrival(new_character, new_character.mind.assigned_role)
 
-	new_character << "You have been fully respawned. Enjoy the game."
+	//Announces the character on all the systems, based on the record.
+	if(!record_found)
+		if(alert(new_character,"Warning: No data core entry detected. Would you like add them to various databases, such as medical records?","Records","Yes","No")=="Yes")
+			data_core.manifest_inject(new_character)
+
+	message_admins("\blue [admin] has spawned [player_key]'s character [new_character.real_name].", 1)
```
Yoshax: Span class.

code/modules/admin/verbs/randomverbs.dm
Yoshax: It seems to me it might be better to perhaps to do as Psi says, or add a third option that allows you to input coordinates to save you the time of moving the mob once it has been spawned.

code/modules/admin/verbs/randomverbs.dm
```diff
@@ -349,111 +349,117 @@ Traitors and the like can also be revived with the previous role mostly intact.
...
-			new_character.real_name = capitalize(pick(first_names_male)) + " " + capitalize(pick(last_names))
-		else
-			new_character.real_name = capitalize(pick(first_names_female)) + " " + capitalize(pick(last_names))
-	new_character.name = new_character.real_name
+		inhabit = 0
 
-	if(G_found.mind && !G_found.mind.active)
-		G_found.mind.transfer_to(new_character)	//be careful when doing stuff like this! I've already checked the mind isn't in use
-		new_character.mind.special_verbs = list()
+	var/equipment = alert(src,"Give equipment? Last job's equipment, or assistant if none.", "Equipment", "Yes", "No", "Cancel")
```
Yoshax: Why not make the 'Yes' option give the same job popup that select equipment does, or even just the call equipment verb, allowing you to select from all pre-existing jobs and equipment load-outs?

Yoshax:
WIll this spawn the selected character's loadout items?

Arokha:
> WIll this spawn the selected character's loadout items?

No, nor did the original, really. Just spawns their job equipment. Of course the original was probably written before loadout items. Or items.

code/modules/admin/verbs/randomverbs.dm
Arokha: What is the 'call equipment' verb? I thought about letting you pick, but then if they were on the manifest and you respawn them as something else their original manifest entry stays. I could then clean THAT up but now we're getting away from a simple 'just spawn this character' proc into something more complicated.

code/modules/admin/verbs/randomverbs.dm
Arokha: S'how the original was, was just copypaste, but what is a better proc to use? I don't really recall having to use span classes in other places with admin messages, I thought it auto-colored them. Is there some other, better admin-message proc?

code/modules/admin/verbs/randomverbs.dm
Arokha: I guess I'm confused about the use-case. You want to spawn a mob in a specific, but arbitrary location that you are not present in, but know the coordinates of?

code/modules/admin/verbs/randomverbs.dm
Yoshax: This here: https://github.com/PolarisSS13/Polaris/blob/802b9652592324cf38247424ac09c40d2a82d121/code/modules/admin/verbs/debug.dm#L388

It allows you to pick from the job list and does the default equip for that job. It feels like it would be a nicer, and more versatile solution.

Arokha: Thanks, I'll use that.

Yoshax: Sure, just thought it might help you. Although, I don't believe it's +ADMIN or +FUN for the Spawn Character verb.

code/modules/admin/verbs/randomverbs.dm
Arokha: Hm, I added an option to spawn you at job spawn locations, but after the round starts and everyone spawns, it seems that it deletes all the spawn points. So, going to say out of scope on that one. Just aghost where you want them and spawn them.

Arokha:
Added some of the suggested features. Both 'job-specific spawn points' and the 'select equipment' menu are currently impossible and would require refactoring of those parts of the code. Instead you can pick a job for them when you spawn them, if you want.

Yoshax:
Mmm. Would be nice to have those features, but i see it would require a lot more work.

SpadesNeil:
This appears to be not working right. https://github.com/PolarisSS13/Polaris/issues/2568

PR #2042, opened 2016-07-05
Planetary Walls

Neerti:
Adds a new type of wall, which makes atmospherics pretend there is actually something beyond the wall that has gas to merge with.  This is used by surrounding your "planet" with this wall.  It's suggested you find more natural ways to wall off the area aways from this final wall (like using trees).
I added some presets for normal air, as well as all of the terrestrial planets of the Vir system.

Thanks to Mloc for telling me that unsimulated walls can do this, so I don't need to use a horrific invisible atmospherics machine.

### DIFF
--- a/code/game/turfs/unsimulated/planetary.dm
+++ b/code/game/turfs/unsimulated/planetary.dm
@@ -1,0 +1,47 @@
+// This is a wall you surround the area of your "planet" with, that makes the atmosphere inside stay within bounds, even if canisters
+// are opened or other strange things occur.
+/turf/unsimulated/wall/planetary
+	name = "railroading"
+	desc = "Choo choo!"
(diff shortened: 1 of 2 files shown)


Neerti:
Upon initialization (when first made), they'll have their atmospheric characteristics, but after atmos starts processing, the walls will cause the turfs to shift towards the wall's settings.  If you were to spawn one of these inside the phoron tank in atmospherics, the tank would lose phoron very quickly, until no phoron remained.

Yoshax:
Utter magic, I blame Mloc.

Hubblenaut:
A concern that came up: Even with the outside turfs slowly equalizing to a set gas distribution, it would still be possible to set the whole area in flames if phoron were released into it and immediately ignited.

PR #2044, opened 2016-07-05
Gives heavy & thunderdome armor armor values.

Cameron653:
Gives the thunderdome armor and the heavy armor the same values as the heavy armor vest, as they had no armor at all previously.


SpadesNeil:
Originally reported here: https://github.com/VOREStation/VOREStation/issues/170

PR #2046, opened 2016-07-05
Fix inconsistency with alien whitelist

Arokha:
The alien whitelist had some _interesting behavior_. Such as that the way new_player.dm was written, all things not "Human" were whitelisted. Regardless of the SPECIES_IS_WHITELISTED flag. Which is unfortunate. Because that's what that flag is for.

This simplifies new_player and redoes is_alien_whitelisted, and removes the unnecessary is_species_whitelisted (probably part of the confusion, since is_alien_whitelisted was used when is_species_whitelisted should have been used). Also adds is_lang_whitelisted because passing languages or species to the same place is unusual.

Oh also there was no exception for gear, such that if say, the Tajaran fur clothing thing wanted to be spawned, even if Tajaran was not a whitelisted species, it'd not let you spawn with it. Doesn't make much sense.

### DIFF
--- a/code/game/jobs/job_controller.dm
+++ b/code/game/jobs/job_controller.dm
@@ -342,7 +342,7 @@ var/global/datum/controller/occupations/job_master
 						else
 							permitted = 1
 
-						if(G.whitelisted && !is_alien_whitelisted(H, G.whitelisted))
+						if(G.whitelisted && !is_alien_whitelisted(H, all_species[G.whitelisted]))
 							permitted = 0
 
 						if(!permitted)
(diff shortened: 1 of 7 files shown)


SpadesNeil:
This may not have fixed anything. Reported issues with alien whitelists.

Yoshax:
I don't understand the issue and it works fine for us. We have alien races whitelisted and it functions just fine.

SpadesNeil:
As I said on Discord (but will repeat here just for reference) is that isn't the problem.

We have the opposite problem. You have alien races whitelisted. There are some alien races that **we don't** want whitelisted, but they are whitelisted just by existing not as a human race.

I'm sleepy and hope that makes sense.

PR #2054, opened 2016-07-07
Adds new non-alcoholic drinkd

Yoshax:
Adds:
- Shirley Temples
- Roy Rogers
- Collins Mixes
- Arnold Palmers

Left to right as list is above.
https://i.gyazo.com/fef9f76b9d6f4e737257b63854a17709.png

### DIFF
--- a/code/modules/economy/price_list.dm
+++ b/code/modules/economy/price_list.dm
@@ -42,6 +42,18 @@
 /datum/reagent/drink/milkshake
 	price_tag = 2
 
+/datum/reagent/drink/roy_rogers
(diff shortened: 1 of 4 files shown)


Anewbe:
"Tastes just like the old man." :+1:

Do we have sweet tea yet, or any non-13 Loko energy drinks?

Yoshax:
We have iced tea? And I don't believe we have any other energy drinks. 

I was/am planning to do more drinks, but I put this up as is because I don't know how long/when I'll do the rest.

PR #2059, opened 2016-07-08
The Beruang can now be sent to visit the station by admins

EmperorJon:
Once per round, with traders gathered, in the same manner as the admin ERT call!

**Summary**
Whilst waiting the indefinite amount of time for a structure upon which we can do our generic ship visiting events, I decided we could make use of the ERT call concept for the trader ship. It's worth noting this isn't intended to be the framework for or even necessarily a step towards these events, but a suitable stand-in for now. This means admins can now "Dispatch Beruang Trader Ship" and players can now "Join Trader Visit". Traders are an antagonist type, however it clarifies they're not an actual antagonist, in the same way that ERT are clarified as "anti-antagonist". 

**Details**
Traders spawn in cargo clothing, on the empty rooms on the CentComm level, which now have a bunch of wardrobes as a mirror. I'm looking for feedback on what should be default and/or available for them before this goes through. They also spawn with a common channel headset. They also have the standard ERT/Heist/Merc style appearance and name editing choices, and they have the same ID-"passport"-in-wallet-plus-some-cash as the Heist Raiders do. **The first person to join is told they are the captain**, however this isn't acted upon in any way yet, so this can be ignored if they so wish. Traders have an ID which uses the "permit" sprite to differentiate it from NT IDs somewhat. 

**ID/Access changes**
This ID has only the new access 160, which all the Beruang's doors which were previously access 150 (syndicate access) now use. As a result, Traders can't cause accidental interesting issues if they're around in the same round as (for example) Mercs. The shuttle's main console is also locked to access 160. As a result, ERT, the Captain, antagonists, and such shouldn't be able to steal their ship or open their doors without cooperation or force.

**Stuff to think about**
Like the ERT, the Beruang can only be triggered once a round. Whilst the Traders themselves can fly back and forth between the Northern Star and the empty area on the spaceport z-level at will, you can't trigger a new call for Traders from ghosts, as this just seemed to make sense. There should maybe however be a way for the Traders to leave the round once they return to their "base" without just ghosting, such as a cryopod or similar, just for cleanliness' sake.


EmperorJon:
Well I forgot to run the mapmerger, so there's a start...

EmperorJon:
I decided I'd pretty up the spawn area. 
Various closets (still open to input on specific things that might be wanted), as well as a little bathroom area (mainly for the mirrors) and a closed-off warehouse for admins to spawn specific things for the Traders if they wish.

image

EmperorJon:
Right, changelog done, erroneous sink/mirror in the above screenshot removed. This can be worked on and improved later but I'm happy with it as-is and would like to trial it out soon!

PR #2060, opened 2016-07-08
Cheap Man's hub password

Yoshax:


EmperorJon:
Cry every time.

Seriously though the sooner we can get this working at the flip of a switch the better. I wish I could help more...

Neerti:
I feel bad for our downstreams.

Yoshax:
Well, I have a better way to do so, but I need Sparky.

PR #2063, opened 2016-07-08
Implements Smart Magazines

Yoshax:
Smart Magazines are advanced technology that use an attached power cell to produce ammunition. They are recent technology, and are far from efficient, or commonplace. With this PR, they are not available anywhere as discussion is needed.

Details:
- Smart Magazines require a power cell
- Smart Magazines can scan loaded firearms with magazines or loaded magazines.
  - Once scanned the Smart Magazine saves the calibre and ammo type, then producing the bullets needed.
- Smart Magazines once scanned can be loaded into the applicable gun.
- It takes three seconds to attach and remove the power cell from a Smart Magazine.
- Smart Magazines cannot be emptied unless they are emagged.
- Smart Magazines always have a max of 20 bullets. They will always produce up to their limit.
- It takes 3 seconds to produce one bullet.
- Numbers are balanced currently as so it uses the entirety of a 10k power cell to produce 20 762 rounds.

They are not quite finished, or perfected, but I have put the PR up to get feedback and allow discussion.


code/modules/projectiles/ammunition/bullets.dm
```diff
@@ -115,6 +129,7 @@
 	icon_state = "pshell"
 	projectile_type = /obj/item/projectile/bullet/shotgun/practice
 	matter = list("metal" = 90)
+	matter = list(DEFAULT_WALL_MATERIAL = 90)
```
Techhead0: Duplicate var definition.

Hubblenaut:
I don't knooow, considering the impossibility of having an automatic lathe in a tiny magazine.

PR #2068, opened 2016-07-09
Adds material stack variants with base 50 amount for adminspawn etc

Yoshax:


PsiOmegaDelta:
If you use these with the cargo supply packs you can remove the snowflakey amount handling.

PR #2075, opened 2016-07-10
Makes clicking on airlocks work again.

Neerti:


Anewbe:
Thought that might have happened, sorry I didn't catch it before the merge.

PR #2082, opened 2016-07-10
Fully removes floor-length braid instead of leaving it as a stupid meme

Yoshax:
Whoever left it in as a meme deserves to be flogged.


Anewbe:
:+1:

PR #2087, opened 2016-07-10
Removes the Robot Console's ability to blow up borgs

Anewbe:
It's a weeeeee bit unethical to be able to explode a sapient thing on a whim, yeah? Plus it's probably cheaper to find and correct the borg than to remake/train a new one.

### DIFF
--- a/html/changelogs/anewbe.yml
+++ b/html/changelogs/anewbe.yml
@@ -1,0 +1,36 @@
+################################
+# Example Changelog File
+#
+# Note: This file, and files beginning with ".", and files that don't end in ".yml" will not be read. If you change this file, you will look really dumb.
+#
+# Your changelog will be merged with a master changelog. (New stuff added only, and only on the date entry for the day it was merged.)
+# When it is, any changes listed below will disappear.
+#
+# Valid Prefixes: 
+#   bugfix
+#   wip (For works in progress)
+#   tweak
+#   soundadd
+#   sounddel
+#   rscadd (general adding of nice things)
+#   rscdel (general deleting of nice things)
+#   imageadd
+#   imagedel
+#   maptweak
+#   spellcheck (typo fixes)
+#   experiment
+#################################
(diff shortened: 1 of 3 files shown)


PsiOmegaDelta:
Unless Polaris has changed it, it's technically only the chassis that's blown up.
The brain itself remains unharmed, if disoriented.

Anewbe:
Still rude.

EmperorJon:
Are there ever circumstances where you can blow a borg but not lock them down?

Anewbe:
Not as far as I can tell, no. Traitor borgs were already supposed to be immune, this just extends that courtesy to regular borgs.

html/changelogs/anewbe.yml
```diff
@@ -0,0 +1,36 @@
...
+# Optional: Remove this file after generating master changelog.  Useful for PR changelogs that won't get used again.
+delete-after: True
+
+# Any changes you've made.  See valid prefix list above.
+# INDENT WITH TWO SPACES.  NOT TABS.  SPACES.
+# SCREW THIS UP AND IT WON'T WORK.
+# Also, all entries are changed into a single [] after a master changelog generation. Just remove the brackets when you add new entries.
+# Please surround your changes in  double quotes ("), as certain characters otherwise screws up compiling. The quotes will not show up in the changelog.
+changes: 
+  - rscdel: "Removed joy"
```
PsiOmegaDelta: This is a rather meaningless changelog.

Anewbe: Oh hey, wrong changelog, one moment.

EmperorJon:
I don't really see why we need to keep boom given that you can lock them down, and we specifically made it so you can't lock down traitors anyway because that's (rather correctly) deemed to be too easy a "permastun".
*shrug*

Neerti:
Gets a :+1: from me.

PR #2088, opened 2016-07-10
Anatomic Panacea clears ingested reagents

Anewbe:


EmperorJon:
Bonus points if it cleared them by forcibly vomiting them everywhere but I suppose this will do.

PR #2089, opened 2016-07-10
Removes Assault Carbine from uplink

Anewbe:
The 5.56 Bullpup Assault Carbine and its magazines are no longer purchasable from the antag uplink.

Still randomly shows up as Renegade or Raider, and as such the ammo remains in hacked autolathes.

### DIFF
--- a/code/datums/uplink/ammunition.dm
+++ b/code/datums/uplink/ammunition.dm
@@ -41,7 +41,7 @@
 /datum/uplink_item/item/ammo/a556/ap
 	name = "10rnd Rifle Magazine (5.56mm AP)"
 	path = /obj/item/ammo_magazine/a556/ap
-
+/*
 /datum/uplink_item/item/ammo/a556m
 	name = "20rnd Rifle Magazine (5.56mm)"
 	path = /obj/item/ammo_magazine/a556m
@@ -51,7 +51,7 @@
 	name = "20rnd Rifle Magazine (5.56mm AP)"
 	path = /obj/item/ammo_magazine/a556m/ap
 	item_cost = 4
-
+*/
 /datum/uplink_item/item/ammo/c762
 	name = "20rnd Rifle Magazine (7.62mm)"
 	path = /obj/item/ammo_magazine/c762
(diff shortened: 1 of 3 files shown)


EmperorJon:
Was this discussed at the meeting or something? I don't know if this was something that was decided to be done or just on your own initiative.

Anewbe:
I believe it was discussed during the meeting, or shortly afterwards. I recall Mangled bringing it up at some point.

Yoshax:
I would rather see re-balancing done than a removal. It could be made to cost more TC and be there as a more expensive weapon for those who really need to shoot something.

EmperorJon:
The Assault Carbine is the old Bulldog, correct? And was added to the uplink as well as the STS?

Yoshax:
Incorrect. The Bulldog is its own gun. The Assault Carbine is a 20rnd Bulldog, essentially, without the grenade launcher. It does more damage than the STS, and as such, should cost more. I would say 10TC vs the current 7TC.

EmperorJon:
I still find it odd that the 5.56 is more deadly than what is apparently 7.62 but I'm not about to start that argument again.

Yoshax:
It's more hassle than it's worth to fix, as we've seen before.

Anewbe:
The bullet swap isn't that difficult, but people have gotten testy in the past.

ParadoxSpace:
May I ask why?

Anewbe:
It's numerically the best weapon available on the uplink, for the cost. It simultaneously makes the Bulldog and the STS superfluous, and as a result it's become the go-to weapon for antaggery, because it doesn't really have a downside.

SpadesNeil:
Remove the Assault Carbine from the uplink, yea. It was supposed to be removed after we went back to the original STS damage values.

EmperorJon:
I marked it as Controversial because people were still discussing what was going on with it. I'm assuming that's all over now.

Neerti:
Since Spades pretty much agrees that the gun should be removed from the uplink, gonna press the big merge button.

PR #2090, opened 2016-07-10
New Limb HUD System, from Bay

Neerti:
Manually ports https://github.com/Baystation12/Baystation12/pull/13530, and https://github.com/Baystation12/Baystation12/pull/13561 from bay.
Adjusts the scaling on the coloration, since it seems Bay has different shading than us for species, otherwise some species would be dark, and others would be superbright.
Credit goes to Zuhayr for the code.


EmperorJon:
Should probably have a changelog, especially for credit purposes. (We need to get into the habit of changelogging basically every addition)

Neerti:
My brain is dumb and titled the commit 'changeling', woo.

EmperorJon:
Nah leave people confused. :p

Yoshax:
Are those left-in world messages deliberate?

Neerti:
I really need to sleep.

PR #2099, opened 2016-07-11
Halves the effect of Tricordrazine

Anewbe:
Requested by staff

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine.dm
@@ -112,9 +112,9 @@
 
 /datum/reagent/tricordrazine/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
 	if(alien != IS_DIONA)
-		M.adjustOxyLoss(-6 * removed)
-		M.heal_organ_damage(3 * removed, 3 * removed)
-		M.adjustToxLoss(-3 * removed)
+		M.adjustOxyLoss(-3 * removed)
+		M.heal_organ_damage(1.5 * removed, 1.5 * removed)
+		M.adjustToxLoss(-1.5 * removed)
 
 /datum/reagent/cryoxadone
 	name = "Cryoxadone"


Yoshax:
I disagree. I am not a fan. :-1:

If you have someone come into medical, and they have a broken limb. You have to do surgery on the limb. You cannot do surgery to repair a bone when damage is above x. In rounds where there is no chemist you do not have bicaridine. Your only options are:

- Apply a trauma kit and wait
- Tricordrazine

Now. Let's do some math.

The organ_regeneration_multiplier is set to 0.5
Here is the code section
I disagree. I am not a fan. :-1:

If you have someone come into medical, and they have a broken limb. You have to do surgery on the limb. You cannot do surgery to repair a bone when damage is above x. In rounds where there is no chemist you do not have bicaridine. Your only options are:

- Apply a trauma kit and wait
- Tricordrazine

Now. Let's do some math.

The organ_regeneration_multiplier is set to 0.5
wound_update_accuracy is set to 1
Here is the code section for wound healing:

```		// slow healing
		var/heal_amt = 0

		// if damage >= 50 AFTER treatment then it's probably too severe to heal within the timeframe of a round.
		if (W.can_autoheal() && W.wound_damage() < 50)
			heal_amt += 0.5

		//we only update wounds once in [wound_update_accuracy] ticks so have to emulate realtime
		heal_amt = heal_amt * wound_update_accuracy
		//configurable regen speed woo, no-regen hardcore or instaheal hugbox, choose your destiny
		heal_amt = heal_amt * config.organ_regeneration_multiplier
		// amount of healing is spread over all the wounds
		heal_amt = heal_amt / (wounds.len + 1)
		// making it look prettier on scanners
		heal_amt = round(heal_amt,0.1)
		W.heal_damage(heal_amt)
```

Now. I'm not sure what the threshold is. But if we say someone has 45 damage. Let's plug in some values and run through some code.

var/heal_amt = 0

`heal_amt = heal_amt * wound_update_accuracy`
heal_amt = 0.5 * 1
heal_amt = 0.5

`heal_amt = heal_amt * config.organ_regeneration_multiplier`
heal_amt = 0.5 * 0.5
heal_amt = 0.25

`heal_amt = heal_amt / (wounds.len + 1)`
heal_amt = 0.25 / (2)
heal_amt = 0.125

`heal_amt = round(heal_amt,0.1)`
heal_amt = round(0.125,0.1)
heal_amt = 0.1

`W.heal_damage(heal_amt)`
W.heal_damage(0.1)

Therefore, by only waiting for a wound to naturally heal, it only heals 0.1 per tick. Meaning the 45 damage wound takes 450 seconds to heal, or 7 and a half minutes. Now, that is a somewhat short time, but it's a time that's stupid to have to just make a person sit around.

Now, perhaps we can compare to Tricord.


The important line is:
`M.heal_organ_damage(3 * removed, 3 * removed)`
Removed is 0.2

M.heal_organ_damage(3 * 0.2, 3 * 0.2)
M.heal_organ_damage(0.6,0.6)

That means that it heals 0.6 damage of both burn and brute, per life tick. Meaning the tricord would take 75 seconds to heal the 45 damage wound. With your proposed changes this would take 150 seconds. Now, that's a pretty large difference. That's making it from just over a minute to two and a half. I do not believe this is a good change, and will not add to anything, it will just end up with people sitting around waiting more while adding nothing interesting.

Again, :-1:

Anewbe:
The hope is that you'd have chems around. In a lot of high damage rounds without a chemist, I've seen the medical crew break in and make it themselves. IF that's not going to happen, there's other fixes for damage lying about. The burn kits have kelotane, at least, and the blue ones have dexalin.

For that big sort of damage, where you need the damage down fast you can toss them in cryo. The base cryo chem has (currently) nearly triple the healing of tricord, in everything but oxyloss. That'll drop 45 seconds down to like 15? The sleep's not going to be an issue if they're already headed into cryo, you just roll them west and slap on the tank.

Yoshax:
I've stated my thoughts. Changing something and having part of your reasoning behind it be that people can just powergame to step around it, sure. As for cryo... I really don't see how you can propose a nerf to Tricordrazine when cryo is even worse, and makes even less sense, at least, in my opinion.

EmperorJon:
At least cryo can't be used in the field. I'm going to say I'm actually for this, personally.

Spookerton:
Tricordrazine *by volume* is almost as effective as all of the primary chemicals except for oxyloss (which doesn't tightly represent anything to begin with unless you're in hardcrit), and works for all four damage types at once.

It's current state represents an aging but often talked about unlisted buff that happened when kelenius made reagents in general a little saner last year - which about doubled its prior effectiveness by volume.

**I'd be open to making its consumption per tick higher to compensate for it being slower**, but in it's current state the medical vendors contain 24 bottles of tricord in quickly combined dylo/inap bottles, and there's about 30 bottles worth of total components in medical alone. Each bottle is worth 180/180/180/360 - which adds up to 5400 brute, burn, and tox healing each on roundstart, not including magic sleeper and chemistry reagents.

EmperorJon:
> not including magic sleeper and chemistry reagents.

By which you mean pumping people full of infinite tricord.

Spookerton:
Yes

Yoshax:
Probably because, as I noted, half the time you don't have a chemist or they're an incompetent ass so your options are abuse the hell out of tricord or hope you aren't dealing with someone who's actually injured so you can wait the seven and a half minutes for them to regen.

Spookerton:
I'll be sure to tell you the next time I regenerate from massive trauma injuries in seven minutes with no medical intervention whatsoever.

Yoshax:
If you read my prior post you will see the numbers. A 45 damage wound takes 7.5 minutes to heal. Unless you're making a reference to real life here, if so, then please do keep in mind that realism doesn't necessarily equate to fun. If our medical experience was realistic it would be eons longer and 90% of that would be eating toast in a bed with nobody else around.

SparklySheep:
You want to nerf Tricord? The stuff is already pretty worthless as it is and you want to debuff it into complete and utter uselessness. In terms of medical balance, tricord is not the thing you should be focusing on, fuck.

EmperorJon:
10u of Tricord + passive healing will heal over 30 damage in a minute during which time you can obviously walk off and wait for it to work. That's not "pretty worthless".

Yoshax:
I don't see why you bother to bring up passive healing, I showed the math, and it's pretty insignificant like it should be. It only heals 6 per minute.

SparklySheep:
This would also nerf the crisis borg a bit.

As well, what's bad about getting people out of medical faster? Waiting around in medical isn't that fun and lets you get back to the game. Pretty much every other medicine available is twice as good (if not more so) as the current tricordrazine is. Fun trumps realism in this certain scenario, we already have way too many bloody surgeries as it is.

Anewbe:
To quote AlabasterTempest, who could not be dragged to Git despite my prodding, "this change is going in because staff has talked about and more or less agreed up this nerf multiple times, but no one's made the PR yet".

Yoshax:
Well, that's nice. It's a really unnecessary, and frankly, stupid change.

PR #2107, opened 2016-07-11
Widens the arrivals dock by 4 tiles

EmperorJon:
Shifts everything to the right like the hotel, dock 4, solar array, etc. right by around 4 tiles. This is so that the three main docks are all the same size and shape for future ShuttleThings(tm) coming Soon(tm).


EmperorJon:
DNM pending me making sure I haven't broken any piping or wiring.

Yoshax:
Before & After?

EmperorJon:
Note my images don't have cavegen turned on so they'll look somewhat different.

Fat arrivals shuttle (a bit of a stand-in for now, and could do with another light at the bottom)
image
image

Emergency storage and the surrounding area is a little different given the insertion of 4 tiles more hallway between it and the bar cut-through.
image
image

Medbay-hotel maintenance has been more or less inverted by the change.
image
image

And finally, the far bottom right of the map is shifted somewhat to the edge of the asteroid as a result.
image
image

PR #2110, opened 2016-07-12
Speeds up toxin

Anewbe:
Ups the speed at which toxin based reagents work by a factor of 5.

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Toxins.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Toxins.dm
@@ -6,7 +6,7 @@
 	description = "A toxic chemical."
 	reagent_state = LIQUID
 	color = "#CF3600"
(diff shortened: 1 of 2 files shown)


Yoshax:
Seems like a very large jump. Can we get some number comparisons on times it would take to die, etc?

Anewbe:
Totes.
Damage formula is (strength*removed), so the base toxin takes 50 units to kill, ignoring hardcrit. This remains constant.
Current toxin metabolism = REM *

Anewbe:
On my phone, can't edit last post, continuing here
Metabolism = REM*0.05, or 0.01. This means the aforementioned 50 units toxin does ~0.04 tox damage per check, and it takes 5000 ticks to process out.
With this PR, metabolism = REM*0.25, or 0.05. You will take ~0.2 tox per tick, and it'll take 1000 ticks to process out. It'll still take basically a whole bottle of toxin to kill you, but now it won't linger for ages.

Anewbe:
Put a little multiplication sign after those REMs, formatting ate them.

Yoshax:
Works for me.

PR #2112, opened 2016-07-12
Adds player age req to Chemist and Atmos

Anewbe:
3 days each
Surprised this wasn't already a thing


Yoshax:
Probably because the other jobs aren't. If you're playing Station Engineer or Medical Doctor, you're basically playing Atmos Technician and Chemist.

Anewbe:
It's more I'm surprised that no one has added a time-block for these two jobs, which have easy access to phoron and other grieftastic things.

Yoshax:
You can't really phoron grief without at least a little atmos knowledge anymore. It's not just two pumps and fill the station.

Anewbe:
Griefers aren't always dumb, though.

PR #2115, opened 2016-07-12
Ensures security robots can be properly emagged

Yoshax:
In addition, they move a little faster when emagged.

### DIFF
--- a/code/modules/mob/living/bot/secbot.dm
+++ b/code/modules/mob/living/bot/secbot.dm
@@ -85,6 +85,15 @@
 			declare_arrests = !declare_arrests
 	attack_hand(usr)
 
+/mob/living/bot/secbot/emag_act(var/remaining_uses, var/mob/user)
+	. = ..()
+	if(user)
+		user << "<span class='notice'>The [src] buzzes and beeps.</span>"
+	emagged = 1
+	patrol_speed = 3
+	target_speed = 4
+	return 1
+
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/bot/secbot.dm
```diff
@@ -85,6 +85,15 @@
...
 	attack_hand(usr)
 
+/mob/living/bot/secbot/emag_act(var/remaining_uses, var/mob/user)
+	. = ..()
+	if(user)
+		user << "<span class='notice'>The [src] buzzes and beeps.</span>"
+	emagged = 1
+	patrol_speed = 3
+	target_speed = 4
+	return 1
```
PsiOmegaDelta: I note that there is nothing that prevents you from emagging the same not multiple times, with no other effect than costing you charges.
It's also a very anonymous override with should  perhaps use visible_message instead of user <<.

code/modules/mob/living/bot/secbot.dm
```diff
@@ -85,6 +85,18 @@
 			declare_arrests = !declare_arrests
 	attack_hand(usr)
 
+/mob/living/bot/secbot/emag_act(var/remaining_uses, var/mob/user)
+	. = ..()
+	if(!emagged)
+		if(user)
+			user << "<span class='notice'>\The [src] buzzes and beeps.</span>"
```
mwerezak: You might want to consider making this an `audible_message()`

PR #2125, opened 2016-07-13
Fixes 3 runtimes due to a null

Yoshax:


Neerti:
Kinda funny how a tiny error can make all the difference.

PR #2128, opened 2016-07-13
Allows FBPs to potentially have different age limits.

Neerti:
Cyborgs have their minimum be equal to their species' minimum, and maximum be their species maximum plus twenty years.
Posibrains have a minimum of 1, and a maximum of 220, as discussed by EmperorJon.
Digital brains (drones) have a minimum of 1 and maximum of 150, as discussed by Spookerton.

Also adds procs to test for someone being an FBP if that someone is a pref file, and to get their brain type.


Anewbe:
:+1:

PR #2132, opened 2016-07-14
Makes biogenerator be accurate about a few product costs

Cameron653:
Edit: 
Changes what shows up on the biogenerator screen to be accurate to the actual costs of the nutriment products.

~~Fixes the E-Z-Nutrient, Left 4 Zed, and Robust Harvest costs from the biogenerator.
They were listed as 60/120/150 for one, 300/600/750 for five even though the formulas for them (And what showed up on the biogenerator screen) suggested they should only cost 10/20/25 each.
Formulas:
([round(10/build_eff)])
([round(20/build_eff)])
([round(25/build_eff)])
Build_eff starts off at one.~~


Anewbe:
So... last I checked these bottles hold 60 units, versus the ten they did when that thing was made. The larger cost seems appropriate.

Cameron653:
Ah, that makes sense. I'll update the formula and what shows on the screen for them instead of reducing the cost to fit the current formula, then.

PR #2136, opened 2016-07-15
Security Armory Nerf

Mark9013100:
* Removes the WT-550 SMG from obj/random/projectiles, also removes all SMG ammo from the Armory.

* Removes the combat armor from the armory, but leaves the lockers incase the Warden ever orders it/other equipment.

<s>* There's now only two sets of ablative + bullet resistant armor.</s>

* Also replaces the chairs in the solitary cells with stools.


Verarch:
i won't object removing the combat armor
but i'd like for the other sets to remain untouched

Finalsong:
I can toast to this :ok:

ParadoxSpace:
Please leave the other sets untouched, yes.

Mark9013100:
I'll amend the ablative/bulletproof armor changes in the morning/afternoon tomorrow.

Spookerton:
Stools in solitary is bad, mkay. They're chairs so that they can't be used to bash the door.

Atermonera:
Stools can also be applied to the head

Mark9013100:
So then it's useful for either suicide or attacking security, I don't see the problem with that.

Anewbe:
Oh, right :-1: , don't like the changes.

Yoshax:
nty

Neerti:
Alrighty, waited long enough, guess it's time to give my input.

Short version:
:-1: 

Long version:
Don't really care about the SMG staying or going, but everything else is fairly bad.

I'm not sure why you and certain other persons feel that combat armor is the most unholy of all evils and needs to be purged from the game.  If you look back in time, the 'generalist' armor was tac armor, which had been nerfed twice and then replaced with combat armor, making security's 'general' armor choice go from 75/75, to 65/65, then to 50/50 with combat armor.  I made a comment to myself saying that I expect that those that had cried for these nerfs would make another cry for another nerf/removal.  I could pull various sayings and adages right now.  I know that this will not be the last PR for this sort of thing.

I would debate your reasoning for why you want to remove them, but you haven't provided any, so I can't do that, so instead I'll move on to the third thing, the stools.

If someone gets put in solitary, they've done something that necessitates them being inside (or else the sec officer who had put them there gets a nice talking to by admins).  Generally this involves trying to trash their cell or fighting their inmates, both of which generally means that they are going to take advantage of that stool in their new cell, by breaking their cell door or their skull.  I don't see how enabling that adds to the experience for anyone, and just annoys security and the engineer who has to go fix the door.

All in all, do not want.

Mark9013100:
>I'm not sure why you and certain other persons feel that combat armor is the most unholy of all evils and needs to be purged from the game. If you look back in time, the 'generalist' armor was tac armor, which had been nerfed twice and then replaced with combat armor, making security's 'general' armor choice go from 75/75, to 65/65, then to 50/50 with combat armor. I made a comment to myself saying that I expect that those that had cried for these nerfs would make another cry for another nerf/removal. I could pull various sayings and adages right now. I know that this will not be the last PR for this sort of thing.

I've said it before and I'll say it again, Security doesn't need sixteen sets of armor. Infact, As of now, Security has 5 variations of armor, the standard vests(including the heavy vests), riot armor, ballistic armor, ablative armor, and combat armor. I see no need for Security to have that much armor. One of the more common arguments I've seen in Combat Armor's defense was how it falls into a jack of no trades area, which, if so, then what's the harm in removing it?

>If someone gets put in solitary, they've done something that necessitates them being inside (or else the sec officer who had put them there gets a nice talking to by admins). Generally this involves trying to trash their cell or fighting their inmates, both of which generally means that they are going to take advantage of that stool in their new cell, by breaking their cell door or their skull. I don't see how enabling that adds to the experience for anyone, and just annoys security and the engineer who has to go fix the door.

The problem is, without the stool, then there's literally nothing to do outside of ghost and leave the round. Which contributes nothing. Second, I find the 'annoyed security/engineers' defense a bit weak, considering I rarely see engineering do anything outside of personal projects and vending machine maintenance. If the sec team thinks the stool is going to cause trouble, they can go and remove it at roundstart.

SpadesNeil:
So why are we removing SMGs? I distinctly recall people fussing about how security basically has no chance against antagonists and my previous (accidental) change to weapon damage values just made that worse. This also seems to make that a lot worse.

I don't think this is good idea.

I say removing combat armor is fine. Force players to be a little more creative when they choose their armor. Though I think that security having "too much" choice in armor is a poor argument. I just don't like players picking the lazy solution.

Mark9013100:
>So why are we removing SMGs? I distinctly recall people fussing about how security basically has no chance against antagonists and my previous (accidental) change to weapon damage values just made that worse. This also seems to make that a lot worse.

I don't think this is good idea.

Security doesn't need SMGs. They have a 90% winrate against antagonists, and if anything, security should be at the disadvantage at all times.

Yoshax:
Security wins because the people who play security play security every round, and the people playing antag half the time have never played before, or have no idea what they're doing. It's like putting a group of army recruits up against a group of special forces. Furthermore, security ends up winning half the time because they get away with being uncaring, unrealistic, stupid pieces of crap with huge justice boners. Antags rarely play to win, but sec does every round.

The problem isn't the items, it's the players.

SpadesNeil:
>The problem isn't the items, it's the players.

Good point. It's a different cause, but the same symptom.

zekesulastin:
You could just remove security like I've been saying in every single goddamned thread on the subject for the past two years because it's by FAR the easiest way to solve every single possible grievance on the subject.  Hell, it's better than having a department whose literal only job is to be cannon fodder so that the antagz can feel special for shooting a redshirt.

I have a much longer rant, but that would be too much for a damn GitHub post.

PR #2140, opened 2016-07-15
Pubifies the server

Neerti:
I feel bad for our downstreams.


Anewbe:
May God have mercy on us all

PR #2149, opened 2016-07-18
Item State Cleanup

SinTwo:
Quite literally the title plus some reorganizing of a few other clothing .dmi files.


Anewbe:
If there's a thing somewhere that makes it auto-check the icon_state for the item states, I'm pretty sure this'll work as intended... Otherwise darkwraith

SinTwo:
It tries to use item_state_slots, if that isn't defined then it tries to use item_state and if that isn't then it will use the icon_state, so yes it does work as intended.

https://github.com/PolarisSS13/Polaris/blob/741706e9e3ee911a38c7d558cfe7e3ffbcee0f59/code/modules/mob/living/carbon/human/update_icons.dm#L927

PsiOmegaDelta:
Those are some meaningful and worthwhile comments you've left in the code.

Anewbe:
Gonna merge this if no one has further thoughts.

PR #2152, opened 2016-07-20
Removes the HoS trench from the security uniform crate

Mark9013100:


Anewbe:
Still in the locker, at least?

Neerti:
He hasn't touched the locker file, just the crate, so yes.

PR #2154, opened 2016-07-20
Makes secret not bad

Yoshax:
#haitusbreaker

### DIFF
--- a/code/game/gamemodes/changeling/changeling.dm
+++ b/code/game/gamemodes/changeling/changeling.dm
@@ -12,7 +12,7 @@
 		certain though... there is never just one of them. Good luck."
 	config_tag = "changeling"
 	required_players = 2
-	required_players_secret = 5
+	required_players_secret = 2
 	required_enemies = 1
 	end_on_antag_death = 0
 	antag_scaling_coeff = 10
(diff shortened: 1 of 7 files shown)


Hubblenaut:
GLORY FOR YOSHAX

SparklySheep:
This is a long overdue change. :+1:

Yoshax:
If this isn't wanted, then I propose my counter-proposal which is removing extended from secret. Having both changes together would be best, imo. Right now, any round before like 8pm GMT ends up being voted secrets which ends up being extended 75% of the time, or is a dice roll between extended and auto-traitor. If there was a chance to get something a tad more exciting than a coin flip for extended, people might ready more.

People don't vote secret to play extended. A lot of votes end in secret because extended has a chance of win and everyone voting for other modes than dogpiles together to have secret win.

EmperorJon:
For the ones which require one player, does that include the antag? I can't remember.

Yoshax:
I am not sure. I believe it's without.

EmperorJon:
What I don't want to see is 1 person ninja, wiz etc. where the only person joining becomes the antag. Not only is that entirely pointless but it also means if 1 person votes secret and joins they have 100% chance of being forced to be the antag when that happens.

Yoshax:
Then it's simple. We remove extended from secret, and the ninja requirements etc stay as they are.

Chnkr:
:+1: x :100:

PsiOmegaDelta:
Alternative suggestion at this point: Remove the required_players_secret var.

PR #2159, opened 2016-07-21
Technomancer Additions/Fixes

Neerti:
Moves spell HUD so it doesn't cover instability warning.
Makes Instability radiate out towards other humans at sufficently high amounts.  People radiating instability will glow purple light.  It is adviced to keep your distance or you will get instability as well.
Mark can no longer be cast on the admin z-level.
Adds Gambit, a spell which will randomly give you (almost) any spell that exists.
Shield now causes a bit of instability when blocking.
Recall will bring anyone who is grabbed along for the ride when cast.
Disposable Teleporters now function correctly.
Instability decays faster.


Yoshax:
What spells can't Gambit give you and why?

Neerti:
It can't give you the base spells used for inheritance purposes, (e.g. the projectile launching one, aura base, summoning base, etc), since they're worthless on their own and would probably make bugs happen if someone tried to cast those.

Yoshax:
So it's a matter of just ensuring you don't get something you're not meant to get codewise, rather than restricting the actual selection?

Neerti:
Yes.

Yoshax:
Coolio. Just to not I also raised the issue that Technomancers have no indication of what their spells cost to use, or in instability other than in their book which makes it kinda clunky, since you never know when you can use something or not, or are safe to use.

PR #2160, opened 2016-07-22
Energy weapons use charge_cost instead of max_shots

Anewbe:
Swaps all of the energy weapons to use charge cost, and a new, weapon specific cell. That cell doesn't do anything as of now, besides be a small powercell.

As a side effect, stunbatons and guns will now charge faster.


code/modules/projectiles/guns/energy/pulse.dm
```diff
@@ -24,11 +24,11 @@
 /obj/item/weapon/gun/energy/pulse_rifle/destroyer
 	name = "pulse destroyer"
 	desc = "A heavy-duty, pulse-based energy weapon. Because of its complexity and cost, it is rarely seen in use except by specialists."
-	cell_type = /obj/item/weapon/cell/super
-	fire_delay = 25
+//	cell_type = /obj/item/weapon/cell/super
+	fire_delay = null		//Aspawn only gun doesn't really need to be balanced against the station/antags
```
Hubblenaut: Not sure why you're doing this. Just because it's not obtainable in the game right now doesn't mean it should be made spammable - especially not for potential downgrade servers, a scenario where an admin is spawning it for the players, or when this gun will indeed be made constructable by a future commit.

code/game/objects/items/weapons/power_cells.dm
```diff
@@ -21,21 +21,58 @@
...
+	matter = list(DEFAULT_WALL_MATERIAL = 350, "glass" = 80, "phoron" = 50)
+
+/obj/item/weapon/cell/weapon/infinite
+	name = "infinite-capacity weapon cell"
+	desc = "What're you even doing with this thing? Put it back in your weapon and get back to work."
+	icon_state = ""
+	origin_tech = list(null)
+	maxcharge = 6000
+
+	check_charge()
```
Hubblenaut: Absolute type pathes here.

Hubblenaut:
I like it, was about time. Remembering I wanted to do something like this back when adjusting rechargers.

code/modules/projectiles/guns/energy/pulse.dm
Anewbe: From my memory, asking the Polaris staff about this weapon (and the SAW) and the initial change was met with roughly "Who cares, it's never spawned". I can't say I've ever seen this thing spawned on Polaris. I'm highly dubious that it will ever be made accessible in regular situations. If a downstream decides to give the fire delay back, it's not exactly difficult.

code/game/objects/items/weapons/power_cells.dm
Anewbe: If I understand what you mean by that, it already is. If I ever get to Phase 2: Reloading, having weapons grade cells all be children of cell/weapon simplifies things.

code/modules/projectiles/guns/energy/pulse.dm
SpadesNeil: I don't think this change should be made. Yes, it's an adminspawn gun, but just because it's adminspawn doesn't mean it should be brokenly overpowered when it wasn't before.

The SAW is powerful, but it's not obscene. It doesn't fire 60 damage bullets at a billion rounds a second because loladminboose. It's just a powerful gun that is used in some events, more powerful than what is normally available.

If I want to admin abuse with the intention of fucking shit up I'll spawn a meteor gun.

Yoshax:
I really don't understand the cells, or their sizes.

SpadesNeil:
Well all cells default to 240 cost, and 2400 charge. That means 10 shots until recharge is required.

Yoshax:
The numbers are ridiculously arbitrary and messy.

Anewbe:
It's for clean multiples. 2400 is divisible by a lot.

Yoshax:
Why not make it properly clean and make it work off tens, or suchlike? Like normal power cells.

code/modules/projectiles/guns/energy/pulse.dm
Yoshax: I don't see why you're changing it. It's pointless to change it for no reason other than "lol why not".

Anewbe:
Could you elaborate on that, please? Honestly don't know how that would work.

SpadesNeil:
Want a gun to have 10 shots? 240 per shot.
Want a gun to have only 3 shots? 800 per shot.
Want a gun to have 8 shots? 300 per shot.

Going by multiples of 10 is too limiting.

mwerezak:
> Why not make it properly clean and make it work off tens

Why?

Yoshax:
Just seems to sort of complicate things when figuring out costs and suchlike. *shrug*

Anewbe:
SpadesNeil described more or less my reasoning, yes.

Yoshax:
I don't get it. Unless the cells are removable and interchangeable or whatnot. The numbers are more complicated :|

Anewbe:
Ideally they will be, at some point.

Anewbe:
Didn't need that DNM, sprites won't matter until they can be removed.

Yoshax:
That's going to cause huge balance concerns. I sort of like that idea, but I'm not really a fancy of powercells just for guns, it kinda complicates things. I know our power cells suck and are dumb, but it's simple :/

code/game/objects/items/weapons/power_cells.dm
Hubblenaut: It should be `/obj/item/weapon/cell/weapon/infinite/check_charge()`, as it's not a proc call, but a proc definition.
Likewise for the other line a bit below.

SpadesNeil:
We should worry about balance when the time comes. What if we want a gun that cannot be recharged with a charging bay, but can be popped open and have its battery replaced? It'd be a very neat feature.

Anewbe:
The balance concerns are why that's not in this PR.

Hubblenaut:
The issue is that the pwoer cells we have right now are the size of car batteries and have huge capacity, which is not needed for smaller devices like stun batons. The huge capacity is also the main reason why rechargers take half an eternity to recharge a gun, as it is the only variable that would make sense to adjust.

Yoshax:
They're only the size of car batteries because you're saying they are. I do not want, and don't care for different sized power cells for every little thing. It's a needless complication.

Hubblenaut:
They will need to be differently sized as the original power cells have a relatively large capacity, meant for usage in APCs and large pullable devices that take a lot of power, like portable pumps, space heaters and similar. In the sprites of these devices you can even see the size of the power cell relative to the character - and it is a size it makes **sense** to have considering its capacity and its industrial use.

Most people already found it ridiculous that the stun baton was enabled to somehow hold these cells. And those energy based guns already made use of custom cell capacity and will now simply need a proper item to reflect it.
Think of these like some kind of capacitor clips for your weapons.

Yoshax:
Having 1 battery type compatible with everything simplifies things and makes it more convenient. It's nice. Adding different types will branch into more types, which is clutter.

Hubblenaut:
It's not hard to separate battery types into industrial size, device sized and ammunition clips, to be quite frank here.

Yoshax:
We don't need multiple sized batteries, it's bloat and confusion. I'm really not in favour of this for the sole reason that it does so.

Spookerton:
Strong :-1: 

Energy weapons all almost universally represent an internal charge, and a cost per shot. I don't want to dig to look for where you've found that this isn't the case, but it's an exception, not a rule. Adding an extra level of abstraction to that is, as others have said, needless bloat as a single action.

Further, the obvious following intent is to make the batteries swappable. This is a problematic goal, in that energy weapons represent an existing utility trade - they are not reloadable, but their beams ignore a fairly substantial portion of normal obstacles, like windows. With the recent recharging tweaks, they're also not a pain to refill.

I understand where you're going cool-wise, but this is a needless change on its own, and with its only reasonable outcome in mind, not positive.

Anewbe:
The way energy weapons currently work is this: Gun is spawned. Gun then checks its max_shots var, multiplies that by charge_cost, then creates a new cell that has that much charge.

This PR drops that down to: Gun is spawned. Gun makes a cell of predefined type.

Don't see how that's abstraction, honestly. I also don't see how 4 new items is bloat. If the bloat's such a big deal, I can remove the larger cells.

Swapping batteries is indeed a future goal, but not within the scope of this change. **If** I ever get around to making battery swapping a thing, the balance will be discussed and tested thoroughly. But in my mind that issue will be dealt with when it needs to be dealt with.

Recharging still takes a long while, from what I've experienced. Yes, it doesn't take epochs anymore, but it's still long and something I'd rather not have to wait around for while I could be doing literally anything else.

Anewbe:
Can't unfuck this as is, closing for now. Will bring up for discussion elsewhere/when.

PR #2163, opened 2016-07-22
Communicator Texting for Ghosts

Neerti:
Ghosts can now text communicators in round, and can be texted by those back.
This is obviously logged, just in-case funny business occurs.
Adds verbs to display a log of messages that a ghost has sent and received, as well as a verb to send a message to a communicator of their choice.

### DIFF
--- a/code/game/objects/items/devices/communicator/communicator.dm
+++ b/code/game/objects/items/devices/communicator/communicator.dm
@@ -222,6 +222,7 @@ var/global/list/obj/item/device/communicator/all_communicators = list()
 
 /mob/observer/dead
 	var/datum/exonet_protocol/exonet = null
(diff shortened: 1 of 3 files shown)


code/game/objects/items/devices/communicator/communicator.dm
```diff
@@ -731,8 +739,11 @@ var/global/list/obj/item/device/communicator/all_communicators = list()
 //				Adds them to the list of people that have messaged this device and adds the message to the message list.
 /obj/item/device/communicator/proc/request_im(var/atom/candidate, var/origin_address, var/text)
 	var/who = null
+	world << "woo"
```
EmperorJon: world <<

PR #2166, opened 2016-07-22
Removes the straight jacket and muzzle from medical

Yoshax:
Straight jackets are very old and are rarely used, and when they are it is very unfun on the recieving player because they are unable to be removed without the help of anyone. The only time I see them used another solution could be used, and they are essentially abused when used.

Muzzles are also very old, and again, unnecessary, they are rarely used, and once again, when they are they are basically abused. One of the biggest uses I have seen in the past is people using them on Wizards, which in-itself is silly.

I feel we are better off without them.


serenemelody:
Yes, please. I've always hated those because 9/10 times they're just abused.

zekesulastin:
Meh.  There are still a bunch of ways to accomplish at least gagging with more common materials, and to be quite frank the handful of times I've seen these used the alternative would be permanent sedation for the sake of the medbay staff's sanity.

Besides, I'm sure some antagonist doc (medicult etc) can make good use of them, if they haven't already.

Yoshax:
If medical needs permanent sedation on someone, it's probably an OOC issue.

zekesulastin:
Doesn't really change any of the other points, or reduce the effectiveness of Medical's shutdown ability.

Yoshax:
There's no reason to not do something because other, just as bad, alternatives exist.

Neerti:
Rather not have this go in, since sometimes you really do need to use these to stop someone from being really shit in the brig.  I recall last hub weekend where that was the case.  Person kept trashing their cell, themselves, and any security who tried to enter, for over an hour, even inside of solitary.

serenemelody:
If you are absolutely opposed to it, then I'd like it to have better code (whenever I've seen it, taking it off seems to be troublesome) and to absolutely only be accessible by the CMO.

EmperorJon:
The straight jacket + muzzle combo is only ever useful for stopping anyone doing at all, even speaking. I'm of the opinion that if someone is going to attack themselves, any security who try to control them, and anything in their solitary cell, it's pretty much an OOC issue.

Yoshax:
If someone is being so shit in the brig that they need jacketed, then it's probably not an IC issue anymore.   Straight jackets are round-enders, and are bullcrap.

Anewbe:
I'm leaning on the side of removal/CMO lockering them. Sometimes they've got a use, but a lot of the times it's a shitcurity tool, an antag SSD trigger, or something that'd be better handled oocly. Admittedly haven't seen it used recently, but I'm not always around.

EmperorJon:
tbh if no decision is made on this in the next few days I'm probably just going to merge it...

Anewbe:
Leave it til the staff meeting, then?

EmperorJon:
Don't think this came up because Dev meeting. Will see what's happening with this later.

PR #2168, opened 2016-07-22
Multiplies all Telecrystal stuff by ten to setup for fine cost adjusting

Yoshax:
10x more to allow for fine cost tuning. E.G. having something cost 35 instead of 3 or 4.

Basically changes nothing other than numbers.


Yoshax:
This compiles, I'm not sure what's wrong: https://i.gyazo.com/1dcf0e977b4c3665970c629b52a9ca86.png

SpadesNeil:
This is nice because I always found some of he prices too high, but also found that if I wanted to change prices, they were suddenly too low.

Like...
Item A: 3
Item B: 3
Item C: 2

Item A is slightly better than B but B is much better than C. If I lower B, it causes imbalance, but if I raise A, then A becomes not worth it.

This little change helps allow potentially fixing that.

Yoshax:
Yes, I have plans to adjust costs but also introduce new, exciting things that aren't quite possible at the moment. This is just a PR to lay the groundwork and allow it to happen.

code/game/objects/items/weapons/cards_ids.dm
```diff
@@ -90,8 +89,12 @@
...
 	return 1
 /obj/item/weapon/card/emag/attackby(obj/item/O as obj, mob/user as mob)
-	if(istype(O, /obj/item/device/telecrystal))
-		src.uses += default_uses/2 //Adds half the default amount of uses which is more than you get per TC when buying, as to balance the utility of having multiple emags vs one heavily usable one
+	if(istype(O, /obj/item/stack/telecrystal))
+		var/obj/item/stack/telecrystal/T = O
+		if(T.amount/2 < 0)
+			usr << "<span class='notice'>You are not adding enough telecrystals to fuel \the [src].</span>"
+			return
+		src.uses += T.amount/2 //Gives 5 uses per 10 TC
```
Techhead0: If you give it 5 TC, does it add 2.5 uses?

Yoshax: I actually have no idea. I suppose I assumed that it would be handled by whatever.

PR #2173, opened 2016-07-22
Adds telecrystal teleporting

Yoshax:
Attacking yourself, or someone else will now teleport them to a random, safe location within 14 tiles. Thanks to Neerti for the Technomancer spell proc used.

### DIFF
--- a/code/game/objects/items/stacks/telecrystal.dm
+++ b/code/game/objects/items/stacks/telecrystal.dm
@@ -16,7 +16,7 @@
 		safe_blink(target, 14)
 		use(5)
 	else
-		user << "<span class='warning'>There is not enough telecrystal's to do that.</span>"
+		user << "<span class='warning'>There are not enough telecrystals to do that.</span>"
 
 /obj/item/stack/telecrystal/attack_self(mob/user as mob)
 	if(user.mind.accept_tcrystals) //Checks to see if antag type allows for tcrystals
--- a/code/game/objects/items/stacks/telecrystal.dm
+++ b/code/game/objects/items/stacks/telecrystal.dm
@@ -7,19 +7,16 @@
 	icon_state = "telecrystal"
 	w_class = 1
 	max_amount = 240
-	flags = NOBLUDGEON
 	origin_tech = list(TECH_MATERIAL = 6, TECH_BLUESPACE = 4)
+	force = 1 //Needs a token force to ensure you can attack because for some reason you can't attack with 0 force things
 
-/obj/item/stack/telecrystal/afterattack(var/obj/item/I as obj, mob/user as mob, proximity)
-	if(!proximity)
-		return
-	if(istype(I, /obj/item))
-		if(I.hidden_uplink && I.hidden_uplink.active) //No metagaming by using this on every PDA around just to see if it gets used up.
-			I.hidden_uplink.uses += amount
-			I.hidden_uplink.update_nano_data()
-			nanomanager.update_uis(I.hidden_uplink)
-			use(amount)
-			user << "<span class='notice'>You slot \the [src] into \the [I] and charge its internal uplink.</span>"
+/obj/item/stack/telecrystal/apply_hit_effect(mob/living/target, mob/living/user, var/hit_zone)
+	if(amount >= 5)
+		target.visible_message("<span class='warning'>\The [target] has been transported with \the [src] by \the [user].</span>")
+		safe_blink(target, 14)
+		use(5)
+	else
+		user << "<span class='warning'>There is not enough telecrystal's to do that.</span>"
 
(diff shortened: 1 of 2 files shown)


SpadesNeil:
Doesn't teleport into walls, does it?

Yoshax:
Refer to the safeblink proc here: https://github.com/PolarisSS13/Polaris/blob/e28c4361f76d2d39b8dce2c154abeb32376e26fa/code/game/gamemodes/technomancer/spells/blink.dm#L16

It's quite elegant in its selection.

SpadesNeil:
That answers that. :+1:

code/game/objects/items/stacks/telecrystal.dm
```diff
@@ -7,19 +7,16 @@
...
-			nanomanager.update_uis(I.hidden_uplink)
-			use(amount)
-			user << "<span class='notice'>You slot \the [src] into \the [I] and charge its internal uplink.</span>"
+/obj/item/stack/telecrystal/apply_hit_effect(mob/living/target, mob/living/user, var/hit_zone)
+	if(amount >= 5)
+		target.visible_message("<span class='warning'>\The [target] has been transported with \the [src] by \the [user].</span>")
+		safe_blink(target, 14)
+		use(5)
+	else
+		user << "<span class='warning'>There is not enough telecrystal's to do that.</span>"
```
Techhead0: Grammar.

SpadesNeil: ew, yea.

Current: "There is not enough telecrystal's to do that."
Correct: "There are not enough telecrystals to do that."

EmperorJon:
Don't traitors now have, what, 120 crystals? 
Five crystals seems a little cheap.

code/game/objects/items/stacks/telecrystal.dm
```diff
@@ -7,19 +7,16 @@
...
 	max_amount = 240
-	flags = NOBLUDGEON
 	origin_tech = list(TECH_MATERIAL = 6, TECH_BLUESPACE = 4)
+	force = 1 //Needs a token force to ensure you can attack because for some reason you can't attack with 0 force things
 
-/obj/item/stack/telecrystal/afterattack(var/obj/item/I as obj, mob/user as mob, proximity)
-	if(!proximity)
-		return
-	if(istype(I, /obj/item))
-		if(I.hidden_uplink && I.hidden_uplink.active) //No metagaming by using this on every PDA around just to see if it gets used up.
```
Iamgoofball: good for removing this

Neerti:
Does this remove the telecrystal adding to uplink functionality?

Yoshax:
It does not. That was code added in my previous PR taken from Bay, and deprecated code to us that wouldn't function. Our code that I added from our own Telecrystal file is below. Our TC amount is stored in the mind, and not the uplink. You use the TCs in hand to add them to your amount.

See here: https://github.com/Yoshax/Polaris/blob/0c7fdd715e8154c4942518c43bec915013ad5b07/code/game/objects/items/stacks/telecrystal.dm#L21

The idea of this PR came about because I was trying to think of a way to have telecrystal teleporting but not as an attack_self, because we already use that. So I decided upon this with the help of a few others.

EmperorJon:
I'd much rather the uplink just have an expensive hand teleporter or something, because I really don't like the idea that every traitor can at any point in time just teleport themselves away from trouble.

Yoshax:
It would give them a way to avoid conflict, or devolve conflict. If you're backed into a corner your choices are:
A - Murder them all
B - Get permabrigged

This adds an alternative. Nobody is going to spend TC on a hand teleporter, because, for one, the hand teleporter is fairly useless and spaces you half the time, and is already easily acquired.

EmperorJon:
Then I'd argue if the handteleporter is useless, make it more useful, rather than adding a different way of teleporting.

Yoshax:
Why give something to everyone? That makes the point of an antag being able to order nice items from their uplink redundant.

SpadesNeil:
>Then I'd argue if the handteleporter is useless, make it more useful, rather than adding a different way of teleporting.

Allegedly the hand teleporter has a pattern to it that can be studied and exploited, and isn't as random as everyone thinks.

Of course this is based on information that is not my own. I just know a player who used to frequent my server could use the hand teleporter to get literally anywhere they wanted with surprising precision.

Neerti:
Grammar issue pretty minor, gonna press the green button, unless you wanna put in a last minute commit to fix the grammar issue @Yoshax ?

Yoshax:
Corrected.

Neerti:
Gonna shove this in.  We can always tweak or revert it if it turns out that this is bad, but giving an option to traitors that allows them to escape a situation without a murderfest is always nice, and it's also a consumable, so it can't be spammed forever.

PR #2177, opened 2016-07-23
New stamps for warden and cargo.

Yoshax:
Add new stamps for warden and cargo.
Replace the stamps in their areas to the corresponding new ones.

Baystation12/Baystation12#13760


serenemelody:
Aw yiss! I've needed this for a while.

Yoshax:
I should note, the PR also adds normal cargo stamps to cargo tech lockers. Meaning they have their peasant stamp, then the QM has his big boy stamp.

PR #2179, opened 2016-07-23
Ports /vg/ get_turf

Yoshax:
Gotta go faster.

Baystation12/Baystation12#13576


Spookerton:
per [the thread](http://www.byond.com/forum/?post=2110095) linked in the bs12 PR:
perf chart

Lummox doesn't appear to mention it being a bug or anything they would change in future, so :+1:  for 5x faster atom2turf.

Issue #1563, opened 2016-05-06
Taping over eyes doesn't blind + initial tape doesn't appear


Verarch:
#### Brief description of the issue
Put tape over eyes and mouth for no reason, and had to remove and readd to get sprite to show, + blindfold doesn't blind
#### What you expected to happen
For blindfold to limit view somewhat and to appear on first applicatoin
#### What actually happened
It didn't do jack shit and had to remove to make it appear.
#### Steps to reproduce
1. Tape up eyse
2. Notice you aren't blind
3. Notice sprite is invisible on mob
#### Additional info:
 - **Server Revision**: master - 2016-05-06
 - **Anything else you may wish to add** no

Yoshax:
Duplicate #1413.

Datraen:
Partial duplicate, points out another bug.
I'll allow it.

Issue #1364, opened 2016-04-06
Heavy-Duty Cell Charger sprite


Anewbe:
#### Brief description of the issue
Sprite on the heavy-duty/industrial cell charger doesn't show up right when the cell is between 25% and 50% charged.
#### What you expected to happen
The second light would light up after 25%, then the third at 50%, and so on.
#### What actually happened
First light is on until 25%, then none of the lights are on until 50%.
#### Steps to reproduce
1. Make battery
2. Place battery in heavy-duty cell charger
3. Observe
#### Additional info:
 - **Server Revision**: Found using the "Show Server Revision" verb under the OOC tab.
 - **Anything else you may wish to add** (Location if it's a mapping issue, etc)

Issue #975, opened 2016-02-13
Removing a welding mask pushed up out of your face while wearing a gasmask doesn't update to show the gasmask


Atermonera:
Only difference between the two images is the removing of the welding mask.
[http://i.imgur.com/bWaOBrw.png](url)
[http://i.imgur.com/JAkvkyT.png](url)

PR #2189, opened 2016-07-24
Icon Fix Fun Time

SinTwo:
Fixes problems caused by incorrect names between mob .dmi's and obj .dmi's as well as fixing several icon related bugs.

Fixes:
#1563 
#1364 
#975 
Teshari Medical Smocks


Yoshax:
To autoclose things have to be on the same line.

Fixes: 1
Fixes: 2

PR #2202, opened 2016-07-25
Adds orderable tank transfer valves (not actual bombs) to the cargo Phoron crate

Yoshax:
They have been unobtainable for the longest time because of BOMBS!! but I feel as though they should be orderable. Because SCIENCE!!!


Mechoid:
This also would make blast-mining, which is now a thing, to be viable. (Just a note).

PR #2204, opened 2016-07-25
Increases accuracy penalties in detrimental states

Yoshax:
As is, you have a 50% chance to strike the chest when blurry eyes and confused, which I feel is too high. With this, you will have 15% chance to strike the chest, which I feel is far more sensible.

You should not be able to fight well in a state like that, but it should leave room to perform other actions.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -159,9 +159,9 @@ emp_act
 	if(user.eye_blind)
 		accuracy_penalty += 75
 	if(user.eye_blurry)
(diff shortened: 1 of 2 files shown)


EmperorJon:
I think 50% is maybe too high, yes, and a reduction could be nice, but I feel 15% is too far the other way. Personally I'd expect somewhere between 25-33%?

Yoshax:
To keep the melee and ranged accuracies the same it would have to be 30%.

Anewbe:
Gonna give this a go-through. If it needs to be dialed down it can, but right now the flash is just an anti-borg tool.

PR #2212, opened 2016-07-26
Some EVA changes

Yoshax:
Screen shot: https://i.gyazo.com/984f1ef6bb51bb2f044dfdc6624ff382.png

- Unmaps breacher
- Maps in rescue rig for medical
- Maps in eva rig for engineering
- Security get a potted plant because no applicable rig for them


Anewbe:
For the moment I'm very eh on this, will need to double check what those two RIGs do, but they're damn good iirc, and I actually really like the Breacher, for some odd reason.
Will post secondary opinion later, if applicable.

Potted plant gets my :+1: though, as long as I can resprite it into a cactus later.

EmperorJon:
:+1: for unmapping the breacher and :+1: for giving Sec the most important and vital equipment they've received in recent months.

Yoshax:
RIGs are basically unused as is, they can be strong, yes. But they exist, and they should be used more.

You can change the icon_state on the plant, there's a lot of them, I was just too lazy.

Neerti:
Bit concerned that I'm fairly sure the impetus behind the removal for the breacher was due to certain events that occurred just earlier today.

Yoshax:
I am going to agree with you Neerti, this PR has my feelings in it. I did fight someone using a breacher, I also won the fight with them. However, I would like to think I can, even with my feelings, be rational. There are many others agreeing with my decision, I asked for opinion in dead chat at the time, and many people there agreed.

It is far too armoured. The slowdown becomes all but irrelevant in most fights because they are ranged. It allows you to tank far too many hits. It has 60+ armor in all it's slots, that is quite ridiculous. To note, the specialised armours such as the bullet resistant have 80 in their respective fields. Getting 60 armour against everything is far too much. It is unnecessary, the item is unnecessary.

martin509:
👍 for adding in the rescue rig, as it's rarely used and very neato.
My thoughts on the Breacher are that it's meant for Unathi, who only do melee combat as a cultural thing, so the slowdown would matter much more in those circumstances.

Hubblenaut:
I don't think potted plans fit into an EVA storage designed to be of purely functional use, to be perfectly honest.

Yoshax:
~Aesthetics~

PR #2217, opened 2016-07-28
Grenade Buff

Woodratt:
Or making grenades great again?

Tweaked the code some, changed the number of fragments spawned from 50 to 63. Damage remains the same, fragments now have a thirty percent armor pen. This also means that how lethal grenades are has gone up by a lot. One landing at your feet is lethal. Standing next to one will seriously hurt. Tile or two back the dropoff is still in effect. 

Basically need some live testing in round with real folk to get a idea of how bad or good this will be. Will likely need to tweak the damage downwards instead of upwards.

Also adjusted the grenade path thingy stuff to select the type of fragments (taken from bay code), on the off chance we add diff types of fragmentation grenades like bay does.

### DIFF
--- a/code/game/objects/items/weapons/grenades/explosive.dm
+++ b/code/game/objects/items/weapons/grenades/explosive.dm
@@ -17,7 +17,8 @@
 //Explosive grenade projectile, borrowed from fragmentation grenade code.
 /obj/item/projectile/bullet/pellet/fragment
 	damage = 10
-	range_step = 2
+	armor_penetration = 30
+	range_step = 2 //projectiles lose a fragment each time they travel this distance. Can be a non-integer.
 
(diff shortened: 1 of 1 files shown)


Anewbe:
Can be tweaked later, merging.

code/game/objects/items/weapons/grenades/explosive.dm
```diff
@@ -32,9 +33,8 @@
 	icon_state = "frggrenade"
 	item_state = "grenade"
 
-	var/num_fragments = 50  //total number of fragments produced by the grenade
-	var/fragment_damage = 10
-	var/damage_step = 2      //projectiles lose a fragment each time they travel this distance. Can be a non-integer.
+	var/fragment_type = /obj/item/projectile/bullet/pellet/fragment
```
dangranos: Is that variable actually used anywhere?

PR #2224, opened 2016-07-29
Fixes two issues on map for Security

SpadesNeil:
First off, this isn't correct. Someone forgot to delete these areas on the left when they moved the escape pods.
capture
So I deleted them.

Second, the lighting in the security briefing room was shit, so I fixed that by adding one light. I moved the bulletin board to the other side of the room and put in a light for best coverage.


Yoshax:
How does 13 tiles of area, a light and a board come out to a diff of 240~ lines?

SpadesNeil:
I have no idea. I did use the map cleanup tool correctly. Hell, I did it both times. Originally I was just going to fix the escape pod.

Yoshax:
Reverted half our map changes that apparently I somehow missed when I had the map open.

SpadesNeil:
Well shit.

PR #2228, opened 2016-07-29
Makes soulstones take halloss into account.

Cameron653:
Soulstones had an error where it added halloss instead of subtracting it.

This resulted in halloss preventing people from being soulstoned, even if they were in critical.

This PR fixes that.


Hubblenaut:
Halloss is or used to be part of the health calculation (needs to be checked), that way by adding it to the player's health, you get their effective health while not taking the pain into account.

That said, subtracting it doesn't make sense if the intention of the code should stay the same.

Cameron653:
Ack, my bad. Going to close this PR down since it's intended.

PR #2231, opened 2016-07-30
Sprays now respect blocking

Anewbe:


Yoshax:
Does it actually work, what with the broken ass shields?

Anewbe:
Worked in testing, at least.

PR #2232, opened 2016-07-30
510 scheduler

SinTwo:
Ports Baystation12/Baystation12#13583 and Baystation12/Baystation12#13844
So basically it's a port of a port.


code/game/machinery/adv_med.dm
```diff
@@ -410,7 +410,7 @@
 			visible_message("<span class='notice'>\The [src] rattles and prints out a sheet of paper.</span>")
 			var/obj/item/weapon/paper/P = new /obj/item/weapon/paper(loc)
 			P.info = "<CENTER><B>Body Scan - [href_list["name"]]</B></CENTER><BR>"
-			P.info += "<b>Time of scan:</b> [worldtime2text(world.time)]<br><br>"
+			P.info += "<b>Time of scan:</b> [stationtime2text(world.time)]<br><br>"
```
PsiOmegaDelta: All lines that say ```stationtime2text(x)``` should likely be ```station_adjusted_time(x)```. You'll note that the proc doesn't actually take or do anything with arguments.
It's something I need to correct for Bay.

PsiOmegaDelta: Actually, I lied.
station_adjusted_time() takes a duration (say, 30 minutes) and returns the final station time.

PsiOmegaDelta: #13844 has the necessary time changes to make autopsy results report correctly (as well as fingerprints, attack logs, and other things).

dangranos: Btw, what is that addition to the exception catching proc? In `scheduler.dm`
at line 329.

EmperorJon:
You removed "needs changelog" but didn't actually add a changelog.

SinTwo:
I accidentally added that label; regardless, changelog added.

Anewbe:
After quick scan, do not see what this might break. Tentative :+1:

EmperorJon:
In that case... witness me.

Neerti:
Due to a rather serious series of bugs that resulted in immortal cats, we're gonna revert this and then try to work out the bugs before a re-attempt.

PR #2238, opened 2016-07-31
Adds "syndicate" dufflebags to the uplink

Yoshax:
Adds them for 10 TC, which I feel is fair for just a bag. It does however, have no slowdown.


Anewbe:
A number of non-duffel bag items are moved slightly by this, is that intended?

Yoshax:
Yes. It's in my attempt to keep the uplink datum files in a semi cost ascending order.

PR #2239, opened 2016-07-31
Uplink pricing adjustments

Yoshax:
- Fixes the surplus crate item worth scaling because I can't math
- Halves the price of voice changer and camera floppy disk thing because 30 is too damn high
- Reduces cigarette kit and toxin beaker from 15 to 10 because they're not very useful and should cost a little less
- Reduces cost of radio keys because nobody ever uses them
- Reduces cost of clerical kit and money briefcase to be a little more sensible
- Reduces powersink cost because it's so unused (Will look into attempting to tweak the numbers at some point in the near future so it drains better)
- Reduces combat knife price a little because 30 seems a little high

### DIFF
--- a/code/datums/uplink/badassery.dm
+++ b/code/datums/uplink/badassery.dm
@@ -61,17 +61,17 @@
 /datum/uplink_item/item/badassery/surplus/merc2
 	name = "Surplus Crate - 240 TC"
 	item_cost = DEFAULT_TELECRYSTAL_AMOUNT * 2
-	item_worth = 360
+	item_worth = 540
 
 /datum/uplink_item/item/badassery/surplus/merc4
 	name = "Surplus Crate - 480 TC"
 	item_cost = DEFAULT_TELECRYSTAL_AMOUNT * 4
-	item_worth = 720
+	item_worth = 1200
 
 /datum/uplink_item/item/badassery/surplus/merc6
 	name = "Surplus Crate - 720 TC"
 	item_cost = DEFAULT_TELECRYSTAL_AMOUNT * 6
-	item_worth = 1440
+	item_worth = 1980
 
 /datum/uplink_item/item/badassery/surplus/New()
 	..()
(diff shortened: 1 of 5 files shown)


Anewbe:
In my experience, the power sink actually drains rather well. Might be alone in that sentiment, but I basically always grab one for shiggles.

Hubblenaut:
The powersink draws 1.5 MW, thus with the usual engine setup drawing about 500 kW from stored power, up to a maximum of 5 billion kW-ticks.

I'd suggest having it draw with 2 MW instead and leaving its price the same. 5 billion kW-ticks is roughly 200 times more than what the main SMES holds.

SpadesNeil:
Power sink also explodes when full. Big kaboom.

Hubblenaut:
Yes, though I don't see that happening with 5 billion kW-ticks capacity.

EmperorJon:
Explosion was either removed or has been bugged for a long time (or so I have been told).

Yoshax:
The power sink is essentially useless at the moment. I am going to look at further tweaks in the future, but I would like to see the price lowered because it is too high, and it is not a huge change.

Hubblenaut:
It's not essentially useless - it drains 50% more power than is generated by the engine, and drains 150% of the power that can be output by the Main SMES at maximum power.

It's doing fairly well.

SpadesNeil:
>Explosion was either removed or has been bugged for a long time (or so I have been told).

You may want to test that because I actively removed explosions from power sinks for being too griefy for my server's uses.

Yoshax:
It's basically unused, and basically useless. It's easily found, slow, and just a poor distraction.

Hubblenaut:
What I just told you is that it is not useless at all, and have given you the underlying facts for that. I have seen it used myself several times within the past few weeks, and its main purpose *is* to be a distraction and unlock ways into secure areas by powering them down.

Yoshax:
Sure, and I feel it is not worth the TC it currently costs. I have played a lot of antag recently, and as such have probably more first-hand experience in the last weeks than anyone else. I have spoken to several others, and they have felt the same. There is no reason, or incentive to buy it.

Anewbe:
Eh, numbers can be tweaked again if needed, merging.

Issue #2141, opened 2016-07-15
Drone Brains are MMI's in code


SilverTalismen:
#### Brief description of the issue

Drone Brains can't be activated to recieve players like a posibrain can be, they work like MMI's and need a human brain
#### What you expected to happen

for it to be activated like a posibrain and not function like a MMI
#### What actually happened

Activating them gives no prompt for ghosts to inhabit it, and from what I've been told code wise they are just MMI's
#### Steps to reproduce

have someone print a drone brain, activate it and watch as you get no prompt to inhabit it as a ghost
#### Additional info:
- **Server Revision**: Found using the "Show Server Revision" verb under the OOC tab.
- **Anything else you may wish to add** (Location if it's a mapping issue, etc)

Anewbe:
No longer pops out a brain, still doesn't have a call for ghosts.

PR #2243, opened 2016-08-01
Drone brains work like posibrains

Anewbe:
#2141


Yoshax:
Seems kinda messy to just duplicate all of the code again.

Neerti:
Might be best to base the 'shove a ghost inside this' code on the /mmi/digital/ part, of which I believe posibrains and circuitbrains both inherit from.

Yoshax:
You could port Bay's Ghost Traps which basically do this 1000% better.

Anewbe:
Porting is hard/scary, was going to do what Neerti said.

PR #2244, opened 2016-08-01
Adds pepperspray and a crowbar to the HoS locker

Anewbe:
Found it a little odd that he doesn't get them.


Hubblenaut:
You could remove the riot shield since you're at it, as the HoS also has the telescopic riot shield now.

PR #2253, opened 2016-08-01
Construction Tweaks and Fixes

Hubblenaut:
**Reinforced Walls**
 - Fixes reinforced walls construction when having one sheet left in the stack

**Windoor Assemblies**
 - Assemblies are now made with two sheets of either normal or reinforced glass sheets, for normal and secure windoors respecting.
 - Assemblies can now be named with a pen, just like airlocks.
 - Windoor assemblies now get placed like windows, and windows take windoor assemblies into account when placed. As a side effect, that part of the code looks much nicer now, as windows and windoors use the same placement algorithm when constructed.
 - Decapitalises windoor assembly names.
 - Shattering windoors will make them drop exactly what they were made of now. (Two shards, a piece of wire and the airlock electronics, optionally two rods if the windoor was secure)
 - Prying emagged windoors will no longer make them shatter on top of returning to an assembly.
 - Cleans up some code and adds its own `update_icon()` proc.

**Airlock Assemblies**
 - Fixes airlocks occasionally not updating their icon correctly.
 - Fixes airlocks causing sparks on spawn as if they were damaged.
 - Fixes airlock not being anchored after being turned into an assembly.
 - Decapitalises airlock names.

**Windoors and Airlocks**
 - Assemblies now show the name they will have after construction in brackets.

**Windows**
 - Windows that get unanchored by force will now correctly have the rotate verb.

*All bugfixes and tweaks thoroughly tested, all is working.*

### DIFF
--- a/code/game/machinery/doors/airlock.dm
+++ b/code/game/machinery/doors/airlock.dm
@@ -412,12 +412,16 @@ About the new airlock wires panel:
 	if(electrified_until && isAllPowerLoss())
 		electrify(0)
 
+	update_icon()
+
 /obj/machinery/door/airlock/proc/loseBackupPower()
 	backup_power_lost_until = backupPowerCablesCut() ? -1 : world.time + SecondsToTicks(60)
 
 	// Disable electricity if required
 	if(electrified_until && isAllPowerLoss())
 		electrify(0)
+
+	update_icon()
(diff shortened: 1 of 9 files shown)


Hubblenaut:
yay

PR #2282, opened 2016-08-07
Item force tweaks

Yoshax:
Many tools and other items have their force set low. For example, toolboxes. They are heavy, metal boxes yet they did 5 damage.


Kelenius:
Let's break bones with two smacks of fire extinguisher? There is a reason why these were nerfed.

Yoshax:
Changed? Nerfed? They haven't been touched since pre-2012.

Kelenius:
I believe that a lot of items were nerfed from their tg values to what we have now a long time ago.

Yoshax:
Perhaps. But I do feel some were a tad heavy-handed, so I have made a few more adjustments.

EmperorJon:
I don't really have an issue with these. :+1:

PR #2283, opened 2016-08-07
Embed Tweaks

Anewbe:
Fireaxes should no longer auto-embed
Adds a var that will disable embedding in melee weapons/items

### DIFF
--- a/code/game/objects/items.dm
+++ b/code/game/objects/items.dm
@@ -20,6 +20,7 @@
 	var/list/origin_tech = null	//Used by R&D to determine what research bonuses it grants.
 	var/list/attack_verb = list() //Used in attackby() to say how something was attacked "[x] has been [z.attack_verb] by [y] with [z]"
 	var/force = 0
+	var/embed = 1	//Whether or not the item can embed
 
 	var/heat_protection = 0 //flags which determine which body parts are protected from heat. Use the HEAD, UPPER_TORSO, LOWER_TORSO, etc. flags. See setup.dm
 	var/cold_protection = 0 //flags which determine which body parts are protected from cold. Use the HEAD, UPPER_TORSO, LOWER_TORSO, etc. flags. See setup.dm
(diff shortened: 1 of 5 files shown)


Hubblenaut:
Your logs are a bit confusing. So what this actually does is:
 - Removing additional chance of sharp objects to embed.
 - Making energy-based melee weapons never embed.

Why is the first one necessary?

Yoshax:
I don't really see the point in the first. Energy weapons like energy swords shouldn't ever embed, but the other embed chance stuff is fine. Changing them to be less likely to embed will actually really upset some balance.

Anewbe:
A more accurate way of putting that first point would be "Strong sharp weapons no longer auto-embed", in that it removes something that forces sharp weapons above a certain threshold to embed. I think the only things it actually effects are fireaxes and spears with really high damage. Anything below that threshold (including knives and cultblades) will continue to embed at the same rate as now, while the stronger things will embed with the same formula, but different numbers.

Yoshax:
So what're the actual chances for weapons and suchlike to embed?

Anewbe:
Current:
var/embed_chance = weapon_sharp? damage/I.w_class : damage/(I.w_class*3)
var/embed_threshold = weapon_sharp? 5*I.w_class : 15*I.w_class


 if((weapon_sharp && damage > (10*I.w_class)) || (damage > embed_threshold && prob(embed_chance)))

New:
Same embed_chance and embed_threshold

if(damage > embed_threshold && prob(embed_chance))

It seems that the damage a wielded fireaxe does is 42, so the embed chance is 10.5, and the embed threshold is 20. Right now, since the damage is 42 (Greater than 10 times the axes weight class), it always embeds on a non-armored target.

This PR makes it so the axe uses the same formula as everything else, with an embed chance of 10.5% on every single hit it makes.

Anewbe:
Gonna close this, do it better, and reopen it later.

PR #2284, opened 2016-08-07
Rewrites xenoarch

Kelenius:
See changelog

#1410


Yoshax:
This doesn't really change the way it's played, does it? Curious as to whether or not a new wiki page or suchlike is required.

Kelenius:
The big changes are that suspension generator doesn't have modes - so you'd need to remove that part; and that finds are now in a cluster - pretty obvious at the first glance but worth mentioning. Mostly it's the same.

Yoshax:
Would you be willing to write out the modifications or suchlike required? I'm not sure we have anyone else particularly familiar with Xenoarch. Other than that, I think this is good, especially seeing as it was ready for merge last time.

Kelenius:
Sure.

Kelenius:
Resolved.

PR #2288, opened 2016-08-08
Fixes lung rupturing

Yoshax:
I am not particularly pleased with how this works, or what it does. However, I have been asked when I mentioned it before to put my fix up because it's better than it not working, so here you go.

I have given it a 5% chance because it's rather low and fair, but high enough at the same time.


Screemonster:
Strictly speaking, if it were to be realisms, a lung rupture would be related to pressure rather than oxygen content. Specifically a sudden loss of external pressure - hence why one of the first things they teach you when learning to scuba dive is never, ever, under any circumstances, hold your breath.

Yoshax:
Slightly less bad code. Yay.

Anewbe:
Good enough/10

PR #2298, opened 2016-08-11
Pet shuffling + changes

Yoshax:
- Unmaps Pun Pun, because having a monkey makes no sense, and it isn't the early 1900s where having a smoking monkey is cool, its also unhygienic because it probably shits all over the floor.
- Moves Runtime from the CMO's office to the Psychiatrist office because for one, it's again unhygienic and the psychiatrist can have it for cat animal therapy or whatever.


EmperorJon:
I'd rather Runtime was in the staff room so I can steal her all the time but okay :P

Kelenius:
It's unhygienic in CMO's office but fine in psychiatrist's office?

Yoshax:
Yes.

sabiram:
more medical care takes place in the psychiatrist's office than in the CMO's. the only reason it's more 'hygienic' is because nobody plays psychiatrist and so the office is just an empty room.

Yoshax:
Okay. Allow me to just modify this.

ursamedium:
I dunno. I feel like cats have some precedence as use in therapy, as do dogs -- more well known with dogs used in mental health, ofc. Medical break room is actually less hygenic, imo, because the medical doctors who need to be cautious of being clean and keeping sterile are closer -- a psychiatrist's job is less hands-on than the medical doctor job, while they need to remain clean as any doctor should, they aren't going to be potentially treating open wounds or examining orifices, etc., I've actually visited several over the years irl that have an animal in-office. 
Granted, less logic-wise, I'm more for break room Runtime because more likely to have someone in there to pet, or retrieve Runtime to bring to medbay patients in emotional distress that could use cat pets or something.

Hubblenaut:
Long live the queen of pets.

SpadesNeil:
Good, I hate Pun Pun.

Anewbe:
Toss this one to the Forums, please.

PR #2308, opened 2016-08-12
Adds a verb for toggling specific pieces of underwear

Hubblenaut:
Does what it says on the tin.
Original idea was to be able to take off the undershirt and socks, but due to the nature of underwear code it turned out to be a verb that works with all pieces of underwear.

Can restrict it to only undershirt and socks if it's so wished.

### DIFF
--- a/code/datums/underwear/underwear.dm
+++ b/code/datums/underwear/underwear.dm
@@ -9,27 +9,34 @@
 *************/
 /datum/category_group/underwear
 	var/sort_order		// Lower sort order is applied as icons first
+	var/display_name	// For displaying in text
+	var/gender = NEUTER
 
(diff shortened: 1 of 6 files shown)


Hubblenaut:
No Party?

Anewbe:
Would prefer it as undershirts and socks, but I'm a relic.

Yoshax:
:+1:

Hubblenaut:
Lewd player took off their panties. (JMP) (BAN)

EmperorJon:
Definitely needs admin logging with a quick ban link.

EmperorJon:
In all seriousness though, nobody seems to have opposed this, so

PR #2309, opened 2016-08-12
Refactor of admin fax handling 

Yoshax:
Ports Baystation12/Baystation12#13795

Ontop of that, adds the ability to select between the Nanotrasen and SolGov logo for the fax header.


Anewbe:
Tested and functional?

Yoshax:
Tested and functional.

Yoshax:
The only withstanding issue is that the logo insertion is not primed for future expansion due to it being an alert and that you only get one chance to select the logotype. But, I'm happy enough with at as is and those problems can be resolved when we actually have a 3rd logo.

Hubblenaut:
Long live the fax queen

PR #2313, opened 2016-08-13
Fixes dirt

Yoshax:


html/changelogs/Yoshax - dirt.yml
```diff
@@ -0,0 +1,36 @@
...
+#   rscdel (general deleting of nice things)
+#   imageadd
+#   imagedel
+#   maptweak
+#   spellcheck (typo fixes)
+#   experiment
+#################################
+
+# Your name.  
+author: N3X15
```
PsiOmegaDelta: Author name?

Hubblenaut: Yosh pls

code/game/turfs/simulated.dm
```diff
@@ -79,7 +79,7 @@
 	tracks.AddTracks(bloodDNA,comingdir,goingdir,bloodcolor)
 
 /turf/simulated/proc/update_dirt()
-	dirt = min(dirt++, 101)
+	dirt = min(dirt+1, 101)
```
Techhead0: People need to learn the difference between pre- and post-decrement. And for these purposes, the fix is better.

Hubblenaut:
Long live the dirt queen

PR #2314, opened 2016-08-14
Taj cleanup

Anewbe:
Game words should match the lore now, skipped the Location file.


zekesulastin:
Send an IRC poke when this gets accepted so we can switch the whitelist entries to the new name.

EmperorJon:
:+1:  Will merge when Zeke is around.

PR #2317, opened 2016-08-14
Removes Chaplain morgue access

Anewbe:


Kelenius:
Wait what why.

Neerti:
Relic from /tg/, and the Chaplain can't even reach the Morgue now due to the current map design anyways.

PR #2318, opened 2016-08-14
New Medical Belt Sprites

Hubblenaut:
**Before:**
Image

**After:**
Image

They are mostly white now, with just a tint of blue color the hue of the EMT labcoats. This way they will visually fit to most, if not all, medical uniforms.


Kelenius:
But why? Security belt is not red, engineering belt is not yellow. All belts are secondary uniform colors.

Yoshax:
Because the orange/pink/yellow or whatever you want to call it is an eye-sore.

:+1:

Anewbe:
Leave the old belt sprite as like medbelt_old or something, I like it for reasons. But yeah, white one looks nice.
Might want to take this one to the community, but it gets a :+1: from me.

SinTwo:
Make sure to update the in-hand icon.

Hubblenaut:
Good call, updated.

EmperorJon:
Looks much less "what?" than the original. Seconding Yosh on the weird salmon colour we have right now... :+1:

PR #2321, opened 2016-08-15
Fixes everyone and their dad having QM access

Yoshax:


SpadesNeil:
It amazes and amuses me how long this has been an oversight.

PR #2322, opened 2016-08-15
CMO and Doctors spawn with Health Analyzer

Hubblenaut:
Health Analyzer will spawn in suit storage instead of the penlight.


zekesulastin:
I cannot thumbs this up enough.  Doctors used to get an analyzer in the white kits they spawned with, but red kits do not include them.

PR #2332, opened 2016-08-16
Makes pierce damage heal correctly when bandaged

Cameron653:
Makes pierce damage heal correctly when bandaged.

For some reason it didn't heal before. I presume this was an oversight, as it means that if you got a pierce wound that had more than 15 damage nothing less than bicaridine, tricordazine, cryo, or some other brute damage healing chemical would fix it.

If this is intended, feel free to close the PR. It just seems extremely odd that every other brute wound type will heal when bandaged with the exception of pierce damage.

### DIFF
--- a/code/modules/organs/wound.dm
+++ b/code/modules/organs/wound.dm
@@ -92,7 +92,7 @@
 
 	// checks whether the wound has been appropriately treated
 	proc/is_treated()
-		if(damage_type == BRUISE || damage_type == CUT)
+		if(damage_type == BRUISE || damage_type == CUT || damage_type == PIERCE)
 			return bandaged
 		else if(damage_type == BURN)
 			return salved


Yoshax:
No, this is working as intended. Because of the lines:
```	proc/can_autoheal()
		if(src.wound_damage() <= autoheal_cutoff)
			return 1```

wound_damage takes into account the number of wounds you have. Punctures wounds do not merge, so it is unlikely for you to have a high damaging one, and if you do, it is offset by number of wounds. However, I will defer this to the others.

@Neerti @Anewbe @EmperorJon

Cameron653:
Ah, alrighty. I'll close it down since it's intended,

Yoshax:
I'm not sure about that. I'd like to leave it to the others, because it seems like it should be the way proposed here.

Yoshax:
I guess nobody else cares, so all is good.

PR #2338, opened 2016-08-17
Tweaks Sec Voidsuit armor

Anewbe:
Drops 10 points from melee for 20 in bullet, so from 60, 10, 30 to 50, 30, 30. Should make the suit about as protective as the Officer vest in a gunfight.


SpadesNeil:
This seems like a good, fair change. I am of the personal opinion that spaceworthy armor should be slightly less good than its space-faring counterpart, but that's just me.

Either way this is more consistent than the nonsense it was before.

PR #2339, opened 2016-08-17
Tweaks the Skrellian Voidsuit armor values

Anewbe:
Matches the actual voidsuit, rather than the random values it had before.

### DIFF
--- a/code/modules/clothing/spacesuits/alien.dm
+++ b/code/modules/clothing/spacesuits/alien.dm
@@ -1,8 +1,8 @@
-//Skrell space gear. Sleek like a wetsuit.
+//Skrell space gear. Sleek like a wetsuit. //TODO: MAKE THIS AN ACTUAL VOIDSUIT, WITHOUT THE CYCLER FUNCTIONALITY
 /obj/item/clothing/head/helmet/space/skrell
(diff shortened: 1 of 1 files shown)


Yoshax:
I was always a fan of it being different and unique, especially with its radiation shielding *shrug*.

Anewbe:
I mean, my issue is with the high laser/energy values... 100 rad also seems like a bit much to me, but eh.

Yoshax:
If you're going to turn it into just a resprited voidsuit you'd be as well just removing it.

ParadoxSpace:
They were fine before this imo.

Hubblenaut:
They were added in a time voidsuit modulation for different races wasn't a thing. Now they seem to be just better alternatives than the other voidsuits. I wouldn't mind if they go.

redstryker:
I like the idea that different voidsuits have different armor values, since obviously the technology of different species would be tailored differently depending on how their physiology is, but *shrug*.

Anewbe:
In hindsight, since I plan on making the item into a functional voidsuit eventually, I'm going to close this until the mapping PRs are merged or closed, will rethink it then.

PR #2344, opened 2016-08-18
Makes hyperzine not make you move sanic speeds

Cameron653:
Hyperzine makes you fast. Due to some change a while back, it now makes you 3X as fast as it used to. This results in players moving _at the speed of fucking light_ instead of being a chemical that was helpful but not insanely over powered.

Showing how fast you are on hyperzine: https://i.gyazo.com/7d52bc71b74d435f8f45c74ddd3d0591.mp4

Hyperzine is easy to make (Literally 3 clicks in the chem dispencer), can be made into 14.88U pills to prevent ODing, and lasts bloody ages to fully metabolize.

This PR makes it so hyperzine makes you a _little_ faster, but not speeds that puts sonic to shame,

### DIFF
--- a/code/modules/mob/living/carbon/human/human_movement.dm
+++ b/code/modules/mob/living/carbon/human/human_movement.dm
@@ -14,7 +14,7 @@
 		return -3 // Returning -1 will actually result in a slowdown for Teshari.
 
 	if(CE_SPEEDBOOST in chem_effects)
-		return -3
+		return -1
 
 	var/health_deficiency = (maxHealth - health)
 	if(health_deficiency >= 40) tally += (health_deficiency / 25)
--- a/code/modules/mob/living/carbon/human/human_movement.dm
+++ b/code/modules/mob/living/carbon/human/human_movement.dm
@@ -14,7 +14,7 @@
 		return -3 // Returning -1 will actually result in a slowdown for Teshari.
 
 	if(CE_SPEEDBOOST in chem_effects)
-		return -1
+		return -3
 
 	var/health_deficiency = (maxHealth - health)
 	if(health_deficiency >= 40) tally += (health_deficiency / 25)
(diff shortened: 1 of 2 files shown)


Yoshax:
Change was made by @Neerti in a Technomancer PR, so deferring to him.

Screemonster:
I don't mind a turbo-sanic-speed-with-no-tangible-downsides chem existing, but maybe not so easily obtainable. 1x speed boost is great for miners and diona without being overpowered for everyone else, and faiiiirrrly reasonable for a thing that's producable at roundstart. 3x turbobastard chems would be better suited to shit like those zoom pills that cause brain damage, or something gated behind materials.

Neerti:
Reason it was set to -3 was due to that at the time, -1 was quite possibly the fastest obtainable speed, back when it was made.  With the advent of taj and now teshari, they will actually move slower than their default speed at -1 instead of -3, so I figured I would fix that oversight while adding the ``force_max_speed`` variable for Technomancer.

Also, I'm fairly sure going from -1 to -3 is a linear change and not a 'three times more fast' change.  Regardless, I don't really mind it going back down to -1, but perhaps it would be better to change the chemical itself to metabolize faster or something if it's such a pressing concern.

Yoshax:
If -1 makes Tajaran and Teshari slower than their default speed then it shouldn't be at -1, imo.

Anewbe:
-1 should make Taj go their normal speed and Tesh go slower, last I checked. Speeding the metabolism seems like a better change in my mind.

Cameron653:
I'll change it back to -3, but should the metabolism be bumped to to say... 2 per tick? So a 14u dosage (About the highest you can get without ODing) will just last 7 ticks then fall off?

Anewbe:
7 ticks is nothing, try like REM * 1.5, maybe.

Cameron653:
Alright. That'll give 46 ticks of speed using a 14u pill.

Anewbe:
Going to close this, get a more moderate number in a PR. If it needs further tweaking, it can be tweaked further, but I've not seen Hyperzine abused before or since the speedboost change, and the proposed number feels too heavy-handed.

Issue #2340, opened 2016-08-17
Sawing a double-barrel shotgun with any shell in it = perpetual pain


zaredman:
#### Brief description of the issue

  Za_Redman/(Donovan Lowstetter)(JMP): if you saw the barrel on a double-barrel with ammo still in it, it'll always go off
  Za_Redman/(Donovan Lowstetter)(JMP): even if the ammo is already used
#### What you expected to happen

It would only go off twice.
#### What actually happened

It goes off more often than a rioter in Milwaukee
#### Steps to reproduce

1) Take double-barrel shotgun
2) Take any ammo
3) Ensure the ammo is in the shotgun
4) Obtain saw
5) Attempt to saw
6) Embrace the pain, as it will always go off as long as there is any sort of shell in there.
#### Additional info:
- **Server Revision**: master - 2016-08-16

Issue #1829, opened 2016-05-31
Loaded double barrel shotgun can be sawed infinitely for infinite shooting in face for no damage.


SpadesNeil:
#### Brief description of the issue

There's something wrong with the (already very hack) double-barrel shotgun code.
#### What you expected to happen

When I saw the gun, it's supposed to shoot me in the face for messing with it, and it's supposed to only do that once, maybe twice. I say once for every round loaded in the gun as long as the round(s) have BBs in them to fire.
#### What actually happened

![capture](https://cloud.githubusercontent.com/assets/6484985/15663853/3a0dce90-26cd-11e6-9793-4856878d907c.PNG)
Note that I only got damaged twice in all my spamming of the saw button.
#### Steps to reproduce
1. Obtain unmodified double barrel shotgun.
2. Leave the ammo inside.
3. Saw the weapon. Nay, spam saw it.
4. Wonder why you're not dead and why the shotgun suddenly has infinite ammo.
#### Additional info:
- **Server Revision**: https://github.com/PolarisSS13/Polaris/commit/dbcf3b18c1f392a7913fb2349f59108eb8797287
  - **Additional info**: Also I'm pretty sure it should make a buzz saw noise when you start cutting it. It doesn't.

PR #2346, opened 2016-08-19
Sawing shotguns now shoots you correctly

Anewbe:
#2340 #1829 
Shoots if the bullets aren't spent, otherwise just clicks.


SpadesNeil:
Closed? Why closed?

PR #2347, opened 2016-08-19
Tweaks the lowest level of heat damage to be less insane

Anewbe:
Lowest tier of fire should act more like a warning than horrific firey death, giving you an opportunity to get into a less toasty place.


Yoshax:
Is this the temperature level for stuff like smashing a bottle over someones head and setting them on fire?

Anewbe:
I think that one moves into the second tier, but as is the moment you cross that threshold you drop 5 health per tick, which is more than I thought it was when I tweaked it the first time.

Yoshax:
So... You're nerfing to to be worse than it was before you buffed it...?

Anewbe:
Where was it before, 2?

Yoshax:
Yes.

Anewbe:
Back to 2.

PR #2348, opened 2016-08-19
Maybe Perhaps Last Major Technomancer PR

Neerti:
Adds ability to sort the spells section of the catalog into categories.  The categories available are 'All', 'Offensive', 'Defensive', 'Utility', and 'Support'.  An example is here;
Category example
Removes preset section on catalog as it was unused.
Projectile spells now have a sound when fired.
Haste lasts five seconds instead of three.
Repel Missiles lasts for five minutes instead of two.
All healing spells work five times as fast, healing in five seconds instead of twenty five seconds.  The amount of instability and healing done has been multiplied to remain consistent.
Passwall can now be used on more than just walls, if there's something dense on the same tile.
Force Missile is now 25% cheaper to cast, and has a cooldown of .5 seconds instead of one second, and does 5 more damage.
Beam's damage is increased by 10, and energy cost decreased by 25%.
Lightning's warm-up time is now one second instead of two.
Overload now costs 10% of total energy instead of 15%, and damage scaled with 4% of their current energy reserves, and 5% with the Scepter, instead of 3%/4%.  Additionally, instability per shot is lowered to 12 from 15.
Track now costs 25 points instead of 30 in the catalog, because roundness.
Fire Aura should look more impressive without a Scepter.
Fixes bug where shooting Lightning made you motionless for twenty seconds.

### DIFF
--- a/code/game/gamemodes/technomancer/catalog.dm
+++ b/code/game/gamemodes/technomancer/catalog.dm
@@ -1,8 +1,13 @@
+#define ALL_SPELLS "All"
+#define OFFENSIVE_SPELLS "Offensive"
+#define DEFENSIVE_SPELLS "Defensive"
+#define UTILITY_SPELLS "Utility"
+#define SUPPORT_SPELLS "Support"
+
 var/list/all_technomancer_spells = typesof(/datum/technomancer/spell) - /datum/technomancer/spell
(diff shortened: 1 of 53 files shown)


Anewbe:
It merges the PR or else it gets the hose again...

PsiOmegaDelta:
"Emits a protective shield ***fron*** your hand"

Issue #2146, opened 2016-07-17
Can't deconstruct bodyscanner


EmperorJon:
#### Brief description of the issue
It just spawns a frame and leaves the scanner still there.

#### What you expected to happen
More components than just a frame.

#### What actually happened
Just a frame.

#### Steps to reproduce
Follow deconstruction steps for the bodyscanner in medical.

#### Additional info:
 - **Server Revision**: master - 2016-07-15
84b353785551d4aca6728835bf0c6d8f26ffa9c1

Issue #2049, opened 2016-07-06
Spectrometers don't accept coolant


EmperorJon:
#### Brief description of the issue
Research radiocardon spectrometers don't accept coolant.

#### What you expected to happen
Add coolant option, adds coolant.

#### What actually happened
Add coolant option, states I've added the entire bucket (120u) of coolant, but the bucket is still full and the spectrometer is still empty.
Edited purity to 1 and amount to 120 in VV and it all worked as expected.

#### Steps to reproduce
Use the coolant tanks next to them. Fill bucket, click on spectrometer.

#### Additional info:
 - **Server Revision**: master - 2016-07-06
8c8dbf55fd90d84738a1c51d2370c34a5e842eea 
 - **Anything else you may wish to add** (Location if it's a mapping issue, etc)

Issue #1283, opened 2016-03-29
Mech Rechargers can be placed in crates/lockers


Mark9013100:
#### Brief description of the issue

Mech Rechargers can be placed in crates/lockers.

#### What you expected to happen

For the mech recharger not to be put in the crate/locker.

#### What actually happened

The mech recharger could be put in a crate/locker.

#### Steps to reproduce
* Get a crate/locker.
* Find a mech recharger.
* Open and close the crate/locker ontop of the recharger.
* Realize that life is ruined now.

#### Additional info:
 - **Server Revision**: Found using the "Show Server Revision" verb under the OOC tab.

Server revision: master - 2016-03-27
ddf4cceb963cf24e2f36f502b4c9c3d2eac651d0 

 - **Anything else you may wish to add** (Location if it's a mapping issue, etc)

PR #2218, opened 2016-07-28
Asset Cache Update, NanoUI Overhaul, and tgui

SinTwo:
Ported from Paradise and Baystation and adapted for our code base.

## What does it do?:
Adjusts how assets are sent to the client and should eliminate the interface lag when you first join as well as changing over to a new form of NanoUI that is very similar to tgui.

Also ported in this is tgui with a few UI's accompanying.

## Images:
http://imgur.com/a/7lrB8

### Asset Cache
ParadiseSS13/Paradise#1807
ParadiseSS13/Paradise#2824
Baystation12/Baystation12#12117  -  Newscaster

### NanoUI
ParadiseSS13/Paradise#3013
ParadiseSS13/Paradise#3030
ParadiseSS13/Paradise#3079
ParadiseSS13/Paradise#3121
ParadiseSS13/Paradise#3173
ParadiseSS13/Paradise#3230
ParadiseSS13/Paradise#3280
ParadiseSS13/Paradise#3322
ParadiseSS13/Paradise#3329
ParadiseSS13/Paradise#3363
ParadiseSS13/Paradise#3412
ParadiseSS13/Paradise#4626

### Animated UI
ParadiseSS13/Paradise#4509
ParadiseSS13/Paradise#4602

### tgui
Baystation12/Baystation12#12344
Baystation12/Baystation12#13689
Baystation12/Baystation12#13721

### datum fix
Baystation12/Baystation12#12464
tgui port broke datums in some capacity, above fixed it.

PR #2357, opened 2016-08-23
Frame Cleanup + Machinery Code Cleanup

SinTwo:
Redo of #2218 with needed fixes + cleans up a large amount of code in the machinery folder.

Makes frame code both cleaner and more modular as well as fixing a few problems.

Fixes: #1283
Fixes: #2049
Fixes: #2146


Yoshax:
Does it fix the sleeper issue?

SinTwo:
First thing I fixed with this.

PR #2363, opened 2016-08-27
Hardsuits/Rigs now have suit coolers

Anewbe:
Hardsuits/RIGs now have an inbuilt suit cooling system.

- You click the button on your HUD, it turns on. You click the button again, it turns off.

- Only cools you if you're actually using the RIG, cannot be used with say, a voidsuit.

- Drains directly from the suit's power cell, but is a bit more energy efficient than the normal unit as a tradeoff.

### DIFF
--- a/code/modules/clothing/spacesuits/rig/rig.dm
+++ b/code/modules/clothing/spacesuits/rig/rig.dm
@@ -345,7 +345,8 @@
 			var/obj/mecha/M = H.loc
 			return M.return_temperature()
 		else if(istype(H.loc, /obj/machinery/atmospherics/unary/cryo_cell))
-			return H.loc:air_contents.temperature
+			var/obj/machinery/atmospherics/unary/cryo_cell/cryo = H.loc
+			return cryo.air_contents.temperature
 
 	var/turf/T = get_turf(src)
 	if(istype(T, /turf/space))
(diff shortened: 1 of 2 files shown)


code/modules/clothing/spacesuits/rig/rig.dm
```diff
@@ -302,8 +310,96 @@
...
+	if(cooling_on)
+		turn_cooling_off()
+	else
+		turn_cooling_on()
+
+/obj/item/weapon/rig/proc/turn_cooling_on()
+	if(!cell)
+		return
+	if(cell.charge <= 0)
+		usr << "<span class='notice'>\The [src] has no power!.</span>"
```
Neerti: You can eliminate usage of usr by adding a user argument to both this and ```turn_cooling_off```.  ```toggle_cooling``` already has an argument for user that you can just feed into those two procs.

code/modules/clothing/spacesuits/rig/rig.dm
```diff
@@ -302,8 +310,96 @@
...
+	cooling_on = 0
 
+/obj/item/weapon/rig/proc/get_environment_temperature()
+	if (ishuman(loc))
+		var/mob/living/carbon/human/H = loc
+		if(istype(H.loc, /obj/mecha))
+			var/obj/mecha/M = H.loc
+			return M.return_temperature()
+		else if(istype(H.loc, /obj/machinery/atmospherics/unary/cryo_cell))
+			return H.loc:air_contents.temperature
```
Neerti: Colons should be avoided whenever possible, and I'm fairly sure this will compile just fine without it.  You can add an if statement if you're paranoid as well.

PR #2367, opened 2016-08-28
Examine Cigarette Lore

redstryker:
Adds lore to cigarette packets, cigars, and pipes that can be seen on the 'examine' tab after examining the item.  
Adds some cigarettes to the cigarette dispenser. 

Commented out the fuel of AcmeCo cigarettes.


polaris.dme
```diff
@@ -546,7 +546,7 @@
 #include "code\game\machinery\requests_console.dm"
 #include "code\game\machinery\robot_fabricator.dm"
 #include "code\game\machinery\seed_extractor.dm"
-#include "code\game\machinery\sleeper.dm"
+#include "code\game\machinery\Sleeper.dm"
```
Anewbe: Was this supposed to change?

redstryker: I don't think it was, I'll change it.

PR #2373, opened 2016-08-29
Adds Changeling Magboots

Anewbe:
- Adds magboots to the space suit and armor
- Armblade has suicide text


Yoshax:
since they're magboots will they equip over shoes? and if so, will that not cause problems when toggling it off?

Anewbe:
They do, and not as far as I know. In testing it worked just fine.

Issue #2382, opened 2016-08-29
External Airlock Opaqueness Inconsistency


Hubblenaut:
#### Brief description of the issue
The red external airlocks spawn without opaqueness, which then gets reset as soon as they get closed.
#### What you expected to happen
I expect external airlocks to be either consistently opaque, or consistently non-opaque.
#### What actually happened
Airlocks spawn non-opaque and then become opaque after usage.
#### Steps to reproduce
Look at an external airlock, then open and close it.
#### Additional info:
 - **Server Revision**: 10d48d74209842cb365e15fc6b030fa3ae5be941 
 - **I really preferred the opaque external airlocks anyway**

PR #2383, opened 2016-08-30
Fixes external airlocks

SpadesNeil:
Fixes an inconsistency in external airlock behavior, in that they start without opacity, but when you open and close them, they become opaque.

Also changed the sprite so the little window is actually transparent. This looks much better in my opinion.
doorext

Fixes https://github.com/PolarisSS13/Polaris/issues/2382


Anewbe:
Looks fine to me.

PR #2385, opened 2016-08-30
Allows snow overlay to be shoveled

SpadesNeil:
Working on a snowfield away map and figured this might be useful for hiding things under snow, like tiles to a base, or wires, or certain items.

Could be useful for events, or that snow map that someone's working on.


Yoshax:
May be better with a short do_after like mopping the floor or mining. Is a bit more immersive.

SpadesNeil:
Done.

PR #2390, opened 2016-08-30
Fixes sprite not updating to not be floating when leaving non-gravity

### LINKED ISSUE #657: Your sprite is stuck in the floating animation sometimes (or maybe always) after leaving space.
On two occasions, I have seen an Engineer go out to wire the solars and then come back in and still look like they're floating (without gravity.) Once it was me, and another time it was another player, so it's not a client side issue. Thoughts?

Yoshax:
Fixes #657


SpadesNeil:
\o/

PR #2392, opened 2016-08-30
Changeling stasis fixes

### LINKED ISSUE #860: Regen stasis doesn't replace lost limbs
In a recent ling round, I had my both my legs sliced off, yet I couldn't get new ones from rapid regen nor regen stasis.

Yoshax:
Fixes #860
Fixes another issues where having a missing limb caused regen to runtime due to trying to undislocate a missing limb.


Neerti:
I recall you saying in Discord asking me if it was okay to use the rejuv proc for reviving.  I assumed you were gonna do that, but it seems you haven't.  Did it cause more problems or issues?

Yoshax:
I didn't test it. I found the issue where I went with that solution. This works. It also seems like rejuv wouldn't do some of the stuff what is done already with the changeling revive. It was easier to do it this way.

PR #2395, opened 2016-08-31
Adds vomit as an emote

Yoshax:
Because our players are sensible and trustworthy and need to vomit.

### DIFF
--- a/code/modules/mob/living/carbon/human/emote.dm
+++ b/code/modules/mob/living/carbon/human/emote.dm
@@ -594,11 +594,10 @@
 			return
 
 		if ("help")
-			src << {"blink, blink_r, blush, bow-(none)/mob, burp, choke, chuckle, clap, collapse, cough,
-cry, custom, deathgasp, drool, eyebrow, fastsway/qwag, frown, gasp, giggle, glare-(none)/mob,
-grin, groan, grumble, handshake, hug-(none)/mob, laugh, look-(none)/mob, moan, mumble, nod, pale, point-atom, raise, salute, shake, shiver, shrug,
-sigh, signal-#1-10, smile, sneeze, sniff, snore, stare-(none)/mob, stopsway/swag, sway/wag, swish, tremble, twitch, twitch_s, vomit, whimper,
-wink, yawn"}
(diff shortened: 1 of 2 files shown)


redstryker:
Finally.

Techhead0:
>Because our players [...] need to vomit.

Anewbe:
:confused: Not sure I really want this, but at the same time I've wished for it on occasion.

code/modules/mob/living/carbon/human/emote.dm
```diff
@@ -589,12 +589,15 @@
 		if("swag", "stopsway")
 			src.animate_tail_stop()
 
+		if("vomit")
+			vomit()
+			return
```
Fox-McCloud: This `return` is trailing and unnecessary.

SpadesNeil:
Does this actually produce puke on the floor? Cuz if so, I see this being abused lol.

Yoshax:
It calls vomit, yes. If it's abused then staff can deal with it.

SpadesNeil:
It seems like very bad practice to put something in that entices people to abuse it, then have the staff punish people for using it.

I don't think this is a good addition. Any time something is added that entices abuse because it's funny or grants an advantage should not be resolved by telling the staff to punish people for using it.

SpadesNeil:
Or, actually, how about requiring a certain nutrition level for vomit to even work?

If you don't have enough food in your gut, you can only dry-heave. This also produces a penalty for abusing vomit, which prevents flagrant abuse without having to be deliberately **intending** to abuse it.

Yoshax:
That is already a thing: https://github.com/PolarisSS13/Polaris/blob/b78be20b75e6ab854970caaae86a16fee7f3ae2b/code/modules/mob/living/living.dm#L818

SpadesNeil:
Oh, there it is! Odd, I've never seen this happen before.

Still I suggest a cooldown to forced vomiting. There should be no reason to vomit more than once every five minutes.

Anewbe:
Yeah, a cooldown on the voluntary emote would be nice.

Yoshax:
That adds more unnecessary code and complexity. It works, and is fine as is. Staff should do their job and deal with 'abuse' rather than hardcoding it.

SpadesNeil:
Or there should not be a feature that entices abuse, but that's just my opinion.

PR #2397, opened 2016-08-31
Projectile flash rounds will now do the same as a flash when hitting a target

Yoshax:
As of now there is little to no reason to use them, as they are useless. This is basically lazy copy pasta of flash code.


Anewbe:
Oh shit yeah, wanted this for ages.

Mechoid:
As a note, this likely will affect borgs, though not in normal borg-flash manner.

Yoshax:
`if(ishuman(M))`

Mechoid:
OH wait I'm blind. Alright

PR #2401, opened 2016-08-31
Adds the ability to wear ponchos as an accessory, also adds the ability to attach ponchos to your suit

Yoshax:
Ponchos are an accessory. They can be attached to uniforms and suits. They can also still be worn as a suit. You cannot attach a poncho to a poncho.

You can now attach it to your armor, and wear it over it.

https://i.gyazo.com/e3ea7c21c3732a0a7c13eabb51d72dcc.png


Neerti:
Perhaps it would be better to rename the ```"poncho"``` entry in ```valid_accessory_slots``` to something more generic for something that goes over a suit/armor item?

Yoshax:
Was going to but didn't really have an idea, so just left it ungeneric. There was a lot of genericifying that could be done here, particularly examine code.

I'll rename it though, I suppose.

Yoshax:
Renamed.

Issue #2408, opened 2016-09-01
Elevator does not spawn dirt or metal when moved--just space.


SpadesNeil:
#### Brief description of the issue
Shuttles between Z-levels was replaced with elevators for some applications on the station. There should not be a gaping hole to the void of space when an elevator is moved. Rather, it should be replaced with metal. Namely, /fixed tiles that cannot be damaged or destroyed by players trying to rip them up.

#### What you expected to happen
When an elevator moves to another Z-level, the empty zone should have steel tiles left over, or at least dirt.

#### What actually happened
![capture](https://cloud.githubusercontent.com/assets/6484985/18155072/da99665a-6fd7-11e6-96ad-1b10f79751d8.PNG)

#### Steps to reproduce
1. Move an elevator.
2. Observe the results on the spot you left behind.

#### Additional info:
 - **Server Revision**: https://github.com/PolarisSS13/Polaris/commit/b83750a804dad7e0ce1b92e61955005db11e0898
 - **Anything else you may wish to add**: You'll notice that the map has dirt where the elevator's destinations are. These also get deleted when the elevator moves back again.

Neerti:
This is not a bug, as those are meant to visually be elevator shafts.  You can change this behaviour by changing the ```base_turf``` type on the area.

SpadesNeil:
>This is not a bug, as those are meant to visually be elevator shafts.

That doesn't make any sense. An elevator shaft has to go up or down. You can see down into space in both cases.

>You can change this behaviour by changing the base_turf type on the area.

So it's a mapping issue, then?

PR #2410, opened 2016-09-01
Fixes Elevator terrain behavior.

SpadesNeil:
Because both ends of the elevator shaft shouldn't be bottomless voids into space. This is how it's mapped on Polaris-5.dmm (until shuttle moving punches holes in the ground), so this is how it's probably supposed to be.

Also fixed the cryo shuttle while I was at it, because that should probably also not be a void when the shuttle leaves since it flies out a tunnel. Remind me to fix some CentCom shuttles later as well. Different problem with those, but it's fixed in a similar way.

Fixes https://github.com/PolarisSS13/Polaris/issues/2408


Neerti:
Is the turf supposed to be there on both the "top" and "bottom"?  The areas you changed the base turf to are inherited by both destinations for all the shuttles you did.

SpadesNeil:
That was intentional, yes. If you want to get fancy with it, the "top" part can be altered to be iron sands or steel plates or something but I wasn't sure if that would work correctly as base turf, since this kind of turf can generally be deconstructed.

SpadesNeil:
Soooo is this being merged or..?

PR #2417, opened 2016-09-01
Makes diona slightly less slow

Yoshax:
14 Tiles
With PR - Slowdown 5 - 00:00:11:20
Currently - Slowdown 7 - 00:00:13:52

Needed change. Painful to play Diona due to size of the station. About 20-30% faster in practice. Still sluggish.


Anewbe:
Spook Approved, Sonic Splinters go.

PR #2422, opened 2016-09-02
Flat Cap Changes and Hair Bow

redstryker:
Adds an actual side-facing sprite to the flat cap because before it didn't actually look like a flat cap because of it.

Adds a white variety of flat cap that can be colored on the loadout with the datum.

Adds a hair bow (pretty) that can be colored on the loadout with the datum.

Sprites:
Bow:
http://puu.sh/qXCL2/127e948db2.png
http://puu.sh/qXAx6/4d6b5b853e.png
Flat Cap Side-View:
http://puu.sh/qXAIS/96e248a2e4.png


Anewbe:
Are there in-hand sprites for the bow?

redstryker:
I don't think they are needed since bows are so small.

Anewbe:
Could use the pill in-hand, certain people want them to present as frequently as possible, is why I ask.

PR #2424, opened 2016-09-03
Ninjas leave no trace

Anewbe:
The explosion was silly
This post, a haiku


Yoshax:
That's pretty silly. The description also makes little sense. There's no need to change it, I've seen it create a lot of interesting situations and RP. That does not need replaced with silliness.

Neerti:
The ninja exploding, from what I've seen on here and on other servers, primarily serves as an OOC 'fuck you' button, to kill your opponents or even the medical team trying to save you in a way that the other people can't really defend against without metagaming (like keeping your distance).  Exploding your opponents/saviors also is capable of permakilling them due to the risk of gibbing.

Making the ninja disappear in smoke still accomplishes the goal of making sure the station doesn't get their hands on the RIG, but gets rid of the 'I lost, so fuck you all' option, which is just fine to me.

Yoshax:
You've replaced a remotely interesting thing that generates RP with a cliche kung-fu flick trope. Good job.

nachomeep:
I'd rather just remove the 'thing that happens on death' over removing the bomb.

Issue #1892, opened 2016-06-07
'Random Medicine' item does not turn into a random medicine


martin509:
#### Brief description of the issue
There is an item that is normally meant to spawn a random medicine at roundstart (for instance, kelotane pills), used in medical lockers. It does not spawn the random medicine, instead appearing as a non-pick-up-able item with the icon of gauze titled Random Medicine.
#### What you expected to happen
I expected a random medicine.
#### What actually happened
[http://i.imgur.com/0VUpqaA.png](http://i.imgur.com/0VUpqaA.png)
#### Steps to reproduce
Look in medical lockers at round start.
#### Additional info:
 - **Server Revision**: Found using the "Show Server Revision" verb under the OOC tab.
 - **Anything else you may wish to add** (Location if it's a mapping issue, etc)

Arokha:
Might be from f75d509e7fe8e364d5db5782a3dd1ced183a1dfd

martin509:
Update: the medicine is spawned but the 'random medicine' item is not deleted.

PR #2427, opened 2016-09-03
Medical related fixes

SpadesNeil:
- Fixes https://github.com/PolarisSS13/Polaris/issues/1892 by undoing a change made by https://github.com/PolarisSS13/Polaris/commit/f75d509e7fe8e364d5db5782a3dd1ced183a1dfd


Anewbe:
Pictures for mapping changes would be appreciated.

SpadesNeil:
capture

Anewbe:
Is that rack gone or present?

SpadesNeil:
It's still there! Right next to the APC.

SpadesNeil:
Hang on let me check something before we merge this.

code/game/objects/random/random.dm
```diff
@@ -11,11 +11,9 @@
 	..()
 	if (!prob(spawn_nothing_percentage))
 		spawn_item()
-
-/obj/random/initialize()
```
SpadesNeil: I'm told that apparently initialize() is supposed to be called by the base New() proc, but for some reason that isn't happening with these items. My way works, the old way doesn't work, and I don't know why.

Anewbe:
Had some discussion, was decided that Medical doesn't need a cycler there, seeing as there's already one in EVA.

Yoshax:
Don't need the suits there either, they should be in EVA.

SpadesNeil:
Okay gonna remove the cycler. Can remove the suits later.

PR #2433, opened 2016-09-05
Improves Blood Drain rune

Anewbe:
As of first pass, now heals internal damage. May require number tweaks.
TODO: Add bone repair. Not sure what the best way to do that would be.


Anewbe:
Worked in testing

PR #2434, opened 2016-09-06
HAZMAT Suits

redstryker:
Adds a HAZMAT variety of the engineering voidsuit.  It is a direct child of the suit and can be created by cycling a voidsuit with Engineering or Atmos access.  Emags will also allow you to make your suit into it.

Sprites:
http://puu.sh/r1obu/f1710f4deb.png


SpadesNeil:
Neat as it is, it looks to be merely a reskin.

redstryker:
That's... what it is?  I'm not denying that.

redstryker:
Reskins aren't always bad when they provide diversity and options to players.  This is only the first PR of many that will add more coloring options to voidsuits.  There will also be some 'antag voidsuits' that can only be obtained with an emagged cycler.

SpadesNeil:
Oh ok then! Neat. Carry on.

PR #2438, opened 2016-09-07
Tweaks skirt pathing

Anewbe:
Makes it so the short skirts no longer cover torso and arms. The skirts that actually cover the torso and arms still do so.


html/changelogs/Anewbe - Skirts.yml
```diff
@@ -0,0 +1,36 @@
...
+#   rscdel (general deleting of nice things)
+#   imageadd
+#   imagedel
+#   maptweak
+#   spellcheck (typo fixes)
+#   experiment
+#################################
+
+# Your name.  
+author: N3X15
```
PsiOmegaDelta: Author?

Anewbe: Ah, good catch, thanks.

PR #2442, opened 2016-09-08
Construction Voidsuits

redstryker:
Adds a 'construction' variety of the engineering voidsuit. It is a direct child of the suit and can be created by cycling a voidsuit with Engineering or Atmos access. Emags will also allow you to make your suit into it.

Was trying to go with the same style as the hazard jacket with this one.

Sprites:
http://puu.sh/r3vNJ/56270c0323.png


Yoshax:
The yellow stripes are incredibly bland... It's just a block of one colour, it doesn't look very good.

PR #2443, opened 2016-09-09
makes people bleed gud

Yoshax:
people bleed quick quick faster x2 spood


Hubblenaut:
Blood for the blood god.

PR #2445, opened 2016-09-09
Biohazard Voidsuit

redstryker:
Adds a Biohazard variety of the Medical voidsuit. It is a direct child of the suit and can be created by cycling a voidsuit with Medical access. Emags will also allow you to make your suit into it.

Sprites:
http://puu.sh/r5zFh/e1a3cb45db.png


nachomeep:
What's the point of this, exactly?

Just asking- what situation would you use this in?

redstryker:
It's a reskin, so there is really no mechanical difference with it.  The point is to add different variations of the voidsuits so that players have something else to use besides the current voidsuits.  I suppose it could be used for like.. a blob or something, but since there's no real mechanical difference, *shrug*.

nachomeep:
This also brings up minor concerns that I have- heisters start with suits of certain departments. If we were to add variations to, say, the mining suit, it'd be harder for them to fake being a crewmember- something I actually see done very often.

redstryker:
I do plan to have a variation for the Mining suit.  I have plans for variations for all of the voidsuits.

nachomeep:
So, will heisters be given a suit cycler and access to these variations to make up for it?

redstryker:
I don't know much about heist, TBH.  If they currently have a suit cycler?  I suppose so.  But if they don't, who knows?  Mercs will definitely have access to all of these variations, however.

Anewbe:
Heisters should already have a suit cycler, if memory serves.

PR #2452, opened 2016-09-12
Water is now more useful on burning people

Anewbe:
It should now be twice as effective at dousing people who have been lit on fire.


Neerti:
Closed/opened to wake Travis.

Yoshax:
Why are we further nerfing fire after there was complaints for the longest time that it was useless?

Anewbe:
It still hurts a lot, but it doesn't basically ignore fire extinguishers anymore.

PR #2454, opened 2016-09-13
Splints are no longer reusable

Anewbe:
Removes a minor exploit in which it was possible to convert ghetto splints into medical splints, simply by removing the splint.


Techhead0:
On Bay, you get back whatever splint you used in the first place.
See: https://github.com/Baystation12/Baystation12/pull/13481

PR #2455, opened 2016-09-13
Station Naming Inconsistency Fixes

SpadesNeil:
The naming of things has been really inconsistent across the game's files. This has always annoyed the shit out of me.

CentComm and Centcomm and Centcom are now all CentCom, specifically with that capitalization. Why one M instead of two M's? Because Comm with two 'M's = Communications. Hence, Telecomms, NOT Telecoms. Telecoms is incorrect. CentCom was also chosen because CentCom with one M and this
casing is most found throughout the game's files. Yes, I know the word Command has to M's too, but all real life instances of Central Command say CentCom, with one M. Take the U.S. Central Command, for instance. http://www.centcom.mil/ Do you know what happens when you put CentComm into Google? It says, "Showing results for CentCom, ~~you idiot~~."

Speaking of Telecomms, I corrected this inconsistency too. Most examples use Telecom**m**. The only file where I deliberately did not change this inconsistency was **camera/presets.dm**, because it never appears to the player, and changing it will fuck up maps. I can change it anyway if we want that consistent too and just fix the maps.

Likewise, Nanotrasen has been changed to NanoTrasen. Nanotrasen only appears 20 times, where NanoTrasen appears 62. NanoTrasen is clearly the preferred, correct naming.

**So to recap:**
Telecomm, not Telecom
CentCom, not Centcom, not Centcomm, and _certainly_ not CentComm.
NanoTrasen, not Nanotrasen.


Anewbe:
Did you test this?

SpadesNeil:
The code compiles and the code runs if that's what you mean. I don't know what else there is to test. Every proc where CentCom is mentioned is renamed appropriately and automatically. I've already done this change in the past on old Vorestation code and there were never issues regarding it.

https://github.com/SpadesNeil/VOREstation/issues

SpadesNeil:
Can confirm; still compiles just fine. I don't know what else to test. There should be nothing wrong with this based on what I changed. If there is, feel free to scold me.

SpadesNeil:
Tested the shit out of this just to try to find something worth mentioning that might be broken. Still nothing. I don't know what more needs testing. Shuttle controller works, fax doesn't have any problems, telecomms is fine, what exactly do you want me to test? It's just a name change.

PR #2456, opened 2016-09-13
Custom Circuitry Machines

Neerti:
Gif: https://puu.sh/r4q4L/82812c085a.gif
This adds a new activity for science and maybe engineering, in which they can build custom machines utilizing components called circuits, that are linked with other circuits to exchange data and potentially interact with the outside world, being able to use logic and data to intelligently affect the outside world, if the machinist is good enough.

## How Does It Work?
A 'circuit' is a self-contained piece designed to do a specific task, like adding numbers from its inputs and outputting the result, or a ticker outputting a pulse signal every few seconds.  These machines are fairly useless on their own, but when wired together, wonderful machines can be made.

Circuits possess 'pins', which both hold 'data', and transfer this data to other pins on other circuits.  There are three classes of pins, however two of them are exactly the same, where as the third class is special.  The first and second are input and output.  Both hold data for the circuit, where as the third is called an 'activator' pin, which can send or receive 'pulse signals'.  Pins can be wired to an arbitrary amount of other pins, so long as all the pins are contained inside the same machine.  A circuit can also be wired into itself, such as an output pin going into the same circuit's input pin.

When a circuit receives a pulse signal, it does what it was built to do, for example adding all the inputs and pushing the outputted data down the line to the next circuit.  Some logic circuits have a second activator pin which sends a pulse out when the evaluation it does is true.  The careful management of activators is how machine creators manage what gets done when.

Input pins are generally on the left side of the interface, output is on the right, and activator pins are on the bottom side, and are colored red.

Input and output pins hold 'data', which can be a string (text), a number, a reference to something outside the system (like a person, or a wall), or nothing (null).  A lot of circuits require that data be in a specific type, however if it is supplied the wrong type, nothing really bad happens, except that it won't give a result.

Pins are wired together by using a special 'wirer' tool in your hand and using an interface, shown below.  You attach and detach wires by clicking two pins to link or unlink them.  The input pins are always on the left, the output on the right, and the activator pins on the bottom.

Interface
The activator pins are always colored in red, and they are essentially how the component knows when to do something.  There's various ways to get a pulse to hit them, from things like pressing buttons, putting data in a numberpad, evaluating a true result in a logic gate, or using a ticker to pulse every few seconds.

## Is This Safe?
Some may feel that this is bordering on NTSL levels of insecurity, but that's not a valid comparison.  This is not a programming language, and the component-nature of this means, plus having only three types of variables to use between all of the components (text, numbers, and refs) means that the worst that can happen is a runtime if I messed up somewhere, and not crashing the server like with NTSL.  Besides, all the components and sometimes BYOND itself will make sure that the input it gets actually makes sense.  Shoving text into an addition circuit will just result in it ignoring the text, and not a runtime.

## What Can This Do?
Well, ideally when I add the rest of the 'manipulation' components I intend to, the sky would be the limit.  Right now there's a lot of things you can do with logic and numbers, however they're not too useful, but I think it's interesting in it's own right, and releasing this now would let people get to learn the somewhat complicated system right away.


Anewbe:
I regret neverything.

PR #2457, opened 2016-09-13
Press Vest

redstryker:
Adds Press Vest
Sprites:
http://puu.sh/ramiR/97c9c199e5.png


code/modules/clothing/suits/armor.dm
```diff
@@ -323,6 +323,12 @@
 	icon_badge = "detectivevest_badge"
 	icon_nobadge = "detectivevest_nobadge"
 
+/obj/item/clothing/suit/storage/vest/press
+	name = "press vest"
+	desc = "A simple kevlar plate carrier. This one has the word 'Press' embroidered on patches on the back and front."
+	item_state_slots = list(slot_r_hand_str = "armor", slot_l_hand_str = "armor")
+	allowed = list(/obj/item/ammo_magazine,/obj/item/ammo_casing,/obj/item/device/flashlight/maglight,/obj/item/weapon/pen,/obj/item/device/camera_film,/obj/item/device/camera)
```
Anewbe: So, as a Press Vest, I don't see the need to have this able to carry ammo mags or casings, seeing as it's not for combat.
Also, could be wrong, but I _think_ you can just include flashlight on that list, so they can tote around the blue ones instead.
Third thing: I know they don't work correctly right now, but the Universal Recorder might be a good thing to add, for when those're fixed.

Yoshax:
How much different can it be from a normal vest that it can't hold the same things as it?

redstryker:
Considering that I based it off this:
http://images.vestguard.co.uk/1/1382345362_87.jpg
I am considering removing storage on it entirely, since there are no pockets on this.

PR #2460, opened 2016-09-14
Robot arms/legs/extremities no longer increase bodytemperature

Anewbe:
In response to a weird bug that was cooking Taj with prosthetics.
Body temp in FBPs is very nearly the same as before, but people with a single robo-part will no longer have body temps in the 300K range.

### DIFF
--- a/html/changelogs/Anewbe - Robobody.yml
+++ b/html/changelogs/Anewbe - Robobody.yml
@@ -22,7 +22,7 @@
 #################################
 
 # Your name.  
-author: N3X15
+author: Anewbe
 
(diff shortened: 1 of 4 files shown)


html/changelogs/Anewbe - Robobody.yml
```diff
@@ -0,0 +1,36 @@
...
+#   rscdel (general deleting of nice things)
+#   imageadd
+#   imagedel
+#   maptweak
+#   spellcheck (typo fixes)
+#   experiment
+#################################
+
+# Your name.  
+author: N3X15
```
Neerti: I should find a way to get Travis to check for this.

PR #2461, opened 2016-09-15
Fixes some Advanced Egun oversights

Anewbe:
It's bigger than a pistol, so it should take up more space. Also fixed an oversight in which the Advanced Egun and Burst Laser were firing more slowly than they should have been, due to inheritance.

### DIFF
--- a/code/modules/projectiles/guns/energy/laser.dm
+++ b/code/modules/projectiles/guns/energy/laser.dm
@@ -6,7 +6,7 @@
 	item_state = "laser"
 	fire_sound = 'sound/weapons/Laser.ogg'
 	slot_flags = SLOT_BELT|SLOT_BACK
-	w_class = 3
+	w_class = 4
 	force = 10
 	origin_tech = list(TECH_COMBAT = 3, TECH_MAGNET = 2)
 	matter = list(DEFAULT_WALL_MATERIAL = 2000)
@@ -71,6 +71,7 @@ obj/item/weapon/gun/energy/retro
 	projectile_type = /obj/item/projectile/beam/heavylaser/cannon
 	max_shots = 1
 	fire_delay = 30
+	w_class = 4
 //	requires_two_hands = 1
 	one_handed_penalty = 6 // The thing's heavy and huge.
 	accuracy = 3
--- a/code/modules/projectiles/guns/energy/nuclear.dm
+++ b/code/modules/projectiles/guns/energy/nuclear.dm
@@ -29,6 +29,7 @@
 	item_state = null	//so the human update icon uses the icon_state instead.
 	fire_sound = 'sound/weapons/Taser.ogg'
 	max_shots = 21	//7 trigger pulls
+	fire_delay = 6	//This one's not a handgun, it should have the same fire delay as everything else
 
 	projectile_type = /obj/item/projectile/beam/stun/weak
 	origin_tech = list(TECH_COMBAT = 4, TECH_MAGNET = 2, TECH_ILLEGAL = 3)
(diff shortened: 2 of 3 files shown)


Yoshax:
That's not a bugfix. That's a nerf. You're nerfing it into oblivion.

code/modules/projectiles/guns/energy/nuclear.dm
```diff
@@ -51,6 +52,8 @@
 	origin_tech = list(TECH_COMBAT = 3, TECH_MATERIAL = 5, TECH_POWER = 3)
 	slot_flags = SLOT_BELT
 	force = 8 //looks heavier than a pistol
```
SpadesNeil: It does appear to be heavier than a pistol according to this comment--but it's still a pistol. Just a fat pistol.

code/modules/projectiles/guns/energy/nuclear.dm
```diff
@@ -51,6 +52,8 @@
 	origin_tech = list(TECH_COMBAT = 3, TECH_MATERIAL = 5, TECH_POWER = 3)
 	slot_flags = SLOT_BELT
 	force = 8 //looks heavier than a pistol
+	w_class = 4	//Looks bigger than a pistol, too.
```
SpadesNeil: But this makes it as heavy as a shotgun. This can't be correct. https://github.com/PolarisSS13/Polaris/blob/fa74226e9901e080d762ed01dc5e6b7756cc27f1/code/modules/projectiles/guns/projectile/shotgun.dm#L7

code/modules/projectiles/guns/energy/nuclear.dm
```diff
@@ -51,6 +52,8 @@
 	origin_tech = list(TECH_COMBAT = 3, TECH_MATERIAL = 5, TECH_POWER = 3)
 	slot_flags = SLOT_BELT
 	force = 8 //looks heavier than a pistol
+	w_class = 4	//Looks bigger than a pistol, too.
+	fire_delay = 6	//This one's not a handgun, it should have the same fire delay as everything else
```
SpadesNeil: Normally this would re-balance it in exchange for its weight change, but given its infinite ammo, this seems like a poor decision.

Yoshax:
Jesus Christ. These are horrifying changes. What is this actually meant to achieve? What is the goal? Other than changing random numbers with no semblance of sense?

code/modules/projectiles/guns/energy/laser.dm
```diff
@@ -71,6 +71,7 @@ obj/item/weapon/gun/energy/retro
 	projectile_type = /obj/item/projectile/beam/heavylaser/cannon
 	max_shots = 1
 	fire_delay = 30
+	w_class = 4
```
SpadesNeil: This I am actually completely okay with. The heavy laser is SUPPOSED to be big!

code/modules/projectiles/guns/energy/laser.dm
```diff
@@ -6,7 +6,7 @@
 	item_state = "laser"
 	fire_sound = 'sound/weapons/Laser.ogg'
 	slot_flags = SLOT_BELT|SLOT_BACK
-	w_class = 3
+	w_class = 4
```
SpadesNeil: This, not so much. This shouldn't be changed.

Neerti: w_class 4 can still fit in a backpack.

SpadesNeil: Are you sure? Because I'm pretty sure shotguns can't, and they're w_class 4.

Neerti: w_class 4 can fit in bags, and has been able to since the inventory rework that occurred awhile ago.  w_class 5 can't fit in bags, which the LWAP/PTR-7 has.

SpadesNeil: Oh. Weird. I guess I haven't tried shoving a shotgun into a bag in some time, then. Still, I don't think the advanced E-gun is as bulky as a shotgun.

PR #2463, opened 2016-09-15
Health analyzers detect appendicitis

Anewbe:


nachomeep:
Wouldn't it be Inflammation (two Ms), or is that a regional spelling?

Anewbe:
You are correct, it should be inflammation. I'll fix it when I get back to my computer.

PR #2464, opened 2016-09-15
Custom Circuit Additions & Fixes

Neerti:
Adds Med Scanner and Advanced Med Scanner circuits.
Adds Local Locator circuit, which is vital for the above to analyze a specific person.  It stores a ref of whoever is holding the device when activated, if any.  More locators will be made in the future, such as the planned Adjacent Locator and Visible Locator.
Clock circuit is now functional.
Five second delay will delay for five seconds and not one second.
Renaming circuitry now respects ability to do so (e.g. being alive).
Firing circuit can now be obtained by science normally.
Examining inside assembly cases now requires adjacency.

-Additions/Edits-
You can now cancel inputting on a text or numberpad.
Zero is now displayed properly on the UI and not be shown as null.
The small and medium assemblies now have real sprites.
Total health percentage for med scanners and advanced med scanners now reports a proper percentage and not 1 or 0.
The constant memory chip now can be set by using it in hand, making it usable.
The constant memory chip now accepts refs as data to store.  To use, select 'ref' when using inhand, then hit it against the thing you want to store.


code/modules/integrated_electronics/input_output.dm
```diff
@@ -37,7 +42,7 @@
 
 /obj/item/integrated_circuit/input/numberpad/ask_for_input(mob/user)
 	var/new_input = input(user, "Enter a number, please.","Number pad") as num
```
PsiOmegaDelta: Mostly curious why one isn't allowed to cancel out of picking the number.

Neerti: Oversight that should be fixed now.

PR #2467, opened 2016-09-15
Lasercannon tweaks

Anewbe:
Lasercannon now has 4 shots and 20 fire_delay, but its shots do 80 burn with 50 armor pen.


SpadesNeil:
Good, I didn't like how the laser cannon became a glorified energy AT weapon that wasn't even worth it. This is how it should be.

PR #2468, opened 2016-09-15
Hawaiian shirt

Anewbe:
Ports most of Baystation #14089
Couldn't figure out the colors, but most of the code's still in there, in case some brave future soul decides to play with it.


SpadesNeil:
I could try screwing with this later. I know there's already code in the game that does something similar.

PR #2470, opened 2016-09-15
Bloodloss tweak

Anewbe:
A bit too fast right now, hopefully this will help.

### DIFF
--- a/code/modules/organs/blood.dm
+++ b/code/modules/organs/blood.dm
@@ -129,7 +129,7 @@ var/const/BLOOD_VOLUME_SURVIVE = 40
 			if(!(temp.status & ORGAN_BLEEDING) || (temp.robotic >= ORGAN_ROBOT))
 				continue
 			for(var/datum/wound/W in temp.wounds) if(W.bleeding())
-				blood_max += W.damage / 20
+				blood_max += W.damage / 30
 			if (temp.open)
 				blood_max += 2  //Yer stomach is cut open
 		drip(blood_max)


Yoshax:
Too fast how? Provide examples?

Anewbe:
It's about 3 minutes from a single revolver hit, two minutes for two shots, three if you're stabbed four times, and ~seven if you've got a knife stuck in you. Woodrat tested with a few things and they all feel very rapid.

SpadesNeil:
Yeeeaa I wasn't fond of the change to blood loss recently. This is better.

Yoshax:
Because bloodloss is fast? 3 minutes is more than enough time to recieve medical treatment. It'either fast, or it's not a threat at all.

alabasterTempest:
It's a bit too fast, as it is. This is fiddling with the balance of it. There may or may not be future adjustments to it, depending on how it plays after this.

Yoshax:
Everyone I spoke to, and provided examples with math thought it was a good, positive change.

alabasterTempest:
At the time, we thought that. This is a balancing fix, and is due to change as we figure stuff out. We're not nerfing it into the ground, we're not undoing your work, we're trying to find a happy medium that works for most players.

Neerti:
Several controlled tests were done with the rate on the server right now, with various weapons from revolvers to pistols to knife tosses, and then some uncontrolled fighting occurred by antags, and people bled out really fast, as in people dying in less than two minutes, which is a bit much in my opinion.  Most people won't carry bandages to stop the bleeding, and our map is quite big, meaning it takes a lot time for medical to get to the bleeding person.

PR #2481, opened 2016-09-17
CE RIG gits gud.

Anewbe:
Adds magboots and insulated gauntlets to the CE_RIG.
Mapped out the now-redundant magboots from the CE Office.


Yoshax:
Diff is off.

PR #2482, opened 2016-09-17
Powersink fix

Anewbe:
Never reached the 5 billion units needed to boom, this should hopefully fix its explosion, without negating its use in disrupting the power. If needed, we can tweak it some more.

### DIFF
--- a/html/changelogs/Anewbe - Powersink.yml
+++ b/html/changelogs/Anewbe - Powersink.yml
@@ -1,0 +1,36 @@
+################################
+# Example Changelog File
+#
+# Note: This file, and files beginning with ".", and files that don't end in ".yml" will not be read. If you change this file, you will look really dumb.
+#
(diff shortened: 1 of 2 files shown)


SpadesNeil:
I never cared for the kaboom but I seem to be the only one who feels that way.

Still, good that it works now. The power sink is a very useful traitor item when it works properly, whether it kabooms or not. The kaboom is just icing on the syndicake.

Yoshax:
Does it actually explode now? And how long until it does?

Anewbe:
Does explode, takes just under 30 minutes.

html/changelogs/Anewbe - Powersink.yml
```diff
@@ -0,0 +1,36 @@
...
+#   rscdel (general deleting of nice things)
+#   imageadd
+#   imagedel
+#   maptweak
+#   spellcheck (typo fixes)
+#   experiment
+#################################
+
+# Your name.  
+author: N3X15
```
Hubblenaut: Top

PR #2484, opened 2016-09-17
Allows robots to be constructed with prosthetic limbs as well as borg…

MagmaRam:
… limbs.

Does exactly what it says on the tin. We've had enough people confused about why they can't build robots, let's just cut that knot.


Anewbe:
Tested?

nachomeep:
Yes. Thank you.

MagmaRam:
It's tested.

PR #2489, opened 2016-09-18
Lasers, energy projectiles, and muzzle flashes now produce light.

MagmaRam:
It's a bit wonky due to lighting only updating once every half-second,
but it's very much functional, at least on my desktop.

I swear to God, it will work this time.

### DIFF
--- a/code/modules/projectiles/effects.dm
+++ b/code/modules/projectiles/effects.dm
@@ -11,7 +11,8 @@
 	if(istype(M))
 		transform = M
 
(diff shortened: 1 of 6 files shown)


Yoshax:
You could have saved a lot of lines by just defining the stuff at a higher level rather then duplicating it 3+ times,

Furthermore, why is there balance changes stuffed in here?

MagmaRam:
Balance changes are stuck in there because mistakes were made.

PR #2494, opened 2016-09-18
Heavy armor values reworked for FUN™

SpadesNeil:
Just to clarify, **I'm not talking about mercenary heavy armor vest.**
I'm talking about **/obj/item/clothing/suit/armor/heavy** which currently appears nowhere in the game, at all.

At one time, this armor had no values at all, because it was never used. Someone who noticed the thunderdome armors and this armor also had no values decided to copy the values from the merc heavy armor over to this--but this armor also has a slowdown, so why would you ever use it? Answer: You wouldn't.

So I made this heavy armor suit actually serve a purpose again--albeit imbalanced, and _that's okay_.

Well, okay, not totally imbalanced. I used to have this available in old Gateway missions on Vorestation. It was popular for fighting mobs but useless when you got back to the station. Here's why:

```
/obj/item/clothing/suit/armor/heavy
 name = "heavy armor"
 desc = "An old military-grade suit of armor. Incredibly robust against brute force damage! However, it offers little protection from energy-based weapons, which, combined with its bulk, makes it woefully obsolete."
 armor = list(melee = 90, bullet = 80, laser = 10, energy = 10, bomb = 80, bio = 0, rad = 0)
 w_class = 5 // massively bulky item
 slowdown = 5 // If you're a tank you're gonna move like a tank.
```

As you can see, this is a very specialized armor.

Considering this armor can only be admin spawned, I say "Why not?" If used, it'd be interesting to see players adapt to fighting someone in this armor. Again, it's basically useless against energy weapons.


SpadesNeil:
Well that didn't happen right.

PR #2501, opened 2016-09-19
Technomancer Tweaks

Neerti:
Cost for various spells and equipment adjusted greatly.  In general, things are cheaper, and everything should now be in multiples of 25, so no points are wasted.
The default jumpsuit for Technomancers is now heavily insulated, protecting them from tasers, batons, and perhaps unfortunate lightning strikes.
Instability between 31 and 50 made less harsh.
Wards made with a scepter of enhancement will break the cloak of anyone invisible that it can see.
Fire aura buffed, increased rate of heating as well as temperature cap for both non-scepter and scepter effects.
Reflect spell made more forgiving for the Technomancer, lasting longer before expiring.
Projectile spells should be able to hit people more reliably.

### DIFF
--- a/code/game/gamemodes/technomancer/clothing.dm
+++ b/code/game/gamemodes/technomancer/clothing.dm
@@ -9,27 +9,30 @@
 
 /obj/item/clothing/under/technomancer
 	name = "initiate's jumpsuit"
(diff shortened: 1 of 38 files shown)


Anewbe:
:+1:

PR #2505, opened 2016-09-20
Sports

SinTwo:
Adds baseball mechanics and football.

baseball3


redstryker:
One question: can you hit back heads and grenades with the bat?

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -315,6 +324,11 @@ emp_act
 			var/shield_check = check_shields(throw_damage, O, thrower, zone, "[O]")
 			if(shield_check == PROJECTILE_FORCE_MISS)
 				zone = null
+			else if(shield_check == PROJECTILE_REFLECT)
+				//var/obj/item/weapon/B = O
+				var/obj/item/weapon/ball/B = O
```
Neerti: It would probably be best if reflected() was made a atom/movable proc instead of being specific to balls, so that PROJECTILE_REFLECT can be used on those and it won't break if for some reason it gets fed that with something that's not a ball.

Mark9013100:
[https://www.youtube.com/watch?v=ITL4NqOKmzk&feature=youtu.be&t=2m3s](url)

code/modules/sports/baseball.dm
```diff
@@ -0,0 +1,97 @@
...
+	name = "baseball"
+	desc = "A ball."
+	icon = 'icons/obj/sports.dmi'
+	icon_state = "baseball"
+	w_class = 2
+
+/*
+	Reflection on being hit
+*/
+/obj/item/weapon/reflected(mob/user)
```
PsiOmegaDelta: Any particular reason the code below isn't defined in the base ```/atom/movable/proc/reflected(mob/user)``` proc and that ```mob/user``` isn't any general ```/atom/a```?

SinTwo: It was intended to keep projectiles from being reflected so you couldn't use a baseball bat as a shield against a gun. The second part, no reason, just didn't think of it.

SinTwo: Edited with suggestions.

Anewbe:
Just merged a thing that turns w_class into a defines thing, it'd be awesome if this conformed. Sorry 'bout that.

PR #2476, opened 2016-09-16
Ported ventcrawling from vg.

Yoshax:
Baystation12/Baystation12#14022

Works fine in testing :ok_hand:

PR #2507, opened 2016-09-20
Revert "Ported ventcrawling from vg."

Neerti:
Reverts PolarisSS13/Polaris#2476
An extremely damaging bug was discovered, which is caused when someone (human, other mobs unsure) opens a nanoUI interface associated with atmospherics (pumps, pressure regulators).  When doing so, their camera gets screwy, and the game allows them to noclip through any atom, density be damned.
@Yoshax has been informed, but I'm reverting this immediately since it's a very damaging bug, which is also easily found by anyone who plays atmospherics for five minutes.


Bedshaped:
Hey there. Zuhayr fixed this bug on Bay12. I think it was due to `relaymove()` not just running off mob.loc.

https://github.com/Baystation12/Baystation12/pull/14054/commits/51ddc0122d8fc912845107e6add9debc5c9ef55d

PR #2508, opened 2016-09-20
Re-adds a nice, useful macro

Yoshax:
Courtesy of Psi, was added in the reverted Ventcrawling PR, and I prefer it to the normal method.


PsiOmegaDelta:
Well, I'm unsure of it being either nice or useful but if someone one day wants to port goonchat this simplifies that.

PR #2514, opened 2016-09-21
Gun rebalancing

Yoshax:
Gun sizes have been readjusted and balanced across the board. There are now 4 distinct tiers.

You have:
w_class = 2, these are small firearms like the holdout pistols and the derringer, they can fit into your pocket. On this note, the holdout pistol no longer becomes w_class = 3 with a silencer attached.

w_class = 3, this is the most awkward tier, this holds all weapons from pistols, to SMGs, to carbines. Why? Because pistols should be smaller, but making them w_class = 2 causes more issues than it solves. Carbines have been shifted into this class because they are small, and lightweight according to their description and definition.

w_class = 4, this holds rifles, both bolt-action and automatic, along with shotguns and includes some other weapons like the Laser Cannon, which remains in w_class = 4 instead of being bumped up due to balance concerns.

w_class = 5, this holds the largest weapons including pneumatic cannons, crossbows, heavy machine guns, and suchlike.

Furthermore, other changes have been made. Most bullets now have a small amount of armor penetration, small calibre pistol rounds have gotten 5 armor penetration, strong rounds have 15. Laser projectiles have also gotten the same treatment, on that note, emitter beams have been given armor penetration and the penetration on the LWAP has been increased. Also on the note of bullets, many larger calibres and such have been given the penetrating var which is what the existing large calibre rifle rounds have that allow them to punch through their target and objects.

In addition, the shotgun rounds provided in the antag uplink have been adjusted slightly, no longer do you buy a large magazine of rounds that does not fit into your firearm, instead you get a box of rounds, this however, means that you get slightly less rounds. Instead of getting 24 rounds, you now get 16. It has been made 16 due to it being enough to load a combat shotgun twice. Also, the normal boxes of shotgun rounds found in the armory and suchlike have been bumped up just a little to provide 8 rounds instead of 7, because shotguns hold 8.

There is quite a bit off diff due to moving some pieces of code into more sensible places, however, most of the moved things have not been actually modified.


redstryker:
🔥💯👌

SpadesNeil:
I know you might think the w_class 3 section might be screwy but I think it's beautiful. :+1:

code/modules/projectiles/guns/energy/pulse.dm
```diff
@@ -10,6 +10,7 @@
 	charge_cost=100
 	max_shots = 20 // This is cut in half by "DESTROY" mode.
 	sel_mode = 2
+	w_class = 4
```
SpadesNeil: Can the pulse rifle still melt walls? If so, I'd go so far as to make this w_class = 5 but that is just my opinion. You only get half the shots with the weapon set to "DESTROY" anyway so this doesn't matter a lot.

Yoshax:
It's fine for the carbines and SMGs, the issue is having the pistols in the same category, because pistol aren't really the same size as carbines, or such, but they aren't the same size as w_class = 2 items, they're more like 2.5, but that's not a thing.

code/modules/projectiles/guns/energy/pulse.dm
Yoshax: I believe so? All weapons can melt walls. I'm not sure if the pulse rifle has any special code for it. I just bumped it up to 4 because it's a rifle like the other rifles. I don't really see any need to make it 5 and balance it, it's not really a commonly usable thing. The people who do have it, probably should just have it as 4 due to the circumstances.

SpadesNeil:
Well pistols can be holstered by using the holster command.

Carbines and the like cannot be holstered.

Although honestly I don't see an issue bumping regular pistols down to w_class of 2. The derringer for example is sprited so that when you have it in your hand, it doesn't show up. It's concealed. Its item_state = "concealed" which is a blank icon. The other holdout pistol can have the same thing done to it if it hasn't already.

Yoshax:
The issue with bumping all pistols down to 2 is it disturbs the balance of the holdout and the derringer, which are pistols that can fit into your pocket, among other things. I think it works as proposed here, and see how it goes. If all is well, then maybe we can try it with pistols at two, or at least, some of them. Or refluff/remodify the pistols, or even add more of varying sizes.

SpadesNeil:
Fair enough!

Anewbe:
Here’s things I’d add, no questions asked:
The shotgun ammo box thing is an excellent change, and the accompanying cleanup of the uplink is also a huge plus.
Cleanup on the lasers and their passthrough is appreciated, not really needed but it does make the code look better.
Moving the various ‘special’ energy weapons into their own files feels like a good change, couldn’t see any changes, so if you tweaked any of those please say so.
 
Here’s some minor oversights that I noticed:
You assigned some values to w_class or max_storage space in launcher/pneumatic.dm and projectile/shotgun.dm, those should be the constants now. Along that same line of thought, the lasercannon comment references w_class 5 and 4 as numbers, those should also be the constants, if said comment remains.
Not sure why launcher/alien.dm changed, doesn’t look like anything really changed in there.
 
These changes aren’t something I’m comfortable with adding, overall:
Adding penetrating to the shotgun slugs and revolver bullets isn’t something we want.
Armor pen on bullets doesn’t need to be upped, in my opinion. They’re already pretty good at screwing people up, and making armor even less useful isn’t a good thing. When we add in damage_soak on armor, these will probably be revisited.
I like that you’ve upped the pen on the LWAP (to 30, from 10) and pulse (to 50, from 30), and the emitter beam probably should be screwing up the poor saps who stand in front of it. The weaklaser, however, is only used by the laser carbine’s suppressive mode, and it’s specifically for hunting things that have no armor, namely hostile mobs. I don’t think it should have any pen, period, because the normal mode is at 10.
Some things were resized, mostly weapons. The changes to the Pulse weapons, Rocket Launcher, and SAW are fine in my books. I thought someone else was planning to change the size of the Silencer anyway, so you saved them a step. I’m not a fan of shrinking the Laser Carbine, Burst Laser, Advanced EGun, and Assault Carbine. The carbine’s at least as big as a shotgun, the Burst Laser is obscenely powerful right now, the AEGun recharges, and the assault carbine is mathematically the most dangerous and unbalanced weapon we have, which is why it was taken off of the antag uplink.
There’s also a decent number of weapon slot flag changes. Most of them seem good to me, but the Burst Laser does not need to holstered, and I don’t understand why you felt it necessary to remove the Back slot from the C20-r.

Yoshax:
None of the moved files or portions of code were changed.
`alien.dm` was changed because its path was changed from `guns/alien.dm` to `guns/launcher/alien.dm` due to it being more like a crossbow and launchers, and individual guns or such don't belong in `guns`.

I don't see why penetration isn't wanted, the ones it has been added to are powerful rounds, deserving of it. All rifle rounds already have it.

5 armor penetration is almost negligible, and for more flavor and oomph behind the bullets/projectiles. A round with 5 armor penetration against a target with 30 bullet armor (normal vest) will only do around 0.75 more damage per shot, or an increase of 7.1%. The only time it remotely matters is when you commit 9 shots, where the current values end with around 94.5 damage, and the ones with penetration do 101.25, of course, these values can be higher or lower given the variance armor does. They're hunks of metal flying through the air at 825+ft/s.

The changes to the sizes of the weapons are insisted upon, as it makes the whole thing more sensible. It specifically states in a number of their descriptions, and by definition of a carbine, that they were compact and lightweight weapons. They are not the same size as a shotgun or other rifle, and shouldn't be represented as such. As far as I am aware neither the Burst Laser nor the Assault Carbine are actually obtainable, and if there is such problems with their balance then that should be changed. The laser carbine is out of balance being a size larger than the energy gun, as they are both very similar weapons. With the carbine being one size larger there is very little reason or motivation to use one when an energy gun is available, due to it also having a stun option, and being holsterable. The AEG is quite clearly pistol sized, or similar, and is not larger than any of the SMGs we have, it simply cannot be. It recharges, yes, but is also difficult to obtain and only really obtainable through science. It is a prototype weapon. It recharging itself does not make it that powerful, it is a nice add-on, yes, but it is not huge. Most of the time the recharging will be irrelevant, as a fight will be over before it comes into play.

The burst laser gaining holstering was a mistake, and an oversight from giving the flag to energy guns. However, if the energy gun is currently holsterable, which it is, then it would be no surprise if the burst laser was also. The back slot was removed from the C20 because it is a SMG like the others, and none of them go on the back, it's for consistency.

SpadesNeil:
>Burst Laser nor the Assault Carbine are actually obtainable

I am unsure about the burst laser but I know for a fact the Assault Carbine is, similar to the SAW, admin aboose rifle.

PR #2517, opened 2016-09-21
Makes cigs branded too

Yoshax:
Description of cig/cigbutt would reveal its brand for dastardly murdersolving revelations.

Baystation12/Baystation12#14261

### DIFF
--- a/code/game/objects/items/weapons/storage/fancy.dm
+++ b/code/game/objects/items/weapons/storage/fancy.dm
@@ -134,7 +134,7 @@
 	storage_slots = 6
 	can_hold = list(/obj/item/clothing/mask/smokable/cigarette, /obj/item/weapon/flame/lighter)
 	icon_type = "cigarette"
-
+	var/brand = "Trans-Stellar Duty-free"
 /obj/item/weapon/storage/fancy/cigarettes/New()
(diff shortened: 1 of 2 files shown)


code/game/objects/items/weapons/storage/fancy.dm
```diff
@@ -186,11 +191,13 @@
 	name = "\improper DromedaryCo packet"
 	desc = "A packet of six imported DromedaryCo cancer sticks. A label on the packaging reads, \"Wouldn't a slow death make a change?\""
 	icon_state = "Dpacket"
+	brand = "\improper Dromedary Co. cigarette"
```
Anewbe: These descriptions aren't consistent with themselves, some have cigarette at the end and the others don't. I at least would prefer that they all use one or the other.

Yoshax: I feel they read fine as is, its flow better with the cigarette on the two its on.

redstryker: Yeah, they read pretty well, IMO.

Anewbe: *shrug

PR #2525, opened 2016-09-22
Centralizes weight class / inventory definitions

Neerti:
A lot of new defines are now in inventory_sizes.dm, which contains;
All the size identifiers (the thing that tells the game if something is bulky, or w/e).
Storage costs for all the sizes, which are exponents of two, as previously.
A few constants for inventory size.

Also changes all storage item's capacity definitions by basing it off of how many 'normal slots' exist for it.  This allows one to change the definition for all of the defines in the file, and everything will follow along without needing to change 500 files.  In testing, I made all ITEMSIZE_COST_* defines doubled, and nothing had broke.

The benefit of doing all of this is that it makes adding new weight classes in the future much simpler, and makes knowing how much space a container has easier, as seeing ITEMSIZE_COST_NORMAL * 7 means it can hold seven normal items.


Anewbe:
#BraceForBugs

PR #2527, opened 2016-09-23
Fixes crate exploit

SpadesNeil:
Plastic sheet crates can be bought for 10 supply points.

I can then make 5 crates out of 50 sheets.

Each crate is worth 5 points.

5 x 5 + 5 = 30, allowing me to sell for a profit of 20 points per box of plastic sheets.

This stops that exploit by making plastic sheet crates more expensive to buy. I don't know how this wasn't discovered sooner. The new cost per crate is 30, making it really not worth the time. It should probably be 40 but I figure if I put it any higher people will complain.


Yoshax:
I do not like this fix, it solves the problem by creating another. You've made it not worth buying at all, not just for exploits. While plastic is not terribly useful and there is not very much you can do with it, you've made it overly expensive to acquire, making it even less useful. There is no reason with metal and plastic being the same price to order metal over plastic but increasing the price threefold increases this problem greatly.

Anewbe:
^

SpadesNeil:
I figured this solution was shit but it was my first choice. Here's my others:

Plan B: Lower the point return on crates from 5 to 1?
Plan C: Raise the cost of everything.
Plan D: Remove the ability to build plastic crates.

Neerti:
B might be the best plan in my opinion.

Yoshax:
Please do not do so, you will be severely nerfing cargo and making it a lot less fun to play. Cargo is also starved for points, please do not make it more difficult to do so.

Neerti:
Actually I'd prefer only plastic being worth 1 instead of 5, with normal crates still being 5.

Yoshax:
Keep in mind that several crates you can order come in plastic crates.

SpadesNeil:
> Actually I'd prefer only plastic being worth 1 instead of 5, with normal crates still being 5.
A discussion was also done about improving the passive point income.

Plan E it is. Closing this.

PR #2529, opened 2016-09-23
Bone Sound V2

redstryker:
Does the same thing as the last one, this time with four different sounds that all randomly play on a bone break.  
All sounds provided by yours truly. (Broke celery.)

https://www.dropbox.com/sh/5he0lri1x64j0jb/AADm9_Q6ovHcOvit-fM3GiPTa?dl=0


SpadesNeil:
I like 2 and 3.

I don't like 1 and 4.

But 2 and 3 are still good without making my ears bleed.

Fox-McCloud:
these should be `ogg` not `wav`

Anewbe:
Changelog would be nice. May need to poll around and see if the sounds don't freak people out, but they're fine by me.

SpadesNeil:
Well I think the new sound effects are less grating, but that's just my opinion.

redstryker:
I don't actually know how to make a changelog.

Anewbe:
Go into your repository, open the html file, make a copy of the example.yml file, rename that as a .txt, open it up, and there should be instructions inside. When you're done there, save it, rename it to a format  like [YourName] - [YourChanges].yml, and commit it the same way as any other file.

Yoshax:
You can edit .yml the same as a .txt
Changing the format and all is just another step, it's unnecessary. You can just open the example and edit it, then save as `yourfile.yml` may save you a short amount of time

Anewbe:
http://ss13polaris.com/forums/viewtopic.php?f=18&p=11717&sid=15c196140870b0f8ed1c39c7138c8186

PR #2533, opened 2016-09-24
Handbags

redstryker:
Adds several different bags to the loadout.  Added two no vars to backpacks.dm.  Doesn't make sense yet, but Anewbe has something planned with it.

Sprites:
http://puu.sh/rlRtZ/782be2d720.png


Anewbe:
To answer the unasked question, I'm planning to let you flip bags like satchels to the opposite shoulder, spritewise.

PR #2540, opened 2016-09-25
Energy weapons can be reloaded

Anewbe:
This changes the way energy guns work, instead of having to recharge them between uses, you can now also swap batteries on the fly, much like a ballistic weapon's magazine.
 - The battery in question is a specific type of power cell, called a device cell. A device cell is a small item, and other cells will not fit in weapons, they're too big.
 - Stunbatons will also use device cells, and you can now replace their cells without a screwdriver.
 - Clicking on an energy gun with an empty hand removes the power cell, clicking an empty gun with a device cell loads that cell.
 - Self-recharging weapons, such as the advanced egun, still self-recharge. They cannot swap cells, and cannot be placed in chargers, but when they do recharge it is always a flat 20% of the maximum cell capacity. Additionally, they will stop recharging for a short period after they've been fired.
 - HoS, Warden, and Sec Officers get an extra power cell in their lockers. If more cells are needed, they can be printed in the protolathe, or from a hacked autolathe.
 - Cargo crates that include energy weapons have a few spare device cells added in. Prices may be or need to be adjusted accordingly.
 - The map has been tweaked slightly. On the main level, multiple device rechargers have been added, notably in the first-aid stations. The Merc and ERT bases have had spare device cells mapped in.
 - Most weapons still have the same number of shots. However, on weapons that can fire both lethally and non-lethally, lasers drain twice as much power as tasers. Practically, this means that the egun has 5 laser shots before you need to reload, but it still has 10 taser shots.
 - Temporary sprites have been added for weapons that have had their cells removed, I hope to replace these with more lovingly crafted ones as soon as I receive them.
 - There is now a proc that admins can call on a weapon to make it self-recharging.

This is a rather large change, but it's been requested of me multiple times, and I think it's a fun idea. Balance changes will likely need to be made after introducing this, but if worse comes to worst this can be reverted.


Yoshax:
This was attempted previously, which has been failed to be mentioned, and can be found at #2160, along with the feedback and comments at the time.

Anewbe:
This is done more properly and completely, so the comments on that PR that are still relevant can and should be written here, rather than forcing people to read through that PR for changes that may or may not be included here.

SpadesNeil:
This will supposedly badly screw up balance but I've also wanted to make some reloading laser weapons for a while.

My only real gripe is this:

>they can be printed in the protolathe, or from a hacked autolathe.

No one will use the protolathe to print these if the autolathe has them. There's a protolathe _and_ autolathe in the Research lab, and it's a lot easier to just hack the autolathe. Just make them printable in the protolathe only. You can't print normal batteries in the autolathe, why print these in there? It's inconsistent.

Anewbe:
They're in the protolathe because that's where you get normal batteries, and they're in the autolathe because that's where you get normal ammunition (and because engineering/cargo is generally more populated than RnD). It's more options, rather than having it locked into one department that's not all that frequently played. Theoretically the device cell can/will be made to work in other device-style things, so I'd rather have it more obtainable than less.

SpadesNeil:
Yea but it's really not normal ammunition, and I was always under the impression that the inability to get extra magazines was part of way in which lasers got balanced.

Yoshax:
I'm not a fan of this, at all. For several reasons, some of which were pointed out in the previous PR.

Firstly, this completely removes and changes the whole balance of a lot of things. I'm going to detail what I believe are some of the pros and cons of the types of weapons.

**Projectile**
Pros:
- Reloadable (Different ammo)
- Cannot be EMPed
- In general you get more shots
- Burst Fire
- Chance for internal damage / Broken bones

Cons:
- Most things have higher bullet armor
- Cannot pass through windows
- Chance for internal damage / Broken bones
- Higher chance of stray projectile breaking windows / other things

**Energy**
Pros:
- Pass through windows
- Lower energy armor on most things
- In general they have higher damage
- Some have toggleable firing modes
- Higher scaling, Laser Cannon is top end vs Assault Rifle/Shotgun

Cons:
- Can be EMPed
- Cannot be reloaded

Looking at these we can see what one of the biggest features is reloadability. Removing it severely upsets the balance, and really, just makes the split less interesting. The reloadability is the biggest difference outside of damage type between the two types of weapons, removing it makes both of them less interesting.

Furthermore, having to add a new producable power cell, and mapping them in is more bloat that is not really needed.

And, as Spook said in the last PR:
`Further, the obvious following intent is to make the batteries swappable. This is a problematic goal, in that energy weapons represent an existing utility trade - they are not reloadable, but their beams ignore a fairly substantial portion of normal obstacles, like windows. With the recent recharging tweaks, they're also not a pain to refill.`

Furthermore, increasing the prices of things in cargo is not a good change, seeing as cargo already struggles for points when there is no mining and they're not allowed to use exploits. Weapons are already expensive, and it takes a long time to naturally get enough points for even the least expensive of things.

In addition, you've also made it so the energy gun only gets 5 lethal shots, which seems like a very strange change and I see no reason for making the lethal lasers use twice as much power. It further upsets the balance of these weapons, and does not at all put them in line with anything else. Making it a weapon that only has 5 shots is very poor, in my opinion.

I do not feel as though this is a positive change, and I feel as though it will be an overall negative to the server. It will upset the balance, make weapons less interesting, reduce the variance, all while adding more unnecessary things into the map and code, along with making energy weapons compared to everywhere else more convoluted.

Anewbe:
Gonna go point-by-point through those pro-cons. For the most part I find them inaccurate.

**Projectile**

**Pros:**
 
 - **Reloadable (Different ammo);** Huge. Not only can you reload and jump right back into action, but you can also trade out for AP ammo, or various shot types for a shotgun. Versatility's good.
 - **Cannot be EMPed;** Certainly a plus.
 - **In general you get more shots;** You almost always get more shots. The weapons that don't have 10+ shots are either the Sec pistol or do more damage than any common laser.
 - **Burst Fire;** This is massive, more bullets means more damage, means more chances to cause internal damage.
 - **Chance for internal damage / Broken bones;** You are undervaluing this. A lot. A single bullet can kill you, if the numbers work out. If it doesn't flat out organ damage you to death, it can still embed, break a bone, or otherwise cripple you. Dying on the way to medical is far more common with ballistics.
 - **!Bleeding!;** I added this one. Yes, burn damage now causes blood removal, but that's one and done, and generally secondary to the burn itself. Bullets can cause both external and internal bleeding, meaning a single bullet can cause you to bleed to death, even with the staunching thing. That exact thing happened to me fairly recently. FBPs don't care, sure, but they're still affected by the point above.
 - **!Penetrating!;** Oh hey, one more thing, the penetrating var. Hit one target, and the guy behind him. Dude behind a wall? Thermals and penetrating. Lasers have a wee bit of trouble doing that. Pretty sure this var lets you reach out and smack mech pilots, too.

**Cons:**

 - **Most things have higher bullet armor;** Not really, no. Other than Merc and ERT, most armors are either much weaker to bullets, or equally weak to bullets and lasers. Ninja and cultist armor stand out in that regard, and I've seen more ninja/cult than recently than the other two.
 - **Cannot pass through windows;** This is indeed a con, but with a bit of time you can shoot them out, assuming the target's not in actual space.
 - **Chance for internal damage / Broken bones;** This was listed as both a pro and a con. I'm assuming this means in regards to non-lethally taking someone down, rather than being here by mistake. Generally once someone's switched to the bullets that do this reliably, they're going for the kill.
 - **Higher chance of stray projectile breaking windows / other things;** Accurate, but not as dangerous here as it was on the Exodus.


**Energy**

**Pros:**
 
 - **Pass through windows;** Yeah, that's a plus. No argument.
 - **Lower energy armor on most things;** The majority of damaging energy shots are beams, which checks laser armor. Stun rounds check siemens. As mentioned above, this is still incorrect, correcting for those factors. Additionally, the specialized laser armor can bounce shots away from the target, as can Durasteel walls. Furthermore, lasers don't have the ability to trade out for Armor Piercing rounds, like ballistics can. Sure, the Xray gun exists, as does the Lasercannon, but those are relative rarities, where AP ammo can be printed from any autolathe after some hacking.
 - **In general they have higher damage;** They also have far fewer shots, and don't (with a singluar, aspawn only exception) have burst fire. Burst fire is very strong, more shots downrange means more possible DPS.
 - **Some have toggleable firing modes;** It's stun to lethal, or lethal to weaklaser. I don't see this as a very strong pro.
 - **Higher scaling, Laser Cannon is top end vs Assault Rifle/Shotgun;** The lasercannon is fairly rare, and if you buy it from the uplink you'd better have a safe place to charge it, or it's a big, low shot-count paperweight. Revolvers are far more common (and only 10 TC more expensive), there's at least two shotguns on station _per round,_ and ammo can be printed fairly easily.

**Cons:**
 
 - **Can be EMPed;** Yeah, this is a pretty big con. Being hit with an EMP drains _at bare minimum_ 50% of every cell involved, and if you're at the point-blank, it's the entirity of the cell. This PR doesn't change that. EMP's not all that uncommon, either. There's two ion rifles and a box of grenades in the armory at roundstart. Antags get tons too. They can be bought in the uplink, cultists have a rune, Technomancers have Pulsar and Energy Siphon (which is similar), Changelings have a scream, and the Ninja has emp grenades in the suit (despite being vulnerable to them). Additionally, you can make EMPs with Uranium and Iron, if you're lucky and/or desperate.
 - **Cannot be reloaded;** This con is the whole reason behind this PR. Running out of shots mid fight is much easier with a laser than it is with a ballistic weapon, which is part of the reason people tote around multiple lasers for fights, dealing with wildlife, etc. Those pistols in the armory are better than a taser in most situations, despite the melee armor check, because they do roughly the same amount of stun (plus 5 brute), because you can stick those 7 extra emp-proof shots (that you get in you sec locker) in a pocket, ready to go.
 - **!Long-Term Damage!;** Lasers have very poor long term damage. Infections are a joke, and the blood-evaporation thing is (as mentioned above) usually secondary to the burn damage inflicted. If you're in bad enough shape that medical needs to be giving you blood from a burn wound, you're already at a pretty hefty amount of damage. An infection can run the entire 3 hour round and only tick off a heat warning for a regular old human. Last I checked, FBPs don't care about either of these things, either.

Reloadabilty is the biggest feature here, yes. I don't think that makes the different weapons any more or less boring than they are. Additionally, as mentioned above, the damage type difference is a massive factor. Brute is amazing, burn is less than exciting.

I've not added a new cell. I've changed a var and mapped some things. Wasn't a lot of code.

Energy weapons aren't so much a utility trade as they are a Space Management exercise. For those big scary fights people are packing multiple guns.

Cargo needs some tweaks, or a full rework, as a general thing, and this isn't the best place to argue that point. Increasing the cost of those crates is sensible, because when someone (probably me) gets around to it, points will generate faster, and the contents of said crates are valuable.

Yes, this will affect the balance, but balance isn't a static, immutable thing. I think this change will shake things up for the better, in the end. Tweaks will need to be made, I've already said that. What sort of tweaks, precisely, will be seen, but if it's absolutely necessary, and there is no simpler way to correct it, this can be reverted. It's not exactly a lot of code.

SpadesNeil:
Here's what I propose. What if we make it so energy weapon batteries can't just be yoinked out and slapped in like a magazine, but rather, required screwdriver and/or other tools to remove it? Stuff that would take time to do, so you can't just reload a weapon on the fly. You have to actually sit down, put the weapon down, open the cell compartment, remove the cell with a screw, put in a new cell, screw it in, and THEN you can shoot.

This seems like a good middle ground.

Anewbe:
If that's what it takes, the whole thing'd be useless. You can still just slap the gun in a charger, which is probably faster and doesn't take any sort of special tool.

SpadesNeil:
That simply isn't true. Chargers are in limited supply. Screwdrivers are not. I'm not going to drag a charger with me everywhere I go. That isn't practical. If I'm in maintenance and I have a moment of reprieve, I'll take a chance to swap cells.

Anewbe:
I've added several chargers across the station in this PR. They're not hard to find, either.

SpadesNeil:
If there aren't chargers in literally every room, then I don't see your point.

Yoshax:
`Most things have higher bullet armor; Not really, no. Other than Merc and ERT, most armors are either much weaker to bullets, or equally weak to bullets and lasers. Ninja and cultist armor stand out in that regard, and I've seen more ninja/cult than recently than the other two.`

I may have misread a bit on this one, most armor is equal on both fronts, at least vests and suchlike. 

`!Penetrating!; Oh hey, one more thing, the penetrating var. Hit one target, and the guy behind him. Dude behind a wall? Thermals and penetrating. Lasers have a wee bit of trouble doing that. Pretty sure this var lets you reach out and smack mech pilots, too.`

The issue is though, it's actually fairly unlikely for penetrating to proc, or actually do anything. It really isn't that impacting or useful. It only procs on things that have penetrating, and more than 20 damage, and even then it's only a % chance as the damage to penetrate, and this is, I believe, calculated after armor. If the bullet passes through, you also do not get your gun embedded bullet time.

`Burst Fire; This is massive, more bullets means more damage, means more chances to cause internal damage.`

Burst fire isn't the best thing, it's useful, sure. But in some scenarios single fire is better. Burst fire is inaccurate, and wild. It also uses more bullets, which you list as a good thing, but it is also bad. If you only have 10 bullets, using 3 each shot may not be your best bet.

`Cannot pass through windows; This is indeed a con, but with a bit of time you can shoot them out, assuming the target's not in actual space.`

Shooting the windows out takes time, and ammo. If the person you're fighting can shoot through the window, you're screwed.

`Chance for internal damage / Broken bones; This was listed as both a pro and a con. I'm assuming this means in regards to non-lethally taking someone down, rather than being here by mistake. Generally once someone's switched to the bullets that do this reliably, they're going for the kill.`

Listed as both a pro and con because the LTL rounds can break bones and cause internal damage as far as I'm aware, you can beanbag someone and mess them up pretty badly. There's not an equivalent non-lethal option outside of shotgun stun rounds.

`In general they have higher damage; They also have far fewer shots, and don't (with a singluar, aspawn only exception) have burst fire. Burst fire is very strong, more shots downrange means more possible DPS.`

Burst fire is strong, but it can also be innacurate. It's powerful and it's not, it's hard to judge it fully. And yes, they have fewer shots, but it leads onto a point I forgot to make also, as far as I'm aware, burns cause more pain than bullets meaning it takes less shots to drop someone into paincrit, or suchlike. You get more bang for you less shots.

`Some have toggleable firing modes; It's stun to lethal, or lethal to weaklaser. I don't see this as a very strong pro.`

Stun to lethal, or vice-versa is a strong tool. Normal projectile weapons can't do it, you're locked into what you pick, or have to take the time to reload into something less than optimal, such as flash rounds for most weapons if you want to go non-lethal, this means you have to carry the ammo. 

`!Long-Term Damage!; Lasers have very poor long term damage. Infections are a joke, and the blood-evaporation thing is (as mentioned above) usually secondary to the burn damage inflicted. If you're in bad enough shape that medical needs to be giving you blood from a burn wound, you're already at a pretty hefty amount of damage. An infection can run the entire 3 hour round and only tick off a heat warning for a regular old human. Last I checked, FBPs don't care about either of these things, either.`

I don't really agree with this, it applies to some of the things I've said. A lot of the time, a fight will be over and resolved before any lasting damage is relevant. If you get medical treatment, then you're fine. This only really applies to antagonists who don't have a way to resolve their internal issues or suchlike. It's a fair con in that aspect, having issues you can't deal with isn't much fun.


`I've not added a new cell. I've changed a var and mapped some things. Wasn't a lot of code.`
The thing is mapped in now, and producible and all such, it wasn't beforehand. 

`Cargo needs some tweaks, or a full rework, as a general thing, and this isn't the best place to argue that point. Increasing the cost of those crates is sensible, because when someone (probably me) gets around to it, points will generate faster, and the contents of said crates are valuable.`

It does need work, but right now, it's point generation is laughably low, and the reason people turn to exploits to actually get points. Making things more expensive isn't needed.


`Yes, this will affect the balance, but balance isn't a static, immutable thing. I think this change will shake things up for the better, in the end. Tweaks will need to be made, I've already said that. What sort of tweaks, precisely, will be seen, but if it's absolutely necessary, and there is no simpler way to correct it, this can be reverted. It's not exactly a lot of code.`

The issue is that most people do not care, or do not care enough to comments on things. Most people are just content with whatever. If something goes in that people are content with, or just a little displeased, or anything of the sort, it's highly unlikely that it will be reverted. We shouldn't be settling for changes that people are content with. People should be happy, or glad, or suchlike.


I'm not really a fan of this at all, I see it as unnecessary, and upsetting to the balance as has been noted, and reasons given. It removes the variation and difference between the types of weapons, which is a point you seem to have not addressed in your comment. It's central to their identity that they're recharged and not reloadable. It does not need changed.

Datraen:
[Glory to the Vanu Sovereignty?](https://www.youtube.com/watch?v=Lc5L_K2WNIA)

Laser damage needs to be nerfed, or the batteries need to be big, bulky items.
Otherwise we may as well tell people not to bother with projectile weapons, as energy weapons beat out projectile weapons in just about every case.

redstryker:
To be honest, yeah, laser weapons need a nerf if this plans to be a thing.  Adding such a feature gives a significant advantage to them.  I do like Spades' solution of requiring a screwdriver to remove the battery, but there are also other solutions out there.  This just seems to be an easy way to make ballistics irrelevant now that I think about it.

Spookerton:
👎 Has been denied before, will be denied again.

Anewbe:
Redone, as discussed in the 10/6 Staff Meeting.

Here's what's in:

- Energy weapons and stunbatons now use special device power cells, these can still be recharged.
- Energy weapons can be unloaded and reloaded by clicking them with an empty hand or a device cell, respectively. The process of loading a cell takes a few moments.
- Stunbatons no longer require a screwdriver to switch cells.
- Tweaked the order in which stunbatons check for power, they should now visibly power off when their cell hits 0, instead of one hit after.
- Security lockers (HoS, Warden, and Officer) now have an extra device cell in them.
- Protolathe can print device cells.
- Adds start_recharge() proc to energy weapons. When called, this should cause the affected weapon to begin self-charging.
- Weapons that self-recharge won't do so for a short period after firing.
- On weapons that can fire both lethally and non-lethally, lasers drain twice as much power as tasers.
- Lasercannon and LWAP still use the old system, as do self_recharge weapons.
- Map changed to include more rechargers, Merc and ERT bases include extra device cells.

SpadesNeil:
I am okay with this. May as well see how it works.

I'm still against it but I am not _adamantly_ against it, and I'd like to at least see how it plays out.

Neerti:
This was talked about in the staff meeting, and ultimately it was decided to merge it and see what occurs.

PR #2545, opened 2016-09-26
Fixes #2536

### LINKED ISSUE #2536: Plastic sheets can be exploited for exponential supply points
A box of plastic sheets costs 10 points. The box its self is worth 5 points. You can then produce 5 more boxes with the plastic. Now you have six crates, which amount to 50 points you'll get back from cargo when you send back the boxes.

For plastic to cost more, or for plastic boxes to be worth less.

#### What actually happened
Someone abused this until CentCom got onto the command announcement and yelled at the enterprising cargo technician who saw fit to turn the colony into a box manufacturing plant. Also, your pension is gone. CentCom lost a good amount of stock from this oversight in their logistics system.

#### Steps to reproduce
1. Order plastic sheet crate.
2. Build boxes out of plastic.
3. Send the boxes you made back to CentCom.
4. Repeat until company is bankrupt.

#### Additional info:
 - **Server Revision**: https://github.com/PolarisSS13/Polaris/commit/828dacf485f4377d139ae517e8fb737f62f6b9d0
 - **Anything else you may wish to add**: Increasing the cost of the crates was seen as a bad idea in a previous PR. Let's not go that route.

Anewbe:
Fixes #2536. Plastic crates give 1 point each, normal crates give the usual 5.

Also ups cargo point-gain rate by 50%

### DIFF
--- a/code/game/objects/structures/crates_lockers/crates.dm
+++ b/code/game/objects/structures/crates_lockers/crates.dm
@@ -8,6 +8,7 @@
 	icon_opened = "crateopen"
 	icon_closed = "crate"
 	climbable = 1
+	var/points_per_crate = 5
 //	mouse_drag_pointer = MOUSE_ACTIVE_POINTER	//???
 	var/rigged = 0
 
@@ -240,6 +241,7 @@
 	icon_state = "plasticcrate"
 	icon_opened = "plasticcrateopen"
 	icon_closed = "plasticcrate"
+	points_per_crate = 1	//5 crates per ordered crate, +5 for the crate it comes in.
 
 /obj/structure/closet/crate/internals
 	name = "internals crate"
--- a/code/game/supplyshuttle.dm
+++ b/code/game/supplyshuttle.dm
@@ -138,9 +138,8 @@ var/list/mechtoys = list(
 /datum/controller/supply
 	//supply points
 	var/points = 50
-	var/points_per_process = 1
+	var/points_per_process = 1.5
 	var/points_per_slip = 2
-	var/points_per_crate = 5
 	var/points_per_platinum = 5 // 5 points per sheet
 	var/points_per_phoron = 5
 	//control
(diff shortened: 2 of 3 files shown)


Yoshax:
I do not feel this is enough, really, it fails to address the issue. It's an extra 60 points in an hour. That really isn't anything. I kinda assumed when we were upping it that it would be at least by a factor of two. It's very dull to play cargo when there's no mining and you only have the slow point generation. They either need yet higher generation or more ways of getting points. If people actually want to order more than 1 or 2 things, you end up with 0 points.

Anewbe:
I can remove the points gain and make that a separate PR, if you'd rather. This exploit needs to be fixed, though.

Yoshax:
The exploit kind of speaks for the state of affairs. It's sad that you have to go through exploits to get a decent enough amount of points. Me, TS, and Ala had discussed it on TS and Ala agreed that it's an issue that it's so difficult to get points when there's no miners, or even when there is and they don't give you jack. 

I'm not bothered either way, it's a bug, and I suppose it should be fixed. But people resorting to using bugs, as noted above, speaks for the situation as a whole.

Anewbe:
I'm not saying there's not an issue, I'm saying this bug needs to be gone, and the system needs to be tweaked so people don't need to rely on exploits. I just don't have a magic number to toss in there right now that simultaneously gives cargo a bit more leeway without making it a free-for-all on buying shit every round.

Yoshax:
I feel as though it would be better being able to buy too much than too little, but maybe that's just me.

The PR fixes the issue in a decent way. Especially because cargo code sucks massive dick, :+1: on that basis.

SpadesNeil:
There should be some kind of discussion of ways to rework cargo to make it less shit. I feel like certain items _should_ be expensive because they _should_ require work from miners to obtain--but that's certainly a problem because it means less points for everything else.

Possible solutions to the problem include separate pools of supply points, or using cash money to get extra supply points if a boost is needed. Crew who need extra supply during a supply point deficit can just outright deduct it from their paychecks and send hard cash to CentCom.

Those are just two very rough solutions I thought of and I just woke up so they might suck. You see the direction I'm trying to go, however. We can ramble about this on Discord.

PR #2548, opened 2016-09-26
Writer Alt Title

redstryker:


Anewbe:
Author is a bit weird... Maybe Writer?

PR #2552, opened 2016-09-26
Lessens the amount of pain from burns

Anewbe:
Slightly unrealistic, but getting locked in paincrit from relatively little burn damage is very frustrating.

### DIFF
--- a/code/modules/mob/living/carbon/shock.dm
+++ b/code/modules/mob/living/carbon/shock.dm
@@ -10,7 +10,7 @@
 	src.traumatic_shock = 			\
 	1	* src.getOxyLoss() + 		\
 	0.7	* src.getToxLoss() + 		\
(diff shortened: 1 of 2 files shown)


Yoshax:
This removes one of the biggest strengths of energy weapons and is central to the balance. This has all been noted in the battery PR, I see no reason for this change. It further moves to genericify and remove the split and variation between the two different kinds of weapons.

nachomeep:
I'm incredibly against this, personally. Wasn't this, like, the majority of the reason to use lasers?

Anewbe:
When you've switched to lasers, your intent is to kill. If you want nonlethal takedowns, there are other options. Being trapped somewhere stupid, perfectly 'safe' but waiting for your burn to tick down enough to stand up isn't fun.

PR #2530, opened 2016-09-23
Revert "Revert "Revert "Revert "Adds hub passwordu""""

Neerti:
Reverts PolarisSS13/Polaris#2450

PR #2555, opened 2016-09-26
Revert "Revert "Revert "Revert "Revert "Adds hub passwordu"""""

Neerti:
Reverts PolarisSS13/Polaris#2530
Next time I will try science which if successful, will mean we don't need to do this every two weeks.


PJB3005:
Yo dawg....

PR #2557, opened 2016-09-27
armband go onto outer suit

Yoshax:


SilverTalismen:
Maybe make it do as belts do and add a toggle to switch layers?

Yoshax:
verb bloat is bad

SilverTalismen:
True, but it could simply bea right click option and not a object option?

Yoshax:
It's the same and unnecessary, this does the job without having to be messy

SilverTalismen:
True, it was mostly a thought ^^

PR #2560, opened 2016-09-28
Adds code to make species level metabolism rates possible.

Anewbe:
Doesn't actually add any yet, but they should be possible now.
Certain reagents (notably Inaprov, blood, alcohol, and toxin) aren't affected by a species' metabolism, but rather all work at the same rate.

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents.dm
+++ b/code/modules/reagents/Chemistry-Reagents.dm
@@ -19,6 +19,7 @@
 	var/list/data = null
 	var/volume = 0
 	var/metabolism = REM // This would be 0.2 normally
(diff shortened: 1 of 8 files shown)


code/modules/reagents/Chemistry-Reagents.dm
```diff
@@ -19,6 +19,7 @@
 	var/list/data = null
 	var/volume = 0
 	var/metabolism = REM // This would be 0.2 normally
+	var/mrate_static = 0	//If the reagent should always process at the same speed, regardless of species, make this 1
```
Hubblenaut: Worth to use `TRUE` and `FALSE` here to make it clear this is a boolean value.

PR #2573, opened 2016-10-02
Adds Asteroid Cracks

Neerti:
Adds a new type of turf, which is a cracked asteroid floor.  It is functionally a space tile, which means air can escape through the cracks and into the void.  These cracks can be plugged by building a floor over them using rods and tiles.  The cracks are generated randomly along with the asteroid whenever a floor is placed, with a 1 in 10 chance of it being a cracked floor.

Picture of cracks


SpadesNeil:
Neat. This is actually a very nice addition for making breaches dangerous again if left unrepaired.

SkowronX:
Needs better sprites. And "smoothing"

PR #2576, opened 2016-10-02
Automates Changelog Button, Forever

Neerti:
Replaces the changelog button to point to the 'pulse' tab on GitHub, which will always give up to date information on what was added.  This relieves a fairly significant burden on the dev staff, and introduces a larger number of people to our code on GitHub.
The game will still bold the changelog button if the server's revision SHA is different from what they saw last time.

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -15,7 +15,7 @@ datum/preferences
 	var/last_id
 
 	//game-preferences
-	var/lastchangelog = ""				//Saved changlog filesize to detect if there was a change
+	var/lastchangelog = ""				//Saved changlog SHA to detect if there was a change
 	var/ooccolor = "#010000"			//Whatever this is set to acts as 'reset' color and is thus unusable as an actual custom color
 	var/be_special = 0					//Special role selection
 	var/UI_style = "Midnight"
@@ -262,7 +262,7 @@ datum/preferences
 	if(be_random_name)
 		real_name = random_name(identifying_gender,species)
 
-	// Ask the preferences datums to apply their own settings to the new mob 
+	// Ask the preferences datums to apply their own settings to the new mob
 	player_setup.copy_to_mob(character)
 
 	if(icon_updates)
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -393,31 +393,17 @@
 /client/verb/changes()
 	set name = "Changelog"
 	set category = "OOC"
-	getFiles(
-		'html/88x31.png',
-		'html/bug-minus.png',
-		'html/cross-circle.png',
-		'html/hard-hat-exclamation.png',
-		'html/image-minus.png',
-		'html/image-plus.png',
-		'html/map-pencil.png',
-		'html/music-minus.png',
-		'html/music-plus.png',
-		'html/tick-circle.png',
-		'html/wrench-screwdriver.png',
-		'html/spell-check.png',
-		'html/burn-exclamation.png',
-		'html/chevron.png',
-		'html/chevron-expand.png',
-		'html/changelog.css',
-		'html/changelog.js',
-		'html/changelog.html'
-		)
-	src << browse('html/changelog.html', "window=changes;size=675x650")
-	if(prefs.lastchangelog != changelog_hash)
-		prefs.lastchangelog = changelog_hash
-		prefs.save_preferences()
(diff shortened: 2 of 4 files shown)


Yoshax:
This is a very poor change. Changelog entries are often written in a very different format than what you can see in the pulse. The changelog is more presentable and nicer for people to read.

Neerti:
The reduction of workload on a small dev team, I believe, is very much worth more than whatever subjective prettiness that the current changelog may hold.  No one wants to do changelogs, and you even said that you can't be bothered to do them, on Discord.

Anewbe:
I'd gladly do it if I knew how, to be honest.

Yoshax:
I would rather do it than see it replaced with a poor alternative: #2579

But perhaps at some point you would like to find out, or discuss why it is that I, your previously most active and top contributer, along with others, don't have the motivation to do something simple for the server. Which is something I don't care to do in public.

Neerti:
If I lay out how both systems actually work, it should be obvious why my PR is vastly superior in the long term. 
The current system:
[*] Requires the PR author to copypaste a .yml file and fill it out, which is more often then not prone to errors in filling fields like the name out. I considered trying to use CI to make sure the default name isnt used. 
[*] Requires someone to run an old python script locally and then PR what it spits out , which also takes time away, and often its a week or two before the changelog finally reaches the server. 
[*] Not all PRs recieve changelogs. 
[*] The dates are set on the day of the changelogs being generated, not when the feature gets merged. 

Compare to the pulse page;
[*] Requires zero maintenance on our part. 
[*] Will always be up to date. 
[*] All changes are recorded, no matter how small. 
[*] Attribution and dates will be correct. 
[*] Increases awareness of our codebase, which may encourage bugs being reported more often. 

If we were using the new system, or anyone else was, and someone purposes that it gets replaced by a non-automatic system that just has the supposed advantage of looking nicer, do you think that would go well?

The changelog is holding us back.

Yoshax:
`Requires the PR author to copypaste a .yml file and fill it out, which is more often then not prone to errors in filling fields like the name out. I considered trying to use CI to make sure the default name isnt used.`

Isn't really an excuse, it isn't difficult, and takes less than a minute. It also produces a nicer looking result, which is better formatted and presentable. In addition, most of the mistakes made in the yml such as default name are correct by the person running the script, I know I've correct it many a time. Such as today.

`Requires someone to run an old python script locally and then PR what it spits out , which also takes time away, and often its a week or two before the changelog finally reaches the server.`

Rarely has it ever been more than a few days, only recently. I used to run it almost everyday when there was actual entries. And I would rather continue to run it than have to deal with the alternative.

`Not all PRs recieve changelogs.`

They don't, but they should. You should enforce it and make people do it rather than providing a cop out. And considering you're the biggest culprit for this, it's a poor point.

`The dates are set on the day of the changelogs being generated, not when the feature gets merged.`

Is the only actual issue out of the ones posted. It isn't really a huge one either if the changelog is updated often.


Go take a look at the pulse: https://github.com/PolarisSS13/Polaris/pulse
And now take a look at: https://i.gyazo.com/18770e1f289b844b6fbc4bf4427d2a2f.png

I know which I believe looks nicer and more readable. The changelog is written in a readable way for everyone, the pulse is not, neither are PRs or Commits. It's not even close. It's user friendly and readable vs fast and easy. There's a reason providing a changelog to users is all but the universal standard in games and programs. even in open-source games.

redstryker:
:-1: This change is lazy.  An actual changelog is better since it is designed for people to read and quickly understand.  PR names can be whatever people want.

redstryker:
I would like to also note that what a PR says in the title is not necessarily going to be what is going to be merged.  I've had to change things on my PR's before that required a title change.  I do change my titles when I make such large changes, but not everyone does.

Hubblenaut:
>introduces a larger number of people to our code on GitHub.

I laughed. Really. Changelogs are not the problem on Polaris development. People left for entirely different reasons.

comma:
Do the thing that TG does with changelog generated from PR OP?

redstryker:
We found a solution, actually.  This should be closed.

optimumtact:
Changelog is for players
Git log is for developers

Always write your commit messages for developers or big bad wolves will eat you in the night

Chnkr:
Both TG and Paradise have automatic changelog generation based on a tag inside the PR body, should just use that as this is a poor solution.

PR #2577, opened 2016-10-02
Actually Makes Technomancer Votable

Neerti:
Forgot to do this in the other PR.
Also turns Conflux, Lizard, Visitors, and MercWiz back on.


Anewbe:
Science have mercy on us all...

PR #2586, opened 2016-10-04
Adds Departmental Emergency Vests

redstryker:
Adds 'emergency armors' for the Medical, Supply, Research, and Engineering departments.  They are identical to kevlar vests except for the fact that they carry different things.  They can be obtained from a crate in Cargo.

Sprites:
http://puu.sh/rxnUT/db80e69dbc.png


Woodratt:
Not, really a fan. Of the sprites or the concept really... Cargo can already order vests via one of the crates and they are never distributed to any of the departments.

Anewbe:
More or less in agreement with Woodrat. Other than the tiny splash of color, it's a bit difficult to tell which department these are supposed to belong to, and I'm not sure if the departments really need their own vests, if cargo can just order the generic black ones.

redstryker:
*shrugs*  Don't care that much about this one anyway.  Closing.

PR #2593, opened 2016-10-05
Technomancer Bug Fixes & Tweaks

Neerti:
- Cores can now be locked and unlocked by a verb.  Locked cores cannot be removed by anyone.  Cores automatically unlock if the wearer dies or otherwise stops existing for some reason.
- Healing spells now induce a flat instability cost on the healer instead of it being split between healer and healed.  The amount healed is greatly increased if used on someone besides the healer.  Between old system and this new one, instability cost is much cheaper as well, but heals for about the same as before if healing someone else.
- Mend Organs is now called Great Mend Wounds, and now heals broken bones, IB, eye damage, and blood loss in addition to internal organs.
- Radiating instability (the purple glow) now gives a warning to people afflicted.
- Instability discounts should start applying correctly for specific cores.
- Instability now affects all living mobs and not just humans.  This includes borgs and simple animals.
- Phase shift now adds ongoing instability while hiding inside the rift.
- Projectile spells should be logged now.
- Wards last forever, and should not die in vacuum now.
- Restoration aura is now green colored and not blue.
- Simple mobs and cyborgs are now harmed by lightning.

### DIFF
--- a/code/game/gamemodes/technomancer/core_obj.dm
+++ b/code/game/gamemodes/technomancer/core_obj.dm
@@ -86,6 +86,8 @@
 	if(wearer && wearer.mind)
 		if(!(technomancers.is_antagonist(wearer.mind))) // In case someone tries to wear a stolen core.
 			wearer.adjust_instability(20)
+	if(!wearer || wearer.stat == DEAD) // Unlock if we're dead or not worn.
+		canremove = TRUE
 
 /obj/item/weapon/technomancer_core/proc/regenerate()
 	energy = min(max(energy + regen_rate, 0), max_energy)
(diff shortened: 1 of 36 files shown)


Anewbe:
The instability cost for healing seems to have been doubled, and the heal if used on self seems to have been halved, is this intended?

Neerti:
Most of the healing spells previously followed a code of;
```
H.adjust_instability(2.5)
origin.adjust_instability(2.5)
```
(origin is the source of the spell, H is the human being healed)
every second, for five seconds, which would total 10 instability on healer and 10 on healed (or 20 on healer if self-healing).
Now it is just a flat ```origin.adjust_instability(10)``` that is ran once, and on only the healer, which is effectively half of the instability.
Healing was 25 damage for most types (except organ) on both healer and healed.  It's now 30 for healed if they are not also the healer, otherwise it is 10, however the instability cost is not effectively doubled like the old system.

PR #2606, opened 2016-10-07
Makes flashlights use cells

Yoshax:
- Flashlights now use cells
- These can be swapped
- Flashlights are now marginally brighter across the board
- Flashlights now have a selectable brightness level
- Only flashlights, maglights, and lanterns use this
- Flashlights will spawn with a random amount of power between 25% and 100%
- A full flashlight on low will last 66 minutes, one on medium will last 33 minutes, one on low will last 16 minutes. Numbers are open to change and feedback.
- Charge amount is obfuscated and shown as a general description such as, "Average" rather than a full percent when examining the flashlight

Aim is to add more atmosphere and flavor. It would be more fun, and exciting if you are in maintenance looking for something, a changeling, traitor, etc and you lose track of your charge and it runs out.

I realise the low duration is not very realistic, but the aim of it is fun, flavor and atmosphere over realism.

### DIFF
--- a/html/changelogs/Yoshax-Flashlights.yml
+++ b/html/changelogs/Yoshax-Flashlights.yml
@@ -1,0 +1,39 @@
+################################
+# Example Changelog File
+#
+# Note: This file, and files beginning with ".", and files that don't end in ".yml" will not be read. If you change this file, you will look really dumb.
+#
+# Your changelog will be merged with a master changelog. (New stuff added only, and only on the date entry for the day it was merged.)
+# When it is, any changes listed below will disappear.
+#
+# Valid Prefixes: 
+#   bugfix
+#   wip (For works in progress)
+#   tweak
+#   soundadd
+#   sounddel
+#   rscadd (general adding of nice things)
+#   rscdel (general deleting of nice things)
+#   imageadd
+#   imagedel
+#   maptweak
+#   spellcheck (typo fixes)
+#   experiment
+#################################
+
+# Your name.  
+author: Yoshax
+
+# Optional: Remove this file after generating master changelog.  Useful for PR changelogs that won't get used again.
+delete-after: True
+
+# Any changes you've made.  See valid prefix list above.
+# INDENT WITH TWO SPACES.  NOT TABS.  SPACES.
+# SCREW THIS UP AND IT WON'T WORK.
+# Also, all entries are changed into a single [] after a master changelog generation. Just remove the brackets when you add new entries.
+# Please surround your changes in  double quotes ("), as certain characters otherwise screws up compiling. The quotes will not show up in the changelog.
+changes: 
+  - rscadd: "Flashlights now take powercells."
+  - rscadd: "Flashlights now have multiple brightness levels, including low, medium and high."
+  - tweak: "Flashlights are slightly brighter in general."
+  - rscadd: "The cells that spawn in flashlights will spawn with a random amount of charge between 25% and 100%"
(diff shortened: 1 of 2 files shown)


Datraen:
Travis is being a git about formatting comments.

Yoshax:
Ya, I got it. Typo'd a ' as a ;

Techhead0:
Things I noticed:
- `attackby()` should probably end in `else ..()`
- Flares and lamps will spawn with cells (which can be removed and replaced)
- Penlights are noticeably smaller than their cells and might end up exploited as cell storage devices

Yoshax:
Resolved, resolved and resolved.

Anewbe:
I like the concept a great deal, but I have the following concerns:

1. I would like if they used the device cells, rather than the full sized batteries. Feels more believably sized, and it'll keep the times on a refill consistent.

2. Do the items with power_use = 0 still have the ability to switch brightness levels? Strikes me as an oversight if so, because why would anyone ever use anything else if there's no downside?

3. I don't think the initial charge should be randomized. I'd rather not find myself preparing for a blackout, then having to immediately charge a flashlight to make sure it's actually full.

Also might be wise to throw a thread on the forums about this, people might have opinions.

Yoshax:
1. I disagree, to do so you would have to make the device cells far more prevalent and easy to acquire. This would upset the balance of your energy weapons greatly, or it would make acquiring cells for the lights too difficult. Power cells are fairly common, engineering have a lot, tool storage always has a lot, and there's other easy places to get them.

2. They do not, the verb is removed within the parent new() if they do not have power_use != 0

3. That's part of the fun of it. It's rather dull that everything on station is perfect and pristine, it really adds to the feel that many dislike that every shift is super new and stuff. The randomisation could be knocked up a bit to between say 50% and 100%, but I would like it to some degree to add more flavor and a more 'lived-in' feel to things.

zaredman:
If you want guaranteed light in an 'oh shit' situation, grab a flare.

Yoshax:
Or a glowstick.

redstryker:
Yeah, the glowsticks are there for a reason. :P  They burn twice as long as flares.

Anewbe:
Alright then, I'm fine with leaving it as a normal power-cell if you remove the randomization.

Yoshax:
I've noted in the OP that flashlights are margiinally brighter across the case, I thought I had made this change but it seems I never did. It also means my last commit message is slightly incorrect. Which will be corrected in a moment.

Yoshax:
Should be good now.

html/changelogs/Yoshax-Flashlights.yml
```diff
@@ -0,0 +1,38 @@
...
+
+# Any changes you've made.  See valid prefix list above.
+# INDENT WITH TWO SPACES.  NOT TABS.  SPACES.
+# SCREW THIS UP AND IT WON'T WORK.
+# Also, all entries are changed into a single [] after a master changelog generation. Just remove the brackets when you add new entries.
+# Please surround your changes in  double quotes ("), as certain characters otherwise screws up compiling. The quotes will not show up in the changelog.
+changes: 
+  - rscadd: "Flashlights now take powercells."
+  - rscadd: "Flashlights now have multiple brightness levels, including low, medium and high."
+  - rscadd: "The cells that spawn in flashlights will spawn with a random amount of charge between 25% and 100%"
```
Anewbe: Looks like this was changed, yes?

PR #2607, opened 2016-10-07
Generalises shoe storage

Yoshax:
No longer can you only put knifes into your boots. Any shoe that can hold items can now hold any small or smaller item. This operates on a blacklist, with only a few things currently blacklisted, including forensics stuff, and boxes, since I believe those are small (I may be wrong.).

Some old code has been gutted and/or cleaned up, including the removal of the buggy overlay for having a knife in your boot.

Lastly, some more shoes have the ability to hold items, like the leg armor, magboots and galoshes.

### DIFF
--- a/code/modules/clothing/clothing.dm
+++ b/code/modules/clothing/clothing.dm
@@ -380,7 +380,10 @@
 	var/can_hold_item
 	var/obj/item/holding
 	var/list/item_holding_blacklist = list(/obj/item/weapon/forensics/swab, /obj/item/weapon/reagent_containers/spray/luminol, /obj/item/device/uv_light,
-										/obj/item/weapon/sample, /obj/item/weapon/forensics/sample_kit, /obj/item/weapon/storage/box, /obj/item/clothing/shoes) //no you can't boots with boots in your boots
+										/obj/item/weapon/sample, /obj/item/weapon/forensics/sample_kit) //no you can't boots with boots in your boots
+	var/list/item_holding_whitelist = list(/obj/item/weapon/material/butterfly, /obj/item/weapon/material/butterfly/switchblade, /obj/item/weapon/material/knife,
+											/obj/item/weapon/material/kitchen/utensil, /obj/item/weapon/material/hatchet/unathiknife, /obj/item/weapon/material/hatchet/tacknife/combatknife,
+											/obj/item/weapon/gun/projectile/derringer, /obj/item/weapon/melee/telebaton, /obj/item/weapon/melee/energy/sword)
 
 	permeability_coefficient = 0.50
 	slowdown = SHOES_SLOWDOWN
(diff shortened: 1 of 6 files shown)


Anewbe:
Talked with Woodrat and Neerti, a whitelist of approved items is more sensible in this than a blacklist is. There are a very large number of items with ITEMSIZE_SMALL or smaller that don't make sense in a shoe, for believability or balance reasons.

Yoshax:
There are a lot of items that don't make much sense, yeh, but a whitelist is kinda messy. It's a little bit of fun, and shouldn't pose a massive balance concern. It's just another storage slot that can only hold small items. We have a lot of storage as is, one more small slot does not break the camel's back, so to speak. Having a whitelist of items also assumes that people will update it for new items, or suchlike. A blacklist has the same problem though.

It's a little bit of fun, and isn't a huge balance concern. I thought we were in favour of fun features, disregarding their slight unrealistic nature?

Woodratt:
There's fun and there's being able to put a coffee cup or a hand grenade in my shoe.

Whitelist is a hell of a lot better than the lengthy blacklist we'd have to come up with.

Yoshax:
Many things receive dislike for whatever reasons, but we go through with them because code and the repo is not a stale, unchangeable monolith. Where lies the harm in trying it?

Neerti:
There are hundreds of different items on the station, of which a large number of those fit in the ```ITEMSIZE_SMALL```, things which include AI upload boards, circuitboards, flashes, various pieces of food like cake slices, some types of beakers, some tools like crowbars, grenades of all types including flashbangs and frags, default flashlight and maglight, communicator, PDA, bounced radio, signaler, emergency oxygen tank, and a whole lot more.

Making it a blacklist for all of those, searching for other nonsensical items to block, and having to constantly update it whenever a new small item is added is absurd.

It gets changed to a whitelist or it doesn't get merged.

Yoshax:
Alternatively. We allow all tiny items to be put in, and have a whitelist for allowed small items, such as the various blades and other, similar items. Blacklist remains due to not wanting to shove your forensics stuff into the boot while you try to do your job.

Anewbe:
Alternatively, you make it use a whitelist as requested, and don't include the forensics items in it. Problem averted.

Yoshax:
Where lies the issue in allowing any tiny item?

Yoshax:
`Where lies the issue in allowing any tiny item?`

Anewbe:
`It gets changed to a whitelist or it doesn't get merged.`

`Where lies the issue in doing what is requested of you?`

Yoshax:
You could answer the question and explain why something has to be a certain way rather than ignoring it for more than a day. 

And you're not requesting, you're ordering, a request is: `the act of asking for something to be given or done, especially as a favor or courtesy; solicitation or petition.`

PR #2608, opened 2016-10-07
Allows Suit Jackets to be Worn Over Suit Items

redstryker:


Anewbe:
I'm going to ask you to explain your reasoning on this, I don't quite like it and the blank description doesn't fill me with a great deal of confidence.

redstryker:
Basically, it is so that you can wear them over vests and the like, since you will often see people with sweater vests under suit jackets.  I understand that it doesn't exactly make sense to have it over, say, a leather jacket, but I'd rather be able to wear it over a waistcoat or a vest and have to remove it when I put on a jacket than not be able to wear it over a vest.

Anewbe:
That's not a bad reason, honestly, but I don't want this to be wearable over space suits or other ridiculous things.
This is just me throwing up an idea, but it might make more sense to turn the suits (vests and whatever comprises "the like") that would logically be wearable under a suit jacket into accessories, rather than make this sort of blanket change to the suit jackets themselves. It makes sense that a poncho fits over spacesuits, because ponchos are basically big blankets with holes in them, but a suit jacket tends to be a bit more carefully made and fitted... Also gives the benefit of being able to stack the aforementioned items with actual jackets, for more options.

redstryker:
That's a good idea.  Writing that down so that I can get to doing it later.  Closing.

PR #2615, opened 2016-10-08
Batteries again

Anewbe:
Essentially adds another small battery for use in devices that aren't weapons. Has enough charge for 1~2 shots if it is used in weaponry, though.


Hubblenaut:
This does much more than you describe in your OP and commits

PR #2621, opened 2016-10-09
Medical Skeleton rename

Anewbe:
The grammar wasn't working and my usual method of correct it didn't work either, so the name is now in the description.


Neerti:
You can use the \proper macro so examine would go 'That's wilhelm' and not 'That's a wilhelm'.

Anewbe:
Alright, I was backwards then, standby.

Anewbe:
And done.

PR #2622, opened 2016-10-09
Changes the way embedding works

Anewbe:
Now we can make things never embed, by giving them an embed_chance of 0

### DIFF
--- a/code/game/objects/items.dm
+++ b/code/game/objects/items.dm
@@ -49,6 +49,8 @@
 	var/zoomdevicename = null //name used for message when binoculars/scope is used
 	var/zoom = 0 //1 if item is actively being used to zoom. For scoped guns and binoculars.
 
+	var/embed_chance = -1	//-1 makes it calculate embed chance, 0 won't embed, and 100 will always embed
+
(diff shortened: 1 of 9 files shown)


SpadesNeil:
Oh so like if I'm plinking arrows or chucking harpoons at someone, I can make them have a high embed chance.

Though does this take armor into account?

Anewbe:
It only affects melee weapons and projectiles, right now. Thrown items (including things from crossbows and pneumatic cannons) still work as they did, because that doesn't work quite the same way as the rest of the embed code, and is also fairly important to the usefulness of those weapons.

Melee weapons and projectiles do take armor into account.
Melee armor has a chance to 'blunt' the attack, increasing the amount of damage the attack needs to do before it can embed, and decreasing the chance it has of embedding to do so.
Projectiles (meaning bullets, energy projectiles still don't embed) felt like they were embedding through armor too often (not just my opinion), so there is now a secondary armor_check that can prevent the bullet from embedding. Armor_penetration can lessen this, and armor_penetration at 100 makes armor_check return 0, last I checked. Things that have good AP should still embed about the same, things with low or zero AP will embed less frequently, depending on the armor of the target.

SpadesNeil:
This is good. I like this.

PR #2631, opened 2016-10-12
Bloodloss balancing

MagmaRam:
Blood loss would depend on wound type and the limb it's on.

The "baseline", where the blood loss is the same as current, is a cut on the chest. Piercing wounds (a few two-handed weapons and gunshot wounds) bleed more, and bruises bleed less. Limbs and extremities bleed less.

Also, bruises now bleed at 50 damage, rather than 30.

Hopefully, this will keep bleeding fairly scary for wounds where the real danger is bleeding out (being shot or stabbed) while making you less likely to bleed out from being slapped on the hand a lot.


code/modules/organs/blood.dm
```diff
@@ -125,11 +125,23 @@ var/const/BLOOD_VOLUME_SURVIVE = 40
 
 		//Bleeding out
 		var/blood_max = 0
+		var/blood_loss_divisor = 30	//lower factor = more blood loss
+		//This 30 is the "baseline" of a cut in the "vital" regions (head and torso).
 		for(var/obj/item/organ/external/temp in organs)
```
Yoshax: You could possibly cut down on this a lot by changing it to `for(var/obj/item/organ/external/temp in bad_external_organs)`

code/modules/organs/blood.dm
```diff
@@ -125,11 +125,23 @@ var/const/BLOOD_VOLUME_SURVIVE = 40
...
 				continue
 			for(var/datum/wound/W in temp.wounds)
 				if(W.bleeding())
+					if(W.damage_type == PIERCE) //gunshots and spear stabs bleed more
+						blood_loss_divisor -= 5
+					if(W.damage_type == BRUISE) //bruises bleed less
+						blood_loss_divisor += 5
+					//the farther you get from those vital regions, the less you bleed
+					//depending on how dangerous bleeding turns out to be, it might be better to only apply the reduction to hands and feet
+					if((temp.organ_tag == ARM_LEFT) || (temp.organ_tag == ARM_RIGHT) || (temp.organ_tag == LEG_LEFT) || (temp.organ_tag == LEG_RIGHT))
```
Yoshax: It seems really kinda abstract to assume the bleeding reduces. For bruises and such it makes sense, but for the less a lot less so. It would perhaps be more interesting to keep it as is here for bruises but then add the possibility for piercing/cuts to not get reduced, through a `prob` of some sort, because there should still be a chance for other limbs to bleed a lot. It isn't really believably that bleeding just reduces on different limbs.

code/modules/organs/blood.dm
```diff
@@ -139,9 +151,10 @@ var/const/BLOOD_VOLUME_SURVIVE = 40
 						var/min_eff_damage = max(0, W.damage - 10) / 6 //still want a little bit to drip out, for effect
 						blood_max += max(min_eff_damage, W.damage - 30) / 30
```
Yoshax: The 30s here use the `blood_loss_divisor`, if you're making that a var you should probably make them use it too in case it's changed to keep them the same.

code/modules/organs/blood.dm
MagmaRam: Leaving bleeding partially up to RNG was one of the concepts I spitballed around a bit before doing this, but I decided I'd rather not do that after talking with a couple people about it. Leaving lethality up to RNG isn't very fun, and ends up giving Medbay something to tussle with that they can't actually measure in any way.

code/modules/organs/blood.dm
MagmaRam: Good idea.

code/modules/organs/blood.dm
```diff
@@ -125,11 +125,23 @@ var/const/BLOOD_VOLUME_SURVIVE = 40
...
 				if(W.bleeding())
+					if(W.damage_type == PIERCE) //gunshots and spear stabs bleed more
+						blood_loss_divisor -= 5
+					if(W.damage_type == BRUISE) //bruises bleed less
+						blood_loss_divisor += 5
+					//the farther you get from those vital regions, the less you bleed
+					//depending on how dangerous bleeding turns out to be, it might be better to only apply the reduction to hands and feet
+					if((temp.organ_tag == ARM_LEFT) || (temp.organ_tag == ARM_RIGHT) || (temp.organ_tag == LEG_LEFT) || (temp.organ_tag == LEG_RIGHT))
+						blood_loss_divisor += 5
+					if((temp.organ_tag == HAND_LEFT) || (temp.organ_tag == HAND_RIGHT) || (temp.organ_tag == FOOT_LEFT) || (temp.organ_tag == FOOT_RIGHT))
```
Yoshax: Should also be an else if, same as above. If it's arm or left, it can't be hand or foot.

code/modules/organs/blood.dm
Yoshax: Yeh, that's fair enough, although it does kinda just further establish the metagame of only aiming for the chest or the head further. You could make it still RNG, but not measurable RNG by doing something like:

```					if((temp.organ_tag == ARM_LEFT) || (temp.organ_tag == ARM_RIGHT) || (temp.organ_tag == LEG_LEFT) || (temp.organ_tag == LEG_RIGHT))
						if(!prob(W.damage && W.damage_type != BRUISE))
							blood_loss_divisor += 5
					else if((temp.organ_tag == HAND_LEFT) || (temp.organ_tag == HAND_RIGHT) || (temp.organ_tag == FOOT_LEFT) || (temp.organ_tag == FOOT_RIGHT))
						if(!prob(W.damage && W.damage_type != BRUISE))
							blood_loss_divisor += 10```

Where higher damage non-bruises on external limbs have a higher chance to not be reduced in bleeding, meaning big serious cuts or such on your arm or foot would still have a serious chance of being an issue.

code/modules/organs/blood.dm
Techhead0: The other thing you can do is `if(temp.organ_tag in list(HAND_LEFT, HAND_RIGHT, FOOT_LEFT, FOOT_RIGHT))` for brevity.

MagmaRam:
this is slowly becoming the pull of theseus

Neerti:
Magma tells me this is ready to go, so here's to nothing.

PR #2644, opened 2016-10-16
Introduces Europa-style SQLite Databasing (WIP/REVIEW/DON'T MERGE/ETC)

Neerti:
Adds several features which rely on byond's native SQLite support, which include bans, players joining, deaths, and a new 'playtime' statistic that will be explained later below.  SQLite, unlike MySQL, does not require a separate server process, and instead is a thousand times simpler to set up, only requiring some config.txt setup.

Each part of the following can be turned on or off in config.txt
**It is suggested that downstreams wait awhile before adopting this, in case mistakes are made.**

# Ban Table
Ports and tries to finish Zuhayr's ban datum system.  When the server starts up, it queries the SQLite database and creates a ban datum for all the bans recorded in the database, with all the information held on the datum matching the row on the database it was sourced from.  These datums are kept in global lists, and are checked against when the server wants to know if someone is banned.  Functions like expired() and lifted() make it easy to see if someone should still be banned or not.

The biggest changes between this and the MySQL system, besides the fact that it is SQLite, is that bans and jobbans are the same 'type', with the system knowing that it is a jobban based on if the job column is filled or not.  Likewise for permabans, it knows it is a permaban if the expiration_datetime column is empty.

# Death Table
This is more or less the same as the MySQL version, however I added a few new columns including cloneloss and the dying player's species.

# Player Tracking
More or less the same.

# Playtime
Brand new table and system which tracks how long someone has played in a specific department.  This can optionally be used to restrict access to a specific job until the player has played for a specific amount of time, which has the goal of at least guaranteeing that whoever is playing Head of Security has played a few rounds of security already.

The game leverages the job-guessing function I added for the new event system to also try to accurately determine what department someone is playing, in order to increment the correct department.  The minutes get written to the database file every ten minutes, but can write to it immediately if pressing need exists by using an argument in the increment proc.  This should be used if, for example, the client is about to be deleted, or the round is ending.

The system will not award minutes to mobs that are dead or have been AFK for five or more minutes.

The playtime tracking system can be set in the config, and enforcement of jobs using it is a separate option in the config.  It's recommended to leave enforcement off for about a month, so that ideally everyone who is active will get enough minutes to not notice the new restrictions by then.

# Misc
Adds some new helpers, to display a 'shorthand' name for a mob type, which might be useful for admin UIs (e.g. it can display AI instead of /mob/living/silicon/AI, which is preferred when space is a concern).

# Todo
- [x] Make borg modules have their minutes go to the associated department of the module (e.g. sec module = security)
- [x] Clean up lots of comments.
- [x] Make sure no world << creeps in again.
- [x] Figure out what to do with IA.
- [x] Optimize ```sqlite_load_bans()```, because it causes my client to hang for five seconds on a table containing 30 entries, which is quite bad.
- [x] Get rid of debug verbs and move the category for the ban panel to admin.  Make sure to uncomment the ```else``` statement in the admin verbs area.


code/modules/admin/topic.dm
```diff
@@ -813,7 +916,7 @@
 		var/mob/M = locate(href_list["newban"])
 		if(!ismob(M)) return
 
-		if(M.client && M.client.holder)	return	//admins cannot be banned. Even if they could, the ban doesn't affect them anyway
+//		if(M.client && M.client.holder)	return	//admins cannot be banned. Even if they could, the ban doesn't affect them anyway
```
Neerti: Note to self: Remove comment when finializing.

code/modules/sqlite/playtime.dm
```diff
@@ -0,0 +1,129 @@
+/client/proc/log_playtime_to_sqlite(var/department, var/amount)
+	establish_sqlite_connection()
+	if(!sqlite_db)
+		return null
+	if(!config.sqlite_playtime) // Tracking disabled.
+		return
+	if(department == 1)
+		CRASH()
```
Neerti: Todo: Remove

Neerti:
This should be ready for review now.

code/controllers/Processes/playtime.dm
```diff
@@ -0,0 +1,10 @@
+/datum/controller/process/playtime/setup()
```
Chnkr: why use a process for this? Just store the `world.time` once their mob has spawned, and compare that to the `world.time` if they stop playing, disconnect, or die.

Neerti: Someone could change jobs or go AFK during the middle of the round, and having a process ensures that the time recorded is a relatively accurate measurement of someone's time in a department.

PR #2647, opened 2016-10-16
Flashlights use device cells

Anewbe:
Flashlights use device cells for power. Should have roughly the same amount of use time.
Adds a chance for device cells to spawn where regular cells do.

### DIFF
--- a/code/game/objects/items/devices/flashlight.dm
+++ b/code/game/objects/items/devices/flashlight.dm
@@ -11,7 +11,7 @@
 	var/on = 0
 	var/brightness_on = 4 //luminosity when on
 	var/obj/item/weapon/cell/cell
-	var/cell_type = /obj/item/weapon/cell/high
+	var/cell_type = /obj/item/weapon/cell/device
 	var/list/brightness_levels
 	var/brightness_level = "medium"
 	var/power_usage
@@ -27,7 +27,7 @@
 
 		if(cell_type)
 			cell = new cell_type(src)
-			brightness_levels = list("low" = 5, "medium" = 10, "high" = 20)
+			brightness_levels = list("low" = 0.25, "medium" = 0.5, "high" = 1)
 			power_usage = brightness_levels[brightness_level]
 
 	else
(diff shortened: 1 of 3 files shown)


redstryker:
This would raise the concentration of device cells onstation.  There really shouldn't be that many.

Neerti:
Why not?  The device cells inside the flashlights are of much lower capacity than the weapon cells found inside lasers.  I'm told you would only get one or two shots if you tried to load it with a flashlight battery.

redstryker:
It still raises serious issues with balance.

Yoshax:
It needlessly complicates and obfuscates thing, there really doesn't need to be 3+ types of cells. While it may not be entirely believable that a 'big' cell fits into a small flashlight or what not, it's simplicity for the game and players sake.

Neerti:
@redstryker You only get one lethal shot with a non-laser carbine gun, and two with a carbine.  Combine that with having to play inventory tetris, the delay in swapping the cells while having to remain still, and doing the last two while the other guy is shooting at you with a gun loaded with real ammo, swapping to a flashlight battery is probably a bad idea.  This doesn't even take into account that the guy could dodge, RNG decides to make your single shot miss, or could be wearing armor.  If it somehow becomes a problem, we can address it later.

@Yoshax We actually have roughly ten types of cells currently, if you are referring to the absolute number of /obj/item/weapon/cell subtypes.  If you are instead referring to the implicit 'classes' of cells based on their physical size, that remains unchanged, as there remain two 'classes' of power cells, that being the large power cells that are usually depicted as being the size of a car battery, and the smaller cell, which is what Anewbe wants to put inside the flashlight.

The device cell and the weapon cell are part of the same group, as they're both the same size, which makes their difference inside the group a matter of capacity, which isn't really anything new when it comes to batteries, as the large cell group has 10k/15k/30k subtypes.  Adding one new type to the small cell group, which currently only contains weapon cells, doesn't really seem like a leap in complexity to me.

Right now, the power cell tree is generally this, ignoring batteries no one really cares about like the potato battery;
- Big cells (APCs, RIGs, Mechs, and everything else)
  - 5k (apc)
  - 10k (high)
  - 15k (super)
  - 30k (hyper)
- Small cells (Flashlights, Energy guns, stunbatons, possibly more in the future)
 - 2400 (weapon)
 - 480 (device)

I don't see how adding another type to the small cell group complicates or obfuscates anything.  At all.  In fact, there are several "big" cells that we could replace with device cells, them being ```/obj/item/weapon/cell/crap``` and ```/obj/item/weapon/cell/secborg```, which arguably would make things simpler, not more complicated.

PR #2651, opened 2016-10-17
Ling stings and absorb are logged

Anewbe:


MagmaRam:
thank you jesus

Neerti:
I think there's an attack log helper proc that could be used here.

Anewbe:
Code's a tad cleaner, and the shriek powers will be logged as well.

Yoshax:
You could make a generic sting logging proc, and pass the name of the sting into it and use that, rather than having it done in each sting. Would mean for new stings you would only have to call the helper proc. It would mess up the tense on one of three of the logs, but it's better than having to do it all over.

code/game/gamemodes/changeling/powers/shriek.dm
```diff
@@ -97,8 +103,12 @@
...
 
+	src.attack_log += text("\[[time_stamp()]\] <font color='red'>Used Dissonant Shriek.</font>")
+	message_admins("[key_name(src)] used Dissonant Shriek ([src.x],[src.y],[src.z]) (<A HREF='?_src_=holder;adminplayerobservecoodjump=1;X=[src.x];Y=[src.y];Z=[src.z]'>JMP</a>).")
+	log_game("[key_name(src)] used Dissonant Shriek.")
+
 	for(var/obj/machinery/light/L in range(5, src))
 		L.on = 1
 		L.broken()
-	empulse(get_turf(src), range_heavy, range_light, 1)
+	empulse(get_turf(src), range_heavy, range_light, 0)
```
Yoshax: Isn't noted as a change that I can see.

PR #2652, opened 2016-10-17
Lowers the dislocate chance and damage from disarm intent hits

Anewbe:


Yoshax:
So what sort of force/numbers does that change the stuff to? Right now a 15 force is a guaranteed dislocate, what does it look like after this?

Anewbe:
15 force takes 2~3 hits to dislocate, does about 7.5 brute per hit.

PR #2654, opened 2016-10-17
Should fix the tape thing.

Anewbe:


Neerti:
I can't find any issues, and it appears to work in testing.

PR #2655, opened 2016-10-18
Disarm spam is less effective

Anewbe:
There's a timer between chances to actually do anything with disarm intent.

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -8,6 +8,7 @@
 	var/list/hud_list[10]
 	var/embedded_flag	  //To check if we've need to roll for damage on movement while an item is imbedded in us.
 	var/obj/item/weapon/rig/wearing_rig // This is very not good, but it's much much better than calling get_rig() every update_canmove() call.
+	var/last_push_time	//For human_attackhand.dm, keeps track of the last use of disarm
 
 /mob/living/carbon/human/New(var/new_loc, var/new_species = null)
 
--- a/code/modules/mob/living/carbon/human/human_attackhand.dm
+++ b/code/modules/mob/living/carbon/human/human_attackhand.dm
@@ -260,7 +260,12 @@
 						visible_message("<span class='danger'>[src]'s [W] goes off during the struggle!</span>")
 						return W.afterattack(target,src)
 
+			if(last_push_time + 30 > world.time)
+				visible_message("<span class='warning'>[M] has weakly pushed [src]!</span>")
+				return
+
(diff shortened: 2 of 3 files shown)


SilverTalismen:
I'm all for this, ive watched people disarm spam shove people down so they could kill them and keep them down.

Yoshax:
Couldn't you just make disarm intent just have normal attack cooldown, since I'm pretty sure it doesn't.

Anewbe:
In testing it has the same delay as fists, from what I can tell.
Also wouldn't stop the main issue, which is just using disarm until you roll the lucky number and ends the fight. Actual delay for the push is TBD, though.

Yoshax:
It doesn't really resolve the issue. If you push someone down, you don't have to push them again. You just pin them down, and the fight is over.

Anewbe:
This makes it so if you don't knock them down on the first click the next several will be useless.

Yoshax:
I misunderstood the var name.

I'm not sure this is a good idea, disarm stuff can be a huge problem, but I do not believe that tacking on a six second cooldown is the way to resolve the issue. Six seconds is a lot of time.

Iamgoofball:
Completely kills hand to hand combat meta

SpadesNeil:
Disarm spam does piss me off to no end to be honest. I can often disarm a security officer faster than they can stun me.

You get one chance, I say. If you fail, you shouldn't have time to IMMEDIATELY try again.

Anewbe:
@Yoshax Yeah, I'm probably going to drop it to around 3~4 seconds, still need to discuss.

Yoshax:
You could do more interesting options than a straight cooldown such as dynamically calculating the chance to push based on multiple factors including a persons health, pain, mob size, etc. Make it a lower chance than now at perfect health conditions. 

Or introduce a counter using intents, like pushing a person who is on help intent would be easier than pushing someone who is on harm intent, and pushing someone else on disarm intent would be more difficult.

I feel as though a different solution than a flat cooldown would be optimal, something that adds to the feel and the flow of the game at the same time as resolving the issue.

Anewbe:
I could (and would like to) do that, at some point, but I don't have the time and energy right now to go and essentially rewrite hand-to-hand.

SpadesNeil:
3-4 seconds seems reasonable. Even 2 seconds.

Anewbe:
Time's dropped to ~3 seconds.

SpadesNeil:
I like this. 3 seconds is completely reasonable.

It negates the spamming abuse of disarm, while still allowing disarm to be combat effective.

PR #2658, opened 2016-10-20
Waistcoats are accessories

Anewbe:
Additionally, split a few ensemble uniforms into parts, a la the suit_jacket items. These are the detective outfits with waistcoats, and the gentleman's suit.


Yoshax:
You could make them both and retain the current features, whilst adding more.

Yoshax:
Also messes with chameleon kits which cannot into accessories.

Anewbe:
Would that work with the starting_accessories list?

Yoshax:
Not entirely sure. You could make the accessories wearable on the outer suit slot but have them defined as accessories by type, much like ponchos. As for the chameleon kit stuff, a chameleon tie or whatever would be trivial.

PR #2661, opened 2016-10-21
Adds shaved short hair and five o'clock shadow beard

Anewbe:
Fer yer head an' face.


Magneld:
And here I was thinking you were asking me if you should do this to yourself.

Anewbe:
I don't have enough tattoos for it to work, sadly.

PR #2672, opened 2016-10-23
Sec gear tweaks

Anewbe:
HoS coats have the same protection values as normal Sec gear
Detective gear is now as protective as the Forensics coat
Armored Trenchcoat sprite was tweaked slightly


ParadoxSpace:
Sounds controversial.

Anewbe:
It's sensible and I can't remember why I didn't already do this.

SpadesNeil:
Considering how big of a target the HoS is, are you sure that's a good idea to change the HoS coat? Maybe nerf it a bit but not make it as whimpy as regular security armor. I feel like that might cause balance issues.

The rest sounds fine though.

Anewbe:
The HoS really, really should not be on the front, they should be commanding their Officers, who should be on the front. Having that gear with that level of protection encourages being overaggressive, which in turn makes them a bigger/easier target.

ParadoxSpace:
Yes, but at some point they're going to be at the front. I'd personally leave it up to the forums or something.

SpadesNeil:
A fair point, @Anewbe ! I withdraw the statement.

PR #2681, opened 2016-10-25
Adds HUD Aviators

Yoshax:
Ports Baystation12/Baystation12#14433

Code's messier than Bay's cause we don't have whatever makes their eye handling stuff so pretty, and didn't feel like finding, porting and testing it for this little thing.


Anewbe:
Concept's good, code looks good, one quick query: Would it be feasible to repath them as sechud/aviator , rather than sechud/toggle?

Yoshax:
Done.

PR #2682, opened 2016-10-25
NO_POISON or IS_PLANT now grant hallucination immunity

Yoshax:
Also stops hallucinations from preventing passing out from halloss.

### DIFF
--- a/code/modules/mob/living/carbon/human/life.dm
+++ b/code/modules/mob/living/carbon/human/life.dm
@@ -902,7 +902,7 @@
 			Paralyse(3)
 
 		if(hallucination)
-			if(hallucination >= 20)
+			if(hallucination >= 20 && !(species.flags & (NO_POISON|IS_PLANT)) )
 				if(prob(3))
 					fake_attack(src)
 				if(!handling_hal)


Anewbe:
Not certain how this stops the halloss thing, but my bigger question is: What is the reasoning behind this change?

Techhead0:
The halloss thing is a leftover message from when I made this commit on Bay. It doesn't apply here.

Yoshax:
Woops, thought I had removed the holdover message, guess not. Was already fixed here/never like that.

This change is quality of life, and sensible. Mostly for Diona. They can't wear mesons, or anything of the sort that stop hallucinations from the Supermatter, and hallucinate. They don't take radiation damage, so don't need the radiation suit, but can't protect against the hallucinations that really make no sense affecting them in the first place. They live and space, around stars and all that fun, they really shouldn't be experiencing hallucinations from the Supermatter, nor at all. It's very unlikely that a species so different and incomprehensible experiences hallucinations at all, never mind the same as humans.l

Spookerton:
Sounds fine to me.
👍

PR #2688, opened 2016-10-27
Allows everyone to get Sunglasses/Aviators

Mark9013100:
- Everyone can choose Sunglasses, Prescription Sunglasses, Aviators, and the IA Sunglasses.
- Removes the fake Sunglasses/Aviators.
- Removes an un-needed allowed_roles line.

### DIFF
--- a/code/modules/client/preference_setup/loadout/loadout_eyes.dm
+++ b/code/modules/client/preference_setup/loadout/loadout_eyes.dm
@@ -46,7 +46,6 @@
 /datum/gear/eyes/secaviators/prescription
 	display_name = "Security HUD Aviators, prescription"
 	path = /obj/item/clothing/glasses/sunglasses/sechud/aviator/prescription
-	allowed_roles = list("Security Officer","Head of Security","Warden")
 
 /datum/gear/eyes/medical
 	display_name = "Medical HUD (Medical)"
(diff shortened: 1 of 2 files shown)


Fox-McCloud:
This is pretty much any powergamers dream, right here--as it's one of the two things needed for you to be half immune to flashbangs and fully immune to flashes (the dominant incap method of heads of staff).

Mark9013100:
the horror of eyewear

redstryker:
You do realize that there are moderators for a reason, right?  Powergamers usually do other things besides wearing sunglasses, so it can be detected.  Just because someone can potentially use sunglasses to stop from being stunned does not mean that we should all-out restrict them from use by normal people.

SpadesNeil:
Could just make glasses not protect against pepper spray.

It sure doesn't work IRL, why does it work here?

Fully shielded goggles should protect your eyes, but not freakin' sunglasses. We should change how pepper spray works with glasses.

Fox-McCloud:
Well, there's really not much point to wearing non-cosmetic sunglasses aside from powergaming, to be honest.

Cosmetic sunglasses get the RP point across without impacting gameplay (and these are currently an option for players to utilize and select), giving them real sunglasses serves no additional roleplay point other than giving an inherent mechanical advantage to the player (flash immunity and less eye damage from welding).

That's ultimately the point I'm making; there's currently cosmetic shades--which allow you to look cool or wear their for RP purposes ("bright lights give me a headache"), but don't have that powergamey edge.

Anewbe:
We discussed this in our November 6th staff meeting, the general opinion was mostly negative where it was not entirely neutral. Fox-McCloud made most of the points I was going to make, beyond that.

Anewbe:
Been a week with no answer, closing it.

PR #2690, opened 2016-10-28
Adds another mob_size class

Anewbe:
Adds MOB_HUGE, which is equal to the old MOB_LARGE. New MOB_LARGE is size 30. Juggernauts are still the same size (now MOB_HUGE), while construction drones remain MOB_LARGE.


Anewbe:
Self-merge due to lack of devs

PR #2692, opened 2016-10-28
Species var to ignore the effects of item slowdown

Anewbe:
slowdown_fixed = 1 means the species will be unaffected by speed boosts or slowdown from items.


Anewbe:
Self-merge due to lack of devs

PR #2694, opened 2016-10-29
Fixes some surgery bugs

Anewbe:


code/modules/surgery/surgery.dm
```diff
@@ -3,6 +3,8 @@
 /datum/surgery_step
 	var/priority = 0	//steps with higher priority would be attempted first
 
+	var/req_open = 1	//1 means the part must be cut open, 0 means it doesn't
```
comma: Stuff like that is exactly what can_use proc of surgery steps is for :(
Like here https://github.com/Anewbe/Polaris/blob/18b15163f6ca70b2e84f53355b854a020e337983/code/modules/surgery/organs_internal.dm#L13

Anewbe: That's part of surgery__step/internal, I need it to work on every surgery.

PR #2706, opened 2016-11-08
Surplus crates will no longer make more surplus crates

Anewbe:
The chance was fairly small, but it still happens, and that's not desired behavior.

### DIFF
--- a/code/datums/uplink/badassery.dm
+++ b/code/datums/uplink/badassery.dm
@@ -3,6 +3,7 @@
 ************/
 /datum/uplink_item/item/badassery
 	category = /datum/uplink_category/badassery
+	blacklisted = 1
 
 /datum/uplink_item/item/badassery/balloon
 	name = "For showing that You Are The BOSS (Useless Balloon)"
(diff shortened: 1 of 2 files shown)


Yoshax:
There's a blacklisting feature. Use that.

Anewbe:
What would that look like?

Yoshax:
Like this should do it: https://github.com/PolarisSS13/Polaris/blob/master/code/datums/uplink/announcements.dm#L6

Anewbe:
Thank you.

Yoshax:
You should add it to Random Items as well, it does not inherit from single random item, and as such, surplus crates will be able to order many random items.

Anewbe:
Excluded all Badassery items, don't see a reason to have balloons spawn in crates.

Yoshax:
Fine assuming it obeys the blacklist, see no reason it shouldn't. In the logs from the round there were instances of blacklisted items being bought, but I believe that's from the legacy random items purchasable thing that doesn't adhere to the blacklist.

ParadoxSpace:
I don't see why it should be removed if it's such an infinitesimally small chance.

Anewbe:
Staff doesn't want that chance to be a thing, you already get a lot of stuff, cost-wise, from a single crate.

PR #2707, opened 2016-11-08
Fixes borg shocks

Anewbe:
Alright, this should prevent doors and wires from shocking silicons, but the antag abilities that call electrocute_act will still damage them.


Yoshax:
You could define the 0.75 somewhere, and plug it in as a var rather than defining it each time. Means you could adjust it later for all the things while only changing one.

Anewbe:
I'm not entirely sure why I would want to do that.
The var is defined at 0.0 in silicon's electrocute_act so every other call of that proc won't damage them. I'm not certain that changing it to the way you've suggested is easier than adding in the var for the offensive abilities, which, from my viewpoint, are exceptions to the way they're normally shocked.

PR #2708, opened 2016-11-08
Allows more jobs to become certain antags after the round starts

Anewbe:
Adds a system to allow culting/revconverting of many jobs that cannot be converted now.


Yoshax:
You could save a lot of code duplication by doing like...

```/datum/antagonist/proc/draft_antagonist(var/datum/mind/player)
	if(!can_become_antag(player) && !player.assigned_role in restricted_jobs)```

Or so, but it skips the ignore role var, I'm not sure that's used though.

PR #2710, opened 2016-11-08
Fix world/Topic()'s manifest to include synthetics like datacore.dm

Leshana:
* The world/Topic() bot integration API's manifest output uses the same logic as datacore's get_manifest(), except it omits the part that includes synthetics.  The `set_names` variable clearly shows bots are intended to be included, they are just arent.
* Copying the logic from datacore.dm ensures a consistent view of bots and AI.

### DIFF
--- a/code/world.dm
+++ b/code/world.dm
@@ -224,6 +224,19 @@ var/world_topic_spam_protect_time = world.timeofday
 					positions["misc"] = list()
 				positions["misc"][name] = rank
 
+		// Synthetics don't have actual records, so we will pull them from here.
+		for(var/mob/living/silicon/ai/ai in mob_list)
+			if(!positions["bot"])
+				positions["bot"] = list()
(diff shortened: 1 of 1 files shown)


Anewbe:
So in layman's terms this puts the AI and Borgs on the PDA manifest, or what?

Also, going to close and reopen this to get Travis running.

Leshana:
As far as the manifest familiar to most players (the one in the lobby or the one available to ghosts) it changes nothing, because AI and Borgs were *already* on that manifest.
This also does not change the PDA manifest.  I have not checked if AI or Borgs show on that.

What this does is add AI and Borgs to the IRC API manifest endpoint, just updating it to make it consistent with the in-game manifests.

Note: The game code exposes a "web" service API so that external programs can access some data about the game, such as round status and manifest etc.  This has been in the code for years, but it was out of sync with some changes made elsewhere.  This PR just fixes that.
I only did it because on my server we actually use the API to show the manifest on our website.

PR #2713, opened 2016-11-09
Switch floodlights to smaller battery item

Arokha:
It was already just editing the charge to the exact same values as the non-high power cell. Why give it a high-power cell? So silly. And misleading if you take the cells and put them other places.

Fixes https://github.com/VOREStation/VOREStation/issues/698
For you, it just makes things make actual sense.

### DIFF
--- a/code/game/machinery/floodlight.dm
+++ b/code/game/machinery/floodlight.dm
@@ -4,7 +4,7 @@
 	icon_state = "flood00"
 	density = 1
 	var/on = 0
-	var/obj/item/weapon/cell/high/cell = null
+	var/obj/item/weapon/cell/cell = null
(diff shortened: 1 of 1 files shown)


Nerezza:
For future reference I want to note that this was a pain for rookie miners in the past, who'd nab high cap cells from floodlights to use in their mining drills. Except, they'd turn out not to be high capacity high cap cells, but low capacity high cap cells...

Arokha:
Never again shall the tyranny of improperly named batteries repress the miners.

Nerezza:
Unless you get your batteries from the Engi-vend. Its high caps seem to be lower capacity than high cap too. I really don't know how this stuff keeps happening, I'm poking at this one though =3=;;

PR #2717, opened 2016-11-11
Concealed holsters, suits hiding accessories.

MagmaRam:
Two new things that suit-slot clothing can hide - accessories and holsters. Hiding jumpsuits will hide all accessories, and hiding accessories will hide holsters, but I've specified them for clothing that hides jumpsuits, too.


Anewbe:
Okay, I need to read through this in depth, but I think it might be best to just add a variable to accessories that can be hidden, like var/hideable, and have that switched on those two holsters, rather than making a visible holster subclass.

MagmaRam:
It is, accessories with concealed_holster set to 1 will be invisible when a holster-concealing suit is worn. The visible holster subclass is all the holsters that are not concealable under a coat or other suit, since the default holster object is concealable.

Anewbe:
Why does it need to be a subclass? You can just make the var = 1, no?

Nerezza:
You work fast Magma, holy crap! Is there a way that security can search someone for a concealed holster? Other than strip searching, that is.

code/game/machinery/cryopod.dm
```diff
@@ -410,6 +410,13 @@
 				preserve = 1
 				break
 
+		if(istype(W,/obj/item/weapon/implant/health))
+			for(var/obj/machinery/computer/cloning/com in world)
+				for(var/datum/dna2/record/R in com.records)
+					if(R.implant == W)
+						qdel(R)
+						qdel(W)
+
```
Anewbe: Why was this changed?

MagmaRam: That was something that I didn't mean to commit, is why.

Anewbe: Thought so

PR #2718, opened 2016-11-11
Adds teargas grenades, currently aspawn only

Anewbe:
Ported from Baystation

### DIFF
--- a/code/game/objects/items/weapons/grenades/chem_grenade.dm
+++ b/code/game/objects/items/weapons/grenades/chem_grenade.dm
@@ -278,3 +278,26 @@
 		beakers += B1
 		beakers += B2
 		icon_state = initial(icon_state) +"_locked"
+
+/obj/item/weapon/grenade/chem_grenade/teargas
+	name = "tear gas grenade"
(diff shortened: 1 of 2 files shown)


Yoshax:
Did you try using these and testing the effectiveness?

I previously have multiple times, and never was able to have something suitable and appropriate.

Anewbe:
That's why they're aspawn only, it will let us test and tweak before it gets into common use.

Yoshax:
Well, in my experience and testing they do nothing when actually exposed to them until about 60-120 seconds afterwards, wherein your pain builds up till you're slow as hell, which takes another 60-120 second, then after a little bit more you're dropped into paincrit, for about... 400+ seconds.

Anewbe:
Yep.

Yoshax:
That's kinda ridiculous to put in even as aspawn only.

ParadoxSpace:
Pls tweak

MagmaRam:
These aren't going to be added to the proper game until they're properly tweaked and tested, and this'll help with getting that done.

PR #2720, opened 2016-11-11
Defibrilator Kit

Arokha:
Adds a defib kit. Not ported from /tg/, not a giant backpack. It's more like an AED than anything. ~~Click with empty hand or drag onto yourself to take the pads out, and put them into some dead person's uniform slot.~~ Drag the kit onto someone to begin applying the pads.

Currently has ~~3~~ 2 attempts with 75% chance of success with the default battery, as long as they've been dead for less than 10 minutes. Compare to Technomancer revive of automatic success within 30 minutes. Replace battery with screwdriver. Uses a device-battery.

The ~~pads are~~ kit is cabled, so if you walk away with them it yanks the ~~cable out~~ pads off and you have to reconnect them. Patient can't be moving either, and it won't shock living people (someone can code an emag thing for that I guess?).

**(Does not include any map changes that would actually spawn it anywhere. You'll have to do that. If you want, anyway.)**

defib kit example


Anewbe:
I'm not personally opposed, but I'll need to talk to the other staff.
Additionally, for this I'd actually prefer the regular cell type to be use, unless you've got a really specific reason not to.

Arokha:
Didn't have a real reason for the device cell other than it being a device. I mean mostly to prevent people from putting a hyper-cell in it and getting 30+ uses in one cell or whatever. But if that's what's desired, it's changed now.

Anewbe:
Well, that's not a bad reason, actually, but remember that there's a bigger device cell, too.

Arokha:
> there's a bigger device cell

I wasn't aware. Do you mean the 'weapon' one? That's the only child I could find.

Anewbe:
Yeah, but it's 5 times bigger, if memory serves.

Arokha:
Yeah, 'tis. Well, doesn't really matter then I suppose. But, it's using the chain of normal cells now.

Nerezza:
I'm not staff, but I mentioned that we should have defibs a couple days ago in ghost-chat, so I want to toss my input in real quick.

My primary reasoning is that with defibs we'd no longer have to clone someone right off the bat, I think this would greatly help roleplay flow.

Yoshax:
Further trivialises death. Extremely magic and silly. Serve no purpose while cloning exists.

Arokha:
I'm a little confused as to why you'd call an extremely common real-life object that virtually every office building has `extremely magic and silly`.

Yoshax:
Because a defibrillator does not do anything close to what it does here. A defibrillator is used to recover a patient from cardiac arrest by forcing the heart back into a normal, regular rhythm. They don't bring people back to life, especially not people who've been dead for ten minutes. They are common, yes, and important and useful, but they do not bring people back to life, especially not in the manner here.

Arokha:
What you've just said is incorrect, though. Defibrillators can work fine after ten minutes if people have been performing CPR properly. Patients would suffer brain damage after five or so if left alone, but CPR can prolong that quite a bit. The ten minute limit I used is to offset the fact that there's no actual pulmonary system in SS13 that can really simulate the forced circulation from CPR, by adding five minutes.

In terms of 'bring people back to life', I don't quite get your reasoning. This doesn't heal people, if they died due to blood loss, toxin, brute, burn damage, etc they will just convulse and immediately be dead again. So as far as a user is concerned, this is useful in about the same situations.

Yoshax:
Then in what is situation is it actually meant to be useful? I struggle to come up with any scenario where the dead person doesn't just die again due to having already died.

Arokha:
Suffocation, for example, due to space or other things, but also there are situations in which you could potentially correct whatever caused the person to die while prepping the defib. Starting an IV drip (in our case, blood, though often just saline in real life) to increase blood pressure. Stopping bleeding to prevent further blood loss on being revived, etc.

Basically if they died from oxyloss and their brute+burn is under 200 then you can maybe save them with it. If their brute+burn is > 200 then you probably can't, will likely take too long. If their brute+burn is ~100 and oxyloss is ~100 due to bleeding or something you can shoot them up with drugs and try it and hope for the best.

Arokha:
I should also note this already exists as the technomancer revive spell. I mean, I copypasted the code from that. So... useful in about the same situations!

Yoshax:
I really don't see the need, or use for them when cloning exists. Also not a fan of the way of using them, it's pretty clunky and not clear.

Arokha:
None of the medical equipment on the station comes with directions.

If you really want, I can add voice prompts, though.

Nerezza:
I'll kinda go through chronologically here with my counterpoints:

First off cloning itself is unrealistic, which I feel puts it on-par with this defib setup in that regard. However, the problem I personally have with it is it's too perfect a game mechanic, with defibs we have an option that's less perfect where people revived with them will possibly suffer from their near-death experience. (missing limbs and such)

The decision to attempt defib will be up to the doctors. "Can we recover the body enough for the patient to survive if we restart their heart?" would become a new question to ask themselves. Cloning would have less consequences, yes, but this brings about a new alternative for people who are DNC and if defibs stick, we could redevelop the general stigma for cloning we used to have because there'd no be an alternative.

As for directions, first as said above the medical equipment doesn't come with in-game directions. Second off, voice prompts would be okay but if we want to make this a doctor-only thing (imo we should) then it can be assumed ICly that the doctors have training and don't require more than the Green/Yellow/Red indicator light on it already. If we want to make this a flat-out AED, yeah it'd be good to have voice prompts so anyone in the crew can use them but I think that's going too far for the first implementation.

Yoshax:
Most of the stuff is pretty obvious to use, and follows the common use of stuff ingame. You click people to use stuff on them, applies to almost everything in medical. Or you click drag onto them. This kinda breaks all of that and does something different. The issue with the usage is it follows none of the common ways things are used in game.

My issues lies not with the concept, or use, or any of it in isolation. It's with it existing when cloning exists. There's no reason to use this when cloning exists. It's just a waste of time trying to use it, or determine if you can, or it's worth, or anything of the sort when you can just ding dong clonnu.

Arokha:
Saying 'whatever just clone them it's mechanically better and easier' is tantamount to metagaming. It's supposed to be HRP.

Yoshax:
Perhaps that's so. But a huge portion of the game boils down to that. And not something that's exactly easily enforced, it's kinda difficult to go, "Oh, hey, you didn't try to defib that person, ding dong bannu." I don't like this for a multitude of reasons I've previously listed.

Arokha:
I never suggested you ban people for not performing some medical thing. If that were the case, I'd suggest banning everyone who doesn't make the perfect cryomix. That's more banning people for NOT metagaming. This is just another tool for people to use, if they want, as part of the medical job. It can add urgency and drama, even if someone finds a body, it gives paramedics or doctors a reason to rush and check to see if they can do anything, which is realistic.

Nerezza:
The problem seems to boil down to cloning's existence, which is right now a stopgap game mechanic in everyone's mind I'm sure. I think this PR should focus on getting defibs in, and then the overall medical system can develop from there.

We (and medical) will get more leeway to develop the system further (something AT pointed out in Discord) where, for example, cloning or FBP conversion can become a more extreme measure.

Screemonster:
I'm just gonna say that the "oh there's no reason to do this while cloning exists" line is kinda silly, 'cause the same could be applied to doing surgery or any number of medical procedures on the station where just letting the patient die and cloning them afterwards is easier than actually saving them.

Nerezza:
The thing is, that'd be euthanasia which is a no-no.

On the other hand, recently deceased are still dead so cloning would the already did is obviously not euthanasia. The mechanic is easier, and fluff-wise there's nothing made to give cause for using defibs.

We're going to have to make something up ultimately, but after we figure out how they're used in practice.

Arokha:
The legal definition of dead requires that their state be irreversible. So someone who recently stopped breathing is not legally dead.

Yoshax:
There's a difference. One is letting someone die, the other is reviving them in an inferior manner to what else you have available. One one hand, you're preventing something rather than curing it, which is preferable as per the Hippocratic Oath. There's no reason to revive someone and put them through more suffering instead of cloning them to perfect health.

```The legal definition of dead requires that their state be irreversible. So someone who recently stopped breathing is not legally dead.```
If that is the definition, then nobody in the SS13 world is ever dead, all death is reversible. Your point is moot.

Arokha:
Other than CMD? And what if they're DNC?

Yoshax:
From a pov, you could consider CMD to be a positive thing. People not remembering getting butchered is a positive thing, not bad. If they're DNC, which so few people are, they can still be FBP'd or Borg'd, or whatever. You also assume these DNC people wouldn't be DNR.

Arokha:
I do, yes, since many people's reasons for DNC is religious. At least that's what I'm told. Which makes sense to me. Typically people's reasons for DNR are different. Having worked at a long-term-care company for 10 years, I've heard the whole slew of reasons.

And YOU could. I'd rather remember. I wouldn't want to 'lose' an unknown amount of time from my life, and never know what it was. If you can't remember it, how could you know 'good thing, since it was terrible!' It'd just be blank.

Techhead0:
Honestly, I'd make them using them behave like IV drips with drag-to-attach.

The philosophical argument I'm leaving aside, with the exception of mentioning that I think these would be perfect for reviving people who suffocated in breaches/space.

Arokha:
> Honestly, I'd make them using them behave like IV drips with drag-to-attach.

I could do that if that's preferred. Just nix the paddles item altogether.

Nerezza:
I actually like the idea of drag-attach. The stripping interface is pretty clumsy, and we already nix the need to remove uniforms for stuff like surgery because of this.

Yoshax:
In my experience, a whole big portion of DNCs are people who just don't like cloning and stuff, and would rather be dead than cloned.

Gonna have to say I'm further against this than I originally was, mainly due to the possible powergaming issues caused by being able to potentially bring people back to life and have them go, "Urist McTraitor strangled me in the locker room."

Anewbe:
Those people will still be under the same rules as clones.

Nerezza:
In my experience, as someone who puts DNC on all my medical records I just don't like cloning and borgification. I'd rather be defibed than dead, though.

Arokha:
> In my experience, a whole big portion of DNCs are people who just don't like cloning and stuff, and would rather be dead than cloned.

... isn't this ... an argument FOR the defib? If they don't like cloning, that has nothing to do with their like of defib. This prevents them from every getting to the point where cloning would be needed so, ...

I mean I'm already doing this on our server, regardless. I just thought I'd put it here if y'all wanted it. And the drag-to-attach is a good idea so I can do that in a bit.

Woodratt:
Right, being discussed by staff. Most of our staff is interested in it however.

Magneld:
Just going to put here what I put in Staffland.
Ble bla bla defibs do not work that way etc etc. HOWEVER. Game. So sure why not, imagine it wont work on a person who's been burned up, shot up or has his head lying over there, or is otherwise obviously killed to bits by extreme violence so this seems pretty reasonable.

Arokha:
Can you mark DNM for me? I want to implement @Techhead0 's suggestion of the drag-to-attach like IVs but I don't have time tonight or maybe tomorrow.

Yoshax:
May want to use to_chat instead of usr << or suchlike, makes the move to Goonchat less work, if that's ever going to be a thing.

Screemonster:
People on /tg/ complained about the "aaaaa but now bombing genetics doesn't guarantee my target stays dead and they might get defibbed or this or that or..." thing too - for a very short while, before they found that if they're in a position to kill someone to death, especially to kill someone to death stealthily, it's a lot easier to render someone undefibbable than it is to render them unclonable. Even if it's just chopping their head off and tossing it out an airlock, or giving them a few extra bashes with the fire extinguisher to put them over 200 brute, or whatever other steps you'd go through to make sure they're not in a state that could possibly come back.

If Urist McTraitor can strangle you in the locker room, Urist McTraitor can stove your corpse's face in with his emergency oxygen tank.

Yoshax:
You shouldn't have to take the extra power gaming step to ensure that people won't be revived, especially because it opens a whole question of whether it is acceptable, and no doubt people will complain about.

Arokha:
Ok, that change is made. What's that do, exactly? It looks like to_chat is just a #define for <<

Nerezza:
I personally don't know enough to be for sure, but as said above the to_chat this is just in case of a move to goonchat, so perhaps if that ever comes the to_chat you used today will be replaced with a goonchat version, or it will be #define'd to its replacement.

Arokha:
Another PR that was merged modified device.dmi so I had to merge them manually, hence the commit.

Screemonster:
Nerazza, that was mostly in response to Yoshax's comment that traitors would be harmed by people getting defibbed and ratting out their killer, which is.. not something I think is gonna be an issue on either HRP or LRP. Especially given that simply successfully hiding the body for ten minutes will prevent defibbing.

Anewbe:
Tested, works fine, doesn't seem particularly strong as is. Merging.

Arokha:
Just remember it's not actually in the map anywhere.

PR #2722, opened 2016-11-11
Makes tox damage more dangerous

Anewbe:


MagmaRam:
How much tox does it take right now for the health indicator to go into yellow? It seems like tox doesn't give the player much indication that it's happening beyond vomiting, which I don't think happens at 30.

Anewbe:
Roughly 60, which is odd and probably needs to be changed.

MagmaRam:
Do you want to change that in this PR, or merge this now and change toxloss pain levels later?

Anewbe:
I'll edit this PR to give it a better warning.

SpadesNeil:
This makes toxin damage a little exceedingly deadly. I'm not in favor of it. A toxin level of 30 is very commonly exceeded just by drinking a little too much or taking a tad too much medicine, among other things. Making liver damage occur as a result is just punishing.

Nerezza:
Vomiting usually follows a lengthy period of nausea irl, so we could add messages about that in-game so people know they're ill.

As for the PR, 30 seems like too little a level, while 60 does seem like a little too much. 50 might be a small jump, but I think it'd be reasonable for toxins to start damaging the liver.

SpadesNeil:
50 I like.

Anewbe:
Threshold's back up to 50

PR #2723, opened 2016-11-11
Improves bandanas

Arokha:
No more tie verb. Just put it into whatever slot and it (correctly) adjusts the needed stuff to work. Added 'Can be worn on the head or face' to the description to indicate that.

Currently wearing a bandana on your head makes it impossible to eat, or put on a mask, which doesn't make sense, so this fixes that too.

Fixes https://github.com/VOREStation/VOREStation/issues/718

### DIFF
--- a/code/modules/clothing/masks/miscellaneous.dm
+++ b/code/modules/clothing/masks/miscellaneous.dm
@@ -173,62 +173,54 @@
 
 /obj/item/clothing/mask/bandana
 	name = "black bandana"
-	desc = "A fine black bandana with nanotech lining."
+	desc = "A fine black bandana with nanotech lining. Can be worn on the head or face."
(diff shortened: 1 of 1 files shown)


Arokha:
Wow git. You're really good at mangling my diffs.

Well, I replaced all the bandana logic with:
```
/obj/item/clothing/mask/bandana/equipped(var/mob/user, var/slot)
	switch(slot)
		if(slot_wear_mask) //Mask is the default for all the settings
			flags_inv = initial(flags_inv)
			body_parts_covered = initial(body_parts_covered)
			icon_state = initial(icon_state)
			
		if(slot_head)
			flags_inv = 0
			body_parts_covered = HEAD
			icon_state = "[initial(icon_state)]_up"
			
	return ..()
```
That way the stuff is done automagically when you put it in a slot.

PR #2737, opened 2016-11-13
More ap bullets

Anewbe:
- Adds more armor piercing bullets to the uplink.
- Adds in .45 caliber AP bullets, 20 damage and 15 armor pen
- Lowers the damage of the 9mm AP bullet to 15

### DIFF
--- a/code/datums/uplink/ammunition.dm
+++ b/code/datums/uplink/ammunition.dm
@@ -18,14 +18,26 @@
 	name = "Pistol Magazine (.45)"
 	path = /obj/item/ammo_magazine/c45m
 
+/datum/uplink_item/item/ammo/c45map
+	name = "Pistol Magazine (.45 AP)"
+	path = /obj/item/ammo_magazine/c45m/ap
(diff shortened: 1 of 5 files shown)


Yoshax:
The rest of the AP ammo on the uplink is the same price as its standard counterpart. Making the new ones cost more doesn't make much sense, nor does it costing more. It's not a straight upgrade, it's situational.

Anewbe:
It's fairly non-situational; most anyone you're pulling AP ammo to fight will be armored, and the damage decrease is 5 brute on every bullet type, so it's not really a massive difference.

Yoshax:
There's no reason for it to be more expensive, and you're making it less appealing. It was specifically made the same price to make it appealing to actually get. It's situational because you're only going to pull it against foes you expect to be heavily armored, otherwise normal ammo is fine. When your damage, for example, on the pistol ammo is 20 and decreased to 15, that's not 'not really a massive difference', it's 25%.

PR #2742, opened 2016-11-14
Slimes will now dump reagents they can't process.

### LINKED ISSUE #1876: Slimes cannot handle reagents not in their reagent list.

Slime reagent handling object's reagent get clogged when there's something not on the list in it.

To dump the chemical
#### What actually happened

The chemical clogged it
#### Steps to reproduce

Inject slime with chemical it can't handle, I.E. Hydrogen
Cry.
#### Additional info:

MagmaRam:
They won't throw up a runtime and then refuse to process more stuff.

Fixes #1876

### DIFF
--- a/code/modules/xenobio2/mob/xeno procs.dm
+++ b/code/modules/xenobio2/mob/xeno procs.dm
@@ -64,6 +64,10 @@ Procs for targeting
 
 			reagent_response = chemreact[R.id]
 
+			if(!reagent_response)
+				temp_chem_holder.reagents.clear_reagents()
+				return 1
+
 			if(reagent_response["toxic"])
 				adjustToxLoss(reagent_response["toxic"] * reagent_total)
 


code/modules/xenobio2/mob/xeno procs.dm
```diff
@@ -64,6 +64,10 @@ Procs for targeting
 
 			reagent_response = chemreact[R.id]
 
+			if(!reagent_response)
+				temp_chem_holder.reagents.clear_reagents()
+				return 1
```
Datraen: This will remove all reagents instead of properly disregarding erroneous ones.

An alternate approach is to move all of the if checks under an if(reagent_response) check, to prevent runtiming.

MagmaRam: I checked whether chems that actually had a function worked properly or not, and it appears that they did, reagent_response will only be null if the chem doesn't do anything to slimes.

Issue #2691, opened 2016-10-28
[Feature] Hyperzine could use some more downsides.


SpadesNeil:
#### Brief description of the issue

It's a little OP. "Hyperzine is currently only used to get an edge on time and other people / security. There's no reason not to get an IV of it watered down and drag it around with you."
#### What you expected to happen

Someone suggested, "Hyperzine has no downsides. Give it something like consistent brute damage over time and lower it's OD rate to five units. That way, it can still be useful without being openly used."

I think that's stupid tbh, but maybe consuming too much induces seizures! Stuns you, you fall over, and twitch violently on the ground. That is a huge problem in combat if you over-do it because suddenly all that mobility you had is gone, but it's not stupid now.
#### Additional info:
- **Server Revision**: https://github.com/PolarisSS13/Polaris/commit/4130719b3b5a6945bc0b337bf940f2dc561a9f22

Atermonera:
If hyperzine is too useful for combat, then a combat-related malus would make more sense than raw damage over time. Especially if you lower the OD threshold, if someone can get hyperzine, can't they also water it down with medications to heal damage caused by ODing, or whatever raw DOT you introduce?

SpadesNeil:
Oh! Maybe consuming too much induces seizures! Stuns you, you fall over, and twitch violently on the ground.

That is a huge problem in combat if you over-do it because suddenly all that mobility you had is gone.

Atermonera:
I was thinking accuracy penalties and misfires, but that could work too.

SpadesNeil:
![why not both](https://cloud.githubusercontent.com/assets/6484985/19832293/3164d868-9dee-11e6-851e-1a0c62478c27.jpg)

PR #2745, opened 2016-11-15
Adds Stimm, a homemade hyperzine; makes hyperzine toxic

Anewbe:
Stimm is 1 part sugar, 1 part fuel, with 5 fuel as a catalyst. Doesn't last as long as hyperzine, and it hurts, but it should be much easier to make.
Makes hyperzine lightly toxic

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Core.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Core.dm
@@ -162,7 +162,7 @@
 
 /datum/reagent/fuel/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
 	if(issmall(M)) removed *= 2
-	M.adjustToxLoss(2 * removed)
+	M.adjustToxLoss(4 * removed)
 
 /datum/reagent/fuel/touch_mob(var/mob/living/L, var/amount)
 	if(istype(L))
--- a/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine.dm
@@ -311,23 +311,6 @@
 		var/mob/living/carbon/human/H = M
 		H.update_mutations()
 
-/datum/reagent/hyperzine
-	name = "Hyperzine"
-	id = "hyperzine"
-	description = "Hyperzine is a highly effective, long lasting, muscle stimulant."
-	reagent_state = LIQUID
-	color = "#FF3300"
-	metabolism = REM * 0.3
-	mrate_static = TRUE
-	overdose = REAGENTS_OVERDOSE * 0.5
-
-/datum/reagent/hyperzine/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
-	if(alien == IS_DIONA)
-		return
-	if(prob(5))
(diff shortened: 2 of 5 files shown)


Screemonster:
Making hyper slightly toxic is a neat way to discourage lightning chemist roadrunner moments, but anyone with access to hyperzine (outside of zoom pills) can get dylov-

oh. OH. Just noticed the path change under /toxin. Outstanding.

Yoshax:
It doesn't really make it slightly toxic though, it's more like pretty highly toxic. But maybe that's just me. I believe this will make it a lot less used, not that it's used much currently.

Screemonster:
Eh, a full syringe of it will do 5%-ish of an absolute max of 45, so... 2.25 points of organ damage? Not really much more dangerous than getting wasted at the bar and if you can make hyper, you can pop a peridaxon pill once you're done with it.

I'm curious as to why it's organ damage and not straight toxloss though.

Yoshax:
It's not, it's both already. The ..() parent call calls the base toxin affect_blood which does `M.adjustToxLoss(strength * removed)`

Therefore each tick, or every second, I forget how oftten affect_blood happens, it will do a small amount of building toxloss, along with the organ damage chance. take_organ_damage() isn't internal damage anyway, it's external. It picks a random external organ and does `3 * removed` damage. 

It does damage, yes, and is the wrong choice to resolve a non-existent problem.

Anewbe:
It's 2 tox per unit, (which isn't a big deal because it's 30 TOX for a full syringe, and 30 TOX won't kill you) and a somewhat low chance to take minor brute damage. Doesn't cause organ damage unless you somehow pass 60 TOX total.

Yoshax:
I really fail to see the need for it, nor the why. Making something that is rarely used even worse isn't productive to good balance or gameplay. 

Why is this necessary?

Screemonster:
Because #2691 - there's been a small rash of infrequent but _incredibly_ annoying players essentially using it to piss off security and yakketysax away at turbo sanic speeds (mostly along the lines of "hey I've got contraband whaddaya gonna do 'bout it copper?" _nyoooom_), then when we tell them to cut that shit out they scream at us about banning them for using a feature of the game.

Yoshax:
And I'll say the same thing I said elsewhere, if people are being annoying and abusing something to just be annoying, then the staff should do their job and deal with the player, instead of spoiling the feature for everyone else. If they're screaming at you, then slap them around and get some respect out of them.

Anewbe:
We've not had a problem with Hyperzine ourselves, but it's currently very, very strong. As in, it doesn't really have any downsides.

Yoshax:
I'm just not sure I see the need to resolve a problem that does not the exist. The preemptive measures make something already unused, and not entirely appealing even less appealing.

That, or have a stable version that exists also. With all the chemicals we have, with no down sides, It's not very believable that we can produce a chemical to cure all injuries, or brain damage, or anything of the sort but not a stimulant of the same level. I'd either like a stable version to exist also, or for the side-effects to only exist for the ghetto version. But that's just my opinion.

Anewbe:
Removed the brute damage from hyperzine, but it's still too easy to get hyperzine for the zero downsides it has.
Made the damage chance from stimm bigger, and it gives its own message now.
Upped the tox damage from drinking fuel, because it was waaaay too low and relates to this PR as an ingredient.

Anewbe:
Staff polled, results positive.

Magneld:
Damnit why did I go to bed before checking this D:

PR #2747, opened 2016-11-15
Allows emptying of ashtrays into disposals

Anewbe:


code/modules/recycling/disposal.dm
```diff
@@ -106,6 +106,15 @@
 		update()
 		return
 
+	if(istype(I, /obj/item/weapon/material/ashtray))
+		var/obj/item/weapon/material/ashtray/A = I
+		user << "\blue You empty the [A.name]."
```
Yoshax: Span class, notice.

code/modules/recycling/disposal.dm
```diff
@@ -106,6 +106,15 @@
 		update()
 		return
 
+	if(istype(I, /obj/item/weapon/material/ashtray))
+		var/obj/item/weapon/material/ashtray/A = I
+		user << "\blue You empty the [A.name]."
+		for(var/obj/item/O in A.contents)
+			O.loc = src
```
Yoshax: forceMove

code/modules/recycling/disposal.dm
Yoshax: user will be given a message on an empty ashtray when nothing is emptied

Yoshax: all the code is run through and processed upon an empty ashtray too

Yoshax:
can also never dump the actual ashtray into the disposals, recommend checking for empty and continuing on with if its so, else do the emptying

Anewbe:
Should be cleaner now, thank you.

For some reason tossing the ashtray itself into disposals doesn't want to work, got any input?

Yoshax:
Putting the thing into the disposals is the default behaviour. Let it be handled by the rest of the attackby()
Check for contents, if there's contents then empty the ashtray, otherwise just let the proc continue. So have your return within your if and else is needed.

if(contents != 0)
    do empty
    return

Anewbe:
do empty in this case being the for loop to check contents?

Yoshax:
Loop through the contents and put them into the disposals, you'll also have to do the update_icon and update diposals in that if. You want it all done in that if, without an else, so when there's no contents it just skips by that continues on to default behaviour down below.

```dm
if(istype(I, /obj/item/weapon/material/ashtray))
	var/obj/item/weapon/material/ashtray/A = I
	if(A.contents.len > 0)
		user.visible_message("<span class='notice'>\The [user] empties \the [A.name] into [src].</span>")
			for(var/obj/item/O in A.contents)
				O.forceMove(src)
		A.update_icon
		update()
		return
```
that should do it, i think

PR #2749, opened 2016-11-15
Being unable to breathe and having oxyloss now causes brain damage that scales with your oxyloss

Yoshax:
When you cannot breathe, and are taking oxyloss you now have the chance to take brain damage. This amount scales with your oxyloss. It has a base chance of 5% per breathe tick, with is every 4 seconds. The amount of damage done is 2% of your current oxyloss.

This makes actual being unable to breathe a huge threat beyond just being unable to breathe.


serenemelody:
Do you have an estimate on how long it would take before you were mechanically not able to use things like computers in that state?

Yoshax:
Unsure. Not even sure that exists. Was unable to find it, or the numbers for it.

zekesulastin:
There are a couple places it's checked for effects:
https://github.com/PolarisSS13/Polaris/blob/da77d7d511155736b6f4299dc70a875cf4f0815e/code/modules/mob/living/carbon/human/life.dm#L187 (RNG; headaches start at 5, blurry at 15, dropping things at 35, blackouts/weaken at 45)
https://github.com/PolarisSS13/Polaris/blob/19d0fc333823dfbe50f06941f9fe9a9ad05d4f05/code/game/gamemodes/cult/ritual.dm#L111 (brainloss/100 chance to fizzle a rune?)
https://github.com/PolarisSS13/Polaris/blob/5f466c8802a396840789b8ef312802edcc722fa0/code/modules/power/apc.dm#L856 (brainloss/100 chance to forget how to use an APC)
https://github.com/PolarisSS13/Polaris/blob/19d0fc333823dfbe50f06941f9fe9a9ad05d4f05/code/game/machinery/machinery.dm#L230 (brainloss/100 chance to fail machinery use, guaranteed fail at >=55)

As far as I can tell, you are probably going to be dead by the time you'd have to worry about oxyloss causing you to be too brain damaged to use a machine unless you ignore a minor constant oxyloss source for a long time.

Yoshax:
Thanks very much, Zeke, I didn't think to look for the getBrainloss.

I was considering making the brain damage tick up in the same manner whenever you have oxyloss, and not just when you're not breathing, but I didn't think people would like that. 

If this is something people want, I could possibly adjust some of those numbers, particularly the machinery. The scale is kinda really off on that. I could see about adding more for it in general, but I didn't want to do that work and then have it denied. Easier to lay the frame and have it approved then add on later.

zekesulastin:
The GitHub search is a little weird like that if you're used to i.e. grep (also the silver searcher `ag` is probably the bestest thing ever :) http://geoff.greer.fm/ag/ )

The scale is from back when brainloss didn't kill you at 60; there used to be higher effects too (like the speech substitute one but I can't even find that anymore).  I think the last explicit artifact of that is a commented out line in the chaplain bible-naming code which sets brainloss to 100 if you have certain things as the religion.

Changing the scale would probably be a good thing as it's own PR; one low-effort way to increase the effects of brainloss would be to have the CLUMSY mutation automatically applied past a certain point as a lot checks for that.  At the least, it's an inspiration.

Yoshax:
The CLUMSY stuff is a good idea, but I'm not sure about it. It would give various people more work though, which is nice. But without Genetics I believe the only way to fix that would be with the chem which I can't remember. Although, it wouldn't be too difficult to just tie it into the brain damage and make it cure itself like it applies itself depending on brain damage amount.

If this gets in, I'll consider a follow up.

Anewbe:
Rather than CLUMSY, maybe confused, like flashes do?

PR #2753, opened 2016-11-16
Fixes bears

Anewbe:


SilverTalismen:
What exactly is being fixed?

Anewbe:
They weren't doing their damage calculation right, so changing the upper and lower damage vars did nothing.

SilverTalismen:
Aah, Well glad that was fixed, sounds Un'bear'able.

PR #2756, opened 2016-11-16
Lightning spells and bioelectrogenesis will no longer stun outright

Anewbe:
Stunlock was an issue with some of the spells.


SilverTalismen:
Why not just have it not stun them if they are already stunned but can still stun otherwise?

Yoshax:
Or make the stun a chance based upon damage and perhaps other factors.

Anewbe:
@SilverTalismen The stun is part of electrocute_act(), so that'd be difficult, if it's even doable. I'm not sure.
@Yoshax Problem with that is, again, electrocute_act(). Any change that gets made will affect both the antag things, and machinery/wire shocks, so unless you're looking for a sliding scale for all stun shocks, not sure if that'll work.

PR #2757, opened 2016-11-16
Adds lemon juice and milk to the bar.

Nerezza:
Lemonade should actually become a common drink on the station now, and bartenders can now serve milkshakes without having to run up to the cafe for milk from the coffee machine.

The amount of lemon juice and milk from the soda fountain and the booze-o-mat conform with what could be obtained already.

I made a carton icon similar to the OJ and lime juice cartons, and to conform with the naming for those two ('limejuice' and 'orangejuice') I renamed the original 'lemonjuice' icon to 'glass_lemonjuice' and gave the old name to my new carton icon. The old icon was not being used, but I decided to preserve it instead of outright delete it.


Yoshax:
Isn't there already a milk item? The one in the kitchen? Why define a new type?

Nerezza:
Because I wanted it to be as similar as all the other stuff in the booze-o-mat as I could get.

Anewbe:
Don't see a problem with adding a cardboard carton style milk sprite thingy, other than that I don't really have any thought on this.

PR #2758, opened 2016-11-16
CPR can be (pointlessly) done on dead people

Anewbe:
RP things only, won't actually help if they're already dead.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_attackhand.dm
+++ b/code/modules/mob/living/carbon/human/human_attackhand.dm
@@ -55,7 +55,7 @@
 
 	switch(M.a_intent)
 		if(I_HELP)
-			if(istype(H) && health < config.health_threshold_crit && health > config.health_threshold_dead)
+			if(istype(H) && health < config.health_threshold_crit)
 				if(!H.check_has_mouth())
 					H << "<span class='danger'>You don't have a mouth, you cannot perform CPR!</span>"
 					return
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/human/human_attackhand.dm
```diff
@@ -81,12 +81,14 @@
 				if(!do_after(H, 30))
 					return
 
-				adjustOxyLoss(-(min(getOxyLoss(), 5)))
-				updatehealth()
 				H.visible_message("<span class='danger'>\The [H] performs CPR on \the [src]!</span>")
-				src << "<span class='notice'>You feel a breath of fresh air enter your lungs. It feels good.</span>"
 				H << "<span class='warning'>Repeat at least every 7 seconds.</span>"
 
+				if(istype(H) && health > config.health_threshold_dead)
```
Techhead0: Isn't this istype a little paranoid?

Anewbe: Comes after the do_after, so I don't know, maybe?

Techhead0: Normally I'd agree, but:
A) The code after the `istype(H)` check doesn't reference `H` again.
B) `do_after(H, 30)` is already checking `H` every tick.
C) `H` is referenced before the check.

PR #2759, opened 2016-11-16
Removes a frequent runtime

Anewbe:


Nerezza:
That one's super noisy when someone's not breathing, huh? :P

Yoshax:
You have,

`if(!breath...)`

then `if(breath...)`

Anewbe:
The first line is `if(!breath || (breath.total_moles == 0) || suiciding)`, so if the others happen it run-times.

Yoshax:
350-351 never executes with the changes, making lung rupturing never happens.

Nerezza:
I could be reading it wrong, but that might be why 353 is there.

Yoshax:
My professional has suggested you remove 349-352, delete the else, then put 350-351 below 374.

Anewbe:
Why?

Yoshax:
So lung rupturing can still occur in circumstances where `if(breath && breath.total_moles)` is not true, such as where total_moles = 0.

Anewbe:
That's what the else is for?

PR #2761, opened 2016-11-17
Fixes lung rupturing in partial pressure or overpressure

Techhead0:
Also deals with a runtime

Alternative to #2759

### DIFF
--- a/code/modules/mob/living/carbon/human/life.dm
+++ b/code/modules/mob/living/carbon/human/life.dm
@@ -369,9 +369,9 @@
 		else if(L.is_bruised())
 			safe_pressure_min *= 1.25
 		else if(breath.total_moles < BREATH_MOLES / 5 || breath.total_moles > BREATH_MOLES * 5)
-			if prop(8)
+			if (prob(8))
 				rupture_lung()
-				
+
 	var/safe_exhaled_max = 10
 	var/safe_toxins_max = 0.2
 	var/SA_para_min = 1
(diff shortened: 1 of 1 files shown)


Anewbe:
Travis says:
> code/modules/mob/living/carbon/human/life.dm:372:error: missing condition
> code/modules/mob/living/carbon/human/life.dm:373:error: proc definition not allowed inside another proc

Anewbe:
Are you sure this fixes the runtime?

Techhead0:
The runtime was because it would check `breath.total_moles` in situations where `breath` didn't exist. (Line 349) The `return` at the end of `if(!breath || (breath.total_moles == 0) || suiciding)` ensures that if `breath` doesn't exist, then it won't reach the later `breath.total_moles` comparisons.

Anewbe:
If it works better than the other one I'll merge it, I'm just trying to parse the code.

Still not totally sure why the second block of code is placed where it is.

Techhead0:
It is where it is to take advantage of preexisting checks for `should_have_organ(O_LUNGS)` and `L.is_bruised()`.

Anewbe:
Something I've found (that I consider an issue) in testing these PRs is that lungs will frequently pop if the person goes into hardcrit, which I don't think is intended.

Techhead0:
People in crit had been passing a `null` breath to `handle_breath()`. [See this proc.](https://github.com/PolarisSS13/Polaris/blob/0bb15b27324a371804f8f625708cbe8bb03b00e0/code/modules/mob/living/carbon/breathe.dm#L8)

Without a runtime fix, this would cause a runtime (`null.total_moles`), aborting `handle_breath()` before any chance of lung rupturing. With any fix in, lung-popping is a risk of crit.

I'll include a quickfix so that null breaths don't pop lungs but vacuums do.

Anewbe:
That'd be appreciated.

PR #2763, opened 2016-11-17
Armor can hold helmets in suit storage

Anewbe:


SpadesNeil:
Due to the way the armor system works, this is not a good idea.

SpadesNeil:
Wait, do you get armor from clothing equipped in that slot?

Anewbe:
Shouldn't, no. Pretty sure the suit storage slot is basically just a pocket.

SpadesNeil:
Only reason I ask is because I've noticed other stat effects like slowdown being applied when an item is held in open hand for example. I vaguely recall a similar effect when I var-edited an item to fit in a pocket but this was _way_ back when I was still using Bay code.

Anewbe:
Worse comes to worst we fix it.

PR #2766, opened 2016-11-17
Xenobio again

MagmaRam:
Hopefully fixes the thing that Dataren pointed out. Tested with multiple combos of chems that do and don't have effects.


Anewbe:
Is there not a way to make it clear a singular reagent?

Datraen:
@Anewbe It clears at the end of the for loop.

PR #2768, opened 2016-11-18
Technomancer changes

Anewbe:
Drags in the boring parts of https://github.com/PolarisSS13/Polaris/pull/2689


Nerezza:
I think this looks good. I like that summoned things fade away, in particular.

PR #2769, opened 2016-11-18
[WIP] Animates airlock/shutter welding, adds a delay.

Nerezza:
The primary focus on this pull request is animating airlock welding to make it more noticeable, and adding a short time delay (currently 6 seconds, compared to 3 seconds crowbarring shutters) where you have to stand next to the airlock to finish welding or unwelding.

Another included addition is a little animation for shutters to show which ones are being crowbarred by people, since I was already working close to that code anyway.

This is almost completely done and after 12 hours straight both the above features work as intended, but both the code and animations need polish. I invite everyone to compile this locally and run around welding all the airlocks and shutters you can find before commenting since it's fun ;D

---

There is one major flaw/concern in this at the moment. Putting these overlays on airlocks that've been closed for more than 4/10ths of a second does not play the overlay's animation from start-to-finish, but starts somewhere in the middle OR stuck on the final frame. This is a BYOND "feature" from [years ago](http://www.byond.com/forum/?post=180590#comment942801) and the only way around it is to temporarily change the airlock's icon_state var immediately before applying the overlay.

As it currently works, every door's dmi has 'welding' and 'unwelding' movies for overlays, and a copy of 'door_closed' and 'door_locked' renamed to 'door_closed_waro' and 'door_locked_waro' ( 'waro' being short for 'work around'). Before applying one of the overlays, I swap the icon_state to its '_waro' counterpart then swap back once I've removed the overlay.

A more minor concern is that without a few changes it'd be possible to open the airlock while someone's welding on it, something I'm trying to prevent for now by using the 'operating' value already in that prevents airlock usage when true. I need to look deeper into it, and possibly come up with an alternative.

Finally a couple incredibly minor issues: First, when someone walks away while welding, their progress is reset. Since it's only 6 seconds to weld, I figured keeping this as it is would be best. Second, you can weld multiple doors simultaneously. This is something just about everything using do_after suffers from, so I didn't bother looking into a solution.

Lastly I still need to go through my changes and polish everything. For example you'll see that I was trying out a few 'user <<' messages for shutters, but I'll probably leave those out.

I'm eager to see what people have to say about this. Personally instant-welding always bothered me, but to get this to work BYOND had me do strange things in the dark with airlock .dmis that has left me uncomfortable.

Todo:
- [ ] Go over and adjust animations to look nicer.
- [ ] Clean up/rewrite both update_icon
- [ ] Investigate var 'operating'
- [ ] Hope that someone else can figure out a better icon_state workaround


Yoshax:
You should use `to_chat` instead of usr <<

Nerezza:
I've decided to hold off on the animated welding until I get some more experience and know for a fact that there's no other way to do this.

I'm going to put up a PR later with a couple shutter bugfixes, and a little change to shutters I made while doing this.

PR #2770, opened 2016-11-18
Makes Communicators free

Anewbe:
Concept is to give anyone who wants a cellphone their cellphone, for free.

~~Current issues: 
Doesn't save the selected pref properly when you save slot: Goes back to a blank state every time
Doesn't actually give anyone a comm yet: I'm looking for a way to make one bit of blanket code instead of having to put it in every single job
Worded poorly: I plan on adding at least one other color of Communicator and redoing the wording, but I want to make the code work before I start adding in the shiny bits.~~

### DIFF
--- a/code/modules/client/preference_setup/loadout/loadout.dm
+++ b/code/modules/client/preference_setup/loadout/loadout.dm
@@ -21,7 +21,7 @@ var/list/gear_datums = list()
 		if(!use_name)
 			error("Loadout - Missing display name: [G]")
 			continue
-		if(!initial(G.cost))
+		if(isnull(initial(G.cost)))
 			error("Loadout - Missing cost: [G]")
 			continue
 		if(!initial(G.path))
--- a/code/modules/client/preference_setup/loadout/loadout_utility.dm
+++ b/code/modules/client/preference_setup/loadout/loadout_utility.dm
@@ -11,6 +11,7 @@
 /datum/gear/utility/communicator
 	display_name = "personal communicator"
 	path = /obj/item/device/communicator
+	cost = 0
 
 /datum/gear/utility/folder_blue
 	display_name = "folder, blue"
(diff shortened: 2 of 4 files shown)


Nerezza:
Is it possible to make it cost 0 loadout points?

Anewbe:
Tried that, seems to just remove the option, instead of making it free.

Yoshax:
This a hugely unnecessary solution for a problem that isn't a problem. The code presented here is worse than the current solution in almost all ways. It adds more of its own snowflake code in a manner that is worse than what is currently in place. The current solution ties into an already existing system and is very simple, building upon a system that handles all of the stuff, including the presenting of the thing in the loadout, the spawning of the item, etc. This, does not. It adds new things to read and write to and from the sav files, it adds more unecessary logic to various places, adds its own HTML for presenting it and further touches preferences.

Solely from a perspective of code quality I give this a large :-1: Even then, the issue is not solely the code itself, but the fact it exists at all and bloating the other things more. Character setup is already fairly bloated and poor, adding more vars, lists and logic to it does no favours.

Anewbe:
How would you do it, then?

Techhead0:
[Change this line](https://github.com/PolarisSS13/Polaris/blob/0bb15b27324a371804f8f625708cbe8bb03b00e0/code/modules/client/preference_setup/loadout/loadout.dm#L24) to `if(isnull(initial(G.cost)))` and 0-cost things should work.

Anewbe:
Travis doesn't like that.

Techhead0:
[Same problem, different place.](https://github.com/PolarisSS13/Polaris/blob/0bb15b27324a371804f8f625708cbe8bb03b00e0/code/unit_tests/loadout_tests.dm#L12)

Anewbe:
Thank you for the help.

PR #2777, opened 2016-11-19
Fixes stupid naming of weapon crates

SpadesNeil:
This was confusing and dumb. Now it's not.

### DIFF
--- a/code/datums/supplypacks/munitions.dm
+++ b/code/datums/supplypacks/munitions.dm
@@ -4,10 +4,10 @@
 */
 
 /datum/supply_packs/munitions
-	group = "Weapons and Munitions"
+	group = "Munitions"
 
 /datum/supply_packs/randomised/munitions
-	group = "Weapons and Munitions"
+	group = "Munitions"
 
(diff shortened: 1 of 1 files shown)


Yoshax:
Munitions literally means Weapons. The group name being Weapons and Weapons makes no sense.

Anewbe:
http://www.dictionary.com/browse/munitions
Going to agree with Yosh here, in that I think it's fine as is.

SpadesNeil:
:U Fiiiine.

SpadesNeil:
The other corrections are still there. Why is this tagged unwanted feature removal?

Anewbe:
I don't consider it a bugfix, and I don't think it's a feature, and I don't have a tag for an in-between.
Not sure the capitalization on those crate names is appropriate, if memory serves there are issues that come up when you do things like that.

SpadesNeil:
That doesn't make any sense. We have an inconsistent, confusing naming, and the capitalization doesn't make a lick of difference except again not being inconsistent with every other crate name.

Anewbe:
Fair enough.

PR #2780, opened 2016-11-19
Drawing blood takes time

Anewbe:


Yoshax:
When used on yourself, and others, or just others?

Anewbe:
Everyone

Yoshax:
Inconsistent with the other behaviour. Using a syringe on yourself to inject things takes no time, hardsuit or not.

Anewbe:
Drawing blood requires a bit more precision and focus than giving an injection, at least as far as I know.

Yoshax:
The issue is the self vs use on others. Injecting someone else takes time, injecting yourself does not. Here, both take time. They should match in one way or the other, in my opinion. Either both on self takes time, or neither does.

PR #2781, opened 2016-11-19
Makes jumpsuits hide under space suits under right conditions.

Nerezza:
Mostly noticeable on tesh wearing voidsuits, bulky jumpsuits were overlayed by the voidsuit making the jumpsuit 'peek' around the latter's icons.

Now, spacesuits in general will outright hide jumpsuits with exception only to rigsuits. Yeah I'm looking at you rescue suit, you and your ****ed up icons =3=

----

This has been tested thoroughly, as in I put on every single /suit/space in game, and in doing so I found a conflict with the pirate jacket (WHY?) and an unrelated issue with a cult 'space suit', both fixed.

As said above, I'd love to take out the '!rigs' but their icons are weird and stuff.

### DIFF
--- a/code/game/gamemodes/cult/cult_items.dm
+++ b/code/game/gamemodes/cult/cult_items.dm
@@ -110,6 +110,7 @@
 	slowdown = 1
 	armor = list(melee = 60, bullet = 50, laser = 30, energy = 15, bomb = 30, bio = 30, rad = 30)
 	siemens_coefficient = 0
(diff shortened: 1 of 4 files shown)


MagmaRam:
I must have missed these when I did the big holster thing. Whoops. :+1:

Nerezza:
This has been a very long-standing issue, and I only really noticed it when I began playing tesh since everything's slightly bulky on them. Can't blame you for missing it!

PR #2783, opened 2016-11-20
Allows borgs to use toggle lock on closets/crates

Yoshax:


Nerezza:
```
/obj/structure/closet/attack_ai(mob/user)
    if(istype(user, /mob/living/silicon/robot) && Adjacent(user)) // Robots can open/close it, but not the AI.
        attack_hand(user)
```
While you're at it, if you'd like you could change the above lines 291-293 in closets.dm to:
```
/obj/structure/closet/robot_attack(mob/user)
 	if(Adjacent(user))
 		attack_hand(user)
```

Techhead0:
BTW, should be `attack_robot()`, not `robot_attack()`.

Nerezza:
~~But I want a robot attack~~
derp, thanks for correcting me!

Yoshax:
Should be now.

PR #2784, opened 2016-11-20
IV Drips are no longer dense

Yoshax:
its annoying + inconsistent


MagmaRam:
my surgeon experience just went from a 5/10 to an 8/10 overnight

PR #2785, opened 2016-11-20
Misc windoor fixes. Dron (and borgs) can now rename airlock/windoor assemblies.

Nerezza:
This started as me making robots able to rename windoors and turned into something a little more.

- Robots can rename windoors and airlocks. They have to stand next to them and click with no modules enabled, similar to buckling.
- Windoors can now be disassembled with significantly less broken glass. The windoor has to be opened first, then ~~an engineer~~ someone can crowbar the windoor 'out of its frame' to turn it back into an assembly. If the door's closed, then there's still a lot of broken glass involved.
- How windoor assemblies rename themself on each step has been rewritten and actually works properly now.
- Windoors given no name during construction now use the default name instead of a null name.
- Emagged windoors now give broken circuits when deconstructed, instead of resprited airlock electronics.

Tested and working for everything, even caught a couple last minute bugs. Also while I was in the neighborhood I to_chatted with all the usr <<s.


Anewbe:
I like the bottom four, but I'm not sure how I feel about the first.

Nerezza:
I could give all the robot modules (and drons) a pen, but I figured this way was better. I could add a check to see if the robot in question is an engiborg or a dron, probably.

Anewbe:
Misread the first point, thought you meant they could rename active doors, which would be no bueno.

Nerezza:
OH, yeah I'd agree. I'll put a check in for roborg module type anyway though, since engiborg and dron are the only ones who can construct doors. (Ironically, construction robot can't construct doors)

Nerezza:
Added the istypes, tested and works!

Nerezza:
Already fixed

PR #2787, opened 2016-11-20
Uplink guns come DNA Lockable

Yoshax:
- All guns purchases in the antag uplink now with come preinstalled with a DNA Locking chip, allowing them to be DNA Locked. Information further on this can be found in previous PRs, #1726 #2400
- Backend for uplink item spawning has been expanded to allow for extra args to be provided to the item's new. 
- Visible weapons has had /ranged/ added to allow for all of this to fully work
- Lot of changes in visible_weapons.dm due to moving the items into not ranged then ranged, and in TC order

### DIFF
--- a/code/datums/uplink/uplink_items.dm
+++ b/code/datums/uplink/uplink_items.dm
@@ -32,6 +32,7 @@ var/datum/uplink/uplink = new()
 	var/datum/uplink_category/category		// Item category
 	var/list/datum/antagonist/antag_roles	// Antag roles this item is displayed to. If empty, display to all.
 	var/blacklisted = 0
+	var/list/new_args = list()
 
 /datum/uplink_item/item
 	var/path = null
@@ -39,8 +40,6 @@ var/datum/uplink/uplink = new()
 /datum/uplink_item/New()
 	..()
 	antag_roles = list()
-
-
 
 /datum/uplink_item/proc/buy(var/obj/item/device/uplink/U, var/mob/user)
 	var/extra_args = extra_args(user)
@@ -132,7 +131,9 @@ datum/uplink_item/dd_SortValue()
 	return I
 
 /datum/uplink_item/item/get_goods(var/obj/item/device/uplink/U, var/loc)
-	var/obj/item/I = new path(loc)
+	var/list/input_args = new_args.Copy()
+	input_args.Insert(1, loc)
+	var/obj/item/I = new path(arglist(input_args))
 	return I
 
 /datum/uplink_item/item/description()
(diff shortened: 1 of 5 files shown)


Magneld:
Bit 50/50 on this, mind if I toss this somewhere everyone can notice it and talk it over and not just staff once I get home?
This isn't a nope from me whatever you decide I'm just a bit unsure.

Yoshax:
Feel free.

SpadesNeil:
Mixed feelings on this. DNA locking means obviously that only the antag who owns the gun can use the gun. This is both good and bad for the same reasons. Depends which end of the barrel you're on.

Nerezza:
Sec already has guns, and one thing I feel has been a problem is that they pick up powerful guns off of traitors and overpower over traitors with those said guns. This would change that, sec wouldn't be able to pick up rifles off the bodies of traitors to use on other traitors, and on top of that if the gun's set to self destruct it'd be destroying evidence of an agent on the colony.

The way I see it, this would make it so the lighter armed traitors won't be facing whatever the super-heavily-armed guy bought when before he got taken down. On top of that, the self-destructing thing fits a high-tech subversive spy theme.

Yoshax:
They're not the explosive variant, and there's no way (as of yet) to change between non-explosive and explosive. It basically just makes the gun shit the bed and never work if you don't have the DNA.

Nerezza:
I'm fine with that too, I just wanted to throw that in while I was unsure. To me the most important thing is making sure traitor-summoned weapons don't beef security up to steamroll levels.

zekesulastin:
There is a forum thread here as well: http://ss13polaris.com/forums/viewtopic.php?f=18&t=1547

code/datums/uplink/uplink_items.dm
```diff
@@ -132,7 +131,9 @@ datum/uplink_item/dd_SortValue()
 	return I
 
 /datum/uplink_item/item/get_goods(var/obj/item/device/uplink/U, var/loc)
-	var/obj/item/I = new path(loc)
+	var/list/input_args = new_args.Copy()
+	input_args.Insert(1, loc)
```
Anewbe: Looks like this is backwards from what gun/New() is asking for.

Yoshax: No, it's fine. The 1 isn't the argument given to new, it's saying make the 1st element of input_args into the first argument and then shift the other ones up a place to ensure that the loc is inserted at position one.

Anewbe: Ah, alright.

Anewbe:
Is there a way to limit this to certain weapons, rather than all of them?

Yoshax:
There is.

SpadesNeil:
I'd like to see it limited to higher power weapons.

So, other than pistols and the standard laser carbine.

Yoshax:
I wouldn't. Especially as how that leaves it kinda all over the place. Seeing as "high power" is highly subjective.

PR #2788, opened 2016-11-20
Fixes up and implements the experimental welding tool

Yoshax:
- Holds 40 units of fuel, regenerates the fuel
- One to chief engineer
- 2 in each YouTool as premium products
- Replaced welding tool in Syndicate Toolbox
- Can be ordered from cargo, you get 2 and other supplies for 20 points

### DIFF
--- a/code/datums/supplypacks/engineering.dm
+++ b/code/datums/supplypacks/engineering.dm
@@ -55,7 +55,7 @@
 			)
 	cost = 20
 	containertype = /obj/structure/closet/crate
-	containername = "Advanced Welding Supplies crate
+	containername = "Advanced Welding Supplies crate"
 
 /datum/supply_packs/eng/fueltank
 	name = "Fuel tank crate"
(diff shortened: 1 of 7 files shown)


Anewbe:
Not sure it makes a load of sense to have it vending machines and a supply pack if it's experimental, sounds more like something research would make.

Yoshax:
It's for the availability, so people actually have a chance to use them. 9/10 times locking something behind a research wall ends up with it never seeing the light of day. They're in the premium slot of the vending machine to give coins a little bit more of an actual use to give you something. The crate is just for general availability and to give cargo a little more something to have people interested in actually going down there.

Anewbe:
Leaving it in the syndicate toolbox and the CE's belt isn't exactly uncommon, and if you add it to research anyway there's a low but non-zero chance that someone else will make it.
Please change it.

Yoshax:
going to be 100% unseen outside of the CE behind a science wall, like everything else in R&D

Anewbe:
Thank you.

PR #2790, opened 2016-11-20
Fixes #2630

### LINKED ISSUE #2630: Disassembling sleepers with people inside them deletes the mob inside!

Surely this is a bug!

Not be able to deconstruct sleepers when people are in them.
#### What actually happened

Player was deleted.
#### Steps to reproduce
1. Shove ~~victim~~ patient into sleeper.
2. Be dumb engineer borg.
3. Deconstruct the sleeper with a person inside.
#### Additional info:
- **Server Revision**: https://github.com/PolarisSS13/Polaris/commit/85d1ef3062076d89d91c00236bdf579698389d82

Atermonera:
Fixes #2630 
Because I took way too long to get around to this.

Implements Magma's solution, because mine didn't actually gain that much, and would've had some less-than-sensible implications for any future machinery with occupants.

### DIFF
--- a/code/game/machinery/machinery.dm
+++ b/code/game/machinery/machinery.dm
@@ -136,7 +136,13 @@ Class Procs:
 				component_parts -= A
 	if(contents) // The same for contents.
 		for(var/atom/A in contents)
-			qdel(A)
+			if(ishuman(A))
+				var/mob/living/carbon/human/H = A
+				H.client.eye = H.client.mob
+				H.client.perspective = MOB_PERSPECTIVE
+				H.loc = src.loc
+			else
+				qdel(A)
 	return ..()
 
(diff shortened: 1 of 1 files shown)


Anewbe:
If you put Fixes #Bugnumber in the OP it should close the bug if this merges.

Anewbe:
I feel like there may be a single proc to call for that process, but I'll need to double check.

Atermonera:
That's the bit of go_out() used by any machines with an occupant to actually move the mob out of the machine. My original solution was to refactor occupant and go_out to the machinery class so I could call if(occupant) src.go_out(), but because of ai_fixer and xenobio2 machinery, occupant would have to become var/mob/living, and any machine that actually uses the var would be expected to specify what subclass from there can actually enter.

PR #2798, opened 2016-11-21
Adds a preference in character setup that determines whether or not your character can actually die

Yoshax:


Anewbe:
That tox changelog is wrong, my fault. If you've got a way to change that for the master please let me know.

Yoshax:
Fix'd.

PR #2800, opened 2016-11-22
Misc APC tweaks & fixes

Nerezza:
- Ports Bay's soft APC rebooting nanoUI and tweaks it into a 'Polaris variant'.
- Changes the random APC-bluescreening event. It now effectively emags the APCs so tatortots can take advantage of the event, and engineering has a reason to repair the APC.
- Emagged APCs and malf hacked APCs can be fixed by removing the power cell and applying a multitool to 'reboot' the device.
- EMP'd APCs now give Bay's timed (or manual) reboot nanoUI so that important places aren't shut off permanently if engineering can't get around to them. Critical APCs (AI core, engine) stay off for around 30 seconds. Other APCs (hallways, everything else) will stay off for 5ish minutes.
- Nins and twizs no longer **** up APCs when they drain power from the cell completely.
- Nin/twize drained power comes from the powernet primarily, and only draws from the battery if the powernet doesn't give enough (random chance) or if the cable is cut. This also means that if the battery runs dead and the APC is wired to a SMES, it'll continue sucking power. This functions almost exactly like it does in Bay, with an  adjustment to match how our power works right now. I also left out the thing where locked APCs would reduced power drain to 1/4th.
- Bluescreened APC sprite now switches to an open wire panel sprite if a screwdriver is applied.
- APCs with an open wire panel have to have the panel closed to access their UI. This is mostly because the sprite visibly shows the screen off.
- Grid checks put a nanoUI notice on APCs. As it is now, you have to wait for the grid check to end or use the grid checker in engineering before you can manipulate affected APCs.
- Ported a small clean-up to area power states from Bay
- Bashing APCs to pieces was always ineffective but would spam everyone anyway. Now, it only spams the person bashing the APC. This is a placeholder, ideally you should be able to bash a working APC's cover off if you want its battery out but this PR is already big enough as is. If the cover can be bashed off (now only on visibly broken APCs. I changed malf-hacked APCs to need wire hacking to open) it'll act as it did before.
- APC grammar fixes
- More APC grammar fixes
- My god the grammar fixes
- to_chat'd the user <<s again.
----
As a little side note I'd like to add electrical storm changes from https://github.com/Baystation12/Baystation12/pull/12865 eventually, which this PR would facilitate.

All of the things added were tested, tweaked more, tested more, and finally after a whole weekend I finished this up. All that's left to do is:

- [x] Delete that one line I accidentally left in, whoops.
- [x] Make and commit a changelog.


Nerezza:
All done.

I want to note about the new emag-fixing thing. Before, you would fix emagged APCs by removing their electronics and applying a brand new APC frame to 'replace' the old front panel. This bothered me, hence the change to the multitool fix.

Nerezza:
Oh right, I knew I was forgetting something. APC ion_act() isn't being used anymore so I commented it out with /**/, and left it in just in case someone wants to use it again.

PR #2801, opened 2016-11-22
Fixes runtime in inventory.dm

Nerezza:
Clicking an inventory slot with nothing in your active hand would runtime. Now it checks to make sure you have something in your active hand, and if not it returns 0.


Anewbe:
Tested?

Nerezza:
Yeah, I made this early while I was doing the APC stuff and had it compiled in every time I built to test that stuff.

PR #2802, opened 2016-11-22
Adds more lightsout landmarks. Fixes #2796

### LINKED ISSUE #2796: apc_damage and electrical_storm events ONLY affect cargo and nowhere else.
Right now there's only one "lightsout" landmark on the map, and it's located in cargo's lobby. This landmark affects the events electrical_storm and apc_damage meaning these two random events only ever occur around there.

Whoever's touching up the map next, I'd appreciate if you could stick more of that landmark into other areas because I'm currently making changes to apc_damge, and it would be cool if said changes affected more than just cargo and the engineering washroom.

Nerezza:
Cargo no longer takes the full force of every electrical_storm or apc_damage event. There are now 10 more epicenters the two events can choose, placed in varios places on the station except for the bridge. Depending on severity (and rng) the bridge can still be affected.


Anewbe:
Did you use the map merger?

Nerezza:
Derp, nope. Let me figure out how that works then.

Anewbe:
Get a fresh map, use 1prepare_map.bat from the tools folder, make the changes, then use 2clean_map.bat, should do it.

Nerezza:
Already had that one pretty quick, problem is I wanted to ammend but I don't think I can do that with 10,000 lines conflicting. I'll be done in a minute.

Nerezza:
@Anewbe Fixed it

PR #2804, opened 2016-11-22
Inflatable QoL tweaks

Nerezza:
Features:
- Ctrl-click can deflate inflatables now.
- Using inflatables on adjacent tiles deploys them to that tile. Currently this doesn't care about objects in the way, ie computers or welding tanks, but it does respect glass and imo it works fine this way.

Bugfix:
- Bumping an inflatable door no longer freezes it closed for a few moments.
----
inflatables.dm is a mess so I'll probably look into cleaning it up to be more like Bay's sometime after Thanksgiving.

I'm also probably going to port the robot inflatables dispenser if Yosh doesn't do that, so engiborgs and crisis borgs can deploy airlocks where needed. However, I don't think the engineering derpartment should get dispensers like in Bay, hence the turf-deploy in this PR.

Edit: Oh and, this was tested.

### DIFF
--- a/code/game/objects/structures/inflatable.dm
+++ b/code/game/objects/structures/inflatable.dm
@@ -4,14 +4,20 @@
 	icon = 'icons/obj/inflatable.dmi'
 	icon_state = "folded_wall"
 	w_class = ITEMSIZE_NORMAL
+	var/deploy_path = /obj/structure/inflatable
 
-	attack_self(mob/user)
-		playsound(loc, 'sound/items/zip.ogg', 75, 1)
-		user << "<span class='notice'>You inflate [src].</span>"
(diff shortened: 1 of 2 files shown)


Anewbe:
Dispensers, you say?

Nerezza:
The mechanical psuedo-cases in Bay's engineering that lets you turf-deploy and instantly deflate-collect inflatables (I'm sure you already know, but future reference.) Honestly I think that's fine for robots, but I'd prefer the cases we have now plus the turf deploy I put in above.

It just feels like we're going too far with equipping engineers with robotic dispensers, and aside from that I'm a little bothered with how said dispensers wipe out the fingerprints on their inflatables.

In other words, I'm being nitpicky but at least I put together an alternative that'd make engineer-portable dispensers redundant ;P

PR #2814, opened 2016-11-24
Changes the formatting for attack logs for reflex shots.

### LINKED ISSUE #2806: The way that reflex shots are logged is confusing and needs to be less confusing
And if I don't write it down in a place where I'll see it every day then I'll forget to do it.

MagmaRam:
When someone triggers reflex fire, it'll note it in their attack logs. It'll also note in the shooter's attack logs that the gunfire was reflex and not manual. In addition, the attack logs that admins get will be changed to something in the format of "[shooter] shot [victim], who triggered gunfire [weapon] by reflex.

Fixes #2806, since that's my own reminder about this.


Magneld:
THANK. YOU.

PR #2816, opened 2016-11-24
Blood can be put into dead bodies.

MagmaRam:
Right now we're doing this because blood can't be put into dead people, which makes Medical's job aggravating if they try to defib someone who died of blood loss.

Most reagents are still skipped, but blood is, importantly, not.
This needs to be discussed further, as apparently the reason this was this way is because of lag. I could try doing something that, if they're dead, only moves blood from bloodstr to vessel, but this is an easier and cleaner fix.


Anewbe:
Looks good to me

PR #2817, opened 2016-11-24
Defibs should revive people now.

MagmaRam:
This might be an unnecessary fix, I might just have the worst luck. But I couldn't get a defib to res in 10 tries before this, got it on the second after.


Anewbe:
It was working fine for me when I tested it.

MagmaRam:
It might be down to my own cruddy luck, then. I also haven't seen defibs work on the live server, though. If they've worked on live, then I'll close this one.

Anewbe:
I'll test it live sometime today.

Nerezza:
How does time of death track? Is it a timer counting down or the round time that death occurred? I'm still in vacation so I can't check

PR #2819, opened 2016-11-24
Fixes improper rookie lung

Anewbe:


Anewbe:
Self-Merge cause this is important.

PR #2820, opened 2016-11-25
Fixes the vacuum thing again

Anewbe:


Anewbe:
Still borked, still self-merge

PR #2821, opened 2016-11-25
A Fix for Encoded Audio Language, and by Accident, Languages in General

ForFoxSake:
So, I was doing a little testing, and I noticed that I could bypass the `can_speak_special()` proc of EAL.
I set about fixing that issue, and noticed that in fact, I was bypassing `can_speak()` altogether, so I ended up fixing that too.
This fix means that now, organic creatures cannot speak EAL, however they can still learn and understand it.
Along side this fix, Positronic Brains and Full Body Prosthetics that are created mid-round automatically understand EAL, but this does not effect FBPs made in the character setup.

If you have any questions, thoughts, comments or critique, please leave a message or contact me on discord.

### DIFF
--- a/code/modules/mob/language/language.dm
+++ b/code/modules/mob/language/language.dm
@@ -203,8 +203,10 @@
 		if(!(L.flags & NONGLOBAL))
 			if(L == default_language)
 				dat += "<b>[L.name] ([get_language_prefix()][L.key])</b> - default - <a href='byond://?src=\ref[src];default_lang=reset'>reset</a><br/>[L.desc]<br/><br/>"
-			else
+			else if (can_speak(L))
 				dat += "<b>[L.name] ([get_language_prefix()][L.key])</b> - <a href='byond://?src=\ref[src];default_lang=\ref[L]'>set default</a><br/>[L.desc]<br/><br/>"
+			else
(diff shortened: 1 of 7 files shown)


Anewbe:
You test this?

ForFoxSake:
Personally, I tested it as thoroughly as I could, even had to connect twice to a test server in order to make sure adding and removing positronic brains worked.
I would say it's working as intended, but I can't account for every situation and taste.
If there's nothing dire that sticks out, then yeah, it's good to go.

Nerezza:
Can diona learn to speak EAL? I think they should be capable but definitely something to ask their maintainer.

SilverTalismen:
As things stand Diona can only learn one language so even if they could learn it they hav e to pick carefully what they know and don't.

Anewbe:
Gonna toss this about for mild discussion, looks like it works, but people may prefer it as is.

MagmaRam:
I'm fairly fond of humans being able to speak EAL, but then I'm also the sort of person who wouldn't mind datajacks and the like being canon.

Anewbe:
Gonna make an implant to let people speak EAL if it's in their language list, we'll see how that goes. Otherwise, away we go.

PR #2839, opened 2016-12-02
Adds an implant that allows humans to speak EAL.

MagmaRam:
It's available in the loadout for 2 points.

Only like four days late!

### DIFF
--- a/code/modules/mob/language/station.dm
+++ b/code/modules/mob/language/station.dm
@@ -116,7 +116,12 @@
 	space_chance = 10
 
 /datum/language/machine/can_speak_special(var/mob/speaker)
-	return speaker.isSynthetic()
+	var/obj/item/I
+	for(I in speaker.contents) //this is ugly but I can't make the pretty way work
+		if(istype(I, /obj/item/weapon/implant/language/eal))
+			return 1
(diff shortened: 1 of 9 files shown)


code/modules/mob/language/station.dm
```diff
@@ -116,7 +116,12 @@
 	space_chance = 10
 
 /datum/language/machine/can_speak_special(var/mob/speaker)
-	return speaker.isSynthetic()
+	var/obj/item/I
+	for(I in speaker.contents) //this is ugly but I can't make the pretty way work
+		if(istype(I, /obj/item/weapon/implant/language/eal))
+			return 1
+
+	return (speaker.isSynthetic())
```
Zuhayr: ```return ((locate(/obj/item/weapon/implant/language/eal) in speaker) || speaker.isSynthetic())```

Zuhayr: Actually make that
```
var/obj/item/weapon/implant/language/eal/beep = locate() in speaker
return ((beep && beep.implanted) || speaker.isSynthetic())
```
Otherwise carrying an implant in your hand will let you speak the language.

PR #2841, opened 2016-12-06
Fixes water-coolers.

ForFoxSake:
Creating a water-cooler either through admin spawn or by construction currently doesn't initiate it's reagents list, causing the thing to throw a runtime when you try to add or remove water to it.
These runtimes can then be exploited to duplicate the water containers, which can be turned back into glass at an autolathe.

Figured that this might just about be a big enough issue to warrant a one-line PR.


Anewbe:
Tested?

ForFoxSake:
Yep. Took longer to perform the two compiles to test it, than it did to fix it.

PR #2842, opened 2016-12-07
A Few Atmospherics Tweaks (Valves, Vents and Layout)

ForFoxSake:
**Contains a map edit to Atmospherics Department!**

These few little things have been bothering me for a little while now. For starters, digital valves can't be moved or created. I think this was initially intended to stop people from circumventing ID-locked digital valves that were supposed to be a thing. Either way, this fixes the bizarre lack of item sprites for digital valves.
The next issue that kept bugging me was the fact there was no available tool to regulate an area from overpressure, despite the fact that rooms that need this functionality (Atmos tanks, the SM Core, etc) start with air vents mapped in for this exact purpose.
With this, you can now switch air vents from pressurizing to syphoning... Which makes things a lot easier when it comes to constructing an engine room, server room or sauna.

Now that digital valves can be pulled up (by anyone with sufficient access), I've also removed a few of the redundant digital valves from the atmospherics room, and removed a few redundant pipes in the process, hopefully making atmospherics a little easier on the eye, and giving atmos wizards more room to work their magic.
Here's what the area looked like **before**: http://imgur.com/0qnXAE6
And here it is **after**: http://imgur.com/2VjmCrs

I have tested all changes I've made, some twice over. And everything appears to be in order.


MagmaRam:
>With this, you can now switch air vents from pressurizing to syphoning

you are jesus

PR #2846, opened 2016-12-09
Speeds up walking

Anewbe:
Walk and run speed are entirely based on the config file, rather than any sort of in-code constant.

Will require updating the config file to actually work.


Anewbe:
Thinking, based on what the numbers have turned into, that we'll leave the runspeed config where it is (plus one for the initial change, minus one for the shoes) and make the walkspeed 5+the current one (plus seven for the initial change, minus one for the shoes, and another minus one because walking is hella slow right now).

Anewbe:
Given the green light to Self-Merge, should be able to get the configs changed with the update in one swoop.

PR #2851, opened 2016-12-09
Adds Vital Underwear Icon File Overriding

ForFoxSake:
Fixes nothing.
Part 4 of 4.

This change allows for multiple underwear icon files, either for easier expansion downstream or for admin shenanigans mid-round.

### DIFF
--- a/code/datums/underwear/underwear.dm
+++ b/code/datums/underwear/underwear.dm
@@ -67,7 +67,7 @@ datum/category_group/underwear/dd_SortValue()
 	if(!icon_state)
 		return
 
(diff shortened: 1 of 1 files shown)


Anewbe:
Tested?

ForFoxSake:
Everything that was part of the larger PR has been tested. This one particularly has been tested at least twice over because I wasn't expecting the DM language to work the way it does.
But yes, this PR does work, and the existing underwear remains completely unaffected, which is the priority here.

I'd also not really call this a feature, since it doesn't really add anything in itself. But it's not a bug fix either, I understand. Is there a refactor label?

PR #2861, opened 2016-12-17
Adds a crate with defibrilators for 30 points

MagmaRam:
As far as I know there's no way to get more of them yet.


Anewbe:
I _think_ this might leave them unlocked, did you check that?

MagmaRam:
The crates are unlocked, I didn't see any reason to lock a crate full of first-aid supplies. Aren't they generally only locked when they actually contain something extremely dangerous (viruses, guns, emitters, supermatter)?

Anewbe:
Fair enough.

PR #2863, opened 2016-12-17
Circuits should work like posibrains now

### LINKED ISSUE #2141: Drone Brains are MMI's in code

Anewbe:
Fixes #2141
Moves posibrains and circuits into the MMI file

I can't shake the feeling that I missed something, but it seemed to be fine.


MagmaRam:
I can't see any problems with it that jump out at me.

Issue #2609, opened 2016-10-07
Exosuit Grippers can use the same atachment infinatly on mechs.


Edwardc120:
#### Additional info:  This could easily be used for making infinite materials if you can recycle the parts
- **Server Revision**: Found using the "Show Server Revision" verb under the OOC tab.
- **Anything else you may wish to add** (Location if it's a mapping issue, etc)

Edwardc120:
Side Note as well when you try to remove the other two drills you can't there stuck there on the mech.

Atermonera:
This should've been closed by #2864

Issue #1998, opened 2016-06-28
Research Borgs cannot attach/insert prosthetic (and flesh, assumedly) organs/limbs.


Mechoid:
#### Brief description of the issue

Research cyborgs cannot attach prosthetic limbs or re-insert organs they have removed.
#### What you expected to happen

Research cyborgs could re-attach an FBP's leg or head, for instance if it lost either. Or restart a newly repaired FBP by removing and/or re-inserting its cell or brain.
#### What actually happened

The gripper does nothing. Post-surgery change, it now gives the 'you can't find a use for this' message regarding the gripper itself.
#### Steps to reproduce

Be a research cyborg, saw off an FBP's limb, try to re-attach.
Be a research cyborg, rip out an FBP's cell, try putting it back in.
#### Additional info:
- **Server Revision**: 6374bc2c4c342b53bbf2950b9e792ec7e2084495
- **Anything else you may wish to add**

Edwardc120:
Also for gripers the sheet loaders can't load there materials into Protolathes like they can autolathes

Anewbe:
Was this working at any point, or has it been broken since the flesh grippers were added?

Mechoid:
Robotic limbs functioned beforehand, though they likely ceased once they were made into objects of differing type from normal borg-body limbs. With internal organs/parts like the brain and (For FBP's) the minibattery, it's persisted since the gripper was added.

Atermonera:
Unable to reproduce, looks like this's been fixed,

Atermonera:
Ok, this should have been closed when #2864 got merged. I'll look into testing it again, just to double check.

Issue #858, opened 2016-01-26
Scienceborg Portable Destructive Analyzer Bug


McNanite:
The research module's portable destructive analyzer tool cannot seem to be used to increase the research level of R&D. I've tested a little bit, on a few arcade circuits and some choice items around the lab, and nothing seems to do it. I have tried resyncing the database and the tool multiple times, but that failed as well. Overall, this bug prevents the research borg from actually doing anything useful within R&D.

Atermonera:
Unable to reproduce, tested this when bumping research levels up enough to print boards for a ripley while testing fixes for unrelated issues.

Atermonera:
This should also have been closed by #2864

PR #2864, opened 2016-12-17
Gripper Fixes, round one.

### LINKED ISSUE #2671: Research Borgs' grippers cannot insert objects into other borgs.

Research borgs' science grippers cannot insert objects, such as powercells or upgrades, into other borgs. They can take them out, and go through the process of putting it back in, only for the object not to go in.

Research borg takes old cell out, grabs new one, smacks other borg, puts in cell.
OR
Research borg picks up upgrade, uses on other borg or itself. Upgrade goes in.
OR
Research borg removes broken bits from other borg, and puts in new ones using the gripper.
#### What actually happened

Cell remains un-inserted.
Upgrade remains un-inserted.
Parts remain un-replaced.

Other borg is very unhappy.
#### Steps to reproduce

-Be research borg, with other borg.
-Open other borg's panel, use gripper on borg, you now have the cell.
-Hit them with the gripper again to try and insert the cell. Notice cell doesn't go in.

-Be R. borg, other borg not needed.
-Pick up any upgrade, open to-be-upgraded borg's panel, yours or theirs.
-Smack with gripper, get told the upgrade is installed.
-Note it doesn't actually get installed.

-Be R. borg, with broken friend borg.
-Open panel of other borg, remove cell to access parts.
-Crowbar out broken pieces, try using the gripper to install the replacements.
-Note it does the same as the upgrade.
#### Additional info:
- **Server Revision**: 109e7962c4d48743688f52826a97cad8b7874175 
- **Anything else you may wish to add** Gripper bugs continue to compound on one-another. The issue seems like it would have the same reason across all of them, due to their nature.

Atermonera:
Attackby() wasn't being called when wrapped.attack() happened, so I manually called it (Don't know if the order of the two particularly matters, if so then it's a quick fix). Fixes #2671 and allows exosuit grippers to function again. Tested.
#2609 is fixed by this.
#1998 might have been fixed, or was beforehand
#858 couldn't be reproduced when I was testing various stuff after making the fix.

### DIFF
--- a/code/modules/mob/living/silicon/robot/drone/drone_items.dm
+++ b/code/modules/mob/living/silicon/robot/drone/drone_items.dm
@@ -134,7 +134,8 @@
 
 	can_hold = list(
 		/obj/item/mecha_parts/part,
-		/obj/item/mecha_parts/mecha_equipment
+		/obj/item/mecha_parts/mecha_equipment,
+		/obj/item/mecha_parts/mecha_tracking
 		)
 
 /obj/item/weapon/gripper/no_use //Used when you want to hold and put items in other things, but not able to 'use' the item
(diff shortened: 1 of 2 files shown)


Anewbe:
Hey, don't go quoting me in them things, I'm not at all a reliable source of information. (But I'm pretty sure that's right)

Otherwise seems to be fine.

Anewbe:
Also, in seriousness, if you can do a quick search for the issues this fixes (I think there's more than one, iirc) and tag them in the first post, that'll close them so I don't have to manually.

Atermonera:
#2609 is fixed by this.
#1998 might have been fixed, or was beforehand
#858 couldn't be reproduced when I was testing various stuff after making the fix.

PR #2865, opened 2016-12-17
Replaces [stack]/fifty atoms with atoms that spawn the normal stack

MagmaRam:
The stack/fifty objects were really buggy, this should fix all of that in one go.


Anewbe:
Did you test this?

Techhead0:
I'm curious how they were buggy, aside from R&D stuff that checked exact type instead of using istype.

Anewbe:
They weren't working in a lot of cases. The carpet stacks especially were screwed up.

MagmaRam:
let me just fuck with git for 50000 weeks trying to figure this out ok

Techhead0:
So... having read this, if this actually fixes whatever problems you're having with carpets, I'll be impressed.

PR #2893, opened 2017-01-02
Nerfs NPC Pirate

SpadesNeil:
Rapid firing /obj/item/projectile/beam? How about no? That's 120 damage
in a split second. With the new way burn damage affects players; fuck
that!


SilverTalismen:
Where's your sense of- *Gets fried by a single NPc pirate, dies instantly*

PR #2895, opened 2017-01-02
Might fix changeling bans

Anewbe:
For some reason it's not checking properly for bans, not entirely sure if this'll fix it.


Anewbe:
Self-Merge

PR #2896, opened 2017-01-02
Removes an unused gas from the code

Cameron653:
Removes oxygen_agent_b from the code.

This gas has no use, is unobtainable in game (Even with admin powers you'd have to edit a preexisting gas, as there is no canister to spawn it in with,) doesn't do anything special, and there is pretty much no reason to even have it in the code. 

After some testing, everything appears to work as it should. Scrubbing, siphoning, and pressurizing with the vents all works as they should.

### DIFF
--- a/code/ATMOSPHERICS/components/trinary_devices/filter.dm
+++ b/code/ATMOSPHERICS/components/trinary_devices/filter.dm
@@ -40,7 +40,7 @@
 	..()
 	switch(filter_type)
 		if(0) //removing hydrocarbons
-			filtered_out = list("phoron", "oxygen_agent_b")
+			filtered_out = list("phoron")
 		if(1) //removing O2
 			filtered_out = list("oxygen")
 		if(2) //removing N2
@@ -95,10 +95,10 @@
 
 /obj/machinery/atmospherics/trinary/filter/process()
 	..()
-	
+
 	last_power_draw = 0
 	last_flow_rate = 0
-	
+
 	if((stat & (NOPOWER|BROKEN)) || !use_power)
 		return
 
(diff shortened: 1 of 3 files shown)


Anewbe:
Did you thoroughly test this? Odd things happen when atmos code gets touched.

Cameron653:
I did a small bit of testing (Making sure it compiled, spawning into the server to make sure everything didn't suddenly break) but nothing too extensive.

I'm going to do some testing real quick to see if the scrubbers are able to scrub all the gasses correctly and whatnot. I'll edit the post with results after I'm done.

Anewbe:
I'm very wary when it comes to changing that code, is why I ask.

Cameron653:
Alright, after some testing here's my results:

Siphoning Section: Seeing if the gasses are siphoned correctly
O: Works
N: Works
N2O: Works
Phoron: Works
CO2: Works

Scrubber Section: Seeing if the gasses scrub correctly:
O: Works
N: Works
N2O: Works
Phoron: Works
CO2: Works

Pressurizing Section: Seeing if the vents correctly put gasses into the room
O: Works
N: Works
N2O: Works
Phoron: Works (why would anyone ever want to do this)
CO2: Works

I think that's everything. Here's the room I used to pump/scrub/siphon the gasses in: 
https://i.gyazo.com/e8fa9deeda3b3c004cbbec8424efd929.png
And here's the setup of how I got the gasses into the atmos system without having to wait years for it to replace the currently existing gas: 
https://i.gyazo.com/73dc707b3576e92b01e7b1a359b5921c.png

Hubblenaut:
Oxygen Agent B doesn't get used anywhere, removing this is perfectly safe.

PR #2898, opened 2017-01-03
Lowers spider egg implantation rate, warns victim

Anewbe:


MagmaRam:
Thank goodness. Would it be possible to add it to the self-examine (ex. "your left hand is bruised and has something moving under the skin") and health scanners as well, or is that for another time?

Anewbe:
Outside the scope of this lazy fix.

PR #2903, opened 2017-01-05
Make Flashes Great Again

Neerti:
Hand-held flashes now do 30 halloss per flash, if the person lacks flash protection.
Hand-held flashes now can be used 10 times per minute instead of five, due to needing ~4 flashes to down someone.
Flash rounds do 20 halloss if the mob hit lacks flash protection.
Cleans up the hand-held flash code, removes copypasta, removes colons.

### DIFF
--- a/code/modules/projectiles/projectile/energy.dm
+++ b/code/modules/projectiles/projectile/energy.dm
@@ -33,7 +33,7 @@
 					H.confused = max(H.confused, flash_strength + 5)
 					H.eye_blind = max(H.eye_blind, flash_strength)
 					H.eye_blurry = max(H.eye_blurry, flash_strength + 5)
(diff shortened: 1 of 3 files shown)


Spacemanspark:
Jesus, that's quite the change.

code/modules/projectiles/projectile/energy.dm
```diff
@@ -33,7 +33,7 @@
 					H.confused = max(H.confused, flash_strength + 5)
 					H.eye_blind = max(H.eye_blind, flash_strength)
 					H.eye_blurry = max(H.eye_blurry, flash_strength + 5)
-
+					H.adjustHalLoss(22) // Five flashes to stun.  Bit weaker than melee flashes due to being ranged.
```
Anewbe: Any chance/logic to have this scale based on flash_strength? There's a shotgun flash round that uses it too... Not really a big deal, though.

PR #2906, opened 2017-01-07
Adds Automuting

Neerti:
Adds a system adapted from Aurora that will automatically mute someone if they say things too quickly in IC, OOC, Deadchat, Ahelp, or Pray.


Anewbe:
Welp, here we go.

PR #2909, opened 2017-01-10
Adds new Server News

Neerti:
Adds ability for admins with R_SERVER perms to be able to set a persistant 'news' announcement that works similarly to admin memos, but for the public.  This is ideal for telling the public things such as when an event is scheduled, directing people to a forum thread, new lore changes, new policies, etc.
The news window allows to define a title, and the body of the text, using an admin verb.  The author and date are added automatically.
Any players can read the news window in the lobby.  The button will bold itself, and display (NEW!), if the player has not seen the news before.  This is done by comparing a hash of the body that the client remembers verses a hash the current news body.
Example image: Example
It should be noted that only one news piece can exist at a time.


Anewbe:
So... Not more IC news?

Neerti:
It's for OOC things like rule changes, event dates, announcements, etc.

PR #2911, opened 2017-01-11
Applies various chem changes to Tajaran

Anewbe:
Discussed with the lorehead.
- Most chems are now metabolized more quickly.
- Taj get drunk from less alcohol
- Coffee is once again dangerous for Taj
- Hyperzine and Stimm are more dangerous for Taj


Anewbe:
Gonna tweak the way the Hyperzine toxicity works, not sure playing with the strength is the best decision.

PR #2917, opened 2017-01-14
Space carp have a chance to stick around after their events

Anewbe:
1 in 4 chance that any given space carp will not be deleted at the end of a carp event. Most of them will still  leave, though.

### DIFF
--- a/html/changelogs/Anewbe - Carp.yml
+++ b/html/changelogs/Anewbe - Carp.yml
@@ -22,7 +22,7 @@
 #################################
 
 # Your name.  
-author: N3X15
+author: Anewbe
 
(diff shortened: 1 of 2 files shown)


html/changelogs/Anewbe - Carp.yml
```diff
@@ -0,0 +1,36 @@
...
+#   rscdel (general deleting of nice things)
+#   imageadd
+#   imagedel
+#   maptweak
+#   spellcheck (typo fixes)
+#   experiment
+#################################
+
+# Your name.  
+author: N3X15
```
Neerti: N3X15

Anewbe: RAGGLE FRAGGLE FRAGGLE

PR #2925, opened 2017-01-15
Automute tweak

Anewbe:
People have been tripping it by mistake, this should help narrow the focus to spambots


Anewbe:
People keep tripping the filter with IC chatter, merging it.

PR #2927, opened 2017-01-15
Teshari tweaks

Anewbe:
Ports the super hearing from https://github.com/Baystation12/Baystation12/pull/13926
Gives them the Hide verb and they can pass tables like they're not there.


SilverTalismen:
I see no issue with any of this I would assume that Teshari can just run under the tables with how small they are

Anewbe:
Still a bit of a balance concern, but given how rare they are and have been I doubt it'll come up. Still gonna wait, though.

SilverTalismen:
Yeah that makes sense!

PR #2937, opened 2017-01-18
Human mobs spawn in with a random amount of nutrition

Anewbe:
Might give the chef something to do.


SilverTalismen:
Outside of the so many vending machines laying around.

Anewbe:
We'll just pretend those aren't easier.

PR #2943, opened 2017-01-19
Light boxes should be able to hold 24 bulbs/tubes now

Anewbe:


PsiOmegaDelta:
Polaris doesn't have the make_exact_fit() proc?

Neerti:
It does, last I checked.

PR #2955, opened 2017-01-24
Big Ol' Medical Tweaks

Anewbe:
Ports https://github.com/Baystation12/Baystation12/pull/14276

Partial port of https://github.com/Baystation12/Baystation12/pull/15118
1.) Ports a blood clotting chem and a bone fixing chem, renamed and tweaked
2.) Ports Inaprovaline lessening the effects of bloodloss, also slows OxyLoss related brain damage

Lessens the effects of minor heart damage on bloodloss
Prosthetic organs take randomized damage when emp'd, should average out to the old amounts

Spiders now have a special reagent, called Spidertoxin, which is used in making the clotting med. Spiders now produce spider meat instead of xenomeat, the recipes involving xenomeat have been made to work with this instead.
Adds a more powerful anti-toxin, which can cause vomiting but also heals the liver.
Dexalin Plus is weaker, should still be enough in most cases.
Oxycodone and Peridaxon now have side effects. These should be fairly minor, but they may need to be tweaked as we go along.

The suiciding var should now dissipate after some time, assuming the suicidee survives that long.

Death Alarms can be ordered from Cargo.
The clotting med can be ordered from Cargo.

Death Alarms now announce the deaths on general comms
The Supermatter Monitor now announces over general earlier.


Kasuobes:
Hi.

Anewbe:
Yo.

code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Toxins.dm
```diff
@@ -39,6 +39,13 @@
 	color = "#003333"
 	strength = 10
 
+/datum/reagent/toxin/spidertoxin
+	name = "Spidertoxin"
+	id = "spidertoxin"
+	description = "A liquifying toxing produced by giant spiders."
```
zekesulastin: Toxing?

Kasuobes:
Let me know how this works.

PR #2957, opened 2017-01-26
Electrical Revolution Circuit DLC

Neerti:
Lots of new things for integrated circuits.  Biggest feature is the addition of power requirements for most circuitry.  Each assembly now comes with a device cell, which circuits draw power from in order to do things.  Some circuits draw more than others.  The power consumption is listed on the circuit interaction UI.
Circuit Stuff:
- Circuits now require electricity to function.
- Assemblies now start with a device cell.
- Assemblies can be placed in a recharger to charge the device cell inside.
- Large assembly can hold more stuff, specifically double of what the medium assembly can hold, because the large assembly is twice the size of the medium assembly.
- Circuits which are bigger than the assembly itself cannot be placed inside the assembly.
- General reorganization with folder structure for integrated circuits, mostly just removing or splitting files.
- Assemblies now count total space remaining from components' w_class values instead of an absolute part count.
- Changes debug circuit box to spawn all circuits which are obtainable normally.
- Adds new 'smart' category for circuits.  Currently the only circuit that uses it is a new pathfinding circuit.
- Ports toggle button, LEDs, (de)multiplexers from Bay.
- Adds power generation circuits, which include metabolic siphon (for implants), internal energy siphon (for FBP implants), solar cells (for calculators), and two flavors of tesla relays, which draw from APCs nearby.
- Adds power transmission circuits, currently two, which work mostly the same way.  They move power from their assembly's power cell to the target.
- Adds sensor circuit, which when inserted into an assembly, it will output a ref of someone when the holder of the assembly tries to shove the assembly in someone's face.
- Adds video camera circuit, so now drones can be remotely controlled.  Camera is set to the research network.
- Adds very very simple pathfinding circuit, which all it does is output the direction of a target.  If output is fed to a locomotion circuit, a drone will be able to follow you now.
- Adds Not Equals circuit, does what it says on the tin.
- Adds On False Result pin for all logic circuits, which should functionally act as the 'else' part of an if statement.

Other Stuff:
- Access acquiring procs have been changed slightly, to make a future addition easier.  This should have no effect for objects using those procs beforehand.
- Fixes logic bug on Technomancer gloves of regen.


SilverTalismen:
Nice! I should make a character to figure out this circuitry stuff...

PR #2967, opened 2017-01-29
Fixes some things that should make mining more interesting

Anewbe:
Iron and Coal in the walls
Drills aren't stopped by space or cracked sand
Crates will show up from mineral walls, occasionally.
Digging through artifacts doesn't always destroy them


code/modules/mining/mine_turfs.dm
```diff
@@ -483,6 +483,11 @@ var/list/mining_overlay_cache = list()
 				if(prob(50))
 					M.Stun(5)
 			M.apply_effect(25, IRRADIATE)
+			if(prob(25))
+				excavate_find(prob(5), finds[1])
+	else if(rand(1,500) == 1)
```
Techhead0: Next time, instead of `rand(1,500) == 1`, you can just do `prob(0.2)`.

Anewbe: ... Sensible.

PR #2981, opened 2017-02-04
Fixes runtimes

Yoshax:
changes made in #2967 broke stuff due to failing to rename the rest of the things
new null(src) == runtimes

also rand(0.5) == 0
100% of the time

### DIFF
--- a/code/modules/mining/mine_turfs.dm
+++ b/code/modules/mining/mine_turfs.dm
@@ -175,9 +175,8 @@ var/list/mining_overlay_cache = list()
 
 	if(severity <= 2) // Now to expose the ore lying under the sand.
 		spawn(1) // Otherwise most of the ore is lost to the explosion, which makes this rather moot.
-			var/losses = rand(0.5, 1) // Between 0% to 50% loss due to booms.
 			for(var/ore in resources)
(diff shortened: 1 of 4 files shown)


code/modules/mining/mine_turfs.dm
```diff
@@ -175,7 +175,7 @@ var/list/mining_overlay_cache = list()
 
 	if(severity <= 2) // Now to expose the ore lying under the sand.
 		spawn(1) // Otherwise most of the ore is lost to the explosion, which makes this rather moot.
-			var/losses = rand(0.5, 1) // Between 0% to 50% loss due to booms.
+			var/losses = rand(1, 2)/2 // Between 0% to 50% loss due to booms.
```
Techhead0: While functionally the code what the original coder intended from this line, this isn't great in context. It makes ore drops either one half or full, when a wider range is probably desired.

I'd remove this line and use the following below:
```
var/amount_to_give = rand(Ceiling(resources[ore]/2), resources[ore])  // Should result in at least one piece of ore.

Anewbe:
Has this been tested?

Yoshax:
yup. Seemed pretty okay.

PR #2985, opened 2017-02-05
Adds rugged style PDAs

Anewbe:
Styles are from Bay


nachomeep:
These really don't feel like they fit in with the other sprites ingame.

PR #2991, opened 2017-02-07
Rename mecha power cell

Arokha:
Sort of annoyng to have it named the same as an actual power cell, with a different max charge. If it's going to have a unique charge amount, it should have a unique name. High-cap cells are 10k, super are 20k, so this is some weird between-cell with 15k.

### DIFF
--- a/code/game/mecha/mecha.dm
+++ b/code/game/mecha/mecha.dm
@@ -173,7 +173,7 @@
 		cell = C
 		return
 	cell = new(src)
-	cell.name = "high-capacity power cell"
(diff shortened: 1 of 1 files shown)


nachomeep:
Why not just replace the mechbattery with a super-cap rather than keep the snowflake one?

Arokha:
Don't want to mess with balance.

A more "complete" solution would be to make it either a super or high cap and adjust all mech power costs accordingly. But this was much easier and I assume if I changed that many files it would take a lot longer to review and you'd worry I'm trying to mess with yer stuff.

PR #2993, opened 2017-02-08
Re-fix for #2312

### LINKED ISSUE #2312: Possible to clone departed players.

Players that leave via the residential elevator, teleporter or cryopods remain on the cloning database if they've been scanned. On servers with respawn enabled, if they subsequently join as another character and some doctor mashes the clone button, they wind up yanked out of the character they're playing and stuffed into a clone of the character that left the station.

For players to be removed from cloning databases when the character leaves the game.
#### What actually happened

Suddenly in clonebay without warning while previous body drops braindead to the floor.
#### Steps to reproduce

Get scanned.
Leave game via cryo.
Ghost. Respawn. Rejoin as different character/job.
Have a doctor mash clone.
#### Additional info:

Possible fix by having the cryo pod check for the health implants the cloner uses?

Arokha:
The implant is stored as a string-reference, not an actual pointer, so you have to use locate() to resolve it into an actual pointer to see if that matches the implant we're removing.

Fixes #2312
Fixes https://github.com/VOREStation/VOREStation/issues/196

### DIFF
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -413,7 +413,7 @@
 		if(istype(W,/obj/item/weapon/implant/health))
 			for(var/obj/machinery/computer/cloning/com in world)
 				for(var/datum/dna2/record/R in com.records)
-					if(R.implant == W)
+					if(locate(R.implant) == W)
 						qdel(R)
 						qdel(W)
 


Anewbe:
Tested?

Arokha:
Yes, tested. Works like one would expect. Just removes them from the records DB in the computer.

Unrelated, it's rather hard to scan yourself by yourself on a test server...

Mechoid:
Local servers allow you to join the server as yourself, then log out of BYOND, and join again, to have 2 connected clients, for future testing. BYOND is weird.

Arokha:
Huh. Interesting. I ended up getting into the scanner, aghosting, clicking the computer, and clicking scan then quickly un-aghosting back into body before it has a chance to fail. 😄

PR #2994, opened 2017-02-09
Fixes dragging items with storage onto an occupied hand

Leshana:
* If you mouse-drag certain items onto a hand which is already holding something, the item goes into limbo.
* The reason this happens is that the standard code handling drags is directly calling usr.u_equip, despite the comment on that proc clearly saying never to do that.  So it just throws it to limbo and the next line attempting to put it in hand does nothing becuase the hand is occupied.   Better way is to call unEquip, so the item will drop on the ground if it fails.
* Fixes https://github.com/VOREStation/VOREStation/issues/394


Leshana:
In addition to the reported issue of backpacks, I did a search thru code for other places calling u_equip in the same way.  Found a couple others and fixed those too.

PR #2996, opened 2017-02-12
Lock-on changes

MagmaRam:
Starting slow, this does two things. One, fixes a bug that meant guns took time to lock on. Two, shifts times slightly, 2.5 seconds to lock on and 1 second between shots.


SilverTalismen:
So no more human auto turrets?

PR #2997, opened 2017-02-16
Jukeboxes are coded to have hackable wires, but don't respond to wirecutters or multitools.  Fix that.

Leshana:


Anewbe:
Good find.

PR #2998, opened 2017-02-19
Ports MultiZAS

Anewbe:
Not sure if this'll work, but we sort of need it.


Anewbe:
Meh, might look at it next weekend, don't have brainpower to port it properly atm.

PR #3001, opened 2017-02-20
Adds framework for planetary maps

Neerti:
Adds a bunch of surface code for future planetary maps.  More details later when I finish the writing.

Adds a new planet datum system, which can handle various aspects of a 'planet', such as day/night cycle, weather, time, and temperature.
Adds weather system for Sif.  Various types of weather added include snow, blizzards, hail, rain, storms, or perhaps just cloudy.  Specific weather types can change how much sun reaches the surface, as well the temperature.
Available weather types so far;
- Clear, doesn't do anything by itself, and is the default.
- Overcast, makes it slightly darker.
- Light snow, looks pretty and lowers temperature somewhat, but otherwise harmless.
- Moderate snow, which lowers temperature even more and causes outdoor tiles to accumulate snow.
- Blizzard, which makes it much colder and darker, and makes loads and loads of snow to accumulate.
- Rain, which makes everyone outside get wet over time, meaning it reduces their fire stacks.  In the future it might convert dirt tiles to mud.
- Storm, which currently acts as a more potent version of rain.  In the future I might add thunder and lightning.
- Hail, this type will hurt anyone standing outside.  Melee armor will help against the damage.

Adds outdoor tiles of various types, with some code and sprites adapted from Europa code.
Outdoor tiles have a layering system, which tiles will change to different types of tiles by 'promoting' or 'demoting' to different layers.  Generally the layers go 'rock (bottom), dirt, snow/grass/etc'.
Adds water tiles, which will slow anyone walking into them considerably, cause them to get wet, and apply a new overlay to show them standing in the water.  Lying down in water will cause you to drown, unless you have internals, are a skrell, or otherwise don't need to breathe (robot, changeling, etc).
Adds ```water_act()``` proc, that currently only gets called when someone enters a water tile.  In the future it might get called by things such as washing machines and sinks.
Adds system to have specific tiles slow movement when crossing.  Currently no tiles on the station make use of this so far.
Adds new time datum system, that works kinda like the DateTime class in some programming languages.  Just without the date part.  A subtype is used for Sif Time, as Sif has a 32 hour (and ~40 minutes we're ignoring) clock.
Adds new adjacent turf edge overlaying system, borrowed from Europa, for more prettyness.  May be able to replace the asteroid edge thing we currently have.
Adds snow material, currently it's rather useless.  In the future, snowballs and snowmen/skrell/cats/etc will be possible.
Moves ```can_feel_pain()``` to mob/living level from mob/living/carbon level.
Adds species ability to breathe water.  Currently only skrell make use of it.
Adds Steve's big brother, blue glowing trees.


code/game/turfs/simulated/outdoors/outdoors.dm
```diff
@@ -0,0 +1,109 @@
...
+
+// Called by weather processes, and maybe technomancers in the future.
+/turf/simulated/floor/proc/chill()
+	return
+
+/turf/simulated/floor/outdoors/chill()
+	promote(/turf/simulated/floor/outdoors/snow)
+
+/turf/simulated/floor/outdoors/snow/chill()
+	return // Todo: Add heavy snow.
```
Anewbe: Do you have this written down somewhere that we'll remember?

code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -132,8 +132,9 @@ Please contact me on #coderbus IRC. ~Carn x
 #define L_HAND_LAYER			24
 #define R_HAND_LAYER			25
 #define FIRE_LAYER				26		//If you're on fire
+#define WATER_LAYER				27		//If you're submerged in water.
 #define TARGETED_LAYER			27		//BS12: Layer for the target overlay from weapon targeting system
```
Anewbe: Think this may need to be updated?

code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -1116,6 +1118,18 @@ var/global/list/damage_icon_parts = list()
...
 	if(update_icons)   update_icons()
 
+/mob/living/carbon/human/update_water(var/update_icons=1)
+	overlays_standing[WATER_LAYER] = null
+	var/depth = check_submerged()
+	if(depth)
+		if(!lying)
+			overlays_standing[WATER_LAYER] = image("icon" = 'icons/mob/submerged.dmi', "icon_state" = "human_swimming_[depth]")
+		else
+		//	overlays_standing[WATER_LAYER] = image("icon" = 'icons/mob/submerged.dmi', "icon_state" = "mob_submerged")
```
Anewbe: Seems like a bit of dead-end code?

code/game/turfs/simulated/water.dm
```diff
@@ -0,0 +1,102 @@
...
+	edge_blending_priority = -2
+	movement_cost = 8
+	movement_message = "You swim forwards."
+	depth = 2
+
+/*
+/turf/simulated/floor/water/deep/update_icon()
+	..() // To get the edges.  This also gets rid of other overlays so it needs to go first.
+	icon_state = "abyss"
+*/
```
Anewbe: Is this supposed to be commented out?

code/modules/mob/living/living.dm
```diff
@@ -792,3 +793,10 @@ default behaviour is:
 		update_icons()
 	return canmove
 
+/mob/living/proc/update_water() // Involves overlays for humans.  Maybe we'll get submerged sprites for borgs in the future?
+	return
```
Anewbe: Doesn't do anything yet, or is it updated to be a human specific thing elsewhere?

code/modules/planet/weather.dm
```diff
@@ -0,0 +1,268 @@
...
+		)
+
+/datum/weather
+	var/name = "weather base"
+	var/icon = 'icons/effects/weather.dmi'
+	var/icon_state = null // Icon to apply to turf undergoing weather.
+	var/temp_high = T20C
+	var/temp_low = T0C
+	var/light_modifier = 1.0 // Lower numbers means more darkness.
+	var/transition_chances = list() // Assoc list
```
Anewbe: These may need to be tweaked in the future, from just eyeballing it seems like once weather starts it doesn't look like it'll stop.

code/modules/planet/weather.dm
```diff
@@ -0,0 +1,268 @@
...
+			if(!T.outdoors)
+				return // They're indoors, so no need to pelt them with ice.
+
+			var/target_zone = pick(BP_ALL)
+			var/amount_blocked = L.run_armor_check(target_zone, "melee")
+
+			if(amount_blocked >= 100)
+				return // No need to apply damage.
+
+			L.apply_damage(rand(5, 10), BRUTE, target_zone, amount_blocked, used_weapon = "hail")
```
Anewbe: Can we start it at 3-6? Feels a tad high.

code/game/turfs/simulated/outdoors/outdoors.dm
Neerti: Should be on the etherpad now.

code/modules/mob/living/carbon/human/update_icons.dm
Neerti: Merging is fun.

code/game/turfs/simulated/outdoors/outdoors.dm
Anewbe: Excellent.

code/modules/mob/living/living.dm
Neerti: It's there so other things can inherit it, like borgs, and how borg icon code is different from human icon code unfortunately.

code/modules/planet/weather.dm
Neerti: If you mean the base weather datum, that one never gets used.

code/modules/planet/weather.dm
Neerti: Autohealing will probably remove the damage regardless but I can tweak it.  Also remember it only fires once a minute.

code/game/turfs/simulated/water.dm
Neerti: Was an old method which should've been deleted, which I will do shortly.

code/modules/planet/weather.dm
Anewbe: No, I just put that comment there because I'm terrible at decided locations. Basically my concern is that most weather types have themselves in the transition_chances list, or other very similar weather, which could, as far as I could see, mean you get overcast > rain > storm > rain > rain > etc and never go back to "good" weather. Not sure that's a bad thing though, in all honesty.

code/modules/mob/living/living.dm
Anewbe: Deleted the original comment because I took two seconds to look and found the human specific version, feel free to disregard.

code/modules/planet/weather.dm
Anewbe: Oh, that's hella low then.

code/modules/planet/weather.dm
```diff
@@ -0,0 +1,268 @@
+#define WEATHER_CLEAR				"clear"
+#define WEATHER_OVERCAST			"overcast"
+#define WEATHER_LIGHT_SNOW			"light snow"
+#define WEATHER_SNOW				"snow"
+#define WEATHER_BLIZZARD			"blizzard"
+#define WEATHER_RAIN				"rain"
+#define WEATHER_STORM				"storm"
+#define WEATHER_HAIL				"hail"
+#define WEATHER_WINDY				"windy"
+#define WEATHER_HOT					"hot"
```
Anewbe: Neither WEATHER_WINDY or WEATHER_HOT are used, it seems?

code/modules/planet/weather.dm
```diff
@@ -0,0 +1,268 @@
...
+
+/datum/weather_holder
+	var/datum/planet/our_planet = null
+	var/datum/weather/current_weather = null
+	var/temperature = T20C
+	var/wind_dir = 0
+	var/wind_speed = 0
```
Anewbe: Don't think these winds are ever checked or modified either, are we going to implement that at some point?

Neerti: Future stuff, most likely.

code/modules/planet/weather.dm
Neerti: Future plans mostly.

Anewbe: Added to the etherpad.

code/modules/planet/weather.dm
Anewbe: Added to the etherpad.

code/modules/planet/weather.dm
Neerti: The transitions happen in 10 to 15 minute intervals.  The intention with placing themselves in the probability is to act as a 'do nothing, keep raining/snowing/whatever', just so that if it snowing, there's no guarantee that it'll stop in a short period of time.

code/modules/mob/living/carbon/human/update_icons.dm
Neerti: Should be fixed now.

code/modules/planet/weather.dm
Anewbe: Alrighty, suppose we'll see how it goes then.

PR #3003, opened 2017-02-21
Ports a bunch of fluff stuff

Anewbe:
- Adds a hairstyle from Bay
- Ports an undershirt from Bay
- Ports Nitrile gloves from Bay (In the loadout and medical maint)
- Ports High-Top shoes from Bay (Not yet in the loadout)
- Ports cloth footwraps from Eros (In the xenowear loadout)
- Ports the Blue-Edged labcoat from Bay (In the loadout)
- Allows roller beds to be carried on one's back
- Jumpsuits that cannot roll down or roll sleeves should no longer display the verbs


Anewbe:
Map level 5 should be the same overall, map merger wasn't working and I needed to repath something.

PR #3005, opened 2017-02-22
Enhances the Paper Shredder

Leshana:
* Makes the paper shredder constructable and deconstructable.
* Switches to some new sprites which are animated!
* Machine now requires power, icons respond to power on/off etc.

image

Resolves https://github.com/VOREStation/VOREStation/issues/991


code/modules/paperwork/papershredder.dm
```diff
@@ -122,18 +187,3 @@
 		M.drop_from_inventory(src)
 	PoolOrNew(/obj/effect/decal/cleanable/ash,get_turf(src))
 	qdel(src)
-
```
Leshana: Simply re-ordered this code so the type definition precedes its procs.  Looks nicer.

code/modules/paperwork/papershredder.dm
```diff
@@ -92,8 +125,40 @@
 	paperamount--
 	return PoolOrNew(/obj/item/weapon/shreddedp, get_turf(src))
 
+/obj/machinery/papershredder/power_change()
+	..()
+	spawn(rand(0,15))
```
Neerti: Not sure why a randomized spawn() is needed, but that's pretty minor.

Leshana: Several other machines have this random delay (fire, air alarms etc).   It makes it so when power is restored to a room full of stuff it all flickers to life instead of instant-on.  Which is cool and cinematic, so I copied it.

Neerti: Fair enough.

PR #3006, opened 2017-02-22
Don't give synthetics the survival kit

Arokha:
What're they gonna do with it, anyway? Other than just make a pile of them in arrivals.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/species.dm
+++ b/code/modules/mob/living/carbon/human/species/species.dm
@@ -187,13 +187,18 @@
 	return sanitizeName(name)
 
 /datum/species/proc/equip_survival_gear(var/mob/living/carbon/human/H,var/extendedtank = 1)
-	if(!H.isSynthetic())
-		if(H.backbag == 1)
-			if (extendedtank)	H.equip_to_slot_or_del(new /obj/item/weapon/storage/box/engineer(H), slot_r_hand)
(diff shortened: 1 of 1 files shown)


Anewbe:
The box itself is actually pretty nice to have...

SilverTalismen:
Yeah atleast leave them the box, I myself use it as a synthetic for engineering or to store minor stuff in

Arokha:
Ok, lemme do that.

Arokha:
There. So, you get an empty box if you're a synth. And if you wanted to be crazy could add more conditions if there were ever more box types that showed up.

PsiOmegaDelta:
Well, from an OSHA point-of-view it makes some sense to enforce that everyone in an artificially pressurized environment have emergency equipment even if they may not personally have use for it.
You know, if case a synth happens to common across an unconscious meatbag colleague in a room currently experiencing drastic pressure loss.

Anewbe:
I'm fine with it going either way, honestly. I can poll some folks and see who wants what, if y'all want that.

Arokha:
I mean. Shouldn't everyone also get a suit cooler then? Or Vox phoron tanks? You're generally not required to wear PPD for OTHER PEOPLE on OSHA worksites.

nachomeep:
No. Because suit coolers are big and unwieldy, and Vox don't exist as crew-members.

Arokha:
Well. Do what you will with it.

PR #3007, opened 2017-02-22
Fix construction and deconstrution of PACMANs

Leshana:
* Modernized to use modern machine code instead of doing it itself.
* Made it compatible with modern circuits and machine frames.
* Also fixed some color macros to use CSS while I was there.

Fixes https://github.com/VOREStation/VOREStation/issues/989

### DIFF
--- a/code/modules/power/port_gen.dm
+++ b/code/modules/power/port_gen.dm
@@ -10,7 +10,6 @@
 
 	var/active = 0
 	var/power_gen = 5000
-	var/open = 0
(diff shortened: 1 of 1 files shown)


Anewbe:
Seems to be fine at a glance, will do a more thorough review some time later today or tomorrow.

Anewbe:
Did you test this?

Leshana:
I tested deconstructing and reconstructing a regular pacman. I did not re-test on super or mrs.

PR #3011, opened 2017-02-23
Tasers can hurt simple mobs

Anewbe:
Stun weapons can now hurt simple mobs at half the rate that they stun people. Should let Sec handle spiders/carp at range, without needing someone to open the armory.

### DIFF
--- a/code/modules/mob/living/living_defense.dm
+++ b/code/modules/mob/living/living_defense.dm
@@ -92,13 +92,6 @@
 			src.visible_message("\red [src] triggers their deadman's switch!")
 			signaler.signal()
 
-	//Stun Beams
-	if(P.taser_effect)
-		stun_effect_act(0, P.agony, def_zone, P)
-		src <<"\red You have been hit by [P]!"
-		qdel(P)
-		return
-
 	//Armor
 	var/absorb = run_armor_check(def_zone, P.check_armour, P.armor_penetration)
 	var/proj_sharp = is_sharp(P)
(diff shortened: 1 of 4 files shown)


Leshana:
This seems like it will do _actual_ damage to simple mobs from stun weapons, not just stun them.
Also, what I am wondering is why the changes to /mob/living/bullet_act ?

Neerti:
Doing actual damage is the intended effect, I believe.

Anewbe:
The damage is intended. It's supposed to kill the simple mobs, means I don't have to give secborgs an Egun, and that a bog-standard Sec Officer can try to take out carp without going EVA.

The changes to /mob/living/bullet_act() fixes a bug that made the shotgun stunshell not do the actual damage it's supposed to do.

nachomeep:
How much damage does it do relative to an energy gun?

Neerti:
It does half of the agony damage it would have done to a humanoid mob as actual damage.
Eguns fire a weaker version (25 agony) of the hitscan projectile that tasers fire (40 agony), so that eguns aren't straight upgrades from tasers.  To a simple mob, it would deal 12.5 damage per shot from an egun, or 20 per shot from a taser.
Stun revolvers fire a different projectile, one which isn't hitscan, and is a bit stronger (55 agony), meaning it would do 27.5 damage.
Shotguns loaded with stunshells (I love those) do the most agony (80 agony), so it will hit for 40 agony, although using them on carp feels like a waste to me.

This means that in general, it's more efficient to just use a laser, as those do 40 damage always, however if you only have a taser, it's not useless if spiders start trying to eat the crew.  This situation can arise when there's no warden, or if you're a security borg.

Anewbe:
Neerti's correct on the goal here. This isn't trying to make stun weapons the go-to for this sort of thing, but rather to let them act as a weapon of desperation. Because batoning a spider to death is a tad suicidal.

PR #3018, opened 2017-02-25
BlockHeadHair doesn't make players bald.

nachomeep:
Instead, it replaces their hairstyle with short hair [unless their hair was already short enough to not have any clipping issues]. Also replaces a lot of 'BLOCKHAIR' tags with 'BLOCKHEADHAIR', because chaplains should have beards visible when their hoods are on.

Tested. It works. No clipping issues found. Point out any other hair that should be allowed and any that should be disallowed, and any items that need tag updates regarding BLOCKHAIR and BLOCKHEADHAIR.

Mostly ported from - https://github.com/Baystation12/Baystation12/pull/14239

### DIFF
--- a/code/modules/mob/new_player/sprite_accessories.dm
+++ b/code/modules/mob/new_player/sprite_accessories.dm
@@ -51,8 +51,8 @@
 	bald
 		name = "Bald"
 		icon_state = "bald"
-		species_allowed = list("Human","Unathi")
 		gender = MALE
+		species_allowed = list("Human","Unathi")
 		veryshort=1
 
 	short
@@ -544,7 +544,7 @@
 	tightbun
 		name = "Tight Bun"
 		icon_state = "hair_tightbun"
-		gender= FEMALE
+		gender = FEMALE
 
 	coffeehouse
 		name = "Coffee House Cut"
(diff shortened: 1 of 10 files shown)


nachomeep:
Forgot a changelog. Will be commiting in a bit.

code/modules/mob/new_player/sprite_accessories.dm
```diff
@@ -547,12 +562,10 @@
 /datum/sprite_accessory/facial_hair
 
 	icon = 'icons/mob/Human_face.dmi'
-	gender = MALE // barf (unless you're a dorf, dorfs dig chix /w beards :P)
```
Neerti: Not use if removing this actually does anything, but this appears to be an undocumented change.

nachomeep: Oh, sorry. I'll change that back- that was for testing purposes.

Anewbe:
Mind going back through this and not touching any of the gender related flags?

nachomeep:
Not at all.

Anewbe:
Alright, code looks fine now. Have you tested this?

nachomeep:
I have. Haven't found any clipping issues or anything thus far.

PR #3022, opened 2017-02-27
Infection code improvement

Arokha:
**Brain: Based on Meningitis**
- L1: Stiff neck
- L2: Severe headache (+1 confusion randomly)

**Eyes: Based on Conjunctivitis**
- L1: Itchy eyes
- L2: Blurred vision (+10 eye_blurry randomly)

**Heart: Based on Endocarditis** (there are no actual common heart infections)
- L1: Chest tightness
- L2: Chest pain (+25 halloss)

**Kidneys: Based on Pyelonephritis**
- L1: Lower back pain
- L2: Malaise (randomly slow to walking)

**Liver: Based on Pyogenic Abscess**
- L1: Right-side abdomen pain
- L2: Poor blood filtration (capped toxins buildup) (not realistic)

**Lungs: Based on Pneumonia**
- L1: Coughing
- L2: Shortness of breath (+30 oxyloss)

**All External Organs: Based on Staph infections**
- L1: Painful notification
- L2: Sharp pain disables limb briefly (drop/trip/blind)

**Septicemia/Toxic Shock Syndrome:**
When organs reach L3 infection they go numb and become useless
They generate 1 tox per life tick, plus a function of how infected they are (in increments of 0.25). This will rapidly become unmaintainable. Though there's really no reason you should have let your infection get this far.

Advanced Medical Scanner shows necrotic/gangrenous organs, and shows internal organ infection severity more accurately.

Organ infection spreads properly up and down through body parts. Previously it would infect adjacent body parts, but, those parts may never actually begin processing. Then if their germ_level was 1000+ and someone tapped them with a wrench, they'd process and instantly die. This would also mean spaceacillin wouldn't fix them.

It looks like infections were coded incorrectly originally. Seems like they would call up to organ/handle_germ_effects() ... until a certain point, at which point they would stop. Considering organ/hge() handled the fever, this would cause fever to disappear. Now it's all arranged properly.

### DIFF
--- a/code/game/machinery/adv_med.dm
+++ b/code/game/machinery/adv_med.dm
@@ -333,6 +333,8 @@
 					organStatus["splinted"] = 1
 				if(E.status & ORGAN_BLEEDING)
 					organStatus["bleeding"] = 1
+				if(E.status & ORGAN_DEAD)
+					organStatus["dead"] = 1
 
 				organData["status"] = organStatus
 
--- a/code/game/machinery/adv_med.dm
+++ b/code/game/machinery/adv_med.dm
@@ -507,9 +507,9 @@
 						infected = "Acute Infection:"
 					if (INFECTION_LEVEL_TWO + 200 to INFECTION_LEVEL_TWO + 300)
 						infected = "Acute Infection+:"
-					if (INFECTION_LEVEL_TWO + 300 to INFECTION_LEVEL_TWO + 400)
+					if (INFECTION_LEVEL_TWO + 300 to INFECTION_LEVEL_THREE - 50)
 						infected = "Acute Infection++:"
-					if (INFECTION_LEVEL_THREE to INFINITY)
+					if (INFECTION_LEVEL_THREE -49 to INFINITY)
 						infected = "Gangrene Detected:"
 
 				var/unknown_body = 0
@@ -549,9 +549,9 @@
 						infection = "Acute Infection:"
 					if (INFECTION_LEVEL_TWO + 200 to INFECTION_LEVEL_TWO + 300)
 						infection = "Acute Infection+:"
-					if (INFECTION_LEVEL_TWO + 300 to INFECTION_LEVEL_TWO + 400)
+					if (INFECTION_LEVEL_TWO + 300 to INFECTION_LEVEL_THREE - 50)
 						infection = "Acute Infection++:"
-					if (INFECTION_LEVEL_THREE to INFINITY)
+					if (INFECTION_LEVEL_THREE -49 to INFINITY)
 						infection = "Necrosis Detected:"
 
 				dat += "<tr>"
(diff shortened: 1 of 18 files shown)


Arokha:
Ok. I think I'm mostly done. Just got through testing various things. Does anyone see anything they don't like? L3 infections can kill you with fair rapidity but to let an infection get to L3 means you have to have been ignoring pain messages for like 30 minutes.

Neerti:
Everything looks fine so far, and L3 being deadly isn't a problem (since it should kill you at some point).

Arokha:
Ok. Well I'm not sure what else I can add. Infections do 'interesting things' for organs now that make them obvious, and they process properly. I spent some time on a test server with someone treating various levels of different infections. Seems to work like one would expect. It takes a long while to go from the L1 to L3 infections, like 30 minutes, so there should be plenty of warning time.

Wickedtemp:
Can confirm, I'm the someone that was testing the infections with Aro. Stage Three infections aren't an outright death sentence, if you get to Medical ASAP (which you already should have done, unless you're AFK or powergaming and ignoring pain) and get a Cillin OD. I personally think the toxins are strong enough, one gangrenous infection was more powerful than what Dylovene and Tricord could heal, it took adding Arithrazine to lower toxins, but even that only helped for about another 40 seconds before toxins began rising. Cryo did the trick, but I'm sure a gangrenous set of lungs would be much more annoying to deal with than the rotting leg we were testing.

Of course, Cillin OD makes it much easier to deal with. I'm happy with it, I wouldn't mind it being more dangerous, but I am happy with how it is.

Anewbe:
The probs seem a bit low across the board, roughly how often do they pop up?

Also not sure if our necrotic organ rejuvenation surgery is working, did you test that for the dead bits?

Arokha:
The probs actually come up a decent amount. You have to consider the TIME over which this happens. It's a 1/100 chance over like 3000 mob life ticks. Also the external organs are '.' prob which is 1% at L1 and doubles to 2% at L2 so they show up more, then.

I had to fix the necrotic organ surgery, yeh. It was actually two problems in the dropper. One is that the dropper only worked on non-help intent, and the other was that the surgery check would look for peridax _in the dropper after using it_, like it would check before starting the step to make sure you have peridax, then check in the dropper after. So you had to disarm-intent-click with an eyedropper set to have a transfer amount of 1u and have more than 1u in the dropper. Otherwise it wouldn't work. 😄 

I fixed that, yes.

I actually see one more thing I want to change, so I'm gonna do that real quick.

Arokha:
I had forgotten to make ORGAN_DEAD limbs show up with a message in the scanner. They would have been obvious in a visual exam and likely very infected, but if for some reason you spend all the time fixing their infection in all their limbs and forget which ones had gangrene, this makes it easier to figure out which limb you need to ~~cut off~~ surgery. Oddly, I remembered to add it on the printout already.

Arokha:
There. Can you remove the WIP label, too? I don't have anything else to change.

PR #3029, opened 2017-03-01
Make blood loss/heart removal make sense

Arokha:
Right now it gives you 300 toxin damage "to be sure" you die. That's a little odd. Especially on health scanners after someone has bled to death. And you can't do heart transplants. We can do heart transplants now in 2017. 3500 per year. Pretty sure they won't forget how by 2561 or whatever.

This makes having no heart or too little blood do 75 oxyloss per tick. Enough to kill you eight seconds if you have no special care. It becomes 15 per tick with dexP which is fast enough to kill you "pretty fastly" and slow enough you can swap out a heart in surgery.

With dexP+dex+tric it becomes a meager 10ish per tick, but an inevitable 10 per tick as they are literally now full of every oxyloss removing drug in the game. So you can survive a part of a minute, max, with not enough blood. You're still paralyzed and eye_blurry for most of it. You should probably be in surgery.

### DIFF
--- a/code/modules/organs/blood.dm
+++ b/code/modules/organs/blood.dm
@@ -109,12 +109,14 @@ var/const/BLOOD_VOLUME_SURVIVE = 40
 			if(prob(15))
 				var/word = pick("dizzy","woosey","faint")
 				src << "\red You feel extremely [word]"
-		else
-			// There currently is a strange bug here. If the mob is not below -100 health
-			// when death() is called, apparently they will be just fine, and this way it'll
-			// spam deathgasp. Adjusting toxloss ensures the mob will stay dead.
-			toxloss += 300 // just to be safe!
-			death()
+		else //Not enough blood to survive (usually)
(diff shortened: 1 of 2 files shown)


Arokha:
Ok. This is going to make me sound dumb.

@Anewbe How do changelog pls?
Can I just copy someone's and add it in a commit?

Anewbe:
So there should be a bunch of .yml files in a folder called polaris/html/changelogs. Copy the one called example.yml, and there should be further instructions in there.

Arokha:
This is right, yes?

Anewbe:
Yep

Anewbe:
Just remembered something, you'll need to go into either organs/organ.dm and redo how vital organs work, or you need to go to wherever the heart organ is defined and remove the vital tag, otherwise they'll still just die.

Arokha:
Yeah, just brain, I was surprised at that too.

PR #3030, opened 2017-03-01
Fixing runtimes, making things functional for the Xenobiology Console

Dankman:
The old build path pointed towards a circuitboard instead of towards the
machine, meaning when you screwdrivered the frame twice, it would cause
frame.dm to runtime, and then not delete the frame, meaning you could
spam click the console with a screwdriver and spawn infinite boards, and
still be stuck with a non functional machine.

injector.computer = null was causing a runtime because it was attempting
to edit something that no longer existed on destruction.


Dankman:
Slight error. Will resubmit.

PR #3031, opened 2017-03-01
Fixing issues with the Xenobiology Computer.

Dankman:
injector.computer=null was being called after deletion, causing it to runtime. This resolves that issue by placing it inside the operation instead of outside.

The build_path for the computer was pointing towards the circuiboard as the end result instead of the machine, causing a runtime in frame.dm when you screwdrivered the console twice, the computer wasn't reconstructable.


Anewbe:
Tested?

Dankman:
Yeah, tested.

PR #3033, opened 2017-03-02
Fixes descriptions on 3 surgical tools

Arokha:
Someone forgot theeeese. So they showed the "Ahelp this" description.


Anewbe:
... That'd be my bad, thanks for catching that.

PR #3035, opened 2017-03-03
Adds Multi-Z Test Map

Neerti:
Adds an example map to test multi-z code, which includes pipes/disposals/wires/ladders/stairs/etc.  Will have a test turbolift when that gets ported.
Ports lots of multi-z code from Bay.
Ports cable code from Bay, which lets it go up/down and is better than the current cable code.
Makes our ladders act like Bay's.
Made falling down a z-level hurt a lot more and make a sound.
Makes jetpack emissions visible if used indoors.

To use the map, untick northern_star.dm and tick example.dm, then compile.

Yes, the game will call the map the Funhouse.

Pictures:
First floor
first-z-level
Second floor
second-z-level


Neerti:
Should note this also fixes the roundstart runtime involving the fake centcom announcement.

PR #3038, opened 2017-03-03
Ports several Paradise emote sounds

Arokha:
In total this adds or updates with sound: sneeze, cough, rsneeze, rcough, rscream, snap, slap.

Split robo-cough and robo-sneeze and robo-scream into their own *rcough, rsneeze, rscream because we don't really have it as a species, and their logic of only allowing the species to do it wouldn't work. It checks if you're synthetic like the existing synthonly emotes.

They have these as species vars so I did that. If someone wants to get cat sneeze noises for Tajarans or something, you know. Feel free.

The scream sound is commented out for now... the sound just doesn't... sound right for HRP servers? Maybe someone can find a better scream sound. These seem way too long.

No superfart. "Sorry" ~~notsorry~~


Arokha:
You know, this PR came across shitty what with the commit messages. I'm redoing it.

PR #3041, opened 2017-03-04
Revert "BlockHeadHair doesn't make players bald."

Anewbe:
Breaks hair color on FBPs, from what we can tell.

@nachomeep If you could look into that, it'd be appreciated.


nachomeep:
After looking into it, I don't have a fucking clue what went wrong.

PR #3042, opened 2017-03-04
Fix server startup runtimes from uninitaizlied var

Leshana:
* That if statement wasn't heping anyone as it was.
* Strangely however, even an `if(using_map)` doesn't help, as using_map is in some crazy BYOND limbo state at that point in world init.  but checking istype() succeeds in figuring out its not real.

### DIFF
--- a/code/_helpers/names.dm
+++ b/code/_helpers/names.dm
@@ -16,7 +16,7 @@ var/church_name = null
 	return name
 
 /proc/command_name()
-	if (using_map.boss_name)
+	if(istype(using_map))
 		return using_map.boss_name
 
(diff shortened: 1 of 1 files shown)


Leshana:
Just for reference, I mean the runtimes that are thrown during current build's travis checks:
```
world: Error(Cannot read .boss_name (/exception), null)
command name()
/datum/uplink_item/abstract/an... (/datum/uplink_item/abstract/announcements/fake_centcom): New()
init subtypes(/datum/uplink_item (/datum/uplink_item), /list (/list))
/datum/uplink (/datum/uplink): New()
```

PR #3046, opened 2017-03-05
Ports Turbolifts from Bay.

Neerti:
Ports code for turbolifts, and adds them to the test map.
Pictures:
First floor
Second floor
Also fixes the %ETA% shuttle bug.


SilverTalismen:
It would be great if there was a way to sabotage the elevators..hm.. Looks great anyway

Neerti:
You can use them to crush people.

SilverTalismen:
I more so mean in a emag way, l,ike emag and it drops from the highest point it has avaliable to the lowest very quickly or something?

Neerti:
Kinda against the trend of 'everything needs to be emaggable', also this is just straight port (with a bugfix too) so that's a bit outside of scope.

SilverTalismen:
Mostly a thought I had, Not saying it has to xD

Anewbe:
Tested?

Neerti:
Those pictures wouldn't be possible if it wasn't tested.

PsiOmegaDelta:
Bug fix?

Neerti:
Just the %ETA% thing.

PR #3049, opened 2017-03-05
Updates gun sprites.

nachomeep:
Energy pistol's a pistol now, because it can be holstered and is equal in size to the ballistic pistol.
Tazer just looks dank.

http://i.imgur.com/NUrdnur.png

Stolen from: https://github.com/NEV-Unity/NEV-Unity/


Screemonster:
Ew. Not a fan of the egun at all.
The taser... with all the loose trailing wires and such, looks more like some jerry-rigged pirate special than something professionally built that a security force would use. Cool sprite, better to use it for something else.

nachomeep:
Don't really expect this one to get merged, to be entirely honest. I just like the sprites so I threw it out anyway, just in hopes it **might**.

Screemonster:
tbh, it might have been better to just _add_ the sprites as extra resources in case anyone wanted to use them for something, rather than just outright deleting the existing sprites.

nachomeep:
Fair enough. If this is shot down, I have plans to re-use the pistol one anyway.

Anewbe:
Will need to pass this around to the staff.

nachomeep:
Do note I can easily remove the e-gun sprite if not wanted. I am, however, a fan of the taser, especially the inhand.

Neerti:
Screemonster's thoughts mirror my own.  It'd be nice for a gun to give to heisters tho.

nachomeep:
I'll figure something out with them at some point. Maybe.

SilverTalismen:
Why not go the route of the detectives guns and the choose a sprite thing they have?

nachomeep:
I never liked that myself, so I won't be doing that.

PR #3050, opened 2017-03-05
Updates most of the device sprites.

nachomeep:
http://i.imgur.com/91IqbVH.png

Stolen entirely from https://github.com/Endless-Horizon/CEV-Eris


nachomeep:
I lied about commiting less, apparently.

Anewbe:
Will need to pass this around to the staff.

Arokha:
Looks neat!

Screemonster:
I like some of them, not so keen on others, and would _really_ prefer if the existing sprites were left in the .dmi rather than outright deleted. Just rename them to (sprite)_old or something.

nachomeep:
I'm waiting for feedback of which to keep the old ones for. I, personally, don't like keeping old sprites- bloat that isn't needed. I won't be doing that unless a dev tells me to.

Anewbe:
Some of these sprites are nice, but we would like to keep the old sprites in use, at least for now. However, if you can rework this PR to instead just add the new sprites for later use, we may have use for them in some capacity further down the road.

nachomeep:
Righto. I'll go ahead and do that [later im busy now].

nachomeep:
I love these sprites, but some people have been complaining about taking Eris sprites. I'm just gonna close it to avoid their wrath, considering they'd probably end up in limbo like half of suits.dmI anyway.

Arokha:
Aww. I liked a lot of 'em.

PR #3052, opened 2017-03-05
Reworks empulse and emp_act

Anewbe:
EMP now has four levels of effect, instead of two.
- Level 1 is still the strongest effect, should be basically the same
- Level 2 is now the second strongest effect, meaning it'll hurt, but not as much
- Level 3 is fairly close to the old level 2, tends to be roughly half as strong as Level 1.
- Level 4 is the weakest effect. Actual effect is usually pretty minor.

Also added in an ion pistol, has 5 shots instead of 10, otherwise acts like the ion rifle.


Anewbe:
@Neerti If you have places you think it makes sense to not have the highest tier emps, please mark them.

PR #3061, opened 2017-03-07
Adds robotics labcoat.

nachomeep:
Also adjusts wardrobe to be consistent with the roundstart equipment, because they don't get scientists satchels or bags

Stolen from: https://github.com/ErosResearchPlatform/Eros/pull/5

http://i.imgur.com/tBMv6Hh.png


nachomeep:
I knew I shouldn't have made a last-minute change without compiling.

Anewbe:
I'm not sure this is an improvement, actually. The current roboticist setup looks good (in my opinion) because of the contrast between the white labcoat and the red and black of the actual uniform. This would just sort of turn the whole body into a black and red curtain.

nachomeep:
Fair enough. I really don't like contrast, personally. I'll wait for another comment or two telling me its bad before shutting it down.

nachomeep:
Labcoat is no longer mapped or in loadout.

PR #3063, opened 2017-03-08
Changes tank sprite for Vox

Anewbe:
Makes use of and corrects and undocumented sprite change.


nachomeep:
Adjust assemblies.dmi as well.

PR #3064, opened 2017-03-08
WR_SC_Map_Changes

Woodratt:
Changes to some files and additions in order to help with keeping the SC map up to date with the git. Should not change anything with the current map.

- Fixes sprite issues with departmental intercoms.
- Tweaks to the random objects dm
- Copies Bay's flora and warning sign list, as well as ports some of their signs.
- Addition of a couple new floor decals.
- New table preset, wooden reinforced.


code/game/objects/structures/flora.dm
```diff
@@ -205,4 +201,149 @@
...
+	name = "dead potted plant"
+	desc = "This is the dried up remains of a dead plant. Someone should replace it."
+	icon_state = "plant-25"
+
+/obj/structure/flora/pottedplant/decorative
+	name = "decorative potted plant"
+	desc = "This is a decorative shrub. It's been trimmed into the shape of an apple."
+	icon_state = "applebush"
+
+/obj/structure/flora/pottedplant/xmas/New()
```
Anewbe: Don't think this should be a New(). The New() you replaced basically only renamed the item, so it's unnecessary now, as far as I can tell.

PR #3067, opened 2017-03-09
Give internal cell its own icon

Arokha:
Was a little weird to reuse the supercell... there are already WAY TOO MANY "not an actual powercell" powercells. So this one is just repainted blue and joins all the other prosthetic organs in surgery.dmi. The icon line can just be removed because the parent points to this file.

Did this because currently if you print this from a prosfabricator, it prints with no icon, due to the oddness of how prosthetic organs work, with appending _prosthetic to the iconstate. So just did this while fixing that.

image


Arokha:
Bug too, really, since invisible printed batteries.

PR #3068, opened 2017-03-10
New Event: Canister Leak

Leshana:
Somewhere on the station, a gas canister is faulty and ruptures, releasing its contents (no explosion).
* Added as an event_manger event.  This version is nice on low severity and picks a canister with nobody nearby.
* Added helper method to check area occupancy.
I see that there has been work done on the "gamemaster" event manager, so:
* Also added as a gamemaster event. This one doesn't do any mansy pansy checks for people standing near a high pressure canister on a space outpost without being prepared for atmos problems.


SilverTalismen:
All funa nd games until a plasma one ruptures and a atmos tech is smoking near it :) Support from me!

Neerti:
Might want to add a check to ensure the canister contains a sufficent amount of gas before choosing to burst it, as otherwise an empty canister might get picked and it wouldn't do anything.

SilverTalismen:
Another idea, perhaps for the new map later, is that tanks left out in the elements get damaged overtime depending on the weather and have a higher chance to burst?

Leshana:
I've added a check that the canister has some amount of gas in it, so it will only pick ones that will have at least some effect.

PR #3072, opened 2017-03-10
Adds engineering + science crates

nachomeep:
Also enables them for engineering/robotics equipment. 
image

Ported from https://github.com/d3athrow/vgstation13


nachomeep:
Note that even hydroponics has their own crate style already.

Anewbe:
The issue isn't so much with the crates themselves, as much as it is with the sprites in the image. The engineering ones especially are a bit of an eyesore. I think this'd be much better if the departmental crates just had a colored line, like the mentioned hydroponics crate, because they don't really feel like they fit with the rest of the crate sprites.

nachomeep:
Fair enough. What's the opinion on the scicrates, then? I actually really like those ones.

Anewbe:
Those seem fine, to me.

uraniummeltdown:
the engineering ones are actually from tgstation

nachomeep:
So they are. Thank you. I'll be replacing them anyway so it doesn't matter, though.

nachomeep:
New engicrates: image

code/datums/supplypacks/atmospherics.dm
```diff
@@ -11,7 +11,7 @@
 	name = "Inflatable barriers"
 	contains = list(/obj/item/weapon/storage/briefcase/inflatable = 3)
 	cost = 20
-	containertype = /obj/structure/closet/crate
+	containertype = /obj/structure/closet/crate/secure/engineering
```
Anewbe: Not sure this warrants a secure crate.

nachomeep: I don't actually remember why I made it secure, you're right.

code/game/objects/structures/crates_lockers/crates.dm
```diff
@@ -283,7 +283,7 @@
 /obj/structure/closet/crate/rcd
 	name = "\improper RCD crate"
 	desc = "A crate with rapid construction device."
-	icon_state = "crate"
+	icon_state = "engi_crate"
 	icon_opened = "crateopen"
 	icon_closed = "crate"
```
Anewbe: You sure these other two lines don't need to be updated too?

Anewbe: Think these other two lines might need to be tweaked a bit.

nachomeep: ...oh, right.

PR #3074, opened 2017-03-10
"Recipe Dump" admin verb

Arokha:
The idea of updating the wiki list of recipes is daunting, especially getting all the icons and resizing them to look nice on a webpage, so fuuuuuuuuck that.

I made a verb that does it. It dumps upsized (64x64) images of every food icon and a recipes.html file into your cache with a table that contains all the recipes, nicely formatted, in alphabetical order (the -burger and -kebab at the top are because those are [whatever-meat]-burger and [whatever-meat]-kebab).

Then you can just throw them on a website. It will include a recipes.css file if it's in the same dir.
Here's the result on our server:
~~https://vore-station.net/infodump/recipes.html~~
https://vore-station.net/infodump/recipes_food.html
https://vore-station.net/infodump/recipes_drinks.html

EDIT: Because I'm absolutely crazy, I added a separate drinks one that is output as well.

(As a note, the date at the top is not edited in, that's done by the dump)

### DIFF
--- a/code/modules/reagents/Chemistry-Recipes.dm
+++ b/code/modules/reagents/Chemistry-Recipes.dm
@@ -557,13 +557,26 @@
 	required_reagents = list("mindbreaker" = 1, "carbon" = 1)
 	result_amount = 3
 
-
 /datum/chemical_reaction/paroxetine
 	name = "Paroxetine"
 	id = "paroxetine"
 	result = "paroxetine"
 	required_reagents = list("mindbreaker" = 1, "oxygen" = 1, "inaprovaline" = 1)
 	result_amount = 3
+
+/datum/chemical_reaction/neurotoxin
+	name = "Neurotoxin"
+	id = "neurotoxin"
+	result = "neurotoxin"
+	required_reagents = list("gargleblaster" = 1, "stoxin" = 1)
+	result_amount = 2
+
+/datum/chemical_reaction/luminol
+	name = "Luminol"
+	id = "luminol"
+	result = "luminol"
+	required_reagents = list("hydrogen" = 2, "carbon" = 2, "ammonia" = 2)
+	result_amount = 6
 
 /* Solidification */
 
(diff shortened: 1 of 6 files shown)


Arokha:
Oops. Don't merge this yet. There's a math problem on the ingredient counter.
... don't add the label, I mean, I'm fixing it right now. You'll just have to put in MORE WORK to remove it in like 5 minutes.

Arokha:
Why do you do these things you do D:<

Anewbe:
Sadism.

Arokha:
@Anewbe Masochistic more like. I did your thing on the translator (and fixed the math here).

Someone pointed out that xenomeat is just named "meat" in the code so shows up as such in recipes, so I renamed it. No real reason why it should just be called "meat" anyway when it's green and has a special description.

Anewbe:
Thanks, will review later.

Arokha:
Ok because I hate myself, I added a drinks one too. They output to separate files, and take separate css sheets if you'd prefer them to be different. If you want the nanoui ones, just steal the drinks.css and food.css from the same folder from the links in the description.

code/modules/reagents/Chemistry-Recipes.dm
```diff
@@ -557,14 +557,27 @@
...
 
-
 /datum/chemical_reaction/paroxetine
 	name = "Paroxetine"
 	id = "paroxetine"
 	result = "paroxetine"
 	required_reagents = list("mindbreaker" = 1, "oxygen" = 1, "inaprovaline" = 1)
 	result_amount = 3
 
+/datum/chemical_reaction/neurotoxin
```
Arokha: I just moved these up here, they were down in the food section for w/e reason.

Anewbe:
Is this completed?

Arokha:
Yup. The flour thing was because flour was, for whatever reason, lowercase ID, lowercase name, when all other reagents tend to have caps names, it caused a problem, I noticed the problem, fixed the problem, and flour/Flour.

Arokha:
You may want to look at #3083 since if you merge and run this before that, you'll get odd recipes because of the problems that #3083 fixes.

PR #3076, opened 2017-03-11
Fix handcuff hud icons appearing on the left and right hand inventory hud slots when handcuffed.

Leshana:
We define the inventory screen objects for hands to be special in that apply overlays to themselves for handcuffed.  By putting that logic on the screen object itself, we allow it to be expanded to other cases besides handcuffs in the future without conflict.

May not be the super-prettiest, nor did I rewrite the entire screen hud system or update_icons to be better.  But this one should be forwards-compatible.  And its nice to have this back after who knows how long its been broken.

image


Anewbe:
Looks like it works, will have reviewed.

PR #3080, opened 2017-03-12
Adds Two QOL Say/Emote Features

Neerti:
First feature is that if your name gets mentioned ICly, the text gets slightly bigger, and bolder.  AIs will also receive this if 'AI' is heard by them.  This works both when hearing someone talk near you or on the radio.  This feature can be turned off in preferences.
Here's an example;
feature demo 1

Second feature is the ability to format your speech by using specific characters.  You will be able to use * to bold, / to italize, and _ to underscore.  This is usable in talking, emotes, AI/borg binary, and deadchat/emote.  The game determines what to format when it sees two of the same character on the same line, and formats anything inbetween, so typing ```You're *fired!*``` will do "You're **fired!**"
Another example;
feature demo 2

### DIFF
--- a/code/modules/client/preference_setup/global/setting_datums.dm
+++ b/code/modules/client/preference_setup/global/setting_datums.dm
@@ -126,6 +126,12 @@ var/list/_client_preferences_by_type
 	enabled_description = "Show"
 	disabled_description = "Hide"
 
+/datum/client_preference/check_mention
+	description ="Emphasize Name Mention"
+	key = "CHAT_MENTION"
+	enabled_description = "Emphasize"
+	disabled_description = "Normal"
+
 /datum/client_preference/show_progress_bar
 	description ="Progress Bar"
 	key = "SHOW_PROGRESS"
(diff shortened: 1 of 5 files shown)


Arokha:
What's the effect if I type `*something along * these lines*`?

Anewbe:
@Neerti can probably explain it.

Neerti:
It would output "**something along** these lines*", as it only formats if it finds an opening and closing character.

Screemonster:
Is the scaling a preference option? Highlighting when I'm mentioned is pretty neat but I'm not sure I'd want to scale it quite that much.
And how does the highlight-bold play with asterisk-bold?

Neerti:
There's no config option to set how much to scale, but there is one to turn the emphasizing on or off.  The default size for text is 3 IIRC, and the emphasizing name size is 4.
Bolding from highlights gets overshadowed by name mentioning but otherwise doesn't cause any adverse effects.

PR #3085, opened 2017-03-14
Fixes 7 Recipes I Broke

Arokha:
LOOK NOBODY'S PERFECT OKAY

I added /food/ to the path of all the food recipes but these 7 have a completely silly means of actually creating the food items, and it results in that proc never being called because I didn't change THOSE to have /food/ in the path as well.

It's not like dough is important, okay.

And this mechanism of doing this is dumb.

And huff.

~~pleasedonthurtme~~


Anewbe:
How dare you, etc.

PR #3088, opened 2017-03-14
Ports Bot Patrolling back to Polaris (Plus bot fixes)

Leshana:
### Ports the new parts of https://github.com/Baystation12/Baystation12/pull/15298 back to Polaris (mostly patrol capability). ###
Also fixes many things that I found to be broken while testing that... so partly a port and partly fixing the port.

* Changes navbeacons to no longer use radios (Doing so was slower and didn't really help anything) Note: var/freq remains for mapping compatibility until all maps are updated.
  * Fixes navbeacons to actually hide under flooring if you put plating over them.
  * Changed how navbeacons are mapped in.  Don't use the picky "code_txt" variable, instead use appropriate subtype.
  * Made useful reusable subtypes so you don't have to code in a type for every. single. one.
* Updated mulebot to use new navbeacon codes.
* Re-activate patrolling!  With the new navbeacons, bots with will_patrol can follow the patrol navbeacons.
 * Note: various "should_patrol" variables on each bot type unified under /mob/living/bot/var/will_patrol
* Securitrons (beepsky) got an overhaul to the new patrol routing.
   * Note: BayStation replaced the snowflake handcuffs code with acutal handcuffs.  Polaris handcuffs requiere a grab to work, so are more complicated.  Kept our existing attack code.
   * Behavior procs reorganized to use better movement detection and just be more flexible.
   * Baystation beepsky didn't actually work.  Fixed it to actually arrest people without getting stuck in an infinite loop of handcuffing.
* Floorbot fixes & removal of bridgemode
   * Floorbot will now remove & replace broken floor tiles.
   * Floorbot will no longer decide its okay to pave over all of SPACE.
   * For the moment patching hull breaches is disabled, some commented it was causing problems before. Put in easy config to turn it back on as soon as everyone is happy this all works.
* Sundry bugfixes to all bots
  * Use forceMove() instead of setting loc
  * Use "\The [X]" message strings.
  * Pass target to do_after() when doing something to a target.
* Fixed events that were supposed to emag bots to now do so again.
* Wrote a commit message that describes what was actually changed.
* Updated polaris-1.dmm with the new navbeacon mapping types.

## Map Changes ##
All of the navbeacons were upgraded to the new types, but only on polaris-1.dmm.   I have put in backwards compatibility code to let the maps compile and the bots still work without this, so any other maps will still work perfectly fine, but will print reminders to world.log until updated.


Anewbe:
Sorry 'bout the silence, need to find time to review this.

Anewbe:
What would need to be done to the map to get this operational?

Leshana:
Nothing, because I already edited your polaris-1.dmm map.   For any other maps that have nav beacons on them, you just need to remove the `freq` and `codes_txt` values, and set either the direction type for a delivery beacon, or the `next_patrol` value for patrol beacons.

(Even if I had not done this, its backwards compatible except for the warning messages anyway)

Anewbe:
Alright. As far as we can tell we have nothing to lose from merging this, so thanks for making it.

PR #3091, opened 2017-03-15
Update values after equalizing gas mixtures.

Leshana:
Fixes https://github.com/VOREStation/VOREStation/issues/1192
* If we do not do this, zones are left with the wrong total_moles (and therefore the wrong **pressure readout**) after equalizing.
* Equalizing happens when two connected zones get close enough to suspend processing. Thus the effect of this bug would be wrong pressure readings on rooms that are no longer being updated, thus locking in the wrong value!


Anewbe:
Tested?

Leshana:
Yes, see screenshot in the issue for prior behavior.  Now see:
image
It gets it right the first time without a developer having to proccall it into correctness.

PR #3093, opened 2017-03-15
Translators can't output hive or nonverbal languages

Anewbe:
Found it a tad odd that a screen or earpiece could output text /in/ signlanguage... Or Changeling.

### DIFF
--- a/code/game/objects/items/devices/translator.dm
+++ b/code/game/objects/items/devices/translator.dm
@@ -16,11 +16,15 @@
 	if(!listening) //Turning ON
 		langset = input(user,"Translate to which of your languages?","Language Selection") as null|anything in user.languages
 		if(langset)
-			listening = 1
-			listening_objects |= src
-			if(mult_icons)
-				icon_state = "[initial(icon_state)]1"
-			to_chat(user, "<span class='notice'>You enable \the [src], translating into [langset.name].</span>")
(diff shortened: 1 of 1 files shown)


SilverTalismen:
Though you've prolly covered it but Maybe check to see if Diona Global Root xan be heard or output with?

Anewbe:
ROOTGLOBAL has the HIVEMIND tag, so it shouldn't get translated.

code/game/objects/items/devices/translator.dm
```diff
@@ -15,7 +15,10 @@
 /obj/item/device/universal_translator/attack_self(mob/user)
 	if(!listening) //Turning ON
 		langset = input(user,"Translate to which of your languages?","Language Selection") as null|anything in user.languages
-		if(langset)
+		if(langset && ((langset.flags & NONVERBAL) || (langset.flags & HIVEMIND)))
```
Neerti: If someone hits cancel on language selection, does it mean that it will cause it to look like it is on due to adding those checks to the if(langset ...) part?

Anewbe: It might, I'll double check.

Anewbe: Nope, runtimes, fixing now.

PR #3094, opened 2017-03-15
Fix rooms holding pressure when exposed to space.

Leshana:
This happens because edges cease processing when the delta between their zones is small enough.  This is normally a fine optimization, but when it results in a room at 4kPa with a window open to space, it breaks imurshuns.

Two main changes to solve this problem without too much cpu cost:
1) Stop edges from sleeping if one side is a hard vacuum.  This ensures that a zone doesn't freeze at a low-but-non-zero pressure when touching hard vacuum.
2) Prevent #1 from causing the edge to stay alive for ages while pressures asymptotically approach zero as they are repeatedly equalized but only half is dumped to space. (Would happen if ZoneA---ZoneB---Space arrangement exists) by detecting when the total amount of air left is small enough that it would normally sleep anyway, and just setting it to zero.

The end outcome is that behavior is mostly the same as before, except when zones have an open path to unsimulated space, they will reach equilibrium at zero instead of semi-random lowish values.

### DIFF
--- a/code/ZAS/ConnectionGroup.dm
+++ b/code/ZAS/ConnectionGroup.dm
@@ -180,7 +180,8 @@ Class Procs:
 	air_master.mark_zone_update(B)
 
 /connection_edge/zone/recheck()
-	if(!A.air.compare(B.air))
+	// Edges with only one side being vacuum need processing no matter how close.
+	if(!A.air.compare(B.air, vacuum_exception = 1))
 		air_master.mark_edge_active(src)
 
 //Helper proc to get connections for a zone.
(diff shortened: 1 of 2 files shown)


Anewbe:
Tested?

Leshana:
Tests indicate that this is working.  Given the test scenario:

pr 3094 test

Zone 1 was allowed to empty to zero pressure, this worked as normal.
Then the airlock between Zone 1 and Zone 15 was bolted open and we watch what happens.
It shared air between them for six air master ticks, until it got to the point where it would normally sleep the edge.   Then for the next six air master ticks it continued to flow air from Zone 15, thru Zone 1, into space, until the total gas held in Zones 1 & 15 combined dropped below the minimum and were set to zero.    On the next tick after that, none of the zones woke up, and processing ceases with both Zone 15 and Zone 1 at 0 kPa

PR #3096, opened 2017-03-16
Add one inflatables kit to paramedic lockers

Arokha:
Makes sense to meeee. Hopefully it does to you, too. Gotta rescue those people!


SilverTalismen:
Or just recover the corpses when it's fixed.

I'm kidding.

Oh god Tempest is gonna kill me

PR #3103, opened 2017-03-19
Not So Simple Animals

Arokha:
This is basically part 1 of a larger project for more realistic NPCs.

This rewrites simple_animals to behave more realistically in some cases, or to just be better in others.

Previously there were many oddities about simple animal behavior, but this should hopefully bring them all up to one might expect them to behave like. Some can call for help, some can follow other mobs around, etc.

There are various setting variables on mobs that you should be aware of. They are all documented at the top of simple_animal.dm, but I'll mention a few of the more important ones here:

- `hostile` - Controls whether they attack members not of their faction on sight.
- `retaliate` - Controls whether they act in self defense if attacked (by thrown objects, guns, fists, etc)
- `run_at_them` - If set (default: on) they will not use A* pathing to locate a path to targets. This is overridden on human-like mobs (syndicate, pirates, etc) who will use doors and path to you.
- `view_range` - Determines the distance the animal can see, instead of 'always 10'. Now 7 by default.
- `cooperative` - Calls for help from their faction within a certain range (also a variable) who come to assist
- `investigates` - Instead of instantly attacking, if they see you for less than 1 second, they will investigate where they last spotted you
- `returns_home` - Mobs track where they started on the map (or where they were spawned) and if this is set, they will return there after several seconds of not doing anything else
- `specific_targets` - Overrides target finding to only call Found(atom) on the mob, who can then target things however it wants. Runtime does this for mice.

There are also several important procs, some previous ones that have been renamed to make more sense, and some that are new that can be overridden on mobs for interesting features:

- `react_to_attack()` - Called when the mob is attacked by something, can have overridden responses like running away, going to get into a mecha, etc (PR includes a 'dark gygax' simple_mob).
- `AttackTarget()` - Called when we are in STANCE_ATTACKING because we think we're in range. Decides what proc to call based on range, such as the ones below.
- `PunchTarget()` - Called when we are in melee range, even on ranged mobs, to enter into melee combat with the target.
- `ShootTarget()` - Called when we want to fire our gun at the target, assuming we have one and are in range.
- `SpecialAtkTarget()` - Called based on a certain chance set on the mob (`spattack_prob`) to perform a special attack, based on the ranges (also vars). Can be something like throwing a grenade, breathing fire, etc.
- `GiveUpMoving()` - Stops all movement and returns to wandering (or not, whatever they normally do).
- `RequestHelp()` - Called by mobs pursuing a target to their faction friends in range (range is a setting).
- `HelpRequested()` - Called ON mobs when they are being asked for help, usually just runs over.
- `CheckFiringLine()` - Called by ranged mobs when about to shoot at something, they avoid shooting their allies and will step in random directions to try and find a clear firing line.

There are also several new lists one can populate with messages:
- `say_maybe_target` - If they investigate, they say this when they spot you before 'acquiring' you.
- `say_got_target` - Said when they lock onto a target.
- `reactions` - Said in reply to members of their faction (if hostile) or anyone (if not) in response to certain text. For example if you add "Hello" = "How are you?" they will reply to the former message being heard with the latter.

There are a lot more variables than listed above, but I don't want to pollute the PR with every single one of them.

Mobs also understand certain languages (galcom by default) and can be made to only speak in certain languages for more advanced NPC mobs on other planets or whatever you're gonna do with 'em.

[Demonstration of 'MGS' Tactics](http://i.imgur.com/KnXByw1.mp4)
[Demonstration of calling for help](http://i.imgur.com/m7SUaRm.mp4)

In the final version, mobs assisting other mobs immediately attack rather than having the 1-second investigate thing. Also mobs that have been fights recently attack faster. They have an 'annoyed' variable that tracks how annoyed they are and if they will attack instantly or not, so you can't 'distract kite' them around forever.


Anewbe:
So question: If I set retaliate, cooperative, and inquisitive to 1, but hostile to 0, what _should_ happen with that mob?

Spacemanspark:
Interesting.

Arokha:
> So question: If I set retaliate, cooperative, and inquisitive to 1, but hostile to 0, what should happen with that mob?

They will do nothing unless you hit one of them, and then if you hit one of them the others will come over and watch. I considered changing how that works, perhaps having the angry one share their target with the others, but to be honest this was one of those things where it's like "Hey if I make a 90,000 line PR nobody is going to take it" so I want to put that sort of thing off till Part 2 (which is like, them accepting commands from players and such).

Anewbe:
Does Inquisitive have an effect on non-hostile mobs at all?

Arokha:
> Does Inquisitive have an effect on non-hostile mobs at all?

Not really, nor would one be likely to be added to this PR, though I'm open to suggestions on how inquisitive, non-hostile mobs would behave that I could do in part 2.

Anewbe:
Well an inquisitive non-hostile mob could try to stay within a certain range of any player they find for some amount of time, and move towards/away from them as appropriate? Like a shy puppy, or something.

Arokha:
Interesting idea. They might follow them a certain distance from their home, then go home or something.

code/game/machinery/doors/airlock.dm
```diff
@@ -37,7 +37,16 @@
...
+			if(src.locked || src.welded)
+				visible_message("<span class='danger'>\The [user] begins breaking into \the [src] internals!</span>")
+				if(do_after(user,10 SECONDS,src))
+					src.locked = 0
+					src.welded = 0
+					update_icon()
+					open(1)
+					if(prob(25))
+						src.shock(user, 100)
+			else if(src.density)
```
Anewbe: Does this whole green section affect anything other than simple mobs?

code/modules/hydroponics/seed_datums.dm
```diff
@@ -184,7 +184,7 @@
 	display_name = "killer tomato plant"
 	mutants = null
 	can_self_harvest = 1
-	has_mob_product = /mob/living/simple_animal/tomato
+	has_mob_product = /mob/living/simple_animal/hostile/tomato
```
Anewbe: Looks like in other places you've been removing the hostile path, but here you're adding it. Why is that?

code/modules/mob/living/simple_animal/aliens/drone.dm
```diff
@@ -1,39 +1,32 @@
...
 	name = "combat drone"
 	desc = "An automated combat drone armed with state of the art weaponry and shielding."
 	icon_state = "drone3"
 	icon_living = "drone3"
 	icon_dead = "drone_dead"
-	ranged = 1
-	rapid = 1
-	speak_chance = 5
+
+	faction = "malf_drone"
```
Anewbe: I may be wrong, but I think this was specifically **not** a thing, to show that the drones are really broken. Don't personally mind if this changes, but it might bear discussion.

code/modules/mob/living/simple_animal/aliens/mimic.dm
```diff
@@ -1,195 +1,196 @@
...
+
+/mob/living/simple_animal/hostile/mimic/crate/LostTarget()
+	..()
+	icon_state = initial(icon_state)
+
+/mob/living/simple_animal/hostile/mimic/crate/death()
+	var/obj/structure/closet/crate/C = new(get_turf(src))
+	// Put loot in crate
+	for(var/obj/O in src)
+		O.loc = C
```
Anewbe: Might be better as forceMoce()

code/modules/mob/living/simple_animal/aliens/mimic.dm
```diff
@@ -1,195 +1,196 @@
...
+	var/knockdown_people = 0
+
+/mob/living/simple_animal/hostile/mimic/copy/New(loc, var/obj/copy, var/mob/living/creator)
+	..(loc)
+	CopyObject(copy, creator)
+
+/mob/living/simple_animal/hostile/mimic/copy/death()
+
+	for(var/atom/movable/M in src)
+		M.loc = get_turf(src)
```
Anewbe: Might be better as forceMoce()

code/modules/mob/living/simple_animal/animals/corgi.dm
```diff
@@ -1,217 +1,227 @@
...
+					else
+						ian = M
+				else
+					alone = 0
+					break
+			if(alone && ian && puppies < 4)
+				if(near_camera(src) || near_camera(ian))
+					return
+				new /mob/living/simple_animal/corgi/puppy(loc)
+
```
Anewbe: I feel like there should probably be a puppies++ in here...

code/modules/mob/living/simple_animal/animals/goose.dm
```diff
@@ -1,34 +1,48 @@
...
+	min_tox = 0
+	max_tox = 0
+	min_co2 = 0
+	max_co2 = 0
+	min_n2 = 0
+	max_n2 = 0
+	minbodytemp = 0
 
-	faction = "geese"
+	speak_chance = 0
```
Anewbe: I feel like we should give them a highish speak_chance and just have the message be something like "honks!" ...

code/modules/mob/living/simple_animal/humanoids/pirate.dm
```diff
@@ -1,56 +1,66 @@
...
+	health = 100
+	speed = 4
+
+	run_at_them = 0
+	cooperative = 1
+	investigates = 1
+	firing_lines = 1
+	returns_home = 1
+	reacts = 1
+
```
Anewbe: No retaliate on pirates?

code/modules/mob/living/simple_animal/humanoids/russian.dm
```diff
@@ -1,55 +1,66 @@
...
+	maxHealth = 100
+	health = 100
+	speed = 4
+
+	run_at_them = 0
+	cooperative = 1
+	investigates = 1
+	firing_lines = 1
+	returns_home = 1
+	reacts = 1
```
Anewbe: No retaliate on Russians?

code/modules/mob/living/simple_animal/humanoids/syndicate.dm
```diff
@@ -1,163 +1,192 @@
...
+	maxHealth = 100
+	health = 100
+	speed = 4
+
+	run_at_them = 0
+	cooperative = 1
+	investigates = 1
+	firing_lines = 1
+	returns_home = 1
+	reacts = 1
```
Anewbe: No Retaliate?

code/game/machinery/doors/airlock.dm
Arokha: No, it never has. That was the section we discussed about how to make it so simple_mobs could break through inoperative airlocks. It doesn't have an effect on players. attack_generic is a thing bots and simple_animals do.

code/modules/hydroponics/seed_datums.dm
Arokha: I spoke with someone and they said that tomatos do attack you when created, so it made sense to subtype them. In response to your earlier question about why I have the paths, that's purely for admin spawning or mapping ease. If you want hostile mobs, you can just look at the path. That's all. It's not necessary to leave it in if you don't like it, but it's sort of nice to know at a glance what your choices are when mapping for "hostile" things.

code/modules/mob/living/simple_animal/aliens/drone.dm
Arokha: Are you referring to their faction? That is in your current code, I just arranged the vars to match all the other simple_animal in terms of line order. I did not change the faction.

code/modules/mob/living/simple_animal/animals/corgi.dm
Arokha: The last 3 comments are not my code, just shows in the diff because the lines changed and git is too much of a git to realize I didn't change them.

code/modules/mob/living/simple_animal/humanoids/syndicate.dm
Arokha: For the previous 3 comments, all /hostile path mobs have retaliate on by default because it makes sense to have that on already-angry mobs.

code/modules/hydroponics/seed_datums.dm
Arokha: I should also clarify: I removed the hostile path from retaliate-only mobs. Having the path be "hostile/retaliate" is a bit unusual, as they ARE NOT hostile under normal circumstances and pathing them as such is strange. I mov

code/modules/mob/living/simple_animal/animals/goose.dm
Arokha: Commited this change.

code/game/machinery/doors/airlock.dm
Anewbe: Just double checking.

code/modules/hydroponics/seed_datums.dm
Anewbe: Fair enough.

code/modules/mob/living/simple_animal/aliens/drone.dm
Anewbe: Odd.

code/modules/mob/living/simple_animal/animals/corgi.dm
Anewbe: Well yeah, but I just happened to stumble across this little bit while reviewing.

code/modules/mob/living/simple_animal/humanoids/syndicate.dm
Anewbe: If that's the case, then there are ~9 other cases where you can probably go back and remove the var(s) from other files.

Arokha: Done

code/modules/mob/living/simple_animal/animals/corgi.dm
Arokha: Fixed them, because I guess I should think of future generations of spessers.

Anewbe:
RNGsus help me, I'm pressing the button.

PR #3106, opened 2017-03-19
Fix thrown objects hitting windows not in their path

Leshana:
* Fixes windows counting as blocking the entire tile for thrown objects when they don't.
* Same thing for (closed) windoors, which had the same problem.
Fixes https://github.com/VOREStation/VOREStation/issues/1212


Anewbe:
Tested?

Leshana:
Yes, it is tested.   Now I can throw welders into the closet after taking an industrial welder.

PR #3109, opened 2017-03-20
Flooring decals should work more places

Arokha:
Such as shuttles. A good example is the trader ship which has hazard stripes mapped in, which don't appear in the game because it doesn't work there.


Anewbe:
code/game/turfs/flooring/flooring_decals.dm:20:error: /turf/unsimulated/shuttle/floor: undefined type path

Arokha:
Yeaaaaaaaaaah that's weird. What's the path for unsimulated shuttle turfs?

Anewbe:
I'm not sure there is one, the only shuttle I can find on Central has simulated floors.

Arokha:
Oh I see, they use plating in the doorways that's unsimulated to make them not leak. Ooone second.

PR #3110, opened 2017-03-20
Make jukebox controllable by its wires

Leshana:
#### More jukebox enhancements!
Added four extra hacking wires to the jukebox for Play, Stop, Next, and Prev.  A resourceful bartender could now remote control their jukebox.


Anewbe:
This is sort of evil, but it looks like it'll work.

PR #3086, opened 2017-03-14
Adds a fuckton of robot, hologram, and AI sprites.

nachomeep:
Will add screenshots soon-ish.

Ported from Baystation12, Aurora, Paradise, VGStation, TGStation, and Heaven's Gate.

PR #3114, opened 2017-03-21
Allow AI to use custom holograms

Arokha:
Adds a third hologram button choice, allowing the AI player to load their current character slot as the hologram so they can custom tailor their AI character to look like whatever. Also fixes it so that when using existing crew as a hologram you can face all four directions and the picture is not based on their records and thus probably terribly inaccurate and old. To compensate, you can only generate a hologram from crew you can currently track, so presumably you take a few pictures with your cameras or something.

I think the blue looks ugly with this, if you want my advice, comment out https://github.com/PolarisSS13/Polaris/blob/master/code/_helpers/icons.dm#L814 (change not included in PR). We have a whole holodeck that can project solid objects. I'm gonna say color holograms is not stretching the lore. :P Holograms are obviously holograms what with the transparency and scanlines. The ones in AI.dmi are "pre-blue" already so those would still be blue.

### DIFF
--- a/code/_helpers/icons.dm
+++ b/code/_helpers/icons.dm
@@ -793,6 +793,26 @@ proc // Creates a single icon from a given /atom or /image.  Only the first argu
 			alpha_mask.Blend(image_overlay,ICON_OR)//OR so they are lumped together in a nice overlay.
 		return alpha_mask//And now return the mask.
 
+//getFlatIcon but generates an icon that can face ALL four directions. The only four.
+/proc/getCompoundIcon(atom/A)
+	var/icon/north = getFlatIcon(A,defdir=NORTH,always_use_defdir=1)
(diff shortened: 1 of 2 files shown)


Anewbe:
Apologies in advance, there's another PR up that's been under review for a while that touches the same files, so if that one doesn't have issues it'll probably cause conflicts on this one. Brief look doesn't seem to show anything glaring.

Arokha:
@Anewbe Which PR?

Anewbe:
https://github.com/PolarisSS13/Polaris/pull/3086

Arokha:
@Anewbe I have rights on his repo. Merge whichever first and I can fix the conflicts on either.

Anewbe:
Fabulous.

PR #3118, opened 2017-03-21
Continues work on the surface map

Neerti:
Does lots of small things and commits the surface map.  The map name will almost certainly need to have it's z-level number changed in the future.

Stuff this will do:
- Adds two maps to the code that are unticked.  One is for the surface, the other is for the future centcom/ship place.  The centcom level has a nice start on the merc shuttle.
- Adds a new 'voidcraft' type of walls, floors, and doors intended for shuttles.  Voidcraft walls have a stripe that can be custom colored by the mapper.  Sprites are from /tg/
- Adds a ZAS zone divider, thanks to @Leshana 
- Does a few optimizations on planetary processes.  More to come later.
- Adds new wall sprites for wood materials.  Also from /tg/.
- Adds a new type of wood, called sifwood for now.  It's like wood but bluer.  The shade will probably need to be adjusted later.
- Made water crossing less spammy.
- Pool water tile now has a deep variant.
- Adds new water tile that acts as a shoreline.  It generates the 'wave' motion animation dynamically using subtraction blending so that changing the 'underwater' tile appearance or the water tile is easy and you won't need to manually edit 50+ sprites.
- Adds 'blood moon' weather that is admin-only, makes everything red.  Intended for Cult Fun later down the road.


Anewbe:
The movement_message thing, is there not a way to make that only check when it's going from shore > water > deep water, and vice versa? Looks like the whole thing go snipped.

Neerti:
It should still give a message when going from no water > water and vise versa.

Arokha:
I like the Voidcraft walls. Very industrial! Do they automerge together like default wall material walls, or stay as separate entities or...?

Shuttle walls have always been kinda fucky with that sort of thing.

Arokha:
I think once you merge this I'm going to make shuttles nicer (not steal turfs from under corners, auto-connect to each other, etc).

Neerti:
At some point I plan to make a copy of the sprite that was grey/whiteist so that coloring it wouldn't ruin the quality, to use for certain materials that were really strong, like Durasteel.  However that is probably quite awhiles away so if you'd like to do it feel free @Arokha

Anewbe:
Looks good to me, will merge wheneverish.

PR #3119, opened 2017-03-22
Allows in-round flavortext updating with verb

Arokha:
This adds a verb in the Preferences tab that reapplies flavortext from character setup to your mob, so if you have changed it, you can push that down to the mob to update it (lost an arm, got a scar, whatever).

To be honest, I don't know how important this is to you. It comes up fairly often for us as people update (E)RP prefs and flavortexts during our longer rounds for various reasons. It won't hurt anything to give it, though.


nachomeep:
Isn't there already a set-flavour-text verb?

Arokha:
... HUH. Go figure.

PR #3126, opened 2017-03-23
Adds Boxcutters

Anewbe:
Does 6 brute open, 4 brute thrown. Otherwise acts like a knife. Might add to the loadout eventually, if it gets in.


RandomColonist:
Maybe it need, idk, some boxcutting features?

Anewbe:
Pretty sure knives do that already.

PR #3133, opened 2017-03-25
Adds greyscale sprites for heels worn/held/on the ground. 

Schnayy:
Still need coding, intended to be any color. First commit, hope I do it right!

Edit: Here's a preview. Actual sprites have been edited to be white, but my previews are black, fyi.
heels front
heels ground

Intended to be in the loadout.


Anewbe:
Sprites look decent, but as far as I remember that the lighter the sprite is, the closer the end-result will be to the selected color. Also, for reasons I do not quite remember it may be better to keep sprite names in lowercase.

In order to add the item itself to the code, you'll need to edit the code in code/modules/clothing/shoes/miscellaneous.dm.
The end code might look like:

```
/obj/item/clothing/shoes/heels
	name = "high heels"
	icon_state = "heels"
```


To add these to the loadout you'll need to add some code in code/modules/client/preference_setup/loadout/loadout_shoes.dm.
What I would do is copy some of the code from /datum/gear/shoes/skater, and change it to reference the new item. So it might look something like:

```
/datum/gear/shoes/heels
	display_name = "high heels"
	path = /obj/item/clothing/shoes/heels

/datum/gear/shoes/heels/New()
	..()
	gear_tweaks = list(gear_tweak_free_color_choice)
```

Anewbe:
For the held sprites you don't need the icon to be "heelsheld", should just be "heels".

PR #3136, opened 2017-03-25
Updates drone respawn time.

Belsima:
Double the regular respawn time prior to this. Maybe.


Anewbe:
If you're going to change it, may as well make it 5 minutes like every other sort of respawn.

Belsima:
It used to be 1/3 the respawn time, so I made it shorter.

Anewbe:
@Belsima Doesn't quite work, not sure why.

PR #3137, opened 2017-03-25
Adds skintone to FBPs

Anewbe:
Sort of based on what Baystation did.


SilverTalismen:
How well does it work? :o

Anewbe:
Should be the same as a human, worked in testing.

SilverTalismen:
Awesome! Will be great for Veymed- Which I'm assuming is the only brand it works for xD

Anewbe:
Yep.

PR #3141, opened 2017-03-25
Dynamic Shuttle Walls/Landings

Arokha:
tl;dr
This PR ports several /tg/ shuttle sprites (engines, heaters, dark shuttle walls), makes shuttle walls auto-join themselves in appropriate directions, rearranges the shuttle icon DMIs to have them be less insane, and has 'partial' turf corners determine a sprite to underlay automatically, making it easier on mappers to use whatever floors they want. Additionally this applies to external ship corners, so they don't display the ugly 'space turf' sprite when in transit space, and things like that.

Additionally, this PR has every shuttle turf store vital information about the turf it landed on, so that it can be replaced when the shuttle leaves. This allows landing on decorated landing pads and returning them to their previous state after.

Long version:
Some things to know about shuttle turfs:
- `nojoin` turfs don't autojoin, and in DM have red border marks in the middle of each side. This is for that classic 'every turf is a square' look.
- `hard_corner` turfs don't use diagonal corners, and instead make full-turf corners, if you want to put things on those walls, and they have purple corner brackets in DM on each corner
- `join_group` var on shuttle turfs determines what group they'll join with. null prevents joining (and is how the nojoin ones prevent joining). So you can have shuttles in shuttles that don't connect.
- `takes_underlays` is usually set automatically if a wall turns into a partial turf, but you can set it manually on any turf to force it to display an underlay. This is useful for 'glass' for example in the bottom of a shuttle. The turf will display whatever it landed on as an underlay (or transit turfs or whatever).
- `under_turf` is a var for storing a turf you always want takes_underlays turfs to display, and will ignore all auto-detection or landing logic for that turf. If you put a path in this var, if the turf is a corner, it will always display the default icon and iconstate of that turf.
- `/obj/structure/shuttle/window` is the window type you should use, and this is also auto-joining with walls and other windows. You should put shuttle plating under it and a grille. Not to be confused with the more annoying /obj/structure/window/shuttle
- `join_always` is a (static) list on turfs of things they will consider join-able. An example is voidship doors, so the walls connect to them, as this looks better. Also things like shuttle heaters and engines, and shuttle windows.
- `join_carefully` is like join_always, but instead only joins if the turf under the object is a /turf/simulated/shuttle subtype and if the join_group is the same as ours. This includes grilles, for connecting into windows (but avoiding connecting to windows at docked stations).
- `/turf/simulated/shuttle/floor/carry` is a special fully transparent shuttle turf that will 'carry' whatever is on it with it, useful for object-only protrusions from shuttles, areas you want people to be able to stand in OUTSIDE the ship but still get pulled with it. All the engines need one of these. It shows up as a transparent turf with yellow inward pointing marks in DM.
- Please don't use custom icon states for floors. Use the correct paths.

The shuttle turf sprites are now non-directional, as in, they do not have any directions stored in the iconstates. They are also separated into their own .dmis so if you create instances from iconstates in DM it's not such a hassle to find what you want.

The system of numbering the iconstates and autojoining uses the normal byond directions. A wall that connects to the NORTH (1) and SOUTH (2) will use the 3 iconstate (summed directions). A T-intersection between EAST (4), WEST(8) and SOUTH (2) will use the 14 icon state (4+8+2).

Should be pretty easy to keep it working right.

It can be made 'more complete' if someone feels like filling out all the iconstates up to 47 different ones, as this is enough to do 'every combination of diagonals and everything'.


Anewbe:
Travis is mad.

Arokha:
I updated Travis to be more appreciative of my contributions.

Arokha:
Also please don't squash merge this.

Arokha:
I forgot to warn you about something. I'm doing it as a comment so it notifies you so it's not missed.

Shuttles landing and leaving from turfs that have "fancy" update_icons() procs that cut all the overlays, I can't help you with that, you're still going to lose the hazard stripes or whatever you put down. I can rewrite ChangeTurf to fix it but there are actually a few turfs that use this. Like snow cuts the overlays to update footprints and stuff. Plating does it for like, flooring and whatnot. So make sure your landing pad doesn't have an update_icons() proc that cuts all the overlays.

Anewbe:
May RNGsus have mercy on my soul, it looks fine from my scans... Merging.

PR #3142, opened 2017-03-26
Human Scissors tool

Arokha:
Adds a tool to the tools folder called HumanScissors that I wrote. As the readme explains, it takes every iconstate in SpritesToSnip.dmi, and cuts them with CookieCutter.dmi sprites.

So if you load a single 4-direction race sprite into SpritesToCut.dmi and load all the bodypart icons from some existing race into CookieCutter.dmi (which I put in there already since that's probably what people will use it for), you get the output of the race sprite cut up into ready-to-use bodypart sprites. Don't forget to add some 'gore' edges if you want~

The output will contain names based off the names in SpritesToSnip.dmi and CookieCutter.dmi. ([StS_name]-[CC_name])


Anewbe:
As far as I can see this works fine, did you run it at least a few times?

Arokha:
Yes, it was used to make the body markings PR.

PR #3144, opened 2017-03-26
Makes space in the uniforms.dmi file

Anewbe:
For reasons unknown I can't seem to get the yoga pants to work on any non-humans. Dunno what else might have that issue, regular pants, shorts, skirts, etc. all work as intended.


Anewbe:
Alright, after this change all uniform sprites will need to be added to the base file, and that for each species.

I can't help but think that there's a better way to do this, but I don't know that better way, and this worked in testing.

Yoshax:
You could split all rolled down sleeves and rolled down jumpsuit into their own file and switch between them like so. You'd manage to cut the original file by at least half. Would likely be a bit more obvious and intuitive too. Simpler too.

PR #3145, opened 2017-03-26
Added hand sprites to most medical items.

Cerebulon:
Added in-hand sprites to most item lacking them from the Medical department.

(Full list: Body bag, Advanced Mass Spectrometer, Rescue Suit Control Module, Stethoscope, Stasis bag, Defibrillator, Blood Packs, Advanced Trauma/Burn Kits, Hypospray, Biopsy Scanner, Surgical Drill, Fixovein, Retractor, Bone-Gel, Bone Setter, Cautery, Hemostat, Inflatable Wall Box.

Slight changes required to bloodpack code to also properly update hand sprites.


Anewbe:
Did you test this?

Cerebulon:
Yes, and fixed a few issues with alignment and such as a result. The blood pack code also works as intended, doing both exactly what it used to do (Changing sprite based on fullness level) as well as updating hand sprites.
Have you found a problem I missed?

code/game/objects/items/devices/defib.dm
```diff
@@ -4,6 +4,7 @@
 	desc = "This KHI-branded defib kit is a semi-automated model. Remove pads, slap on chest, wait."
 	icon = 'icons/obj/device.dmi'
 	icon_state = "defib_kit"
+	item_state = "defib_kit"
```
Anewbe: Don't need item_state if icon_state is the same thing.

code/game/objects/items/devices/scanners.dm
```diff
@@ -298,6 +300,7 @@ REAGENT SCANNER
 /obj/item/device/mass_spectrometer/adv
 	name = "advanced mass spectrometer"
 	icon_state = "adv_spectrometer"
+	item_state = "adv_spectrometer"
```
Anewbe: Same here.

code/game/objects/items/devices/defib.dm
Cerebulon: In this case it was because the defib icon_state is changed to several very slightly different 'open' versions, but I've since realised none of those versions can actually be picked up, I'll take this line out.

code/game/objects/items/devices/scanners.dm
Cerebulon: Does the /adv version not inherit the icon_state from its parent mass_spectrometer, which is set to "analyser" and need to be overwritten or does icon_state overwrite both inheritances?

Cerebulon: The other option I think would be making a proper hand sprite for the basic spectrometer so there's no need for any item_state at all. I think I'll do that.

Anewbe: No, if I remember how the system works, each item checks for its own icon_state, then makes the item_state the same thing. Might be worth testing, but I've not had problems with it.

Anewbe:
To answer your above question, no, I just occasionally ask people if they've tested things.

PR #3146, opened 2017-03-26
Adds sunflower, poppy, and lavender flower crowns.

Schnayy:
Items coded in, still need crafting coded. Not sure how to do that part myself.


Anewbe:
Would you like some ideas on how to code the crafting, or would you rather I merge this, then make the construction things at a later date?

Schnayy:
If you point me in the right direction I can probably code it.

Anewbe:
So what I would do is start with a base "plant crown" item, like a ring of glass or wood or something. Then on that item I'd put in an attackby() call, and if the item being used on it was one of those flowers, I'd have the flower and turn into the correct sort of crown.

Schnayy:
I've tried like 50 times to update to this PR, I'm just going to remake a PR with the hopefully conflict-free files.

PR #3152, opened 2017-03-26
Adds Body Markings

Arokha:
So you can customize your character with tattoos or fur colors for Taj or whatever. I'll let Anewbe work out what he wants to whitelist other than what I did already.

The icons are sliced up with the HumanScissors program I wrote, so they are per-organ. So, lop off a limb, put it on someone else, they have the tattoos and whatnot. It's part of DNA so it comes with when cloning.

They are defined in the normal sprite_accessories file and the icons are all in markings.dm in the format of "markingname-organtag" so if you have a marking that spans both arms and torso you need "thatone-torso", "thatone-l_arm", "thatone-r_arm" icon states.

image
A catguy with black tiger stripes and a green right arm tat having shot himself and being cloned.


Anewbe:
I'm absolutely _not_ stalling for time while I try to make sure this is as solid as it looks, but it also looks like the Unathi Scutes body marking is on their front. I'm pretty sure those belong on the back, at least from what I know of reptiles. Could be wrong.

Arokha:
It's like dragon belly plate kinda things. Those are scutes. As far as I know anyway.

Anewbe:
Scutes are the hard parts on the back, as per my google search. If that sprite is meant to represent those, it may be better to have it on the other side of the mob. Or, if it's used to represent the softer underbelly, then we can leave it be.

Arokha:
`A keratinous, chitinous, or bony external plate, as on the shell of a turtle or the underside of a snake.` Between the two of these, it's meant to be more like the latter, the overlapping large parts like on a dragon or whatever.

Anewbe:
Eh, if it really bugs me when it's in game tweaks can/might be made.

Fox-McCloud:
Paradise has an entire system for this; may want to check it out--you guys may be able to modify/use it for similar purposes.

Arokha:
@Fox-McCloud I believe these sprites are from Paradise, maybe? I'm not 100% sure. They are ported from a port from a port from a port... maybe from Para. Someone said that on Para they still display when you lose limbs and such though.

PR #3153, opened 2017-03-27
'Improved' solution to the full uniform.dmi problem

Yoshax:
Hi folks. This is proposing a 'better' way to do #3144
I thought the excess of files was a little messy and had a better way, in my opinion at least, on how to do it.

This PR contains 3 things.
- Reverting #3144
- A DmiSplitting Tools
- The changes

Before I get onto those explanations, this removes 167 icons from uniform.dmi (the full one pre-#3144.)

The tool is quite simple, and likely badly coded it. It's a very simple program that allows you to fill a dmi file then split the icons inside based on criteria provided. For example, you fill your dmi and run it, use the verb, and provide _d_s to it, having filled the dmi with your full uniform,dmi. What it does is find all the icons with _d_s and then split them off into a new file, renaming them to just be thing_s. It may never have another use, but it's tiny and I coded it, and it works, so I've included it.

So! The actual changes.
It's quite simple really. 
All rolled down uniforms are now in `uniform_rolled_down`
All rolled up sleeves are now in `uniform_sleeves_rolled`

This is handled in the code by setting the icon_override on the uniform to point to one of those files when you roll your jumpsuit down or roll your sleeves up. Because all the states are renamed to no longer be thing_d_s and are thing_s it works perfectly by just doing that.

It's about 4am as I write this description for the PR, and there are a few small problems with it. Providing this is something that is desired, I'd like to request a short DNM on other PRs modifying any of the post-#3144 clothing files. I'll be able to finish and resolve the problems within 24 hours of posting this. Resolving those conflicts would be real suffering.


Anewbe:
None of those files are being touched by open PRs, as far as I can tell, so you should be good on conflicts.

I will say this, though: I'm not sure this is the most intuitive way to do this, but it's admittedly probably better than my solution. However, I know for a fact that I had to recover a number of sprites for the PR this improves, and if you started by reverting that you'll need to readd those sprites.

Yoshax:
What sprites?

Anewbe:
Most of the ones in military.dmi vanished before the PR, yogapants had no sprites, and both chaps sets were broken because they were missing the _s. Not sure what else might've needed fixes, but those ones are the ones I was told about when I started.

Neerti:
This won't break custom item icons, right?

Yoshax:
Manually readded all the military sprites, cut 'em up using the tool in the PR. They're all in the appropriate files now. Added the yoga pants too. Fixed the chaps, and the customs suits which had the same issue of missing _s. 

I've tested a bunch of clothes, and they all work fine. 

As for custom items @Neerti, let me get back to you on that.

Yoshax:
Okay so, I actually have no idea how to setup the custom item stuff to test it. But there's no reason they won't work. This doesn't touch any of the base icon stuff that's used by that. The only point of overlap is rolling down/sleeve rolling but there's no custom items that use that. Regardless, the logic in my last commit should take care of that if it ever exists.

I haven't tested all the possible items of clothing, because there's literally hundreds and it's not exactly feasible. But I've tested a lot. As stands, as far as I can tell and know, this is good to go.

Anewbe:
Can custom items not roll down?

Yoshax:
I honestly have no idea if they currently can. None of the current ones have the icons for it. With my setup in this PR, I can see little reason why it shouldn't work. It uses the correct files for finding the icons through the if else if statement.

Anewbe:
Alrighty.

Yoshax:
I've found no issues with it currently. It works, or at least has for me. Might take a little getting used to for some folks, but I think it's neater than having to use all the different files and species ones. 

Feel free to merge, or not. If you do and any bugs crop up, I can try and take care of them provided I'm made aware of them. I believe I got all the missing sprites, but as said, testing every item of clothing isn't too feasible. It would be a good place for a unit test really.

Anewbe:
So, to double check on this, if I make the sprites the way I normally would, in uniform.dmi, then use the tool, it should split it for me?

Yoshax:
Somewhat, yes. I only included the tool because it's what I used to do all the heavy-lifting. You have to put the icons into `DmiToSplit.dmi`, then when you run the program and use the verb it asks for a criteria could provide anything. 'Dress' for example. It would then take all icons with 'Dress' in their name and place them into a new dmi it prompts you to save, and another dmi which is the Original Dmi minus the icons placed into the new dmi.

Like I said in the OP, it may never have another use. But I included it because it may be useful to someone. Maybe to other servers who've hit the same problem.

If you're adding a new jumpsuit for example, it would be so much faster to just place the normal sprite and rolled down one into the correct dmi, which are named quite appropriately. 

You could run it on another dmis too for whatever purpose. Like you could plug in everything from `suit.dmi` and split all the icons with labcoat in the name out. Just as an example.

As a demonstration here's the military.dmi file I had to take from your other PR to get the icons:
https://i.gyazo.com/ca0a72926e644758a7c7db70561d0afe.png

I put them into the DmiToSplit.dmi and provided the criteria of `_r_s` , it gave me:
https://i.gyazo.com/2c128907be33087524dd809f2a9bfdd5.png

I then did it again for `_d_s` using the other dmi it provided to me and I was left with:
https://i.gyazo.com/e242243337a7b2fafaecc02bd2d92c10.png
In one file and all the rolled down icons in another file.

On something of that scale it's not exactly useful, and would be quick to do by hand. But with uniform.dmi there were over 150 rolled down/rolled sleeve icons, and I didn't want to miss any or mess it up, so I quickly wrote the IconSplitter to do it for me.

Yoshax:
Found a slight logic issue I'm trying to resolve so things can work smoothly if someone puts icons into the wrong dmi or such.

Anewbe:
Alrighty.

Yoshax:
Might be good now. I'll have to take a look tomorrow when it's not 3am.

But basically. The icons can be put into either `uniform.dmi` or the appropriate rolled down one, and it'll still work fine. This should also allow custom items that roll down to work fully, along with any sideloaded admin-muckery.

Yoshax:
Okay. Looked over it when I'm not dying of sleep deprivation and stuff. All seems well. Gets a little more complicated in picking sprites to use when rolling stuff down now, but that's because it'll work whether they're being loaded from custom items, an override, another file, or stuff.

Anewbe:
Alright, will do one last look over when I've got some time.

code/modules/clothing/under/miscellaneous.dm
```diff
@@ -351,16 +344,8 @@
 /*
  * wedding stuff
  */
-/obj/item/clothing/under/wedding
+/obj/item/clothing/under/wedding/
```
Anewbe: Is that last / needed for something?

code/modules/clothing/under/shorts.dm
```diff
@@ -98,18 +91,16 @@
 	name = "khaki short shorts"
 	icon_state = "khaki_shorts_f"
 
+/obj/item/clothing/under/shorts/loincloth
+	name = "loincloth"
+	desc = "A piece of cloth wrapped around the waist."
+	icon_state = "loincloth"
+
```
Anewbe: This should actually act as a skirt, for grammar reasons.

code/modules/clothing/under/solgov.dm
```diff
@@ -230,40 +212,33 @@
 	worn_state = "greenservice_com"
 
 //Dress
-/obj/item/clothing/under/mildress
```
Anewbe: This should probably stay as mildress, because /obj/item/clothing/under/dress is defined elsewhere, and these almost certainly should not be inheriting from that item.

Yoshax:
Sorry, those were all caused by the revert. I didn't realise there were changes other than all the path changes for the changed icon files in it.

Yoshax:
Let me know if there's any issues or bugs. I can hopefully fix them.

Anewbe:
@Yoshax Will do, thanks.

PR #3155, opened 2017-03-27
Flower crowns with fixed conflict? Maybe? Please?

Schnayy:
Had a Hell Of A Time with github for this. Adds poppy/sunflower/lavender flower crowns, a wood circlet, and the ability to add seeds to the circlet to make the crowns. Might've totally botched this, if so closing the pr and trying again in the morning.


Atermonera:
Looks like everything's the same between the pr's, so it should be fine

Anewbe:
Looks like you've got some sort of indentation error, should be a fairly simple fix.

If I'm reading Travis right you need to check these lines out:
code/modules/clothing/head/misc.dm :  /obj/item/clothing/head/sunflower_crown
code/modules/clothing/head/misc.dm :  /obj/item/clothing/head/lavender_crown
code/modules/clothing/head/misc.dm :  /obj/item/clothing/head/poppy_crown

Atermonera:
Either the indentation on the definition of the wooden circlet is off, or the indentation on everything else is, it looks like

Arokha:
Those lines appear to contain 1 space, then a tab, whereas all the other lines are a single tab.

Schnayy:
Thanks, fixing it now :)

code/game/objects/items/weapons/improvised_components.dm
```diff
@@ -68,3 +68,34 @@
...
+	icon = 'icons/obj/buildingobject.dmi'
+	icon_state = "woodcirclet"
+	w_class = ITEMSIZE_SMALL
+
+/obj/item/woodcirclet/attackby(obj/item/W as obj, mob/user as mob)
+	var/obj/item/complete
+	if(istype(W,/obj/item/seeds/poppyseed))
+		user << "You attach the poppy to the circlet and create a beautiful flower crown."
+		complete = new /obj/item/clothing/head/poppy_crown(get_turf(user))
+		qdel(W)
```
Neerti: Before calling qdel(W), it would be best to call user.drop_from_inventory(W) before deleting it, or some weirdness may occur.

Cerebulon:
(If user.drop_from_inventory(src) is unnecessary that's my bad... It was in the spear assembly code so I thought it might be needed here to definitely-delete the circlet as well as the seed packet)

PR #3158, opened 2017-03-27
FBPs should no longer scream when they collapse

Anewbe:
As far as I can tell from testing, this should only stop FBPs from getting pain messages. Mobs have a proc with an identical name that affects everything else, so they should still be taser-bait, and that worked in testing.
Fancy FBPs (Right now only Vey-Med) are supposed to be as close to human as possible, so they still feel pain.

### DIFF
--- a/code/game/objects/items/weapons/material/shards.dm
+++ b/code/game/objects/items/weapons/material/shards.dm
@@ -90,7 +90,7 @@
 					if(affecting.take_damage(force, 0))
 						H.UpdateDamageIcon()
 					H.updatehealth()
-					if(affecting.can_feel_pain())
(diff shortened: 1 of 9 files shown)


SilverTalismen:
So now Veymed FBP's will feel pain again?

Anewbe:
They already do, is the thing. This'll make everyone else /not/ get the pain messages, or scream when they fall over.

Atermonera:
Also, imo, wouldn't it make sense for designers to include warning messages that'll shout when stuff is breaking (I.e., pain)?

Neerti:
The limb damage indicator on the UI still should display the damage by the limb getting darker.

PR #3160, opened 2017-03-28
Borgs can climb

Anewbe:
Mostly tables

### DIFF
--- a/code/game/objects/structures.dm
+++ b/code/game/objects/structures.dm
@@ -174,7 +174,7 @@
 		return 0
 	if (user.stat || user.paralysis || user.sleeping || user.lying || user.weakened)
 		return 0
-	if (issilicon(user))
+	if (isAI(user))
 		user << "<span class='notice'>You need hands for this.</span>"
 		return 0
 	return 1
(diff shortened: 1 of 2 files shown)


Arokha:
LAW #5 DOWN BOY, DON'T GET ON TABLES

Anewbe:
YOU CAN'T STOP ME

Mechoid:
Law 6: The floor is lava.

Leshana:
Allowing some borgs to do it, but for dogborgs it should instead have another if that prints a "You aren't allowed on the table" message.

Anewbe:
We don't have dogborgs as a separate thing.

Atermonera:
"KERBEROS NO! BAD DOGS!"

Arokha:
> We don't have dogborgs as a separate thing.

Wait do you have dogborgs at all? I was mostly joking but you saying 'separate thing' makes me wonder if you have them as altsprites for medical borg or something.

Atermonera:
There's Tin Can and there's Kerberos, both of which are custom Borg sprites that're based on dogs in the code (Ian and Spice, respectively)

PR #3162, opened 2017-03-28
Adds Department Color Jackets

Schnayy:
Moves/renames red&black jacket to sec_dep_jacket, adds supply, engineering, medical, and science jackets. Puts them all in the loadout.

Have one issue: when the jacket is open there are no handsprites. How do I fix that without duplicating sprites?

image


Atermonera:
When it gets opened, it looks for a different icon_state that's named the same as the base with an added tag or two. You might be able to overload the function and call the parent function to open it, then manually set the icon_state to the original Sprite, but it's probably best conform to the standards.

Also, if you don't make any differences between the icon_states for being open and closed it'll be hella hard to tell which state it's in within putting it on first

Schnayy:
I don't mean the icon for when it's held/on the ground, those are different already, I meant the hand sprites of it in-hand.

Anewbe:
Think you may need to duplicate sprites.

PR #3164, opened 2017-03-29
Changes character setup to use correct in-lore names for brain types.

Cerebulon:
Changes assisted/mechanical/digital to cybernetic/positronic/drone when in reference to brains in character setup to better reflect lore and ingame terms at the request of Elgeonmb.. Other organs use Cybernetic/Mechanical.

### DIFF
--- a/code/modules/client/preference_setup/general/03_body.dm
+++ b/code/modules/client/preference_setup/general/03_body.dm
@@ -574,6 +574,9 @@ var/global/list/valid_bloodtypes = list("A+", "A-", "B+", "B-", "AB+", "AB-", "O
 			else
 				organ_choices += "Assisted"
 				organ_choices += "Mechanical"
+		else
+			organ_choices += "Assisted"
+			organ_choices += "Mechanical"
 
 		var/new_state = input(user, "What state do you wish the organ to be in?") as null|anything in organ_choices
 		if(!new_state) return
--- a/code/modules/client/preference_setup/general/03_body.dm
+++ b/code/modules/client/preference_setup/general/03_body.dm
@@ -539,7 +539,7 @@ var/global/list/valid_bloodtypes = list("A+", "A-", "B+", "B-", "AB+", "AB-", "O
 						pref.organ_data[other_limb] = "cyborg"
 						pref.rlimb_data[other_limb] = choice
 					if(!pref.organ_data[O_BRAIN])
-						pref.organ_data[O_BRAIN] = "cybernetic"
+						pref.organ_data[O_BRAIN] = "assisted"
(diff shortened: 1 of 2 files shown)


Anewbe:
Travis is mad.

Cerebulon:
Is Travis being a nerd about backslashes again....

Cerebulon:
Is Travis being mean about backslashes somehow because I saw something about that in some other code? Because I'm not seeing an issue with the modules/organ/pain.dm file its saying there was a problem with?

Anewbe:
Yeah, should be fixed very shortly, that's on me.

Neerti:
You might want to change ```get_FBP_type()```, ```get_min_age()```, and ```get_max_age()```.

Anewbe:
I feel like this touches way more than just brains, will need to review again.

Cerebulon:
If it's safer I could revert the variable stuff to still be assisted/mechanical/digital and purely change the menu option text?

Anewbe:
Well, I more mean this'd worry me less if it was literally _only_ the brains that were being affected.

code/modules/client/preference_setup/general/03_body.dm
```diff
@@ -200,13 +200,17 @@ var/global/list/valid_bloodtypes = list("A+", "A-", "B+", "B-", "AB+", "AB-", "O
...
+					. += "\tPositronic [organ_name]"
+				else
+					. += "\tSynthetic [organ_name]"
+		else if(status == "drone")
 			++ind
 			if(ind > 1)
 				. += ", "
 			. += "\tDigital [organ_name]"
-		else if(status == "assisted")
+		else if(status == "cybernetic")
```
Anewbe: I'm not sure this will work, the only part that has a "cybernetic" is the brain. Probably need to make the brain part its own thing.

Cerebulon: Oh yeah, I missed that one and another one. Fixed now. Went through the file another time to be sure.

Anewbe:
Alright, are you able to test this code?

Cerebulon:
Wow I made a little bit of a mess still while reverting changes I made since the very first commit. Next commit should all work fully as intended, finally. Sorry for the slight disaster this PR has been.......

Cerebulon:
Ok. New options are purely menu changes, and all link back to the original variable options so any and all methods that relied on them are functioning exactly as they did before. Organ selection functions the same as before, just the display names are different for brains. Brains spawn in peoples heads correctly as before, and prosthetic organs appear inside people as before. No issues that I could find.

Anewbe:
If this breaks we may revert it, but I'mma merge it because it seems to be fine.

PR #3167, opened 2017-03-29
Cigarettes and cigars have a "Burnt but not done" sprite

Anewbe:


MagmaRam:
No changelog?

Anewbe:
Nah. Only doing this in the first place because Kholdstare wanted it; it's very much a minor thing.

PR #3172, opened 2017-03-29
Adds in Jumper Cables for FBP reviving

Anewbe:
If the rest of the code is good, I may map in a pair of them into robotics.
Should be cleaner than removing the heart/brain.


SilverTalismen:
Yay!

PR #3177, opened 2017-03-29
Diona limbs should be easier to remove

Anewbe:
As per player request, this should make it easier to remove limbs from a diona with damage, hopefully giving them some extra survivability overall.


SilverTalismen:
Yay Doona changes!

Anewbe:
Staff response was either neutral or positive.

PR #3179, opened 2017-03-30
Added a few fish as simple animals

Leshana:
* They are simple non-hostile simple_animal's that wander around a bit.
* Wandering code is overriden so they will not leave the water.  And they will take damage as an unsuitable atmosphere if they do.
* Fish sprites improved and animated by Kligor


code/modules/mob/living/simple_animal/animals/fish.dm
```diff
@@ -0,0 +1,73 @@
...
+				if(T && is_type_in_list(T, suitable_turf_types))
+					Move(T)
+					lifes_since_move = 0
+
+// Take damage if we are not in water
+/mob/living/simple_animal/fish/handle_breathing()
+	var/turf/T = get_turf(src)
+	if(T && !is_type_in_list(T, suitable_turf_types))
+		if(prob(50))
+			say(pick("Blub", "Glub", "Burble"))
```
Anewbe: This seems like a very chatty fish.

Leshana: Yes, but only when the fish is *out of the water*.  When in water they say nothing.

Anewbe: Ah, that'd... This is what I get for trying to Dev while in my right mind.

PR #3180, opened 2017-03-30
Gives fish an item drop

Anewbe:
~~Drops carp HP to 20. They show up in massive swarms, and this will let a single taser shot take them out.~~

Gives fish mobs a proper drop.


Leshana:
Good call, seems that was not really a proper type to drop.

PR #3181, opened 2017-03-30
Modernize the automatic pipe layer

Leshana:
* Upgrades the automatic pipe layer to modernize its code a bit towards latest standards and fixing various bugs.
* Make it constructable so it can actually be used by someone!
* Prevent infinate steel generation by limiting pipe recycling to pipes made in this machine.  Otherwise the (free) pipes from the normal dispenser could be recycled into steel!

### DIFF
--- a/code/game/machinery/pipe/pipelayer.dm
+++ b/code/game/machinery/pipe/pipelayer.dm
@@ -1,57 +1,101 @@
 /obj/machinery/pipelayer
-
 	name = "automatic pipe layer"
 	icon = 'icons/obj/stationobjs.dmi'
 	icon_state = "pipe_d"
 	density = 1
-	var/turf/old_turf
-	var/old_dir
-	var/on = 0
-	var/a_dis = 0
-	var/P_type = 0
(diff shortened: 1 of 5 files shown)


code/game/machinery/pipe/pipelayer.dm
```diff
@@ -1,89 +1,121 @@
...
-	var/obj/item/weapon/wrench/W
+	circuit = /obj/item/weapon/circuitboard/pipelayer
+	var/turf/old_turf		// Last turf we were on.
+	var/old_dir				// Last direction we were facing.
+	var/on = 0				// Pipelaying online?
+	var/a_dis = 0			// Auto-dismantling - If enabled it will remove floor tiles
+	var/P_type = 0			// Currently selected pipe type
+	var/P_type_t = ""		// Name of currently selected pipe type
+	var/max_metal = 50		// Max capacity for internal metal storage
+	var/metal = 10			// Current amount in internal metal storage
```
Leshana: You are correct! Changed.

PR #3183, opened 2017-03-31
Overdose is more dangerous, Bicaridine works again

Anewbe:
Turns out the thing was either not working, or working to slowly. Now it's in the overdose for bicaridine.
Also fixes the same issue with myelamine, because apparently that wasn't fixing fast enough either.

Overdose in general does more Tox now, Skrell are resistant to the effects of overdosing.


code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine.dm
```diff
@@ -32,6 +32,23 @@
...
+	M.eye_blurry += (wound_heal)
+	if(ishuman(M))
+		var/mob/living/carbon/human/H = M
+		for(var/obj/item/organ/external/O in H.bad_external_organs)
+			for(var/datum/wound/W in O.wounds)
+				if(W.bleeding())
+					W.damage = max(W.damage - wound_heal, 0)
+					if(W.damage <= 0)
+						O.wounds -= W
+				if(W.internal)
```
Yoshax: You could do this in one if statement because you're doing the same in both. I believe.
Can just have `if(W.bleeding() || W.internal)` or something.

Anewbe: I'll keep that in mind, will probably go back to do it later. Thank you.

PR #3188, opened 2017-03-31
Symbols make sprite accessories freak out

Arokha:
Like plus or ampersand or single or escaped quotes in the name, it goes crazy. You can't remove Socks+Belly because it has a plus, once you've added it. I have no idea why. But here's a quick fix. (The ones with slashes weren't problematic but I'm a sucker for consistency. The only broken one was the one with the plus sign).


Anewbe:
Was about to make an issue report for that.

PR #3197, opened 2017-04-01
Lung damage QOL

Anewbe:
You shouldn't get popped lungs from being choked or drowning (I don't think), and brain damage from OxyLoss only starts after having 20 Oxy.

### DIFF
--- a/code/modules/mob/living/carbon/human/life.dm
+++ b/code/modules/mob/living/carbon/human/life.dm
@@ -352,13 +352,6 @@
 			if(!L.is_bruised() && prob(8))
 				rupture_lung()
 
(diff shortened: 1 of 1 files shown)


Yoshax:
You really should put the Brain Damage stuff outside of the failed breath statement. Means there's no Brain Damage from severe bloodloss or anything of the sort, which should probably happen. Would mean bad heart damage would cause it as well, like it probably should.

Anewbe:
Good point.

Anewbe:
Moved it into handle_regular_status_updates(), don't know if there's a specifically better one.

Yoshax:
That seems fine. :+1:

PR #3200, opened 2017-04-01
Translators and animal languages

### LINKED ISSUE #3161: Translators can understand monkeys and dogs
Ian can talk to people with translators
Barks and growls, it's a dog.
#### What actually happened
... Fluent conversation?
#### Steps to reproduce
1.) Get a dog
2.) Get a translator, turn it on
3.) Hold a conversation with the dog
#### Additional info:

Atermonera:
Fixes #3161 
Dogs, cats, birds, and mice all have their own language, rather than inheriting Galcom from the base simple_animal.
Translators will now only translate languages who haven't had machine_understands set to 0

This was fully tested. I couldn't get the verbs for the new languages to work properly but the syllables were working, and a human without the languages was unable to understand them.
The current verbs the languages being used are the default (says, asks, and exclaims), and I used the exact same syntax as every other language uses, so I honestly don't know why it isn't working, but I also expect this to not be a very big issue.

It also looks like the commit to update my local repo snuck in again. It doesn't add any actual file changes, so all is well.


code/game/objects/items/devices/translator.dm
```diff
@@ -16,7 +16,8 @@
 	if(!listening) //Turning ON
 		langset = input(user,"Translate to which of your languages?","Language Selection") as null|anything in user.languages
 		if(langset)
-			if(langset && ((langset.flags & NONVERBAL) || (langset.flags & HIVEMIND)))
+			if(langset && ((langset.flags & NONVERBAL) || (!langset.machine_understands)))
```
Anewbe: Should probably leave hivemind in there, just to be safe.

PR #3208, opened 2017-04-02
Drones and pAIs can't be antags

Anewbe:


Kasuobes:
Might be better to add flags to mobs to disable antag selection instead of removing them from the pool then removing them.

Baystation12/Baystation12#13545

Anewbe:
Will have a better way to do this, leaving open as a reminder to self.

PR #3209, opened 2017-04-02
FBPs can have numbers in their names.

### LINKED ISSUE #3157: Drones/Posi's and Numbers
Drones and Posi Brains cannot have numbers, such as 16, in their name
To be able to be Drone number 16 or something
#### What actually happened
Had to be Drone number Sixteen
#### Steps to reproduce
Make a posi or drone brained FBP, try to set a number in your name, be forced to spell it out
#### Additional info:
 - **Server Revision**:master - 2017-03-26

Atermonera:
Ok so pushing it did a weird and didn't let me make a new branch, fortunately it appears to have not mattered.

Fixes #3157 
FBPs of any sort can have numbers in their name, per the title. sanitize_name is now also called when reset_limbs() is called, so if someone tries to be sneaky and get a numbername while synthed, then unsynth, it'll give them a random name.
Because of something already existing in sanitizeName(), the first character can't be a number.

### DIFF
--- a/code/_helpers/mobs.dm
+++ b/code/_helpers/mobs.dm
@@ -63,12 +63,12 @@ proc/random_facial_hair_style(gender, species = "Human")
 
 		return f_style
 
-proc/sanitize_name(name, species = "Human")
+proc/sanitize_name(name, species = "Human", robot = 0)
 	var/datum/species/current_species
 	if(species)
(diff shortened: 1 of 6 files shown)


Atermonera:
[Testing](https://puu.sh/v7bif/be8e86f2d4.mp4)

PR #3212, opened 2017-04-02
Make cells show correct icon when they spawn

Arokha:
Initialize is great for cells sitting on a counter or something when the map loads. But not really for cells in guns that R&D prints and stuff. Kinda needs to be in New(). Also the Initialize call was completely pointless, so, removed.

### DIFF
--- a/code/modules/power/cell.dm
+++ b/code/modules/power/cell.dm
@@ -5,9 +5,6 @@
 /obj/item/weapon/cell/New()
 	..()
 	charge = maxcharge
-
-/obj/item/weapon/cell/initialize()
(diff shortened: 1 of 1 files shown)


Neerti:
Unless I'm looking at this wrong, I don't see you adding update_icon() going in New().

Anewbe:
They didn't delete the one that used to be in initialize, or it takes the same space it used to, so the diff doesn't highlight it.

Yoshax:
They entirely removed the `initialize()` call, and update icon goes in the what was whitespace, and becomes part of `new()`

PR #3213, opened 2017-04-02
Fix nanopaste item size

Arokha:
I'm pretty sure this is an oversight, rather than a balance thing. Nanopaste is too big to fit in boxes or anywhere else, but seems to be identical to all other 10-stack items like medical stacks in other regards. It should probably be the same size as the other 10-stack items.

Even the icon seems to suggest this, since right now it is just as big as an entire box, though the icon seems to be more like toothpaste-tube and looks awkward sitting in the huge backdrop inventory space.

### DIFF
--- a/code/game/objects/items/stacks/nanopaste.dm
+++ b/code/game/objects/items/stacks/nanopaste.dm
@@ -6,7 +6,7 @@
 	icon_state = "tube"
 	origin_tech = list(TECH_MATERIAL = 4, TECH_ENGINEERING = 3)
 	amount = 10
-
+	w_class = ITEMSIZE_SMALL
 
 /obj/item/stack/nanopaste/attack(mob/living/M as mob, mob/user as mob)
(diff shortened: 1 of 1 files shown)


SilverTalismen:
I think it was made too big so it couldn't just be pocketed for easy repairs for antag FBP's

drexample:
But organic antags can pocket pill bottles.
Also nanopaste isn't that effective.

Arokha:
> I think it was made too big so it couldn't just be pocketed for easy repairs for antag FBP's

Nanopaste does not repair synths externally on your server, and you cannot perform surgery on yourself, so this does not tie into any sort of balance change.

Anewbe:
Not a balance issue, for us, yeah.

SilverTalismen:
I more so mean back when uh We had IPC's who didn't need surgery to be fixed but could just nanopaste themselves and self repair

PR #3214, opened 2017-04-03
Robolimbs last longer without negative effects

Anewbe:


Arokha:
Thank gawed.

"Oh no I got hit in the arm once and now that arm is just USELESS"

PR #3225, opened 2017-04-05
Make energy nets better

Arokha:
You can imagine them as 8-second-resist office chairs that the mob is buckled to now. You can pull them around in the net, too, since it's baaaasically like an office chair now. They can't pull themselves around like office chairs while in the net, though. Because that'd be hax. You can click them (or resist) to start tearing them up (same from the outside).

Also keep in mind the mob inside can still wreck anyone adjacent (or shoot you). That's how it was before, too. So you can't really run up and drag them to prevent escape without them punching you (shouldn't you be roleplaying all this anyway?). Switching hands interrupts the resist so no reloading, getting other weapons, using PDA etc while resisting.

Also got rid of this SUPER SNOWFLAKE VARIABLE on all mobs used literally only by energy nets.

Netgun I made for vorestation not included.

### DIFF
--- a/code/game/objects/items/weapons/weaponry.dm
+++ b/code/game/objects/items/weapons/weaponry.dm
@@ -80,10 +80,8 @@
 	var/turf/T = get_turf(M)
 	if(T)
 		var/obj/effect/energy_net/net = new net_type(T)
-		net.layer = M.layer+1
-		M.captured = 1
-		net.affecting = M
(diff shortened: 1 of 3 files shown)


Anewbe:
Tested this?

Arokha:
Yup. Works fine assuming you're fine with the mechanics change.

Anewbe:
@Arokha Actually, if you can make a changelog for this it'd be appreciated.

PR #3226, opened 2017-04-05
Adds a LeMat Revolver

Yoshax:
Hi folks! Coded this up for fun/proof of concept when I was bored, and thought I'd be as well PRing it somewhere so that some people may make use of it? Or somethnig!

https://en.wikipedia.org/wiki/LeMat_Revolver

The LeMat Revolver is a 9 shot revolver that for the purposes of SS13 is chambered in .38, because I didn't want to make a new ammo type, and .357 is kinda strong! It has a secondary firing mode accessed by the verb, which lets you use the underbarrel shotgun, which loads normal shotgun rounds of any type.

Haven't added it do anything, or anywhere. Others may want to do that, assuming this is wanted at all. Code could likely be better, as it was mostly just a thing I did for fun. But here it is.

Also still slightly WIP.


code/modules/projectiles/guns/projectile/revolver.dm
```diff
@@ -114,4 +114,85 @@
...
+	var/secondary_max_shells = 1
+	var/secondary_caliber = "shotgun"
+	var/secondary_ammo_type = /obj/item/ammo_casing/shotgun
+	var/flipped_firing = 0
+	var/list/secondary_loaded = list()
+	var/list/tertiary_loaded = list()
+
+
+/obj/item/weapon/gun/projectile/revolver/lemat/New()
+	world << "1"
```
Anewbe: World <<

code/modules/projectiles/guns/projectile/revolver.dm
```diff
@@ -114,4 +114,85 @@
...
+	for(var/i in 1 to secondary_max_shells)
+		world << "2"
```
Anewbe: World <<

code/modules/projectiles/guns/projectile/revolver.dm
```diff
@@ -114,4 +114,85 @@
...
+		secondary_loaded += new secondary_ammo_type(src)
+	..()
+
+/obj/item/weapon/gun/projectile/revolver/lemat/verb/swap_firingmode()
```
Anewbe: Can't you just tie this to the existing code, like the Z8 Bulldog has?

code/modules/projectiles/guns/projectile/revolver.dm
```diff
@@ -114,4 +114,85 @@
...
+	handle_casings = CYCLE_CASINGS
+	max_shells = 9
+	caliber = "38"
+	ammo_type = /obj/item/ammo_casing/c38
+	var/secondary_max_shells = 1
+	var/secondary_caliber = "shotgun"
+	var/secondary_ammo_type = /obj/item/ammo_casing/shotgun
+	var/flipped_firing = 0
+	var/list/secondary_loaded = list()
+	var/list/tertiary_loaded = list()
```
Anewbe: What's the tertiary loaded for?

code/modules/projectiles/guns/projectile/revolver.dm
Yoshax: I could have. But Revolvers have attack_hand to be unload all the ammo in the gun, while clicking it with your other hand removes one casing. I kinda wanted to retain that because it's somewhat necessary for unloading the gun when you're done shooting. Either way, I'd have to either have my own thing, like here, or override one of the attack things to make it unload all the ammo at once when clicked with other hand rather than one.

The Bulldog is probably quite a bit better done, but I've still got a few things to add. Like swapping firemodes automatically when shooting the shotgun underbarrel.

code/modules/projectiles/guns/projectile/revolver.dm
Yoshax: Yeh, I'll get to it. Still a little WIP.

code/modules/projectiles/guns/projectile/revolver.dm
Yoshax: Kinda snowflake but it's a third list used to hold the loaded ammo when swapping between the lists.   It's used in swap_firingmode() to entirely swap between the two different kinds of ammo loaded.

Yoshax:
Mostly done now, I think. Added and fixed a few things.

Anewbe:
So you pushed it back to the first commit, then?

Yoshax:
Yeh, ammended my first commit. Just the finishing touching, and adding a proper examine message so you can see when the underbarrel is loaded, and when it isn't, and what it's loaded with.

PR #3227, opened 2017-04-05
Arokha changelog

Arokha:


html/changelogs/Arokha - Changes.yml
```diff
@@ -0,0 +1,37 @@
...
+#   rscdel (general deleting of nice things)
+#   imageadd
+#   imagedel
+#   maptweak
+#   spellcheck (typo fixes)
+#   experiment
+#################################
+
+# Your name.  
+author: N3X15
```
Neerti: Might wanna fix this.

Arokha: I have no idea what happened. I hit save, I swear, and yet it was the default changelog when I committed. Anyway, it's ammended.

Anewbe: I'm not seeing it amended here.

Arokha:
I think there was some kinda mummy or voodoo curse. I think it's gone now.

PR #3237, opened 2017-04-06
Changes the lemat sprite

Anewbe:
~~Sprite was too small, just flipped the revolver for the time being~~

New sprite c/o Mechoid


redstryker:
You could just use alpha.

PR #3239, opened 2017-04-07
Expands traitor codes to use the phonetic alphabet.

Belsima:
Now uses the entire phonetic alphabet, rather than just A, B, and D. Stolen right from https://github.com/tgstation/tgstation/pull/19866/.


Yoshax:
I know it's just a port, but that kinda thing would be better/neat as a general proc. If it were to be used again somewhere, you'd need that whole, long kinda messy thing again. And changes wouldn't persist.

Belsima:
...I'll look into doing that instead when I get home, actually.

PR #3240, opened 2017-04-07
Gives renegade the LeMat.

Belsima:
I was thinking this was a bad idea, but it doesn't use .357 and they already get a damn strong revolver option anyway.


Yoshax:
It's a pretty weak weapon in comparison to the Revolver or suchlike, so it shouldn't really be a worry. You get (9 * 20) + 50 = 360 vs the Revolver which is 6 * 60 = 360. Only upside really is you have 10 total shots, and of course the swagger that comes with having a Revolver with an underbarrel Shotgun.

Belsima:
Oh, that's not too bad at all. The 'revolver with a shotgun' basic idea just spooked me a good bit.

Yoshax:
You could probably get away with adding it to the uplink as well. Pricing it may be a little difficult though.

Belsima:
I'll add it to the uplink in my uplink PR.

PR #3244, opened 2017-04-08
Tweaks blindfold code slightly

Anewbe:
Moved it to where the blurry vision code is

### DIFF
--- a/code/modules/mob/living/carbon/human/life.dm
+++ b/code/modules/mob/living/carbon/human/life.dm
@@ -992,14 +992,15 @@
 			else if(eye_blind)		  	// Blindness, heals slowly over time
 				eye_blind =  max(eye_blind-1,0)
 				blinded =    1
-			else if(istype(glasses, /obj/item/clothing/glasses/sunglasses/blindfold))	//resting your eyes with a blindfold heals blurry eyes faster
-				eye_blurry = max(eye_blurry-3, 0)
-				blinded =    1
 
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/life.dm
```diff
@@ -992,14 +992,15 @@
...
-			else if(istype(glasses, /obj/item/clothing/glasses/sunglasses/blindfold))	//resting your eyes with a blindfold heals blurry eyes faster
-				eye_blurry = max(eye_blurry-3, 0)
-				blinded =    1
 
 			//blurry sight
 			if(vision.is_bruised())   // Vision organs impaired? Permablurry.
 				eye_blurry = 1
-			if(eye_blurry)	           // Blurry eyes heal slowly
+			if(istype(glasses, /obj/item/clothing/glasses/sunglasses/blindfold))	//resting your eyes with a blindfold heals blurry eyes faster
+				if(eye_blurry)
```
RemieRichards: ```max(eye_blurry-3, 0)``` means its lowest bound is 0, the same as if you hadn't done this check.

Anewbe: The check means it won't run if it doesn't need to.

RemieRichards: The cost of the if won't be much cheaper than just running the code.
Surely you know this?

PR #3250, opened 2017-04-09
Makes wood walls into wooden logs

BillyBangles:
For that romantic cabin feel that you never wanted in your space station 13 map.


Cerebulon:
Wouldn't it be better to add this as an option rather than replace the current wooden walls entirely?

BillyBangles:
I'm not entirely sure how possible that is.

Atermonera:
You could make it constructable or of planks, like the wood bat is

BillyBangles:
I'm not sure it can fit together nicely like walls do if I do it like that. Or at least, I wouldn't know how to do that.

Neerti:
The best way of doing it might be add a new material type called logs, then turn the 'logs' you get from hydroponics from plant fruit to material stacks of the new log material, and have them point to your new log sprites for walls.

Anewbe:
@BillyBangles Are you still working on this?

BillyBangles:
I intend to do something along the lines of what Neerti said, but I probably won't get around to it until we finish our new map. You can close this if you want, and I'll just open a new one later when I have the time, but I will get back to this eventually, either way.

Anewbe:
Gonna close this for the time being, as no progress is being made.

PR #3251, opened 2017-04-09
Fixes Immortal (to melee) Borgs

Neerti:
Should make cyborgs take damage from melee weapons.


Anewbe:
And now I feel dumb, was staring at that exact bit of code...

PR #3260, opened 2017-04-10
Fix 'Spawn Character' DNA

Arokha:
Made spawned characters always use male sprites and not have accurate DNA. Also would runtime if you picked the option to not put a mind in, as it assumed there would be one.


Anewbe:
Tested?

Arokha:
Yup!

PR #3264, opened 2017-04-10
Adds carp welding mask.

Belsima:
Tested. Seems to work fine. Access expanded to QM and CT because they can print masks anyway. Ported from https://github.com/Baystation12/Baystation12/pull/15999.


Belsima:
Log added.

Belsima:
**_oh god merge conflicts_**

PR #3266, opened 2017-04-11
Armor soak now caps at 80% of incoming damage

Anewbe:
Soak no longer flat ignores damage less than it, instead causing "scratch damage".


Spacemanspark:
What about the admin cybernetic armor jumpsuit item

PR #3269, opened 2017-04-11
Process deferral and profiling improvements

Leshana:
* Merges Krausus's https://github.com/ParadiseSS13/Paradise/pull/5472 from Paradise'
See there for details, but highlight of the changes impacting this codebase:
* tick_usage is now checked directly inside the SCHECK macro, so its a lil bit faster.
* Same-tick and cross-tick deferral is now based solely on tick_usage, so its all based on how much time is left in the current tick, with the intent of never going above 100 cpu by deferring work to next BYOND tick.
* By default, a process will sleep(0) for every 20% of a tick it uses, and sleep until the next tick at 90% usage.

Note: This is also a semi-prerequisite to lighting upgrade, so it can be scheduled nicely.

### DIFF
--- a/code/__defines/process_scheduler.dm
+++ b/code/__defines/process_scheduler.dm
@@ -7,13 +7,12 @@
 #define PROCESS_STATUS_HUNG 6
 
 // Process time thresholds
-#define PROCESS_DEFAULT_HANG_WARNING_TIME 	300 // 30 seconds
-#define PROCESS_DEFAULT_HANG_ALERT_TIME 	600 // 60 seconds
-#define PROCESS_DEFAULT_HANG_RESTART_TIME 	900 // 90 seconds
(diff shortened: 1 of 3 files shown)


Anewbe:
Has this been tested?

Leshana:
Yes I have tested it.  While testing I encountered one anomaly that occurs if you use the "Break all lights" secret, but I determined its a pre-existing issue (Same thing occurs on branch master: https://github.com/PolarisSS13/Polaris/issues/3280) I'll add a fix for that issue separately.
In fact that experience showed it works pretty decently, because it cleaned up the lighting effects without lagging movement, and the lighting updated (super slow while server was being bogged by the break-all-lights) smoothly too.   Oh yes, original test was with the smooth-lighting changes, but I tested without that too.

PR #3271, opened 2017-04-11
Quick & Dirty Panic Bunker Port

Arokha:
Tested on our server, works. Doesn't work if you don't have a DB (how would it?).


SilverTalismen:
Panic Bunker?

Arokha:
It prevents never-before-seen players from joining while it's enabled, to help mitigate raids and the like.

PR #3272, opened 2017-04-11
Can remove ID from PDA via Alt-Clicking now.

LorenLuke:
Ported from Baystation12.

### DIFF
--- a/code/game/objects/items/devices/PDA/PDA.dm
+++ b/code/game/objects/items/devices/PDA/PDA.dm
@@ -66,7 +66,7 @@ var/global/list/obj/item/device/pda/PDAs = list()
 
 /obj/item/device/pda/examine(mob/user)
 	if(..(user, 1))
-		user << "The time [stationtime2text()] is displayed in the corner of the screen."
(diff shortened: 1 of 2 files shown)


code/game/objects/items/devices/PDA/PDA.dm
```diff
@@ -68,6 +68,18 @@ var/global/list/obj/item/device/pda/PDAs = list()
...
+
+/obj/item/device/pda/AltClick()
+	if(issilicon(usr))
+		return
+
+	if ( can_use(usr) )
+		if(id)
+			remove_id()
+		else
+			usr << "<span class='notice'>This PDA does not have an ID in it.</span>"
```
Anewbe: Use the to_chat() call, please.

Leshana:
This is epic

LorenLuke:
Changed a lot of << to 'to_chat()'.

Three are left, two of which are browser windows and one was a bit suspect, so I didn't want to touch that and break it (L1145). Otherwise, all changed.

PR #3274, opened 2017-04-11
Resisting out of choking grabs isn't 100% guaranteed anymore. 

LorenLuke:
Tweaked some ling checks to compensate.

Ported from Bay12, via own PR.


LorenLuke:
Mind, because grab_kill isn't mentioned it means you can get out of a choking neckgrab 100% of the time if you can resist at all. Really messes with changelings.

PR #3275, opened 2017-04-11
Grabs can be downgraded.

Anewbe:
Someone requested it.
Doesn't go lower than passive grab.

- [ ] Add a timer in between grab downgrades?
- [ ] Should the grab warning at the lowest tier be black, or red like every other one?


LorenLuke:
See this for comparison:
https://github.com/Baystation12/Baystation12/pull/15160

Anewbe:
Don't feel like playing with grabs right now, closing.

PR #3276, opened 2017-04-11
Weakens bloodloss from burn damage

Anewbe:
75% as strong as before


Belsima:
Another burn nerf? Didn't we already lower pain caused? I feel like we're slowly making ballistics and energy weapons the same.

Anewbe:
The pain nerf was because all it did was lock people in paincrit in out-of-the-way places. This is because it has been decided that the bloodloss from burn damage was too high.

PR #3286, opened 2017-04-12
Port tg/paradise/goon soft-edge lighting

Leshana:
Ports:
* https://github.com/ParadiseSS13/Paradise/pull/6161 
* https://github.com/tgstation/tgstation/pull/25778 
* https://github.com/tgstation/tgstation/pull/25861

The Paradise lighting module is merged nearly verbatim.  Changes outside the lighting module are merged based on our changes.  See that PR for major summary.
On top of that, from tgstation: Optimizes lighting overlays for slow clients.
* Instead of using color matrix for all overlays, fully dark and fully light use static black and transparent icon states without color matrix.
* Also implements LIGHTING_ROUND_VALUE so that lighting values actually will reach 0 or 1.

polaris smooth lighting

Technical changes outside the lighting module relevant to downstream users:
* Replace `opacity = X` with `set_opacity(X)` so that the lighting system can update properly.
* Replace direct access of lighting overlays with T.get_lumcount()
* Rename lighting_use_dynamic to dynamic_lighting


PJB3005:
> no mention of /vg/

Arokha:
~~Who?~~ /s

I think it's primarily because we literally took the commits from Paradise and /tg/. Primarily Paradise, as they were the most similar codebase that has this lighting, with optimizations that /tg/ has that Paradise has not yet ported. Goon is credited because they came up with the idea of color-matrix-based lighting, as far as we know. Could be wrong. Post a comment on this web zone if we are wrong.

Though, you're probably super-familiar with /vg/... do you see any optimizations that you've made that might-could be included?

PJB3005:
No idea as long as you've got TG's stuff.

Anewbe:
Important question: How does this work with mob darkvision? Does it fix anything a la Taj darksight?

Arokha:
Kindasorta! They can interact with stuff within their proper darkvision range now. But as far as I can see, they can't actually really SEE those turfs? Like I'm not able to visually pick out objects on them. Or, maybe our monitor contrast sucks? We're both using laptops.

So. It makes it a little better, at least, or maybe it's amazing and we have sucky monitors. One of those.

Yoshax:
It appears to be basically the same as before, it's all dark: https://i.gyazo.com/068423e08851aa3feee9e7e8b6cf5f4e.png

And you can still hover over things and see the name, and rightclick the tiles. Mesons still give Taj and such nightvision though.

Arokha:
I see, in earlier implementations it worked. In Paradise's current implementation of this, it appears that species do retain their darkvision, however the common complaint from Paradise players when I mentioned porting this was that there is no "absolutely dark" turf, everyone can always see every turf even with no light, just very dimly because the darkest state was not completely opaque.

It seems like the performance optimizations in 959b117 ported from /tg/ might undo that, so it's a performance/feature tradeoff. Paradise does not currently have the optimizations in 959b117 whereas /tg/ does.

959b117 among other things changes from using a png to a dmi that has a few icon_states, including a "completely black" one to avoid doing matrix math on turfs that are completely black. There might be a way around this, though then you move back to 'total darkness is not actually totally dark'.

Anewbe:
At some point I'd really like that code to work again, but for the time being I'll sit back and shush.

PJB3005:
As for reimplementing soft lighting under TG optimizations: it changes the icon state of the lighting overlay so it might be possible to change that icon state to set to the soft lighting values.

Leshana:
Reflomphed to resolve merge conflicts.

Leshana:
@PJB3005 is correct, changing the icon state to match the soft threshold works.  I left both icon states in there and made it a #define option to choose which one so it can be put back later if we decide we don't like it, but hey its now here again.

Anewbe:
This apparently works fine, I'm going to stare at it in utter terror for another day or so, then if I still can't find anything it'll probably be merged.

PR #3288, opened 2017-04-12
Hardhat tweaks

Anewbe:
Hard hats now give the same ear protection as helmets, but can no longer fit in pockets.


SilverTalismen:
And uh, what does ear protection do?

Anewbe:
Effects from flashbangs, mostly. Have some other stuff I might make it do sooner-or-later.

PR #3291, opened 2017-04-13
Kills carp lag

Anewbe:
Server was slowing down when the carp wave 'left', this means more carp but less lag. Will need to be discussed.

### DIFF
--- a/code/modules/events/carp_migration.dm
+++ b/code/modules/events/carp_migration.dm
@@ -41,9 +41,11 @@
 		i++
 
 /datum/event/carp_migration/end()
-	for(var/mob/living/simple_animal/hostile/carp/C in spawned_carp)
-		if(!C.stat)
-			var/turf/T = get_turf(C)
-			if(istype(T, /turf/space))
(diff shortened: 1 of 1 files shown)


Leshana:
You also could sleep after deleting each carp, thus the deletions would be spread out over about 2 seconds and not lag anything.

Anewbe:
Yeah, that was the backup idea.

Anewbe:
Tried that, hung the event controller.

Leshana:
I forgot to mention you'd need to do it in a spawn to avoid that.
```
/datum/event/carp_migration/end()
	spawn(0)
		for(var/mob/living/simple_animal/hostile/C in spawned_carp)
			if(!C.stat)
				var/turf/T = get_turf(C)
				if(istype(T, /turf/space))
					qdel(C)
			sleep(1)
```
Ultimately the super-fancy way to do it would be to queue it in the scheduled event system so it would actually be SCHECK'd and stuff.  But that seems like overkill for a small number of objects like this probably?

PR #3292, opened 2017-04-13
Changes Shields

Neerti:
- Meteors now will respect wall health when colliding with walls.  This generally means r-walls are a lot more protective than previously.
- Shields should be more robust in stopping harm from meteors.  When a meteor hits a shield, the shield and meteor compare numbers.  If the shield is strong enough, it'll kill the meteor immediately.  Otherwise the shield dies and the meteor plows into the station.
- To compensate for the above changes, meteors have been beefed up somewhat.
- Shields have a new hexagon-based sprite inspired by Bay's shield sprites.  They also glow blue, and have a neat pulse effect when touched or damaged.
- Shield generator now has a shiny unused sprite from RUST sprites.  Capacitor sprites would be nice later.
- Shield generators can accept multiple capacitors now.
- Shield generators now have multi-Z support.  The option to shield across zlevels only shows if the map supports it.
- Broken shields are now visible instead of being invisible.
- Ports Bay's Shield Diffusers, both the stationary and hand-held varients.  Note ours will just delete the shield.
- Shields can now be hurt from melee weapons.

Pictures:
hexagones
shield animation


code/_helpers/unsorted.dm
```diff
@@ -1299,3 +1299,16 @@ var/mob/dview/dview_mob = new
...
 
+// Displays something as commonly used (non-submultiples) SI units.
+/proc/format_SI(var/number, var/symbol)
+	switch(round(abs(number)))
+		if(0 to 1000-1)
+			return "[number] [symbol]"
+		if(1e3 to 1e6-1)
+			return "[round(number / 1000, 0.1)] k[symbol]" // kilo
+		if(1e6 to 1e9-1)
+			return "[round(number / 1e6, 0.1)] m[symbol]" // mega
```
Leshana: Lowercase 'm' is "milli" not "mega"   Same with the other prefixes actually, but m/M is the one that is most misleading.  Consider the difference between  50 mW and 50 MW being 9 orders of magnitude.

Arokha:
PRETTY!

Anewbe:
I feel like the handheld should just be a melee weapon that does waaaay more damage to shields, unless it was that thing we discussed giving to the CE.

PR #3295, opened 2017-04-13
Adds a lever-action rifle

Anewbe:
... for bears and such.


Cerebulon:
And old-west cyborgs.

MagmaRam:
hell yes my man

This does make me wonder if we should add a new full-size rifle cartridge for the bolt-action and lever-action rifles. Both rifle calibers we have seem to be intermediate cartridges, judging by the fact that they're used in assault rifles. Still, that's something for another PR.

ps put in a changelog

Anewbe:
Rifle isn't put in game anywhere, so a changelog will just get people asking where it's at. When we roll around to actually having it somewhere that people can get their hands on it I'll changelog it then.

PR #3296, opened 2017-04-13
Scrubber Phoron Fixes

Yoshax:
- Scrubbers will now scrub Phoron by default instead of just CO2, this is a little silly and it's not a realistic goal to expect atmos to do the thing on the console for every single one every round.
- Air Alarms will now have the first danger zone at any amount of Phoron instead of 0.2, second danger zone remains 0.5. Note this is apparently kPa.

Do note, in a relatively small room, a small amount of Phoron, 0.5 moles, that wasn't triggering the alarm when set alight burned up to a cosy 200c

- Also fixes attacking alarms with your ID when trying to unlock due to a missing return.

### DIFF
--- a/code/ATMOSPHERICS/components/unary/vent_scrubber.dm
+++ b/code/ATMOSPHERICS/components/unary/vent_scrubber.dm
@@ -19,7 +19,7 @@
 
 	var/hibernate = 0 //Do we even process?
 	var/scrubbing = 1 //0 = siphoning, 1 = scrubbing
-	var/list/scrubbing_gas = list("carbon_dioxide")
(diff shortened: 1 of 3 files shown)


Wickedtemp:
Now add an audio file whenever spilled phoron is detected in chemistry. 
"HEY DUMBASS, DON'T IGNITE THIS SHIT, GOT IT?" on an endless loop.

Leshana:
I'm glad to have that missing return fixed! That was *almost* annoying enough to investigate it!  Of course by having them filter automatically there is even less need for atmos to fix spills anymore.  I'll have to see how that plays out in gameplay.

PR #3304, opened 2017-04-15
Make wingrille spawners usable

Arokha:
Makes the spawner object itself block air and prevent active edges before the game starts, so you can actually use these in maps now without creating tons of active edge warnings and making ZAS zones merge everywhere. Should save a loooot of time for mappers picking those window directions.


Anewbe:
Tested?

Arokha:
Yup, using it on our new planetary map, and the phoron atmosphere already let me know that wingrille spawners DIDN'T work before, and now they do. 😸

PR #3307, opened 2017-04-15
Allows blast doors to be attacked like airlocks.

LorenLuke:
Allows blast doors to be attacked and broken like regular airlocks.

### DIFF
--- a/code/game/machinery/doors/blast_door.dm
+++ b/code/game/machinery/doors/blast_door.dm
@@ -146,12 +146,15 @@
 // Parameters: None
 // Description: Opens the door. Does necessary checks. Automatically closes if autoclose is true
 /obj/machinery/door/blast/open(var/forced = 0)
-	if (src.operating || (stat & BROKEN || stat & NOPOWER))
-		return
-	force_open()
-	return 1
+	if(forced)
+		force_open()
+		return 1
+	else
+		if (src.operating || (stat & BROKEN || stat & NOPOWER))
+			return 1
+		force_open()
 
-	if(autoclose)
+	if(autoclose && src.operating || (stat ~& BROKEN && stat ~& NOPOWER))
 		spawn(150)
 			close()
 	return 1
--- a/code/game/machinery/doors/blast_door.dm
+++ b/code/game/machinery/doors/blast_door.dm
@@ -12,6 +12,7 @@
 	desc = "That looks like it doesn't open easily."
 	icon = 'icons/obj/doors/rapid_pdoor.dmi'
 	icon_state = null
+	min_force = 20 //minimum amount of force needed to damage the door with a melee weapon
 
 	// Icon states for different shutter types. Simply change this instead of rewriting the update_icon proc.
 	var/icon_state_open = null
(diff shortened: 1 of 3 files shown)


code/game/machinery/doors/door.dm
```diff
@@ -24,7 +24,7 @@
 	var/maxhealth = 300
 	var/health
 	var/destroy_hits = 10 //How many strong hits it takes to destroy the door
-	var/min_force = 10 //minimum amount of force needed to damage the door with a melee weapon
+	var/min_force = 20 //minimum amount of force needed to damage the door with a melee weapon
```
Neerti: Will this apply to normal airlocks?

LorenLuke: This does, I meant to have this be for the blast doors. I'll have it changed in a bit.

code/game/machinery/doors/blast_door.dm
```diff
@@ -123,14 +128,29 @@
...
 
 // Proc: open()
 // Parameters: None
 // Description: Opens the door. Does necessary checks. Automatically closes if autoclose is true
-/obj/machinery/door/blast/open()
+/obj/machinery/door/blast/open(var/forced = 0)
 	if (src.operating || (stat & BROKEN || stat & NOPOWER))
 		return
 	force_open()
+	return 1
```
Neerti: There's code below this line which will never get run due to returning early.

Anewbe:
code/game/machinery/doors/blast_door.dm:157:error: ~: missing comma ',' or right-paren ')'
code/game/machinery/doors/blast_door.dm:157:error: ~: missing comma ',' or right-paren ')'

code/game/machinery/doors/blast_door.dm
```diff
@@ -78,7 +79,10 @@
 // Proc: force_toggle()
 // Parameters: None
 // Description: Opens or closes the door, depending on current state. No checks are done inside this proc.
-/obj/machinery/door/blast/proc/force_toggle()
```
Anewbe: Do regular doors not have code for this already?

code/game/machinery/doors/blast_door.dm
```diff
@@ -123,15 +128,33 @@
 					usr << "<span class='warning'>You don't have enough sheets to repair this! You need at least [amt] sheets.</span>"
 
 
+		else if(src.density)
+			var/obj/item/weapon/W = C
+			user.setClickCooldown(DEFAULT_ATTACK_COOLDOWN*2)
```
Anewbe: Why is this doubled?

LorenLuke: Because of the fact that I wanted to strike a balance between the changeling arm_blade and the fact that these were hitherto impenetrable blast doors.

With the changeling arm blade, it takes 20 (600/(40/2)) swipes, meaning they can cleave through it in about 30 seconds, which is rather fast IMO for a blast-resistant door. Conversely doubling the number of strikes just makes it, again IMO, more tedious to break through, and spams comparatively more attack logs. The extra delay only affects attacking the doors, keeps the number of strikes the same and still pads out the amount of time it takes for them to break through the blast door, allowing someone attentive to respond to witness/call backup.

Anewbe: You've got a thing lower down that lowers the damage the door takes, please remove this line.

LorenLuke: Doing so would halve the time, and we're back to them breaking out in 30 seconds, just with fewer strikes.

Anewbe: Then lower the melee damage further.

LorenLuke: Resulting in more strikes, which I don't agree with. However since you insist, I'll do that then.

PR #3308, opened 2017-04-15
Tweaks Changeling Respecs

LorenLuke:
Allows changelings to 'bank' up to 3 ability respecs/re-adaptations at any one time, and changelings start with 2.


Anewbe:
Drop it to 2 and 1, please.

Anewbe:
As a notice, please do not fix multiple bugs or change multiple features in the same PR, if you can help it. Also, advertise what you're doing in a message or the main body of the first post, rather than making us search through commits for every little thing that's changed.

LorenLuke:
> As a notice, please do not fix multiple bugs or change multiple features in the same PR, if you can help it. Also, advertise what you're doing in a message or the main body of the first post, rather than making us search through commits for every little thing that's changed.

Will do. Do you want me to roll this back and split it, or just do that going forward?

Anewbe:
Just in the future.

PR #3311, opened 2017-04-15
Fixes ventcrawling for spiderbots/implants/etc.

LorenLuke:


LorenLuke:
Note: HAS been tested with spiderbots, implants, and borers.

I assume the other things work as well.

Anewbe:
I feel like it'd be more sensible to make this into procs for the correct subtype of mob, rather than having the proc check where you have it doing so.

LorenLuke:
@Anewbe It used to be that way, and I aggregated this just to have it be a master proc rather than have a bunch of 
return ..() 
after each one of them, plus it made it easier for me to re-order the events and not necessarily have it try to sort itself.

Additionally, ventcrawling happens so little, there's not massive amounts of overhead in doing this this way.

LorenLuke:
@Anewbe requesting review.

Anewbe:
Did you test this?

LorenLuke:
@Anewbe I have.

PR #3315, opened 2017-04-16
Allows placing tape masks/restraints back on the roll.

LorenLuke:


code/game/objects/items/weapons/tape.dm
```diff
@@ -163,6 +163,15 @@
 	overlays = null
 	qdel(src)
 
+/obj/item/weapon/ducttape/attackby(var/obj/item/I, var/mob/user)
+	if(!(istype(src, /obj/item/weapon/handcuffs/cable/tape) || istype(src, /obj/item/clothing/mask/muzzle/tape)))
+		return
```
MagmaRam: This should be ..() instead of 'return', calling the attackby proc for /obj/item/weapon. As it is, this would make certain items, like garbage bags, not work on duct tape.

PR #3317, opened 2017-04-16
Fix mining turfs permenently unlit.

Leshana:
* Mining turfs generated during random mapgen get their opacity changed
without a call to recalc_atom_opacity() because the game ticker has not
started.  Make sure we call it even before that point.

Note: Not yet tested.  Will test in a couple hours if not already verified.


Anewbe:
I tested this, seems to work, Travis is having a day off apparently.

PR #3321, opened 2017-04-17
No longer shall ye burn from water

### LINKED ISSUE #3320: Water is hilariously flammable
If you go in the pool, or other such water and then apply a flame to yourself, you'll die horribly in a burning inferno. It's hot enough to bur you to death wearing an atmos voidsuit.
To not die in an inferno.
#### What actually happened
Died in an inferno.
#### Steps to reproduce
Get in the water. Get out. Apply flame. Die.
#### Additional info:

Yoshax:
Makes water apply negative fire stacks instead of actual fire stacks. Means you can still expand on it in the future to do like... Phoron liquids you can swim in or something just by doing `water_act(not_a_negative)`

Fixes #3320

### DIFF
--- a/code/game/turfs/simulated/water.dm
+++ b/code/game/turfs/simulated/water.dm
@@ -44,7 +44,7 @@
 		L.update_water()
 		if(!istype(oldloc, /turf/simulated/floor/water))
 			to_chat(L, "<span class='warning'>You get drenched in water from entering \the [src]!</span>")
-	AM.water_act(5)
+	AM.water_act(-5)
 	..()
 
 /turf/simulated/floor/water/Exited(atom/movable/AM, atom/newloc)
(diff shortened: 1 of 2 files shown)


Anewbe:
If I'm remembering this right, you'll instead want to go down to https://github.com/PolarisSS13/Polaris/blob/38598f3aae9430796714a26ff44467c0bd8e4d5e/code/game/turfs/simulated/water.dm#L96
And change that.

water_act gets passed an amount, if you want the fire stacks to change it'd be better to change it in the appropriate call, methinks.

Yoshax:
That's correct. But that always assumes that amount will be negative, which it might not be. You could use the already existing code to do things like other flammable liquids people can wade through. For whatever purposes. That was my thoughtline anyway. All the existing water stuff will use the -5. If you'd prefer `amount` changed to be a negative in `waver_act` then I do that, just thought I'd lay out my thought process for why I did it the way I did.

Anewbe:
If that's the issue, you could make a var for the turf that would be called in Entered(), and have that be negative on a turf that increases fire stacks. 
For the time being I'd rather the amount entered by positive, and the change come in water_act().

Yoshax:
Done.

Anewbe:
Thank you.

PR #3324, opened 2017-04-17
Ports a few fixes to multi-z and related code from Bay

Leshana:
Ports:
https://github.com/Baystation12/Baystation12/pull/16653
https://github.com/Baystation12/Baystation12/pull/16941
https://github.com/Baystation12/Baystation12/pull/16942

#### Multi-Z code formatting and utilities.
* Global procs should start with /proc
* Use to_chat() macro
* Added utility to find all z-levels are connected (transitively adjacent) and utility to test if two levels are connected.

#### Fix crash in turbolifts built on any z-level other than "1"
* Turbolift auto-mapping code assumed that all lift's first floor would be on z=1.  Fixed so it picks the first floor instead of the zth floor from the floor list.

#### De-snowflake thrown objects hitting turfs.
* Actually tell turfs when a thrown object hits them, and let them decide what to do about it!
  * We do this by calling hitby(), which is how it already works for obj and mob, so this makes behavior consistent.
* This allows us to cleanly solve the problem of a thrown object landing on open space without falling.

P.S. Yes, I am porting my own bay PR's to here... It was good to get their input too.

### DIFF
--- a/code/_macros.dm
+++ b/code/_macros.dm
@@ -42,6 +42,8 @@
 
 #define isxeno(A) istype(A, /mob/living/simple_animal/xeno)
 
+#define isopenspace(A) istype(A, /turf/simulated/open)
+
 #define isweakref(A) istype(A, /weakref)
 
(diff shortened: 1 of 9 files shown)


Anewbe:
Tested, etc?

Leshana:
Yup.  I have been using this code on my test server for a few days actually.

PR #3325, opened 2017-04-17
Silenced Weapons tell firer when fired

LorenLuke:


Anewbe:
Make it so people who can see the thing know when it's being fired, please. Makes sense that the wouldn't hear it, but they'd probably be able to see someone pulling a trigger.

LorenLuke:
@Anewbe The issue is that the silencer suppresses the visible message to. I can put in a provision that it's only effectively a 'visible message' to say... three tiles around, or something. Would that work?

Anewbe:
If you can make it work, then maybe? I need to see it before I'll be able to say for sure.

PR #3328, opened 2017-04-18
Adds Makeshift Armor

Neerti:
Adds the ability to craft makeshift armor for the chest and head.  The armor uses the material system, so what material you use in building it heavily influences how protective it will be.

For the helmet, apply wire-cutters to a bucket, then apply a stack of whatever material you want to make the helmet out of to the newly made bucket helmet.
For the chest piece, you need to use a stack of material in your hand, then craft two 'armor plates'.  Apply wire to one of them, then hit the other together.  Both need to be made of the same material to work.

These kinds of armor can break if they get hit enough.  Brittle materials like glass will break on the first hit it receives, and will probably cause shards to imbed into you.

Generally, these armors are inferior to security armor, both due to lesser protection, higher vulnerability to tasers, and heavy slowdown, but they can be made somewhat cheaply (for steel at least), and is generally better than wearing nothing if shit gets real.

Numbers for common materials can be seen here https://hastebin.com/ukecofaduz.css

*Note that the bucket did not hide hair in the picture, it should now.
Chestpiece
Helmet


Yoshax:
Any example of how many hits things can take before breaking?

Neerti:
From my testing, generally the person wearing it would die from being hit repeatively before the armor if made from a reasonable material like iron.  The chance to take damage is based on the material's hardness, just like material weapons.

Yoshax:
Alright, cool! This is really neat.

PR #3329, opened 2017-04-18
Examining FBPs now shows their braintype

Neerti:
FBPs now will tell you what kind of brain is inside (assuming they have one) upon examination, with the format of 'This is X, a Positronic Android', as an example.
Cyborg FBPs show their species instead of 'cyborg' when examines, e.g. 'This is Y, a Skrell Gynoid'.
Vey-Med and future 'lifelike' FBP models are exempt from this, with the same rules as before.
Clothing such as spacesuits also hides this, as it did previously.

Warble Borg
Positronic Example

### DIFF
--- a/code/__defines/mobs.dm
+++ b/code/__defines/mobs.dm
@@ -201,3 +201,8 @@
 #define TASTE_NORMAL 1 //anything below 15%
 #define TASTE_DULL 0.5 //anything below 30%
 #define TASTE_NUMB 0.1 //anything below 150%
+
+#define FBP_NONE	""
+#define FBP_CYBORG	"cyborg"
+#define FBP_POSI	"positronic"
+#define FBP_DRONE	"drone"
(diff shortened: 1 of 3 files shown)


Belsima:
Why though? I like having, say, a positronic in an NT chassis that tries to pretend to e human (but sucks at it). How are you meant to tell what type of brain's in there just by glancing at them? Remember, flavor text is "what you can see", shouldn't this be the same?

Neerti:
Cyborgs already do this.

Belsima:
I would also prefer they didn't.

Belsima:
Could you explain the reasons for adding this here?

Atermonera:
Some people would prefer to keep their braintype anonymous unless directly asked, so making this a toggle of some sort, like gender is, would probably be preferrable

Belsima:
I still don't understand how a crewmember is meant to smell the fact that I'm actually a positronic, considering the rules state flavor text is only things visible.

SilverTalismen:
Not a big fan of this at all, Doesn't make sense that you can from a glance identify what type of synth your looking at.

Sorry

Neerti:
It has been decided that drones will instead receive a card they're required to carry ingame instead.

PR #3330, opened 2017-04-18
Implements footstep sound system and adds sounds to various floor types.

Leshana:
Here by popular demand:
* Adds system for turfs to have "footstep sounds" that play when someone walks over them.
* Adds system for floor types (the kind of turfs where you can put tiles on them) to support the footstep sounds.
* Adds sounds to plating, tiles, wood, and carpet floor types.
* Footstep sound system is configurable in game_options.txt, it can be turned off completely or the volume adjusted.

Footstep sound files sourced from https://github.com/Endless-Horizon/CEV-Eris


Leshana:
Note that this feature is *inoperative* unless turned on in config files.

LorenLuke:
@Leshana 
Without scouring through here,, is there a provision for a more muffled (or none at all) sound for walking?

Leshana:
@LorenLuke Depends what you mean by that.
* There is a global system configuration for adjusting the volume or entirely disabling footstep sounds server wide.
* Different floors do make different sounds, so floors can be silent by having an empty sound list.

If you mean can players turn off hearing footstep sounds? No.
If you mean, can some carbon/human's be silent by various means? Not as is no.  Eris has code that makes wearing certain shoes silence your footsteps.  Specifically syndiegaloshes.  Which while hilarious, wasn't something I bothered to implement so far.

PR #3332, opened 2017-04-18
Controllerized Planets

Arokha:
Controller with deferrals and SCHECKs to be specific. Won't lag while updating the sun, weather, or temperature.

Also moved some vars around. Namely the planet walls are stored on the planet, not in the weather_holder.

Planets now have their own turfs, the controller 'gives' them to the planets each cycle if there are any unallocated turfs in the global lists, to avoid iterating over other planets' turfs if you have more than one, then cuts the lists if you were crazy and some turf added some invalid type. This saves us type-checking in the for() loops later to make them crunch faster. The former operation should only happen once at the start of the game (and maybe very rarely when turfs are added/removed from a map during the game).

With regards to the temperature updates, rebuilding the zone entirely is an intensive operation. Instead we can use this new cheaty proc (credit to @Leshana) to do it from over here. ZAS code outside ZAS oh noooo. Well, the option is to snowflake this case into ZAS which is maybe worse?

Only downside to all this is that if you manually set weather and time it might take between 1-60 seconds for the controller to get around to checking if you wanted to update it. That's not that big a deal. If you really want you can now debug that controller and call doWork on it.

### DIFF
--- a/code/__defines/planets.dm
+++ b/code/__defines/planets.dm
@@ -1,0 +1,15 @@
+#define WEATHER_CLEAR				"clear"
+#define WEATHER_OVERCAST			"overcast"
+#define WEATHER_LIGHT_SNOW			"light snow"
+#define WEATHER_SNOW				"snow"
+#define WEATHER_BLIZZARD			"blizzard"
+#define WEATHER_RAIN				"rain"
+#define WEATHER_STORM				"storm"
+#define WEATHER_HAIL				"hail"
+#define WEATHER_WINDY				"windy"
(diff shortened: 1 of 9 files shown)


Anewbe:
Code looks fine to me, did you test this at all?

Arokha:
Can't say I tested it with Sif specifically to be honest since I don't have a Sif map. I tested it with our own planet, and it worked fine, which should be functionally similar. You don't use a planet map yet though right? So even if it has a bug or two, you should spot them while making your new map and it shouldn't affect anything.

PR #3344, opened 2017-04-21
Adds ion pistol to uplink, allows holstering

Anewbe:


MagmaRam:
Is it supposed to not fit on the belt and only fit in holsters?

Anewbe:
... Oversight, one minute.

PR #3345, opened 2017-04-21
Radios now work again in-hand or in pocket

Neerti:


Anewbe:
Don't feel like a changelog is needed for this.

PR #3346, opened 2017-04-21
Port multi-z cable improvements from Baystation

Leshana:
* Ports Baystation12/Baystation12#15907
  * Cables used to use thier own custom numbering system for up/down cable (11,12).  This changes them to use BYOND's standard UP/DOWN constants (valued 32,16) instead.
* Ports Baystation12/Baystation12#15940
  * Fixes and makes  powernet connection proc for cables multiz aware.
* Also replaced color macros and stream operator with style and to_chat


Leshana:
Test using northen_star works.

Leshana:
Test of multi-z constructing using the multi-z example map also works.

PR #3348, opened 2017-04-22
Fix typo in zpipe proc name.

Leshana:
Port of https://github.com/Baystation12/Baystation12/pull/16977
* The proc obj/machinery/atmospherics/pipe/up/process was supposed to be obj/machinery/atmospherics/pipe/zpipe/process.
* Also changed hard coded integers to use constants while I was editing the file anyway.


Anewbe:
Tested?

Leshana:
Wait one on this please, am discussing with folks on possible improvement.

Leshana:
We have come to the conclusion that this code is fine.  Yes, I did test on example.dmm

PR #3349, opened 2017-04-22
Tweaks bodybag relaymove(). Allows exit from morgue trays while in bodybags.

LorenLuke:


Anewbe:
Did you test this?

Yoshax:
Src is implicit.

SilverTalismen:
As far as I know, you can escape morgue trays by clicking the tray while inside of it

LorenLuke:
@SilverTalismen It doesn't work from inside a bodybag.

Noticed had forgot to commit the changes to the morgue tray itself. Current version has been tested and works, but will _**not**_ work if you're resting or unable to move while doing so.

PR #3353, opened 2017-04-23
Circuit Bugfixes & QoL

Neerti:
Because Feno found a million bugs one day.

Fixes:
Implant casings should now be able to have the cell placed back inside.
Simple pathfinding circuit will now push null to its linked circuit if it can't find a path, telling the locomotion circuit to stop.
Toggle button can now actually show up as an option when using a device.
Reagent pump and reagent injector circuits should now function again.
Implant casings can now have their UI opened by using it inhand, just like the other casings.

Quality of Life chances:
A lot of pins now will tell you what data type is expected to go there, e.g. \<TEXT\> for strings, \<NUM\> for numbers, \<REF\> for references, etc.
A large amount of circuits which do something now have two pins instead of one.  The second pin newly added sends a pulse out when the circuit is finished doing whatever it is meant to do (e.g. converting from string to number, adding numbers, etc).  This is meant to allow you to chain circuits together in one loop instead of needing to rely on delay circuits to make sure everything runs in the correct order.
The circuit UI (specifically the table part) now scales with the size of the window it is in instead of being static.
Concatinators will automatically convert numbers to text before concatinating instead of rejecting them entirely.
Second output pulse pin added to microphone circuit which only pulses if it hears someone talking in a different language than GalCom.
Sensors now have an input pin to ignore scanning storage containers such as bags.  This is set to 1 by default, as otherwise it is rather annoying if you want to put your device in your bag without resetting the ref.
Removed period from the end of EPv2 circuit's address on examine, as a very specific person thought the FULL STOP was part of the address.
Pulsing an activation pin with a debugger will now skip the powercheck for the circuit, allowing you to play with a circuit without needing to put it in a case first.  This helps testing both for the user in experimentation and myself when debugging.
You can examine circuits outside a case to see the UI again.
You can use a new 'Scan with Debugger' button in the circuit and assembly UIs to get a ref to that circuit without needing to remove it from the case first.  The debugger needs to be in your active hand to do this.

Misc:
Makes a lot of circuit code use the get/set_pin_data and activate_pin helper procs, which makes a lot of code shorter and hopefully cleaner.


LorenLuke:
@Neerti On bay one of the things for scanners as suggested for refs so as to not scan bags was to use a click-drag method, wherein you click drag the scanner over the item. Perhaps it might be better used here for container type objects.

PR #3356, opened 2017-04-23
Adjusts Changelings

Neerti:
Dead changelings that are capable of revival will no longer be able to see deadchat or ghost freely.
Changelings should not be able to transform or shriek while they are inside something (disposals, pipes, lockers, etc)
Fabricate Clothing now costs 1 point instead of 2.
Fabricate Clothing can now be made twice as fast.
Both Shrieks now share a ten second cooldown between uses.
Stasis timer normalized to 2 to 4 minutes.
Visible Camo now will block AI tracking while hidden.
Visible Camo will end if the ling gets stunned.
**Recursive Visible Camo no longer grants true invisibility.  It now allows the changeling to run while hidden.**
The chem counter HUD element will now display a blinking exclaimation mark if below 20 chemicals, to warn that dying will render them unable to revive.


code/game/gamemodes/changeling/generic_equip_procs.dm
```diff
@@ -32,7 +32,7 @@
 		return 1
 
 	if(M.head || M.wear_suit) //Make sure our slots aren't full
-		src << "<span class='warning'>We require nothing to be on our head, and we cannot wear any external suits.</span>"
```
Anewbe: The flesh boots are a child of magboots, iirc, so this isn't accurate.

code/game/gamemodes/changeling/generic_equip_procs.dm
```diff
@@ -32,7 +32,7 @@
...
+		src << "<span class='warning'>We require nothing to be on our head, and we cannot wear any external suits, or shoes.</span>"
```
Anewbe: The shoes on the spacesuits should just be children of magboots, if I remember correctly. Means that this last line isn't true for those powers.

Belsima:
Why can't they change their face in a locker?

Neerti:
It's to fix a lot more bugs like transforming while inside things like pipes, since currently you can go from monkey to human inside a pipe with no adverse effect.

PR #3359, opened 2017-04-23
Adds Drone Identification Cards

Neerti:
Drone FBPs now spawn with an EIO issued ID alongside their NT ID card.
Also makes the bartender's permit automatically name itself when created by the job_controller.


SilverTalismen:
Awesome! Any way to customize these ID's, maybe add a who built this drone so we could list their creator.

Neerti:
Unfortunately that involves adding new player preference vars that only show up for drones.

SilverTalismen:
I mean, couldn't it be done with the uhh Faction thing you can select?

PR #3364, opened 2017-04-24
Allow turbolift specific-floor delays

Arokha:
If one floor is 'taller' than the others, or otherwise takes longer to traverse, you can set delay_time on the area to have it wait a different time there. Or, wait even less time, though less than 1 second isn't possible since the controller only fires every second.

Also makes the default elevator panel nanoui window sliiightly bigger. It often had scrollbars for me after a few floors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,9 @@ language: c
 sudo: false
 
 env:
-  BYOND_MAJOR="510"
-  BYOND_MINOR="1346"
-  MACRO_COUNT=986
+  BYOND_MAJOR="511"
+  BYOND_MINOR="1381"
+  MACRO_COUNT=875
 
 cache:
   directories:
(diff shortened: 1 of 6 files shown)


Arokha:
This is failing Travis because Travis is using a version of Byond to compile it that is too old to understand that expression. I'll make a PR to update Travis.

.travis.yml
```diff
@@ -3,9 +3,9 @@ language: c
 sudo: false
 
 env:
-  BYOND_MAJOR="510"
-  BYOND_MINOR="1346"
-  MACRO_COUNT=986
+  BYOND_MAJOR="511"
+  BYOND_MINOR="1381"
+  MACRO_COUNT=875
```
PsiOmegaDelta: Sneaky version update.

Arokha: What do you mean? I left a comment about it, and it has it's own commit. Wasn't trying to hide it or anything.

Anewbe: I knew it was happening, and I thought I saw those mentioned things somewhere.

PR #3367, opened 2017-04-26
Merc and Heist back up to oldish numbers, heist round ends normally.

Anewbe:


Belsima:
You didn't change the spawn_req or spawn_target.

PR #3376, opened 2017-04-26
Advanced medical stacks now heal some damage

Anewbe:
It's per wound, so on average this will heal around 15 of the appropriate damage from a newly dead body, possibly allowing defibrillation.


Arokha:
Also means you can treat simple_animals with them, as that's the only thing that heals them.

PR #3378, opened 2017-04-26
Updates ammo icons.

Belsima:
From: image
To: image
From: image
To: image
From: image
To: image
From: image
To: image
 
Ported from /vg/station, /tg/station, and CEV-Eris.


MagmaRam:
My first instinct on this is that the shotgun shells are harder to click for little gain in visual clarity. I also just don't particularly like how the other two look, but I'm not the final arbiter on that.

Belsima:
After review... I actually agree. I mostly just liked the shading on the .357 and the EMP shell look.

PR #3383, opened 2017-04-27
Several corrections to Undershirts

Bromuzl:
There is one sprite using two different names on the Undershirt list. There is one sprite which does not have a name, and isn't listed in Character Setup. A third feminine shirt was never ported over.


code/datums/underwear/undershirts.dm
```diff
@@ -8,11 +8,6 @@
 	icon_state = "undershirt"
 	has_color = TRUE
 
-/datum/category_item/underwear/undershirt/shirt_long
-	name = "Long Shirt"
-	icon_state = "undershirt_long"
-	has_color = TRUE
-
```
Anewbe: Why is this block being deleted?

Bromuzl:
I'm going to reword some of this and get rid of the push with no content.

PR #3387, opened 2017-04-28
Stasis Bag and Sleeper Changes

Arokha:
This should add a little depth to these two rather shallow machines. Sure sleepers have some drugs and are the only way to do dialysis, but could probably benefit from being more useful, for being a bigass machine attached to the floor in the middle of medical. And stasis bags are a little silly considering they make the occupant completely immune to all harm from any source indefinitely.

**Sleepers:** These now actually deserve the name, and have an adjustable Stasis setting that has 0%, 50%, 80%, 90%, 99% settings. This can be read as "ignore 99% of life() ticks on this mob", or "ignore every other life() tick on this mob" (for the 50% setting), which is how the stasis bags function now (except they ignore every life tick). This allows you to buy time to treat someone, or have them process drugs more slowly so you can handle multiple patients. On the highest (99% slowdown) setting, it also slightly reduces time-of-death decay, and will allow the patient to be defibbed for longer. (Ody Sleeper: Has the same '30%' effects as a Stasis Bag, see below. No features removed.).

**Stasis Bag:** These act like the sleeper with a permanent fixed 66% setting, ignoring 2 out of 3 every life ticks. This is a direct nerf to their functionality, considering they can preserve someone indefinitely for now, without fail, against any harm. To try and make up for this, they now feature a built-in injector system, making them a sort of distant cousin to the sleepers. You can load a syringe with 15u of your favorite drugs and insert it into the cryobag in advance (then just re-fold it back up). Upon being used, the syringe will empty into the patient automatically (and the drugs will process at 30% speed, due to their slowed life() ticks). You can remove the syringe later with a screwdriver if you want to change it. If you're a completely crazy person, you could have a 'brute' stasis bag and a 'burn' stasis bag or something with different drugs.

So now the Sleeper, Ody Sleeper, and Stasis bag are basically tiers of the same concept:
- Stasis bag: Portable, but max 15u drug administration, fixed 33% metabolism rate, one time use (but fits in a pocket)
- Ody sleeper: Portable, advanced drug administration, fixed 30% metabolism rate
- Fixed sleeper: Fixed, advanced drug administration, dialysis, 0-100% metabolism rate, ToD decay delay

Edit: Forgot to mention, you can set the stasis level on a sleeper with no patient in it, it just applies to whoever gets in, so if you are dying and there's a sleeper in a first-aid station or something nearby, you can call for help, set stasis really high, and get in.

### DIFF
--- a/code/game/machinery/Sleeper.dm
+++ b/code/game/machinery/Sleeper.dm
@@ -68,10 +68,12 @@
 	anchored = 1
 	circuit = /obj/item/weapon/circuitboard/sleeper
 	var/mob/living/carbon/human/occupant = null
-	var/list/available_chemicals = list("inaprovaline" = "Inaprovaline", "stoxin" = "Soporific", "paracetamol" = "Paracetamol", "anti_toxin" = "Dylovene", "dexalin" = "Dexalin")
+	var/list/available_chemicals = list("inaprovaline" = "Inaprovaline", "paracetamol" = "Paracetamol", "anti_toxin" = "Dylovene", "dexalin" = "Dexalin")
 	var/obj/item/weapon/reagent_containers/glass/beaker = null
 	var/filtering = 0
 	var/obj/machinery/sleep_console/console
+	var/stasis_level = 0 //Every 'this' life ticks are applied to the mob (when life_ticks%stasis_level == 1)
+	var/stasis_choices = list("Complete (1%)" = 100, "Deep (10%)" = 10, "Moderate (20%)" = 5, "Light (50%)" = 2, "None (100%)" = 0)
 
 	use_power = 1
 	idle_power_usage = 15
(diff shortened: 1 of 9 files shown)


Anewbe:
So the player of a character inside the sleeper will... Not be asleep, or will be asleep?

Arokha:
> So the player of a character inside the sleeper will... Not be asleep, or will be asleep?

Depends on sleeper setting. 0% and 50% settings don't sleep them. You can consider the 50% one "I'm a new medical player and want to RP while trying to treat you" or something. The rest put them to sleep.

Anewbe:
Delicious. Gathering Staff input.

Yoshax:
I really like this! It has great room for even further improvements in the future too. The touch of being able to save your own life via climbing into a slowed sleeper is really cool, and will likely help people! Assuming they remember about it.

Atermonera:
This should probably have a changelog

Anewbe:
Code looks fine to me, staff seems to be mostly for it, my worry is this: Is there a way to get out of the machines that put you into stasis, or are you screwed if you end up in there for dumb reasons?

Arokha:
Kinda screwed. It has a do_after, so someone has to stand still or whatever if you mean doing it for robust purposes. If you mean just doing it to yourself accidentally, after setting stasis high, then no you can't get out since you're asleep. That's how old bay sleepers used to be, actually, used to put you to sleep and you were kinda doomed, I think? Like long ago, 2-3 years, min.

Belsima:
Could you give it a verb like cryo, where you can eject after a couple minutes?

Arokha:
Pro'lly. What about stasis bags?

Anewbe:
I was going to suggest using something like the cryotube verb for all of them, just in case.

Arokha:
That'd break stasis bags, so if the patient hit it accidentally (just like they might accidentally get into a sleeper set super high), the bag would be unusable again after that. You shore?

Anewbe:
As far as I've seen you can get out of a stasis bag on your own now, is that not the case?

Arokha:
You'd be asleep, like the sleeper. Any kinda stasis beyond ignoring 50% of life ticks puts you to sleep. I can alter that, if desirable, to a different value.

Anewbe:
No, I like the stasis bag just fine... Not sure how to word what I'm looking for.

Arokha:
@Anewbe 

Also made it so you can use grabs on sleepers because that was annoying that you couldn't.

PR #3390, opened 2017-04-28
Makes Technomancer use modifiers, modifier changes, fixes a few bugs

Neerti:
New Stuff:
Modifiers can now apply visual overlay effects to (currently only human) mobs they are affecting.
Corona, Repel Missiles, and Haste now use the modifier system instead of using snowflaky objects inside mobs.  For the end user this won't really have a difference, except for the overlay effects.
Evasion and haste variables added to modifiers datum.
Dispel will remove modifier-based technomancer spells as well.

Tweaks:
Repel Missiles lasts 10 minutes, from five.
Instability glow and Darkness spell have beefed up light intensity to punch through ultra-bright hall lights.
Corona decloaks the victim instantly now.

Fixes:
Illusion should function now.
Scepter can now smash cameras.

Picture:
Haste


Anewbe:
Tested, etc?

Neerti:
Yes.

PR #3399, opened 2017-04-30
Kinda reworks Hyposprays/Autoinjectors

Yoshax:
Instead of using snowflake news with removing reagents and a special new() for every type of autoinjector they now use the base new() and have a `filled_reagents` list which details and contains the reagents to be put in. Also they have a `filled` var which determines whether or not the item stays filled. Furthermore there's a `used` var and a `reusable` var that are used for autoinjector stuff.

There is also a new iconstate for auto injectors, which is used when they're empty but not used, it has red highlights instead of green, also the used sprite has full red highlights instead of green.

~~There's a slight problem with getting the icon to update when reagents are removed to use the unfilled reagent but I couldn't quite get it to work as it should. If anyone has any advice, feel free to share. But it all forks functionally fine as is.~~


Anewbe:
So the icon issue is fixed?

Yoshax:
Yep! I forgot that on_reagent_change existed. Was just what I needed. All is well now.

Yoshax:
All good?

Neerti:
Looks fine to me.

PR #3401, opened 2017-04-30
Allows more than one holodeck

Arokha:
The current code does not account for there ever being more than one holodeck. This alters it so that you can make various subtypes of the holodeck computer for other areas, each with different loadable programs.

You could, if you wanted, replace various things with holodecks. Could make the shooting range a holodeck with holocarp for example, or normal cutouts, or holo NPCs or something that cause halloss and shoot back. Or holo surgery centers with various layouts.

***SHRUG*** Do what you will.


SilverTalismen:
Amazing! :D

Anewbe:
Tested?

Arokha:
Yar, using it right now on new testing map, works fine.

PR #3407, opened 2017-05-03
Adjusts some access for Heads

SpadesNeil:
- Removes security door access from all heads of staff (except HoS and HoP obviously); because why does the Research Director, Chief Engie, and CMO need to open these doors other than to irritate the paranoid redshirts?
- Removes Medical access from Head of Security; because you're not a doctor now stop meddling in that department as you can't enforce THE LAW against bleeding to death.

A few people were complaining earlier, so discuss. :U

I have no strong opinion one way or the other.

### DIFF
--- a/code/game/jobs/job/security.dm
+++ b/code/game/jobs/job/security.dm
@@ -14,11 +14,11 @@
 	economic_modifier = 10
 	access = list(access_security, access_eva, access_sec_doors, access_brig, access_armory,
 			            access_forensics_lockers, access_morgue, access_maint_tunnels, access_all_personal_lockers,
-			            access_research, access_engine, access_mining, access_medical, access_construction, access_mailsorting,
+			            access_research, access_engine, access_mining, access_construction, access_mailsorting,
 			            access_heads, access_hos, access_RC_announce, access_keycard_auth, access_gateway, access_external_airlocks)
 	minimal_access = list(access_security, access_eva, access_sec_doors, access_brig, access_armory,
 			            access_forensics_lockers, access_morgue, access_maint_tunnels, access_all_personal_lockers,
-			            access_research, access_engine, access_mining, access_medical, access_construction, access_mailsorting,
+			            access_research, access_engine, access_mining, access_construction, access_mailsorting,
 			            access_heads, access_hos, access_RC_announce, access_keycard_auth, access_gateway, access_external_airlocks)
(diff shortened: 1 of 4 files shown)


Nalarac:
If the head of security loses access to medical, I feel like he should probably also lose access to the rest of the departments as well like engineering, research, and cargo.

code/game/jobs/job/security.dm
```diff
@@ -14,11 +14,11 @@
...
 			            access_forensics_lockers, access_morgue, access_maint_tunnels, access_all_personal_lockers,
-			            access_research, access_engine, access_mining, access_medical, access_construction, access_mailsorting,
+			            access_research, access_engine, access_mining, access_construction, access_mailsorting,
 			            access_heads, access_hos, access_RC_announce, access_keycard_auth, access_gateway, access_external_airlocks)
 	minimal_access = list(access_security, access_eva, access_sec_doors, access_brig, access_armory,
 			            access_forensics_lockers, access_morgue, access_maint_tunnels, access_all_personal_lockers,
-			            access_research, access_engine, access_mining, access_medical, access_construction, access_mailsorting,
+			            access_research, access_engine, access_mining, access_construction, access_mailsorting,
 			            access_heads, access_hos, access_RC_announce, access_keycard_auth, access_gateway, access_external_airlocks)
 	alt_titles = list("Commander")
```
Anewbe: Try updating your thing, I'm 90% sure there's a second alt title for this.

SpadesNeil:
>If the head of security loses access to medical, I feel like he should probably also lose access to the rest of the departments as well like engineering, research, and cargo.

I'd be okay with this. I noticed this while editing, but wanted to first focus on fixing the initial thing people had complained to me about; which is the HoS taking it upon theirself to get in the way of Medical, namely when other officers or the criminals are being treated. I've not seen this issue with other departments, but I'd be okay with removing their access too, since there's really no point in having it.

code/game/jobs/job/security.dm
SpadesNeil: I'll update my branch when I get home.

Anewbe:
Pinged the staff, we don't really see a reason for this.
If you're having an issue with the HoS being in Medbay, yell at the HoS. Same with any other head being where they don't need to be.

SpadesNeil:
Fair enough! Closing this.

PR #3410, opened 2017-05-04
Area alarms should not fight each other for control of fire doors.

Leshana:
* Instead of each individually controlling fire doors, setting or clearing alerts now simply calls firedoors_update() which considers both fire and atmos statuses before deciding to close or open the doors.
* Renamed air_doors_activated to firedoors_closed to be more intuitive.
* Preserved existing "party" alert status for posterity.
* Removed obsolete setting of mouse_opacity, it is always zero on areas.
* Fixes https://github.com/VOREStation/VOREStation/issues/1464


Leshana:
Testing procedure was:
1) Activate fire alert
2) Activate atmos alert
3) Reset atmos alert
4) Activate atmos alert
5) Reset atmos alert.

PR #3411, opened 2017-05-04
Tweaks hypothermia

Anewbe:
No more cold-burns, instead uses oxyloss.


LorenLuke:
So you..  suffocate... from cold...

What benefit does this add over the old system?

Anewbe:
Burns are a terrible way to do it, and it better represents the slowing of the heart and lungs and such.

PR #3412, opened 2017-05-04
Removes the datum pool, as it is not performant.

Leshana:
* Creating new objects is cheap, in fact comparable to the cost of getting it out of the pool, so it doesn't help there.
* Placing items in the pool is far more expensive than letting them garbage collect (resetting of vars and such)

See https://github.com/tgstation/tgstation/pull/23150 and all the other times @Fox-McCloud has talked about this.

Note: At some point we should implement the qdel hints to optimize our garbage collection.


PsiOmegaDelta:
https://github.com/Baystation12/Baystation12/pull/16820 does qdel hints if you want to borrow it.

Fox-McCloud:
Yay, no more pooling.

Will second @PsiOmegaDelta and what he says; qdel hints are great; there's less proc-call overhead with them than `finalize_qdel`.

Anewbe:
Did you test this @Leshana?

Leshana:
I have smoke tested this.   By that I mean that I have compiled it and started up the server then walked around and did a few things.  I tried to test a few things that I know used PoolOrNew (smashing some lights, built and cut a lattice etc).  That worked fine and there were no runtimes.  I did not exhaustively test every single type modified in this PR. 
I did look at every line changed, with particular attention to the cases where 2+ parameters were passed in a list.  The changes are very simple and straightforward, the only way I can really see a bug being introduced is via a typo.

PR #3414, opened 2017-05-05
Starts support for species robolimbs

Anewbe:
Ported from downstream, this just makes it so if I want to make a, for example, Unathi specific brand, I can.


Anewbe:
@Neerti Could you look at this, please?

PR #3421, opened 2017-05-06
More Technomancer

Neerti:
Does more work on the GOLEM.  Still not finished but now it has sprites and the ability to cast spells.
Tesla armor now retaliates against ranged projectiles if the firer is within 3 tiles.
Tesla armor recharges in 15 seconds, down from 20 seconds.
Tesla armor sprite changes to reflect if it is charged or not.
Instability should go away slower.
Fire and frost aura temperatures are now tied to the human species' temperature variables.  This should make them more consistent if we change the thresholds later.  Unfortunately the fire aura is still really weak but I can't fix it without overhauling mob self healing.
Dispel now won't get rid of inserted spells, as they won't exist after this PR.
Gambit now has the potential to give 'rare' spells you can't otherwise get.  The chance to roll for it is based on the core's spell power.  Currently this is only Heal All.
Gambit now has a scepter effect, to give a "random" spell.
Mend Wounds and Mend Burns now combined into Mend Life, and made a modifier spell.
Mend Metal and Mend Wires now combined into Mend Synthetic, and made a modifier spell.
Purify is now a modifier spell.
Mend Organs renamed Mend Internals, and is now instant.
Instability Tap now makes a sound when using it, and the catalog tells you how much you will get by default.
Adds Mend All, a spell only obtainable by Gambit.  It heals all non-internal damage over time.  It functions like Mend Life/Synthetic, but lasts a whole minute.
Updates Repel Missiles percentage in catalog.
Oxygenate made 25 points to buy, from 50.
Chain Lightning made a little stronger.
Added Lesser Chain Lightning, which is more spammable but weaker as a result.
Added Ionic Bolt, which is good for disabling synthetics.
Added Destabilize, which makes an area glow purple for 20 seconds and afflicts everyone inside it with instability.
Modifiers can now have an origin reference.  It's a weakref in order to not interfere with qdel()
Removes all 'insert' class of spells, as the code was terrible, and modifiers are better in every way.


SilverTalismen:
Finally, my life as a healer tree will be easier

code/game/gamemodes/technomancer/core_obj.dm
```diff
@@ -330,6 +330,18 @@
...
+	name = "integrated core"
+	desc = "A bewilderingly complex 'black box' that allows the wearer to accomplish amazing feats.  This type is not meant \
+	to be worn on the back like other cores.  Instead it is meant to be installed inside a synthetic shell.  As a result, it's \
+	a lot more robust."
+	energy = 25000
+	max_energy = 25000
+	regen_rate = 100 //250 seconds to full
+	instability_modifier = 0.75
+
+
```
Neerti: There's no way to buy or otherwise obtain it without adminbus.  When the golem is finished, it'll exist inside the golem mob and there's no way to get it out.

PR #3423, opened 2017-05-06
Lets biological females have facial hair.

Belsima:
Trying to make an FTM character and it's a bit annoying being unable to give him a hint of beard.


Belsima:
what input does this require

Anewbe:
We're debating on whether we want it or not.

Belsima:
Alright, thanks. Do note biological females can grow facial hair too, though.

PR #3427, opened 2017-05-06
Fixes a straightjacket bug

Anewbe:
Now you can't remove it by clicking the space around it.


SilverTalismen:
Great! Bet this was driving you crazy, and without the jacket there was no way to help~ <3

LorenLuke:
@Anewbe Add a way to resist out of it then, if there's no way to take it off yourself.

PR #3431, opened 2017-05-07
Cleans up uniform loadout menu.

Cerebulon:
Suits are now mostly under one-piece (subtypes of 'lawyer') and modular (subtypes of 'suit_jacket') selection menus. Croptops also under selection menu.
Tweaked old woman attire to no longer be a subtype of the lawyer suit since it's not really a suit.


Anewbe:
Did you test this?

Cerebulon:
Oh, I thought I already replied. Yes. One minor commit which fixes the almost completely random capitalisation on the lawyer suits while I'm here and it should be all good.

PR #3405, opened 2017-05-02
Makes the 556 and 762 calibers more appropriate to their naming

Anewbe:
Long time coming, shouldn't have any effect on their actual effectiveness, but now the guns that hit hard say they have a bigger bullet than the ones that hit less hard.

EDIT BECAUSE I'M A TERRIBLE DOCUMENTOR: The bullets that used to do 25 damage were called 762 rounds. These will still do 25 damage, but will be called 556 rounds. The bullets that used to do 35 damage were called 556 rounds. These will still do 35 damage, but will now be called 762 rounds. It looks a lot more complex than it actually is, which is unfortunate.

PR #3432, opened 2017-05-09
Bullet damage fixed, now with realism AND balance.

SpadesNeil:
This basically accomplishes what https://github.com/PolarisSS13/Polaris/pull/3405 was doing, except differently, and with less complicated code changes.

- 5.56mm ammo and 7.62mm ammo had their damage values swapped. Now 7.62 does more damage than the 5.56 rounds. Before you worry about balance being all screwed up, read the rest. I was careful not to break this.
- 556 and 762 ammo costs in traitor uplink had their costs reversed to reflect changes to damage.
- STS (old assault rifle) was replaced with the GP3000 (assault carbine) in traitor uplink and `/datum/supply_drop_loot/ballistics` to avoid screwing up game balance balance. For anyone who doesn't know, the assault carbine uses 5.56 ammo, so it does the same damage as the old space-AK style STS rifle, except now with a spicy new look.
- STS was also renamed from "assault rifle" to "battle rifle" to avoid confusion, in case any admins feel like abusing it anyway.
- Z8 Bulldog has been changed to use 7.62 rounds so its damage output is unchanged from what it previously was. Likewise, its magazine was changed to reflect this. It still only holds 10 rounds max.

Now here's the part where my version is different from #3405 and why this is intentional.

- Bolt-action rifles DID NOT have their ammo switched. This means an increase of total damage with a full magazine from 125 to 175. (For comparison, a shotgun loaded with slugs has a total output of 400.) I think this is fair, as the weapon is no longer completely useless. Previously, a lousy .45 pistol had the same damage output per bullet as the bolt rifles, except with MORE bullets, making a .45 all around a better gun except for no armor pen, and the armor pen DID NOT make up for it! That was just stupid, hence this tweak.
- Lever action rifle had its ammo replaced with .357 ammo, buffing its potential output from 125 to a whopping 300. However, the lever action rifle LOSES its speedloader ability, because that does't make any sense for this kind of gun, and would make it overpowered. It's now basically a slightly nerfed shotgun with Cool™ points and no ammo flexibility. This change I'm a little iffy on, so discuss it. I think it'd be fine with .45 or 7.62 rounds too.

... I also changed some fluff text I didn't like because reasons.


code/modules/projectiles/guns/projectile/automatic.dm
```diff
@@ -344,8 +344,8 @@
 //	update_held_icon()
 
 /obj/item/weapon/gun/projectile/automatic/carbine
-	name = "assault carbine"
-	desc = "The bullpup configured GP3000 is a lightweight, compact, military-grade assault rifle produced by Gurov Projectile Weapons LLC. It is sold almost exclusively to standing armies. The serial number on this one has been scratched off. Uses 5.56mm rounds."
+	name = "assault rifle"
+	desc = "The bullpup configured GP3000 is a lightweight, compact weapon produced by Gurov Projectile Weapons LLC. Uses 5.56mm rounds."
```
SpadesNeil: - Removed edgelord tacticoolness since this isn't an admin abuse rifle anymore.

code/modules/random_map/drop/drop_types.dm
```diff
@@ -61,9 +61,9 @@ var/global/list/datum/supply_drop_loot/supply_drop
 		/obj/item/weapon/shield/riot/tele,
 		/obj/item/weapon/storage/box/emps,
 		/obj/item/weapon/storage/box/flashbangs,
-		/obj/item/weapon/gun/projectile/automatic/sts35,
-		/obj/item/ammo_magazine/c762/ap,
-		/obj/item/ammo_magazine/c762/ap,
+		/obj/item/weapon/gun/projectile/automatic/carbine,
```
SpadesNeil: Oh yea, I changed this too, because I didn't want to leave any loose ends where the OP gun could be obtained.

maps/northern_star/polaris-2.dmm
```diff
@@ -896,7 +896,7 @@
...
 "arn" = (/obj/structure/table/rack,/obj/structure/window/reinforced{dir = 8},/obj/structure/window/reinforced,/obj/item/weapon/gun/projectile/automatic/z8,/obj/item/weapon/gun/projectile/automatic/z8,/turf/unsimulated/floor{icon_state = "dark"},/area/centcom/specops)
-"aro" = (/obj/structure/table/rack,/obj/item/ammo_magazine/a556,/obj/item/ammo_magazine/a556,/obj/item/ammo_magazine/a556,/obj/item/ammo_magazine/a556,/obj/item/ammo_magazine/a556,/obj/item/ammo_magazine/a556,/obj/structure/window/reinforced,/obj/structure/window/reinforced{dir = 4},/obj/item/ammo_magazine/a556,/obj/item/ammo_magazine/a556,/turf/unsimulated/floor{icon_state = "dark"},/area/centcom/specops)
+"aro" = (/obj/structure/table/rack,/obj/item/ammo_magazine/s762,/obj/item/ammo_magazine/s762,/obj/item/ammo_magazine/s762,/obj/item/ammo_magazine/s762,/obj/item/ammo_magazine/s762,/obj/item/ammo_magazine/s762,/obj/structure/window/reinforced,/obj/structure/window/reinforced{dir = 4},/obj/item/ammo_magazine/s762,/obj/item/ammo_magazine/s762,/turf/unsimulated/floor{icon_state = "dark"},/area/centcom/specops)
```
SpadesNeil: This is just so the Z8 for ERT can have ammo.

LorenLuke:
Request inclusion of following since lever action rifles can't use speed loaders.

https://github.com/Baystation12/Baystation12/pull/16229

SpadesNeil:
That's outside the scope of this PR, but I mean, it should probably be a thing regardless of this being accepted or not.

Anewbe:
The only reason the other one hasn't been merged is because I haven't dragged one of my Devs to it yet. This doesn't change the sprites, which is an issue, and leaves the STS in the same Aspawn pit as the carbine does now.

The differences will need to be discussed by the staff, but for now it's outside the scope of what I want done with the code.

SpadesNeil:
>The only reason the other one hasn't been merged is because I haven't dragged one of my Devs to it yet. This doesn't change the sprites, which is an issue,

What's the matter with the sprites? I don't understand.

>and leaves the STS in the same Aspawn pit as the carbine does now.

That's going to happen either way at this rate.

>The differences will need to be discussed by the staff, but for now it's outside the scope of what I want done with the code.

I know, that's why I made my own PR instead. It's nothing against you or what you coded, I just didn't really like the solution so I made my own.

Anewbe:
If you don't change the sprites you end up with guns that look like they're inputting the wrong mags.

You're trading one item from the pit for another. That's not exactly and improvement.

Also, I thought you said you'd be waiting for the other thing to get merged before you starting touching muh guns again?

SpadesNeil:
>If you don't change the sprites you end up with guns that look like they're inputting the wrong mags.

um, what?

The Carbine still uses STANAG mags, and the STS still uses banana mags. The rounds inside the STANAG mags are still 5.56, and the banana mags still use 7.62. The sprites are perfectly fine.

SpadesNeil:
I thought I closed this already. I'll come back to rework this later.

PR #3436, opened 2017-05-10
Subspace jammer item

Arokha:
Jams the three devices that I'm aware of that use subspace communications: Headsets, PDAs, and Communicators.

All three will be unable to communicate if in range of a jammer. The default range is 7 turfs, the battery that comes in the jammer is a weapon cell (just to make it a little harder to find replacements) and lasts 60 seconds. The range and power use are variables.

Speaking through a headset will produce 'soft' or 'loud' static depending on how far you are, allowing you to sort of dead-reckon your way to a placed jammer. You can just carry them, otherwise.

I've got no experience with adding tator items so I'll leave that sort of thing up to you guys.

Credit to Eris SS13 for the sprite! I'd credit the artist if someone tells me who.


Anewbe:
TRAVIS IS DISPLEASED

Anewbe:
Can't currently be accessed in game, right?

Arokha:
@Anewbe Correct

(Sorry, was on vacation)

PR #3438, opened 2017-05-10
Adds Recently Cloned Modifier to Cloners

Neerti:
Makes new clones receive a substantial malus on combat ability, which will last between 25 to 40 minutes.
The malus proposed is:
-40% max health.
10% more incoming damage.
30% less melee damage.
Stuns last 25% longer.
2 more slowdown.
15% easier to hit.

This time can be reduced by upgrading the cloner.  A fully upgraded cloner can reduce the time to 5 to 8 minutes.
Also makes new clones receive a 'cloned' modifier, which doesn't actually do anything besides act as a sort of counter for how many times they've been cloned that only admins can see at the moment.  This will be more relevant in The Future(tm), when modifiers can be shown on the HUD.
Makes modifiers with the 'genetic' flag get copied to cloned mobs.  So far only the 'cloned' modifier uses it.
Adds admin utility/testing/badmin verb to give any modifier to any living mob.

Now you may be asking why this is being added, and it should be obvious.  Security.

It used to be a meme for them to say they get cloned and then immediately charge back into the fray, but lately I've seen things uncomfortably start to trend towards this becoming a reality.  Sometimes the lone sec officer who was cloned and had to go respond to something was actually justified in doing so, so I'm uncomfortable just giving a blanket ban to those who I see do it, so I feel mechanics can help provide an incentive for cloned people to take it easy for a little while, or at the very least play more cautiously if they really do need to fight someone.

For non-sec people who get cloned, most of those maluses won't really matter, except for the slowdown one.

Another benefit is that it also might provide an incentive for Medical to prefer to use the Defib to bring people back verses the cloner when possible, which will hopefully also kill the meme of 'just clone them its easier', which I don't actually see happen but it'll hopefully stop it from ever becoming a real thing.

I wanted to make the permanent 'cloned' modifier also provide a very minor malus (-5% max health or something like that), so dying would objectively be bad from a mechanics standpoint, but this provided rather unpopular, so for now it just does nothing, however I still feel it would be a good addition, especially if we add a 'cloned' Trait later on when/if Traits become a thing.

### DIFF
--- a/code/__defines/mobs.dm
+++ b/code/__defines/mobs.dm
@@ -152,7 +152,7 @@
 #define MODIFIER_STACK_EXTEND	2	// Disallows a second instance, but will extend the first instance if possible.
 #define MODIFIER_STACK_ALLOWED	3	// Multiple instances are allowed.
 
-#define MODIFIER_GENETIC	0	// Modifiers with this flag will be copied to mobs who get cloned.
+#define MODIFIER_GENETIC	1	// Modifiers with this flag will be copied to mobs who get cloned.
 
 // Bodyparts and organs.
 #define O_MOUTH    "mouth"
--- a/code/game/dna/dna_modifier.dm
+++ b/code/game/dna/dna_modifier.dm
@@ -18,6 +18,7 @@
 	var/mind=null
 	var/languages=null
 	var/list/flavor=null
+	var/list/genetic_modifiers = list() // Modifiers with the MODIFIER_GENETIC flag are saved.  Note that only the type is saved, not an instance.
 
 /datum/dna2/record/proc/GetData()
 	var/list/ser=list("data" = null, "owner" = null, "label" = null, "type" = null, "ue" = 0)
--- a/code/game/machinery/cloning.dm
+++ b/code/game/machinery/cloning.dm
@@ -533,7 +533,7 @@
 	incoming_damage_percent = 1.1			// 10% more incoming damage.
 	outgoing_melee_damage_percent = 0.7		// 30% less melee damage.
 	disable_duration_percent = 1.25			// Stuns last 25% longer.
-	slowdown = 2							// Slower.
+	slowdown = 1							// Slower.
 	evasion = -1							// 15% easier to hit.
 
 // Does nothing.
--- a/code/game/machinery/cloning.dm
+++ b/code/game/machinery/cloning.dm
@@ -139,6 +139,25 @@
 	H.set_cloned_appearance()
 	update_icon()
 
+	// A modifier is added which makes the new clone be unrobust.
+	var/modifier_lower_bound = 25 MINUTES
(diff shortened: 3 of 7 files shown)


SilverTalismen:
Sure I'm all for this.

SpadesNeil:
25 minutes seems fair. 40 seems a bit much.

I disagree with a permanent debuff, but that's just me.

Anewbe:
I feel like slowdown 2 is a bit much, and that this seems a bit harsh in general. Will need discussion.

FenodyreeAv:
You can respawn in five minutes. If this goes in you might as well add an 'elevator, shuttle and gateway
 disorientation' modifier as well, that's applied to people who respawn.
This strikes me as an overreaction.

Yoshax:
If they die and are dead, the job slot is gone. That's the point of them dying. Can't respawn if there's no job slot.

Screemonster:
I'm with Ace on this. Permanent debuffs no, but a short period of feeling like shit while you recuperate makes sense.
Surgery, likewise. People hopping off a surgery table and back into the fray? Nah, you get your ass to the recovery ward.

SpadesNeil:
So what's the word on this?

Neerti:
Nobody knows.

Anewbe:
I'll bring it up to the staff again, changes may need to be made.

SpadesNeil:
I'm just gonna restate my position for the record.

Permanent debuff == bad.
Temp debuff == good.

MagmaRam:
Any one of these debuffs alone would be fine as a combat malus, having them all together is just making someone totally useless for up to 40 minutes. Maybe make it so that the more times someone is cloned, the more effects they get?

SpadesNeil:
This has been here since May...

Belsima:
So... how's she going?

Wickedtemp:
Looking at Polaris' population, I don't really think this should be implemented. Neerti's example of "a lone sec officer needing to jump back into the fray" can essentially be applied to almost every round on Polaris.

Instead of a 'recently cloned modifier', I would suggest a 'clone counter' that applies debuffs the more someone is cloned within a round or timeframe. This would allow the one or two sec members some leeway should they be killed once, but if they're cloned twice they begin experiencing debuffs, with every instance of cloning becoming more severe.

zekesulastin:
@Wickedtemp When this was written 5 months ago, we had actual sec regulars to nerf like this.  That's why the entire essay to go with this pull request is all about how it affects sec with one throwaway line about "It won't hurt not-sec except for the slowdown lol".

SpadesNeil:
All I care about is that the effects are **temporary**. Is that the case? Otherwise I'm pretty much gonna immediately disable permanent debuffs when this hits vorecode, 'cuz that's just not fun.

Anewbe:
The effects are **temporary.**

SpadesNeil:
okay good 👍

PR #3440, opened 2017-05-11
Small ZAS Optimization

Leshana:
Merges https://github.com/Baystation12/Baystation12/pull/17138

Minor optimization; eliminates a few proc calls when managing zone connections by turning check() into a macro.

### DIFF
--- a/code/ZAS/ConnectionManager.dm
+++ b/code/ZAS/ConnectionManager.dm
@@ -24,11 +24,15 @@ Class Procs:
 	erase_all()
 		Called when the turf is changed with ChangeTurf(). Erases all existing connections.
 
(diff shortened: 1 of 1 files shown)


skull132:
Old PR but. @Leshana I noticed this is roughly the second ZAS optimization PR in the past few weeks. You might be interested in this: https://github.com/Aurorastation/Aurora.3/pull/2308

It swaps ZAS over to use cached variables whenever possible, and negates a whole metric ass tonne of proc calls while rebuilding zones. Which apparently speeds it up a lot. (And by a lot, I mean: a megazone we had went from 13 second rebuild to ~3 second rebuild.)

PR #3441, opened 2017-05-12
Diona regen fix

Eimras:
Small bugfix to prevent a runtime error from a null pointer and allow Dionaea to actually regenerate.


Anewbe:
You put the disfigure fix in the check that deletes organ stumps and replace them. It needs to be outside of that, or you need to re-order the ifs.

Anewbe:
Did you test if the new code fixes the null pointer you made the original code to fix?

Eimras:
Tested and working properly.

PR #3442, opened 2017-05-12
Diona feel no pain

Eimras:
Adds a check for Diona on pain(I think), and a check to electrocution to prevent Diona from being electrocuted. May they work on wiring and airlock repairs again!

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -966,10 +966,8 @@ default behaviour is:
 /mob/living/proc/update_water() // Involves overlays for humans.  Maybe we'll get submerged sprites for borgs in the future?
 	return
 
-/mob/living/proc/can_feel_pain(var/check_organ, var/mob/living/carbon/human/H)
+/mob/living/proc/can_feel_pain(var/check_organ)
 	if(isSynthetic())
-		return FALSE
-	if(H.species.flags & NO_PAIN)
 		return FALSE
 	return TRUE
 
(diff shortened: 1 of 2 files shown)


Anewbe:
... Also makes them immune to certain technomancer spells, which may or may not be ☹️ . Will need to discuss with staff.

Neerti:
Might be better to replace the hard Diona checks with ```species.flags & NO_PAIN``` instead of ```species.get_bodytype() == "Diona"```.

Eimras:
Scoped down the shocking immunity to only the hands of species with the IS_PLANT flag and changed the pain checking to check NO_PAIN flag.

Anewbe:
You test the zapping part?

code/modules/mob/living/living.dm
```diff
@@ -966,9 +966,11 @@ default behaviour is:
 /mob/living/proc/update_water() // Involves overlays for humans.  Maybe we'll get submerged sprites for borgs in the future?
 	return
 
-/mob/living/proc/can_feel_pain(var/check_organ)
+/mob/living/proc/can_feel_pain(var/check_organ, var/mob/living/carbon/human/H)
 	if(isSynthetic())
 		return FALSE
+	if(H.species.flags & NO_PAIN)
```
Neerti: Think it'd be better to override the human's ```can_feel_pain()``` and check for species instead.

Eimras:
The zapping part was tested and works as intended. The can_feel_pain proc is also in human, and that one does already check for NO_PAIN so my change to living is a bit redundant.

PR #3444, opened 2017-05-13
Adds code to allow shoes to influence speed on snow or in water

Anewbe:
There's probably a better way to do this, but I'm not sure what it is.

### DIFF
--- a/code/modules/clothing/clothing.dm
+++ b/code/modules/clothing/clothing.dm
@@ -384,6 +384,9 @@
 
 	var/shoes_under_pants = 0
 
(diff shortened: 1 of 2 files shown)


SilverTalismen:
Does that mean certain shoes are better in the snow then others?

Anewbe:
Not yet, but eventually.

SilverTalismen:
Aah, understood!

SpadesNeil:
should totally add snow shoes

Anewbe:
Actual Note to Self: Check for turf before checking for shoes.

Belsima:
are you checking for turfs yet

PR #3458, opened 2017-05-21
EZ Map Datum Setup

Leshana:
### Adds an easier and more maintainable way of setting up the map datum.
* Settings for specific z-levels are defined in datum types, allowing the lists of Z levels to be automatically configured.  This solves the problem of being unable to use the z-level #defines in associative lists.
* Converted the example and northern star maps to use the new system.
* 100% Backwards compatible and optional.  The old way will still work, this just populates the old lists.
* This paves the way for more map configuration down the road as more multi-z capabilities are added.


Anewbe:
How will this affect a map that's currently being made? Will the mapper need to change any files to compile said map?

Leshana:
No, it is 100% backwards compatible.  If the mapper decides not to use it, they don't have to.   But personally I think it has its advantages, so I'd recommend using it instead.   For one thing, you get to use your defined constants everywhere and don't have to hard code z level numbers as strings.

PR #3465, opened 2017-05-22
Ports radiation rework and stuff from Bay

Yoshax:
Baystation12/Baystation12#15090
Baystation12/Baystation12#15268
Baystation12/Baystation12#15293
Baystation12/Baystation12#15407
Baystation12/Baystation12#15897
Baystation12/Baystation12#15941

With a bonus of Baystation12/Baystation12#15013

Things appeared to work in testing as expected. Radioactive things are radioactive. The SM is more radioactive radiation is blocked by the walls, doors and blast doors. See linked PRs for full details and stuff.


Anewbe:
The original PRs don't really explain all that well what they do... So what does this actually do?

Yoshax:
Essentially radiation is changed from looping through mobs in view and applying an equal radiation effect to them to creating a radiation hotspot with emits radiation from there and has falloff. Not only that, but it is now affected by walls, doors, blastdoors and suchlike.

For example. If you set up a SM with Phoron and give it a solid amount of shots there's enough radiation to hit the monitoring room and into the low of the hallway above. That's because radiation ignores everything in its way. With this, as long as the blast doors are closed the radiation never leaves the room. If you were to open them, or a door though the radiation would be open to go where was opened.

Also adds geiger counters which let you monitor the radiation present.

As such this just converts most of the appropriate things to use these radiation hotspot things. Such as the SM, Xenoarch artifacts, certain meteors, singulo, and such.

Anewbe:
Neat.

Yoshax:
It is rather! Some things will really need live testing for balance and suchlike, particularly the SM radiation. I took it straight from Bay but enjoyed the ideas so kept all the numbers. Going into the engine room without a radiation suit on a high-power SM is now really quite dangerous (like it should be), but of course, the radiation no longer leaks outside. OSHA!

I think I will add geiger counters to the autolathe though, if people are agreeable to that.

Leshana:
Interesting.  I like how it makes geiger counters possible.  That's one unfortunate bit about the current system.   I also like how it makes walls help block radiation.  Although its pretty expensive if nobody is around to get irradiated.
One quirk is that it appears you are immune to radiation sources if you don't move!

Yoshax:
I'm not sure about the cost of the system, I don't really know much about that sort of thing. A lot of optimisations were made to it though, which are included within this PR, and I don't think there's any huge concerns with it that I know of.

Yoshax:
So apparently this is super expensive and there's better ways to do it. @Leshana is working on something, so don't merge this and I'll probably close it when Leshana finishes and PRs the thing.

PsiOmegaDelta:
Yeah, you do not want the current Bay implementation.

Yoshax:
Leshana said she has something in mind/in progress/done that's ridiculously faster.

Yoshax:
Okay! We have Leshana's rewrites and optimisations now. This should be good for review.

Leshana:
I agree with @PsiOmegaDelta. On my test server the old code was running at 164ms per run with just the supermatter setup.  On average the server would be spending 8% of every second running radiation controller.   Don't ask what happened during a radiation event.
with supermatter
With supermatter running the optimized code uses around 2-3ms, so... 5000% faster?  It will slow down if more people were standing in the engine room, but during most rounds that won't be the case.
perf

BlueNexus:
@Leshana Would you mind porting your optimised version to bay?

code/controllers/Processes/radiation.dm
```diff
@@ -7,37 +7,43 @@
...
+	if(length(linked.sources) > 0)
+		for(var/thing in registered_listeners)
+			if(deleted(thing))
+				continue
+			var/atom/A = thing
+			var/turf/T = get_turf(thing)
+			var/rads = linked.get_rads_at_turf(T)
+			if(rads)
+				A.rad_act(rads)
 		SCHECK
```
PsiOmegaDelta: If each of these steps were to be moved into their own separate procs it'd be easier to tell, using the profiler, where the bottlenecks are.

Leshana: Actually I can already tell you where the bottleneck is, its get_rads_at_turf()  Everything else in those loops is a fraction of that proc's cost.  But if you think I should put them into procs I will.  Three more proc-call overheads every 2 seconds is no big deal.

Anewbe:
Has this been tested?

Leshana:
I took @PsiOmegaDelta advice and split the three main process steps into their own separate procs.   I also noticed while testing that the hard-`del()` performed by `qdel()` on adding sources was slow.  Once we implement qdel-hints that will be easy to solve, but I put in a clearly-commented temporary fix until we port the tg garbage collector.
I also made the radiation cutoff configurable.  Changing that to 1.0 instead of 0.35 would make very little gameplay difference but would save some CPU if we want to tighten it even more.

PJB3005:
I was about to comment "RIP CPU" but I guess you re-worked it to only apply radiation effects when a change happens or something?

Nice either way. Now it's actually worth porting.

PR #3469, opened 2017-05-24
Prevent Gravity Failure event from disabling gravity on planet z levels.

Leshana:


Anewbe:
Tested?

Leshana:
Yes, both the not-thunk and the not-bob

Leshana:
I tested this on the tether map (I don't actually know where sif is).  But it should work the same.

PR #3475, opened 2017-05-26
Path naming of ammo to be consistent, and also fixes some gun things

SpadesNeil:
In summary... Because of how Byond works, we cannot start a datum or item path with a number, so a letter must come first. The reason for this change is that the confusing naming can be detrimental to coders or in-game admins who want these items to spawn. They get mixed up. So now we're doing it consistently.

- a = Ammo, as in individual rounds of ammunition.
- b = Box, intended to have ammo taken out by hand and put into magazines or guns. (Really, these shouldn't exist in my opinion.)
- c = Clips, intended to reload magazines or built-in magazines.
- m = Magazine, intended to hold rounds of Ammo.
- s = Speedloaders, intended to reload built-in magazines.

Some magazines only work for specific guns, so they will be named accordingly. Examples include the m45uzi, or the m45tommy. The reason they are not subtypes of m45 is because they should not be compatible with existing .45 guns. Same thing for m556 verses m556saw magazines.

While I was doing this I also discovered features that were broken by #3405 such as the way sprites change for ammo used by different guns. I have restored this functionality for the STS assault rifle, and GP3000 carbine, so that they again look different depending on what mags are loaded. Ergo, the following changes were also made.

- Small 556 magazines were restored.
- Extended 762 magazines were restored.
- STS and Carbine were restored with the ability to accept large and small mags without sprites looking dumb.
- For balance reasons, Z8 Bulldog rifle cannot accept a 20 round 762 magazine, just like it couldn't accept a 20 round 556 magazine before the #3405.

Small/Extended magazines that I added are not yet available in-game, but they may be in a future PR. Frankly I see no reason not to add a small 5.56 mag as a cheap option for traitors, but I'm already pushing this PR's scope.


SpadesNeil:
That should be everything.

code/modules/projectiles/ammunition/boxes.dm
```diff
@@ -297,34 +275,50 @@
...
 	name = "magazine (5.56mm armor-piercing)"
 	ammo_type = /obj/item/ammo_casing/a556/ap
 
-/obj/item/ammo_magazine/c556/ap/ext
+/obj/item/ammo_magazine/m556/ap/ext
 	name = "extended magazine (5.56mm armor-piercing)"
 	max_ammo = 30
 
-/obj/item/ammo_magazine/clip/a556
+/obj/item/ammo_magazine/m556/small
```
Anewbe: These should probably be named differently so you can tell the difference from the name.

code/modules/projectiles/ammunition/boxes.dm
```diff
@@ -388,41 +382,41 @@
...
+/obj/item/ammo_magazine/m762/ap
 	name = "magazine (7.62mm armor-piercing)"
 	ammo_type = /obj/item/ammo_casing/a762/ap
 
-/obj/item/ammo_magazine/c762/empty
+/obj/item/ammo_magazine/m762/empty
 	initial_ammo = 0
 
-/obj/item/ammo_magazine/s762 // 's' for small!
+/obj/item/ammo_magazine/m762m // Intentionally not a subtype of m762 because it's supposed to be incompatible with the Z8 Bulldog rifle.
```
Anewbe: If we're actually readding this I don't see why it wouldn't work in the bulldog.

SpadesNeil: This was originally done for balance reasons. I don't disagree with you but I didn't want players jamming 20 round mags into the bulldog.

code/modules/projectiles/ammunition/boxes.dm
SpadesNeil: What I really want to do is rework the magazines for 5.56 so that it's consistent with how the formatting of the 762 is laid out.

Smallest mag subtypes first, and work up from there, to have "extended" mags. However, like I said in the other comment, I had balance concerns regarding specifically the bulldog.

Yoshax:
You should include an explanation of the naming and typing scheme in a comment somewhere,  Maybe whatever is the base file for ammunition magazines or such so people in the future may bother to read it any name things in accordance.

SpadesNeil:
Good idea. I'll do that in a bit.

code/modules/projectiles/ammunition/boxes.dm
Anewbe: I don't know, maybe something like "reduced magazine (5.56mm)" or something to that effect. Just so people trying to get it don't expect a full sized magazine.

SpadesNeil:
That should do it.

code/modules/projectiles/ammunition/boxes.dm
SpadesNeil: I'll do that, then.

code/modules/projectiles/guns/projectile/revolver.dm
```diff
@@ -107,13 +107,13 @@
 	name = "\"The Judge\""
 	desc = "A revolving hand-shotgun by Cybersun Industries that packs the power of a 12 guage in the palm of your hand (if you don't break your wrist). Uses 12 shotgun rounds."
```
SpadesNeil: I completely overlooked this. I should fix this later.

"Uses 12 shotgun rounds." Uh, no. Who's the idiot who wrote this? I should give him a piece of my mind.

... OH, wait, _I_ wrote this. Well this is awkward.

Anewbe: Spelled gauge wrong, too.

SpadesNeil: I'm just going to give my usual "shut up I'm sleepy" rebuttal. I was probably sleepy at the time.

PR #3491, opened 2017-06-01
Makes Mechs Fall

Neerti:
Mechs now will fall when crossing an open tile.
Anything below that tile will have a bad day.
Also makes lighting overlays unable to fall just incase.

### DIFF
--- a/code/modules/multiz/movement.dm
+++ b/code/modules/multiz/movement.dm
@@ -23,12 +23,12 @@
 	if(!istype(start))
 		to_chat(src, "<span class='notice'>You are unable to move from here.</span>")
 		return 0
-		
+
 	var/turf/destination = (direction == UP) ? GetAbove(src) : GetBelow(src)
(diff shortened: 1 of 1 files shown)


code/modules/multiz/movement.dm
```diff
@@ -195,4 +217,26 @@
...
+		return
+
+	// Tell the pilot that they just dropped down with a superheavy mecha.
+	if(occupant)
+		to_chat(occupant, "<span class='warning'>\The [src] crashed down onto \the [landing]!</span>")
+
+	// Anything on the same tile as the landing tile is gonna have a bad day.
+	for(var/mob/living/L in landing.contents)
+		L.visible_message("<span class='danger'>\The [src] crushes \the [L] as it lands on them!</span>")
+		L.adjustBruteLoss(rand(70, 100))
```
SpadesNeil: oh god

Issue #3486, opened 2017-05-31
Error dialogue opening when using crew monitoring console.


KepQwet:
#### Brief description of the issue

Relevant image:
![Image of error recieved](https://i.gyazo.com/e52fac5389e672960a0d761c930bcb48.png)

#### What you expected to happen

For the crew monitor to open normally.

#### What actually happened

See the image above; that dialogue would open every time I clicked the monitor, or every time it proc'd in the background, until I closed the monitor or told it to stop creating new messages.  The issue persisted after reconnecting, and also occurred with another, separate monitor console.

#### Steps to reproduce

I am not sure as to the exact conditions that led to the issue. It _might_ have something to do with the fact that I had a laptop in my hand when I clicked the console to open it, but I am unsure.

#### Additional info:
 - **Server Revision**: Server revision: master - 2017-05-31 a412b5063bf9eac1ea83c29ad6f6afe28f0f668e 
 - **Anything else you may wish to add**
The monitor seemed to function fine after clicking "Don't let this page create more messages," but I could not access the tracking map and had to close the error again each time I tried to use it.

Anewbe:
Okay, thought this was just me. It appears to cause a new issue message when people on the console move.

PR #3492, opened 2017-06-01
Prevent returning null from get_authentification_rank()

Leshana:
Make sure we return the no job value if they have an id without a rank that is in a PDA.
Should fix https://github.com/PolarisSS13/Polaris/issues/3486


Anewbe:
❤️

PR #3493, opened 2017-06-01
Port of StonedMC master controller framework

Leshana:
Partial port of @PsiOmegaDelta's Baystation12/Baystation12#16820 and https://github.com/Baystation12/Baystation12/pull/17185
Only ports the StonedMC changes, not the garbage collector (forthcoming in future)

This shouldn't change gameplay or even performance or really anything at all other than adding the "MD" stat panel for admins.   This is just the first step in porting all of our controllers to use Master.
The next step will be to switch to using the /tg garbage collector, which will involve not only swapping out the garbage collector service, but also upgrading every `Destroy` proc to return a qdel() hint.   That is such a big undertaking that I thought it deserved its own separate change so that the important changes in this PR do not get buried under it.


Anewbe:
Has this been tested?

Leshana:
Yeah.  It doesn't do much at this point, but everything still works, so: victory.   We can now start porting over all the process controllers to use this instead.

Iamgoofball:
about half the lag fixes involve qdel though

PR #3499, opened 2017-06-04
Lazyports artifact containers from Bay

Yoshax:
Put artifact inside, be happy no die. All is well in the world. No mapped in cause I dunno where to put and lazy.


code/modules/xenoarcheaology/anomaly_container.dm
```diff
@@ -0,0 +1,43 @@
+/obj/structure/anomaly_container
+	name = "anomaly container"
+	desc = "Used to safely contain and move anomalies."
+	icon = 'icons/obj/xenoarchaeology.dmi'
+	icon_state = "anomaly_container"
```
Anewbe: This sprite doesn't seem to exist.

Yoshax: Seems I forgot to commit the file.

PR #3510, opened 2017-06-06
Ports inflatable dispensers from Bay

FTangSteve:
Also, adds them to engie borgs and gives those borgs a small pool of plasteel.


Anewbe:
How does it work, precisely?

FTangSteve:
What do you mean how does it work? You set whether you want doors or walls on the device. You click and deploy the wall and hit it with the device to suck it back in.

PR #3511, opened 2017-06-07
tg Garbage Subsystem

Leshana:
Continuation of #3493 - the port of tg's garbage collection subsystem, based on @PsiOmegaDelta's Baystation12/Baystation12#16820

### Summary
Replaces the master process controller garbage collector with a StonedMC subsystem garbage collector. Aside from being part of StonedMC, the main differences are:
* Instead of using `finalize_qdel` proc to determine the fate of qdel'd objects,  `Destroy()` is expected to return a qdel hint.  *All Destroy() procs need to be upgraded to do this eventually.*

Hint | Behavior
--- | ---
QDEL_HINT_QUEUE | qdel should queue the object for deletion.
QDEL_HINT_LETMELIVE | qdel should let the object live after calling destory.
QDEL_HINT_IWILLGC | Functionally the same as the above. qdel should assume the object will gc on its own, and not check it.
QDEL_HINT_HARDDEL | qdel should assume this object won't gc, and queue a hard delete using a hard reference.
QDEL_HINT_HARDDEL_NOW | qdel should assume this object won't gc, and hard del it post haste.
QDEL_HINT_FINDREFERENCE | Functionally identical to QDEL_HINT_QUEUE if TESTING is not enabled in _compiler_options.dm. If TESTING is enabled, qdel will call this object's find_references() verb.

* If a Destroy() proc doesn't return a hint, a stack trace is logged so developers can fix it, and its treated as QDEL_HINT_QUEUE for backwards compatibility.
* Previously, any qdel'd `/datum` type would be immediately hard-del'd.  New default is to put in the gc queue like other objects.  *Anywhere in the code that depends on the hard-del reference clearing needs to be replaced by proper reference clearing in Destroy()* This might seem annoying, but avoiding all these hard dels is a good speed boost.
* I searched and destroyed remaining uses of del() that should be qdel()
* I removed as many overrides of the /Del() builtin proc as possible, since @Fox-McCloud says they suck.
* You can now qdel() `/client` due to da̷r͜k̶ ҉m̢a̢gi̛c̨ - `/client` is now a subtype of `/datum`

### Includes code from:
* Baystation12/Baystation12#16820
* Baystation12/Baystation12#17398
* Baystation12/Baystation12#17386

### Status and next steps
I have updated Destroy() procs on quite a few items, but there are many more still to be done.  I have walked around the server deleting random stuff I see all around me in an effort to find the most common cases.  I'd like others to contribute to the effort of eventually overhauling them all as well.


Fox-McCloud:
The only `Del` that should exist, post this, should be `/Client/Del`, because of Client's nature.

Anything else that needs to be hard deleted should return the hint `QDEL_HINT_HARDDEL_NOW` or `QDEL_HINT_HARDDEL` (the latter of which is only utilized for mobs, on TG, as far as I know).

Leshana:
As @Fox-McCloud suggested I have merged in relevant bugfixes from tg's latest garbage subsystem.   In the same vein I've also updated the MC and subsystem code as well, but that will come in a following PR once this one is in.

Indeed, `/client/Del` is the only Del proc remaining, and `/client/Destroy` does return `QDEL_HINT_HARDDEL_NOW` so if someone qdels a /client instance it will work fine.

Fox-McCloud:
@Leshana 

Keep in mind that only works if `Client` is a child of `Datum`. If it isn't, then having a `Destroy` for it is unnecessary as `qdel` will see that `Client` isn't a datum and just hard delete it immediately.

I am unfamiliar with the quirks and oddities (or potential problems) of making `Client` a child of `Datum`, at this point in time, however.

Leshana:
@Fox-McCloud  Yes, this PR sets  `/client/parent_type = /datum here: ` https://github.com/PolarisSS13/Polaris/pull/3511/files#diff-64d6aba98dafae4245c5765723ecd5c9R5
It does indeed seem to be dark magic, but I did ask around a bit.  TG and BayStation have had it this way for a few months at least and they said it was okay.   Its easy to reverse if it turns out to cause problems tho.  I did it for convenience.  With the updates I ported from /tg, even if we unset it as a child of datum, qdel() will still work fine on /clients (like you said) because of https://github.com/PolarisSS13/Polaris/pull/3511/files#diff-d6f59f777ea605d6ea004a768095ba2dR188

Fox-McCloud:
Ayup! Glad to hear this is moving forward for you guys.

PR #3516, opened 2017-06-08
Adds macros to assist with lazily instantiating lists

Leshana:
* For various reasons its best to not instantiate lists until they are actually going to be used, especially if there is a good chance that a given list variable might *never* be used during the lifetime of an object.
* These macros make it simple and concise to add remove and access entries in lazily created lists.
* Converts area.all_doors list to be lazily instantiated as a simple example of using the lazy list macros.


Anewbe:
Tested?

Leshana:
Yes I did test it. Server boots and fire doors operate properly all without runtimes.   area.all_doors isn't going to gain any big speed boost, but its a simple and easily tested example to show the principle.

PR #3522, opened 2017-06-10
XGM init() and lazy list optimization

Leshana:
###  Eliminate init() proc from /datum/gas_mixture
* Initializing a list in the type definition (of any type) incurs a hidden proc call ("init").   Simply by moving the init of the lists to /New() we speed up creation of /datum/gas_mixtures (which happens *often*)
* The list variables still get initialized by the time the instance is created, so this has zero impact on other code, its just starightup faster with no downsides.
### Converts /datum/gas_mixture/var/graphic to a lazy list. 
* The vast majority of gas mixtures _never_ use their `graphic` list.  Prime candidate for making a lazy init list.
* While we are here, add nullchecks to ZAS's use of the graphic list a bit.  /turf/update_graphics was technically already null safe, but its even better to not bother calling it at all right?

### DIFF
--- a/code/ZAS/Turf.dm
+++ b/code/ZAS/Turf.dm
@@ -5,9 +5,9 @@
 /turf/var/datum/gas_mixture/air
 
 /turf/simulated/proc/update_graphic(list/graphic_add = null, list/graphic_remove = null)
-	if(graphic_add && graphic_add.len)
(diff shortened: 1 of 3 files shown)


Anewbe:
Tested?

Leshana:
@Anewbe Yes.

PR #3523, opened 2017-06-11
Nerfs the siemens on most hardsuits

Yoshax:
Hello. This is something I had never noticed. All hardsuits with siemens_coefficient undefined are inheriting from the parent which is siemens_coefficient = 0.2. This makes most of them take 20% agony from stuns, and various other implications.

I propose these changes:

Station:
Industrial - 0.75
EVA - 0.75 (Has 0.0 gloves)
CE - 0.75 (Has 0.0 gloves)
AMI - 0.75
Medical - 0.75
NT Breacher - 0.75

ERT:
Commander - 0.5
Medical - 0.5
Security - 0.5
Engineering - 0.5 (Has 0.0 gloves. This one actually had 0,0 across the board for all pieces.)
HAP - 0.3 (Has 0.0 gloves)

Other:
Merc - 0.5 (Has 0.0 gloves)

### DIFF
--- a/code/modules/clothing/spacesuits/rig/suits/alien.dm
+++ b/code/modules/clothing/spacesuits/rig/suits/alien.dm
@@ -13,7 +13,7 @@
 	offline_slowdown = 10
 	vision_restriction = 1
 	offline_vision_restriction = 2
-
+	siemens_coefficient = 0.75
 	chest_type = /obj/item/clothing/suit/space/rig/breacher
 	helm_type = /obj/item/clothing/head/helmet/space/rig/breacher
 	boot_type = /obj/item/clothing/shoes/magboots/rig/breacher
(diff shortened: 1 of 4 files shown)


Neerti:
Just like to note that ```siemens_coefficient``` acts as a multiplier, meaning something having 0.2 means they won't take 20% less reduced agony/electricity/etc, rather that it would be 80% less, as 100 * 0.2 = 20

Regardless this seems fine for me.

Yoshax:
Apologies, I was aware of that, I meant they'd take 20% of the original, but typo'd it.

SpadesNeil:
The thing about electricity is electricity is going to take the easiest path to the ground. For the same reason the guy at the Boston Museum of Science doesn't get turned into fried poultry while sitting in the bird cage at the lightning exhibit, it doesn't make sense for metal hardsuits to allow their occupant to be electrified. If they get shocked, electricity is going to go around the suit's surface, rather than through the fleshy thing inside of it.

https://www.youtube.com/watch?v=1-mFl_YXqiA I am referring to this.

Yoshax:
Disregarding realism, as it is currently, anyone in a hardsuit is essentially immune to most forms of non-lethal takedown.

SpadesNeil:
Sounds fair to me.

PR #3524, opened 2017-06-11
Makes Examine Panel slightly more useful.

Neerti:
The examine panel can now display different interactions with a few items, using icons, mostly as a proof of concept.

Example


Leshana:
Whaaat? Are you giving away all the secrets?  With this... people would...   People would know how to play the game without the wiki open on their other monitor.

Screemonster:
I fully support using the examine panel so coders can include their interface changes in their code without having to scream at people to update the wiki every time a PR is merged. :v

PR #3528, opened 2017-06-12
Ports Bay/Paradise Defibs

Arokha:
I like the ones I wrote, but these are proooobably better.

They have normal units, which are worn on the back and you can grab the paddles out of to shock people.
Then there's compact units, which are worn as a belt, and serve the same purpose.
And finally there's the combat units, which allow revives through spacesuits, and have no safety so you can zap living people on harm intent (good for antags? CMO?).

These appropriately check for blood, a heart, etc. So, in that way, they're a bit more realisic than my original ones which I was sorta lazy in writing.
Emagging them turns off the safeties, allowing them to zap people who are still alive.
They still take cells and start with a highcap which is 10 uses in a normal unit and 5 in a compact.

I actually cleaned up their sprite choice code a little and added a sprite for the 'combat' one since previously it was just the emagged sprite (flashing exclamation point). Now it's just got a red heart monitor, and the emagged ones keeps the exclamation. They also have a charge bar, so you can see how many uses you have left.

Did make one balance change in that compact ones use twice the power (inefficient small capacitors? or something?).

image

I'm not really sure what to do about the jumper cables defib kits for synths. @Anewbe ?


Anewbe:
I mean, they're not technically a necessary item, so you can comment them away if you want.

Arokha:
They are stubbed right now, and will spawn a normal defib, because it won't compile and pass travis otherwise, since I assume they are on your map.

Anewbe:
They are indeed on my map.

Woodratt:
>  +#define DEFIB_TIME_LIMIT (8 MINUTES) 

If I'm reading this correctly, probably needs to be bumped back up to 10 minutes. 

Aside from the concerns with the jumper defibs. There was also concerns in regards to the burn damage they deal to the person they're attempting to revive. 15 seems a bit high, considering two shocks would means 30 burn damage, three 45 burn with the associated blood loss and all. (If I read this correctly again). Could it be lowered to persay, 5 or so burn? Or would that be bad?

Arokha:
I don't have strong opinions about it! I can change those, sure.

Arokha:
Beep? @Anewbe

Woodratt:
Will beep Anewbe and Neerti when get chance.

PR #3531, opened 2017-06-15
Artifact_harvester and blastdoor fix and changes

Cameron653:
-------------------------------------------**Artifact_Harvester**-------------------------------------------------
- Adds a small delay when the artifact_harvester spawns in before looking for an artifact_scanpad.

Without this delay, the artifact_harvester can spawn first, (depending on the map layout) look for the artifact_scanpad, not find it and give up. Afterwards, the artifact_scanpad spawns but can't be linked, as the artifact_harvester already looked and gave up.

Example:

- Without the delay: https://i.gyazo.com/cb96fe8a563f5868231e968d4ab6c0c5.png
- With the delay: https://i.gyazo.com/548ea3b6521c9c9612ac95cdd7993b0a.png
--------------------------------------------------------------------------------------------------------------
------------------------------------------------**Blast Door**---------------------------------------------------
- Fixes an indent error with blast doors that prevented them from being fixed with plasteel, as plasteel is a child of "/obj/item/stack/" not "/obj/item/weapon/"
- Fixes an indent error that prevented blast doors from being hit by an item.
- Modified the code so non-weapons can also be hit against the door, although it has a much harsher brute modifier (.15 compared to .35 that a weapon has) when attempting to break down the door. 
--------------------------------------------------------------------------------------------------------------

### DIFF
--- a/code/game/machinery/doors/blast_door.dm
+++ b/code/game/machinery/doors/blast_door.dm
@@ -119,7 +119,7 @@
 			return
 
 
-		else if(src.density && (user.a_intent == I_HURT) //If we can't pry it open and it's a weapon, let's hit it.
+		else if(src.density && (user.a_intent == I_HURT)) //If we can't pry it open and it's a weapon, let's hit it.
 			var/obj/item/weapon/W = C
 			user.setClickCooldown(DEFAULT_ATTACK_COOLDOWN)
 			if(W.damtype == BRUTE || W.damtype == BURN)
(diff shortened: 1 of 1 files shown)


Cameron653:
DNM. Doing another quick fix.

Cameron653:
Alright, did some modifications and made it so weapons & nonweapons can correctly hit it now, it can be repaired correctly. This should be ready to merge if there isn't any visible issues.

Anewbe:
Travis Issue.
Also, we prefer that PRs are as small and as specific as possible. This was fine before, now it means we need to double check both parts and decide if we want them.

Cameron653:
Yea, I'm currently working on fixing the compile issue and the description at the moment. Should be fixed in a second.

Cameron653:
Alright. Description changed, compile errors fixed (I always forget to make it have two =='s and two brackets)

If need be, I can split this up into two PR's, but I don't see that being necessary due to this being two bugfixes and one extremely small modification (I don't even think any non weapon items _have_ 20 or more brute/fire damage)

Cameron653:
Going to split this up into two different PR's

PR #3548, opened 2017-06-28
Mutagenic chemicals tweak

Cameron653:
-------------------------------------**Mutagenic  Chemicals Tweak**--------------------------------------------

- Changes amount of mutagen/slimetoxin/aslimetoxin required for a good/bad mutation
- Adds an 8% chance per tick for a cosmetic mutation when you have  mutagen/slimetoxin/aslimetoxin in you. This equates to about 4 cosmetic mutations per 10 units injected.
- Documents the math and original coder's idea in coder comments.

Was doing some experimentation with unstable mutagen and was wondering why I was getting literally 0 mutations. Did the math (100/(0.2 * 0.1)) and found out that it would take _5000_ (!!!) units of mutagen to be injected for a single cosmetic & good/bad mutation.

Went ahead and fiddled around with the numbers for a while before changing it so that there is a 2% per tick for mutagenic chemicals to give a cosmetic & good/bad mutation, and an 8% chance per tick for a purely cosmetic mutation. 

I'm not sure if good/bad mutations are even enabled, but this ensures that if they are enabled then mutagen will actually be useful. This also makes it so it's useful for changing people cosmetically.

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Toxins.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Toxins.dm
@@ -315,8 +315,12 @@
 	if(istype(H) && (H.species.flags & NO_SCAN))
 		return
 
+//The original coder comment here wanted it to be "Approx. one mutation per 10 injected/20 ingested/30 touching units" 
+//The issue was, it was removed (.2) multiplied by .1, which resulted in a .02% chance per tick to have a mutation occur. Or more accurately, 5000 injected for a single mutation.
+//To honor their original idea, let's keep it as 10/20/30 as they wanted... For the most part.
+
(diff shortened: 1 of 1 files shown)


Cameron653:
Any suggestions or anything I missed on this?

Anewbe:
Are these the genetics mutations?

Cameron653:
Yes. They're the ones that change the person's hair color, eye color, etc. And the randmutb and randmutg gives them a bad or good superpower mutation if they're enabled.

zekesulastin:
It would be really nice if we had had a way to fix cosmetic mutations before they became a thing again (since genetics is still broken).  If only I had seen this before it was committed.

Anewbe:
Ah shit, knew I was missing something.

PR #3549, opened 2017-06-30
24-round shotgun mag fixes + making it relevant.

Andromeda-K22:
Adds the AS-24 Shotgun to the traitor uplink for 100 TCs, also makes automatic shotgun magazines (which spawn with shotgun slugs, as a note) costs the appropriate amount (13,000 metal up from 2200)

### DIFF
--- a/code/datums/uplink/visible_weapons.dm
+++ b/code/datums/uplink/visible_weapons.dm
@@ -70,6 +70,11 @@
 	item_cost = 75
 	path = /obj/item/weapon/gun/projectile/automatic/sts35
 
+/datum/uplink_item/item/visible_weapons/automaticshotgun
+	name = "Automatic Shotgun (12g)"
+	item_cost = 100 //it's an automatic shotgun. Has a lot of potential for destruction.
+	path = /obj/item/weapon/gun/projectile/automatic/as24
+
 /*/datum/uplink_item/item/visible_weapons/bullpuprifle
 	name = "Assault Rifle (5.56mm)"
 	item_cost = 7
--- a/code/modules/projectiles/ammunition/magazines.dm
+++ b/code/modules/projectiles/ammunition/magazines.dm
@@ -467,7 +467,7 @@
 	icon_state = "12g"
 	mag_type = MAGAZINE
 	caliber = "12g"
-	matter = list(DEFAULT_WALL_MATERIAL = 2200)
+	matter = list(DEFAULT_WALL_MATERIAL = 13000) //did the math. now fixed the exploityness of this thing. Have fun!
 	ammo_type = /obj/item/ammo_casing/a12g
 	max_ammo = 24
 	multiple_sprites = 1


Woodratt:
If this is the shotgun I think it is, Heavily biased against adding due to how ridiculously strong the auto shotgun is.

Andromeda-K22:
It's the 3-round burst AS-24. It costs 100 TCs for a reason, which is literally going to gimp you completely aside from a few blocks of C4 or a knife.

SpadesNeil:
This is terrifying. Please no. I had a _semi-auto_ shotgun once that had a fixed magazine, and it was _still_ terrifying.

24 rounds of shotgun in a detachable box mag makes the SAW look like a pea shooter.

Andromeda-K22:
Then why do either of these items exist in the code.

_I did not code the gun. I did not code the magazine._ If they're so OP, why was the 24-round magazine even added to the autolathe. Why do these things exist.

SpadesNeil:
~~Admin abuse.~~
~~FUN™~~
Knocking powergamers down a few pegs when necessary.

Andromeda-K22:
I'm going to take a peak into the code - because it is my believe that i have balanced the avalibility of this well. 100 TCs _believe it or not_, is quite a few. You are limited to: two bricks of C4, or a combat knife, or a tactical knife. 

Getting more ammo is also a significant factor, considering the magazines now cost **13000 metal.** 

Current damage values for shotgun slugs however, are absurd. Absolutely absurd. Slugs are 50 damage, with 15 AP. 

Combat armor has 50 bullet protection. ballistics armor rests at a not-inconsiderable eighty. Assuming AP works like i think it does, after applying the AP effect to it, Combat Armor has a mere 35 bullet and ballistics has a less mere 65. 

Slug damage is still absurd. So slugs probably need a nerf, which would make this gun less completely and totally fucking nutso, as well as restricting the magazines to 12-rounds and removing the slug ones and adding in beanbag + buckshot mags. An accuracy nerf when firing in burst mode is probably in order too, but i'l have to take a look at how bad the values are already.

Yoshax:
You want to nerf slugs, and fuck the balance on normal shotguns, to balance this gun?

Andromeda-K22:
Slugs need a nerf period. Did you read anything i said? I'm not messing with the balance of pellets / beanbags. Again. Slugs have 50 damage and 15 AP. 

This would affect _this gun as well as all other shotguns. It is a general nerf on one specific ammo type._
I don't see how this fudges balance when majority of the time it's pellets / beanbags being used. 

A majority of the time slugs are used is when security has literally no other ammo (because who wants to fuck around with producing shells one by one?).

Yoshax:
I read it, yes. What I'm reading now is that 'slugs are op' so 'lets nerf slugs' but 'buckshot is more op' but 'lets nerf slugs'.

Maybe you're not aware but shotguns are meant to hit hard. You're also just seriously overestimating how good they are, especially in comparison with other guns.

SpadesNeil:
Shotguns generally aren't even semi-automatic in this game, severely limiting their firepower.

It's all about DPS, and this autoshotgun should not be accessible normally.

Anewbe:
Yeah, one burst from this thing causes Crit, I'm not really jumping to add it.

The cost fix is fine by me, for the magazines, but that's probably its own PR.

Andromeda-K22:
I might as well go ahead and close this, since there doesn't appear to be a ice cube's chance in hell of it getting through.

PR #3551, opened 2017-07-01
Fixes mimic death()

SpadesNeil:
I'm pretty sure this is not correctly done.


SpadesNeil:
aaaaaaaaaaaaaaaa why'd merging one update do that pls squish merge this

also fixed malf drone because someone was a dumb

PR #3555, opened 2017-07-03
Woodrat breaks map stuff

Woodratt:
Added SC map files for further work. Including hopefully all the related files.

New airlock sprite thanks to Virgo, with a alternate non-see through sprite.

Changes to the space and transit tiles (ported from bay).

Changes done to the NS map that should bring it more up to date with changes (shield defuse stuff and several bug fixes)


Anewbe:
Were you unable to use the map cleaner for the main map, or would that have wiped the SC?

Other than that looks fine from what I can see.

Woodratt:
In theory the map cleaner was used.

PR #3556, opened 2017-07-03
Readds Jumper Cable Kits

Nalarac:
+Readds jumper cable kits to revive FBPs (Robotics will spawn with these
without any map changes but they will be empty so that needs to be fixed
anyways)
+Readds jumper cable kits to cargo orders
-Lowers jumper cable kits in order to 2 from 4 to be in line with the
defibrillator changes
+Adds robot defibrillators to both medical modules
+Adds robot jumper cables to research module
+Adds some new icons for jumper cables to fit in better with the new
defibrillator sprites


Techhead0:
"... and that's when my dad beat me with a length of jumper cable."

Nalarac:
Nice meme. Though, at the moment the "jumper cables" are just slightly resprited defib shock paddles since I feel the functionality is more important than having perfect sprites.

PR #3566, opened 2017-07-07
Removes Forced Antag Draft

Neerti:
Comments out the code to force people into an antag role if they vote for the gamemode, in an effort to have more people join at roundstart.


Belsima:
Changelog?

PR #3570, opened 2017-07-07
Enables unused cypherkey sprites.

Belsima:


Anewbe:
Actually, going to revert part of this, the syndicate and binary keys are **supposed** to look like the default key.

PR #3573, opened 2017-07-08
Assorted map tweaks

Anewbe:
- Adds loaded defibs to the map
- Removes Christmas
- Corrects the security gun spawns


Anewbe:
Fairly small but important change, was given permission to merge.

PR #3577, opened 2017-07-09
Fixes slugs being available in an unhacked autolathe

Anewbe:


Anewbe:
Got permission to self-merge.

PR #3582, opened 2017-07-10
Rolling sleeves on flannel shirts now uncovers your arms

Anewbe:


Anewbe:
Got permission to self-merge.

PR #3583, opened 2017-07-10
Adds a boatload of sweaters.

Belsima:
Mostly ported from Eros Research Platform.

image


Anewbe:
Off topic: Have you checked that the body_parts_covered actually does anything with accessories?

Belsima:
I have not, but it was used for the ponchos so I just assumed it did.

PR #3586, opened 2017-07-10
Fixes up the galoshes.

Belsima:
From: image
To: image

Another minor adjustment, but the old sprite bothered me every time I saw it.


SilverTalismen:
Finally they are 30% less ugly, would be nice if we got some normal slip resistant shoes though.

PR #3592, opened 2017-07-11
Adds a corgi, bird, and more sprites to the pAI.

Belsima:
image

Taken from Paradise Station.

I decided to take more from Paradise Station.
image


Atermonera:
And if you add a comment after a list item, it has to end in a comma or Travis throws a fit

PR #3593, opened 2017-07-11
makes scrubbers filter N2O

Belsima:
"ai, set all scrubbers to filter N2O"
"[src] has been moved to intelligence storage."

### DIFF
--- a/code/ATMOSPHERICS/components/unary/vent_scrubber.dm
+++ b/code/ATMOSPHERICS/components/unary/vent_scrubber.dm
@@ -19,7 +19,7 @@
 
 	var/hibernate = 0 //Do we even process?
 	var/scrubbing = 1 //0 = siphoning, 1 = scrubbing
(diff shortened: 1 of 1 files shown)


SpadesNeil:
Did... did it not already?

Atermonera:
Not by default, no. It has the functionality to scrub anything, but was initially set to just phoron and co2

Anewbe:
I feel like the brig makes a fairly specific use of siphons not auto-filtering N2O...

Atermonera:
Has that system ever actually been used? Particularly by non-antags?

SpadesNeil:
I've never seen the N2O in the brig used in all the time I've played SS13, except by people who just wanted to see what happens. The prisoner population isn't ever enough to justify it.

Anewbe:
I've personally used it several times. It's helpful.

PR #3600, opened 2017-07-13
Gives medical scrubs sleeves, that can be rolled down.

Anewbe:
Gives all of the colorful medical scrubs white sleeves (from an undershirt), that can be rolled down to go back to the usual sprite.


Belsima:
Isn't this what actual undershirts are for?

Anewbe:
Actual undershirts don't have the hiding flags or the same covering stats.

Belsima:
Ah, right.

PR #3602, opened 2017-07-15
Lowers the hallucinations from the conversion rune

Anewbe:


Neerti:
It'd probably be better to do ```target.hallucination = min(target.hallucination + 100, 500)``` to cap it, since the rune generally does that every few seconds, and 500 hallucination takes a really long time to go away anyways, approximately 17 minutes iirc, so perhaps a lower cap would be better.
Also should amend that comment above.

Anewbe:
Good plan, done.

PR #3603, opened 2017-07-16
Caps the negative effects of the high-end drugs

Anewbe:


elgeonmb:
(it looks like myelamine has an incrementor instead of an assignment operator so the cap amount of eye blurry would be hit ridiculously fast)

(but also I'm pretty code-illiterate so IDK)

PR #3605, opened 2017-07-16
Gives IECs some TLC

LorenLuke:
Numerous improvements (see changelog).

Multitool functionality: Click pin name to wire, click data (in parethesis) to input data (cannot input refs). Click connected pin name to sever connection between two pins. Click <PULSE IN/OUT> to pulse activator pin. Use in hand (red state) to cancel current connection. All other tool functions identical to before. Click [Return to assembly] to return to the assembly. Click @ [circuit] on a connected pin to move to that circuit in the assembly.


Atermonera:
Can you still have multiple windows for an assembly? Sometimes in testing assemblies it's annoying when you have to pulse one circuit and then navigate to another circuit to see the response, much simpler when there's a window for both open and you just have to refresh the second circuit.

LorenLuke:
@Atermonera 
>multiple windows

Not presently. I can add it in phase two as a player preference, if needed. One thing of note to make this easier is that when you've two inputs linked, you can click the circuit name after the at sign (@) and it will move the window to the relevant circuit, allowing you to trace data in either direction to see if your circuit is functioning as intended.

Anewbe:
Has this been tested?

LorenLuke:
@Anewbe 
>Tested
Rigorously.

PR #3608, opened 2017-07-17
Changes Spider meat back to Xeno meat

SpadesNeil:
But also changes a bad design choice and makes spider meat a viable alternative to recipes requiring Xeno meat.

Now, for those of us who do still use xenos for things, xeno meat isn't suddenly completely useless, while spiders still produce their own unique meat that is usable in these recipes.


Anewbe:
Did you test this, all the recipes and such actually work with the spider meat?

SpadesNeil:
I have not. I should do that. Kinda rushed this out.

SpadesNeil:
Sorry for the delay. Tested this. Verified it works just as I intended.

PR #3611, opened 2017-07-18
Makes blob light green and dim.

Belsima:
Ported from https://github.com/Baystation12/Baystation12/pull/14599.


LorenLuke:
Maybe add an orange light to the core?

Belsima:
Done.

PR #3612, opened 2017-07-19
Changes 5.56mm to 5.45mm (literally just a name change)

SpadesNeil:
Because the STS is clearly a goddamn Space AK and I cringe every time I see 5.56 on its ammo type.

This change serves to do literally nothing except make gun enthusiasts like me not cry inside, even though it's a fictional space gun and nobody outside gun lovers care what ammo it uses but DAMN IT I CAN USE THAT LOGIC TOO, SO IT USES 5.45mm MUNITIONS NOW SO THERE IT NO LONGER MAKES ME IRRATIONALLY ANGRY AND ANEWBE AND I CAN BOTH BE HAPPY. :U (love you Anewbe)

Also fixes a few oversights like how the assault rifle was still labeled 5.56 (or now 5.45) in the traitor menu, or how the ceremonial rifle said it used the wrong ammo as well.

Oh yea, and the small magazine for the 5.45 gun actually looks like it's the correct mag for the gun. This has the added bonus of freeing up that SVD mag for later use, if we ever want it.

PR #3613, opened 2017-07-19
Changes 5.56mm to 5.45mm, now 20% less stupid

SpadesNeil:
The same as https://github.com/PolarisSS13/Polaris/pull/3612 except this time I'm not a dumbass and accidentally overwriting other changes because the Github client decided not to sync with Polaris.


code/modules/projectiles/guns/projectile/automatic.dm
```diff
@@ -111,7 +111,6 @@
 	force = 10
 	caliber = "7.62mm"
```
Anewbe: Caliber still doesn't seem to match, but I don't know what it actually does anymore.

code/modules/projectiles/guns/projectile/automatic.dm
```diff
@@ -181,14 +180,13 @@
 	force = 10
 	slot_flags = 0
 	max_shells = 50
-	caliber = "5.56mm"
+	caliber = "5.45mm"
 	origin_tech = list(TECH_COMBAT = 6, TECH_MATERIAL = 1, TECH_ILLEGAL = 2)
 	slot_flags = SLOT_BACK
-	ammo_type = "/obj/item/ammo_casing/a556" // Is this really needed anymore?
```
Anewbe: As mentioned above, not sure what precisely this does, but I'm not sure why it's being removed.

SpadesNeil: This _used_ to be code used to fill up the gun with the correct kind of bullets when it spawned. This is completely useless now, as this gun uses a magazine, which already has bullets spawned inside of it.

Anewbe: Alrighty.

PR #3618, opened 2017-07-22
Meteors warning time.

Anewbe:
Meteors now give around 5 minutes warning before they start smashing the station

### DIFF
--- a/code/modules/events/meteors.dm
+++ b/code/modules/events/meteors.dm
@@ -1,6 +1,6 @@
 /datum/event/meteor_wave
-	startWhen		= 5
-	endWhen 		= 7
+	startWhen		= 150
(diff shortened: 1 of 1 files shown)


SpadesNeil:
While I am not opposed to the change (and in fact support it), what prompted the change if you don't mind me asking?

Nalarac:
My guess would be to give engineers a chance to pump up the power on shields.

Anewbe:
A player requested it in our Town Hall, and I thought it was a fair point. The idea is to give people a chance to get towards the center of the station, or to let the Engineers try to do shield things, or pray, as it suits them.

Anewbe:
We may or may not have a better way to do this.

PR #3621, opened 2017-07-23
Makes sure female executive suit has different name

Cameron653:
The modular suit selection looked for all suits under /obj/item/clothing/under/suit_jacket
As both /obj/item/clothing/under/suit_jacket/female and /obj/item/clothing/under/suit_jacket/really_black were named "executive suit" it only showed one of them, which happens to be the female only. By renaming the female one to "female executive suit" it allows both the /female and /really_black to show up under the modular suit selection

- Corrects the name of an outfit to allow both suits of the previous same name to be selected by the modular suit selector.


Cameron653:
Ready for merge. Tested it and it appears to work as expected.

PR #3627, opened 2017-07-23
Fixes an oversight with eguns

Anewbe:
Discussed


MagmaRam:
If this changes the damage eguns do, should it also be labeled "feature tweak" and get a changelog?

Anewbe:
Don't see the need for a changelog.

PR #3630, opened 2017-07-23
Ports TG Medibot Sounds + More

Neerti:
Medibots now have audio sounds files to play when they talk, which I find to be both practical and makes the bots somewhat creepy and/or endearing.  The files are from /tg/station.
Adjusts some code so that all the new sound files can play when applicable.
Adjusts small pieces of code to make slightly prettier.
Re-Adds the Mysterious Medibot, International Medibot of mystery.  Admin-spawn only.
Emagged medibots will do the beepsky rant 25% of the time.
Bonus: Adds integrated circuit that can play all the new medibot voices.


Anewbe:
Looks fine, gonna give it the usual wait.

PR #3634, opened 2017-07-24
Oxyphoron change proposal

silveryferret:
So, water + phoron messes with a lot of stuff. My bad. 

However, hardly any medicines/toxins/etc use both hydrogen and phoron! Even stuff that uses hydrogen + more complicated chemicals, like methyldrate (or whatever the ADD med is called exactly), none of the stuff that's required uses both phoron and hydrogen. So my proposal is that oxyphoron becomes hydrophoron, as well as have an inhibitor of nitrogen so that it doesn't mess with the lexorin recipe, which is the only thing I managed to find that uses hydrogen and phoron. 

This also has an added benefit of being /slightly/ more realistic. It could possibly be fluffed that hydrophoron + water makes heavy water, and that's where the deuterium comes from. We're using a magical chemical already (phoron) so some suspension of disbelief is still required, but... you know.

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Toxins.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Toxins.dm
@@ -64,8 +64,10 @@
 /datum/reagent/toxin/hydrophoron/touch_turf(var/turf/simulated/T)
 	if(!istype(T))
 		return
-	T.assume_gas("oxygen", ceil(volume/2), T20C)
 	T.assume_gas("phoron", ceil(volume/2), T20C)
+	var/datum/effect/effect/system/spark_spread/s = new /datum/effect/effect/system/spark_spread
+	s.set_up(1, 1, T)
+	s.start()
 	remove_self(volume)
 
 /datum/reagent/toxin/spidertoxin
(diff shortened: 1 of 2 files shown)


silveryferret:
If discussion is had, and it turns out I missed some of the advanced/optimal mixes, I've no problem editing it to require like.. Tungsten or something like that. I just wanted to try to keep the suspension of disbelief at a minimum.

code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Toxins.dm
```diff
@@ -45,23 +45,23 @@
...
+/datum/reagent/toxin/hydrophoron/affect_touch(var/mob/living/carbon/M, var/alien, var/removed)
+	M.take_organ_damage(0, removed * 0.1) //being splashed directly with hydrophoron causes minor chemical burns
 	if(prob(10 * fire_mult))
 		M.pl_effects()
 
-/datum/reagent/toxin/phoroxygen/touch_turf(var/turf/simulated/T)
+/datum/reagent/toxin/hydrophoron/touch_turf(var/turf/simulated/T)
 	if(!istype(T))
 		return
 	T.assume_gas("oxygen", ceil(volume/2), T20C)
```
Anewbe: Still want it to create oxygen?

PR #3642, opened 2017-07-25
Makes Thunderdome camera indestructable.

SpadesNeil:
Uses this code:
```
/obj/machinery/camera/ex_act(severity)
    if(src.invuln)
        return
```

### DIFF
--- a/code/game/machinery/camera/camera.dm
+++ b/code/game/machinery/camera/camera.dm
@@ -13,7 +13,7 @@
 	var/c_tag_order = 999
 	var/status = 1
 	anchored = 1.0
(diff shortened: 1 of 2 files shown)


code/game/machinery/camera/camera.dm
```diff
@@ -13,7 +13,7 @@
 	var/c_tag_order = 999
 	var/status = 1
 	anchored = 1.0
-	var/invuln = null
+	var/invuln = 0
```
SpadesNeil: This is just me nitpicking so mappers know they can set a value to 1 for it to be invulnerable.

PR #3659, opened 2017-07-28
Allows Click Dragging for Cryostorage

Nalarac:
**Allows individuals to click drag others into all forms of cryostorage
including robotic storage, the elevator, and the gateway.**

~~-Was unable to fix a small bug where if you moved while putting someone
into storage, they would be placed into the storage device while the
body would remain outside (but still be deleted upon processing) --
Found out this has nothing to do with my code since it was still a
problem when I tested without my changes.~~

### DIFF
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -622,7 +622,7 @@
 	if(!M)
 		return
 	if(occupant)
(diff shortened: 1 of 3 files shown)


Anewbe:
Would prefer that the cyborg tweaks be their own PR.

Anewbe:
Tested and all that?

Nalarac:
Yep, tested.

PR #3663, opened 2017-07-29
Modifies Illegal Equipment Module

Nalarac:
+Illegal equipment module now scrambled equipment module
+Scrambled equipment module now unlocks emagged borg items without
actually emagging the borg, allowing for AI sync, law updates, and being
emagged.
+Makes the miner borg diamond drill appear in all cases of being
upgraded or emagged instead of only if emagged while the mining module
is equipped.

Changelog includes all my previous PRs as well for ease of access to the
changes by the general public. I can trim it down to just this PR if
requested.

### DIFF
--- a/code/game/objects/items/robot/robot_upgrades.dm
+++ b/code/game/objects/items/robot/robot_upgrades.dm
@@ -157,7 +157,7 @@
 
 /obj/item/borg/upgrade/syndicate/
 	name = "scrambled equipment module"
-	desc = "Unlocks new and often deadly module specific items a robot"
+	desc = "Unlocks new and often deadly module specific items of a robot"
 	icon_state = "cyborg_upgrade3"
 	item_state = "cyborg_upgrade"
 	require_module = 1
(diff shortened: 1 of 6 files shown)


Anewbe:
Tested?

code/game/objects/items/robot/robot_upgrades.dm
```diff
@@ -156,17 +156,17 @@
 
 
 /obj/item/borg/upgrade/syndicate/
-	name = "illegal equipment module"
-	desc = "Unlocks the hidden, deadlier functions of a robot"
+	name = "scrambled equipment module"
+	desc = "Unlocks new and often deadly module specific items a robot"
```
Anewbe: Missed an "of"

Nalarac:
I had it tested and then merged it to the wrong branch, so I had to spend some time tweaking it so it was only the single changes I wanted. Probably why I had the mistakes from redoing it quickly. Tests showed it works as intended.

PR #3666, opened 2017-07-30
Tweaks prometheans

Anewbe:
Gives them a few things as we get ready to make them playable.
- Weaker limbs
- Gives them Money
- Hungrier now
- Less resistant to shocks
- Less resistant to brute
- Less able to handle temperature extremes
- Heal slower

Also tweaks the Unathi blurb, I was in the file and noticed an inaccuracy.

### DIFF
--- a/code/modules/clothing/clothing.dm
+++ b/code/modules/clothing/clothing.dm
@@ -69,7 +69,7 @@
 	//Set species_restricted list
 	switch(target_species)
 		if("Human", "Skrell")	//humanoid bodytypes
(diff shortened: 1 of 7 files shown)


Yoshax:
You may wish to look at this: https://github.com/VOREStation/VOREStation/pull/1226

It allows Prometheans to wear the suit that is cycled to whatever species they are transformed into. Not 100% certain it will work straight, but you may wish to look it regardless because it may work.

Anewbe:
I will do so in the next batch of tweaks, probably. Thank you.

Anewbe:
Loremaster approved.

PR #3673, opened 2017-08-01
Overhauls Circuit Pins

Neerti:
This PR changes integrated circuits in many ways.

Instead of making input/output pin types which acted the same way in practice, we now have many different kinds of pins, dedicated to holding specific data types, e.g. strings, numbers, refs, lists, colors, dirs, boolean values, etc.  These dedicated pins only allow their specific data type (or null, for most of them) to exist in their data slot.

To declare a specific pin type, do it like this;
```
inputs = list("pin name here" = IC_PINTYPE_STRING)
```

``inputs_default`` and ``outputs_default`` are now used to assign default values to pins.

Changes most pins to use a dedicated pin.  Memory pins still use the normal (now called ANY) pin since they are expected to be able to hold any data type.

Removes circuit designs from RnD, replacing it with a portable integrated circuit printer, that accepts metal sheets and dispenses any kind of circuit that could be obtained from RnD previously.  Some of the advanced circuits will require an upgrade disk to obtain, which can be made in RnD or ordered from cargo.  Shoving circuits back into the printer will recycle the metal.  A 'clone' upgrade disk exists but currently does nothing.

Removes autopulse, since it was terrible and was literally a race condition.

Removes splitter circuits since they were made for when only one pin could be linked to one pin.  This hasn't been true in a very long time.

Splits input_output.dm into input.dm and output.dm.

Moves responsibility of writing data from the multitool to the pin itself.

Adds a few list circuits.


Anewbe:
Tested, etc?

PR #3676, opened 2017-08-03
Updates Tools

Neerti:
Adds toolspeed var, which is a multiplier on how 'fast' the tool works.  0.5 means it goes twice as fast.
Adds usesound var, which determines what sound is used when a tool is being used.
Changes a lot of code to use those two vars instead.
Adds 'ayyy' tools, which are ported from /tg/'s abductor gamemode.  They're currently admin only but I might make them obtainable by xenoarch later.
Adds powertools, also from /tg/.  CE starts with them in a new toolbelt that spawns in their locker, ported from (you guessed it) /tg/.
Changes welder sprites to look nicer, ported yet again from /tg/.  Modified the blue welder slightly so it can be the electric welder sprite.
Welders now have an eye safety modifier var, which makes certain welders more or less safe to use without protection.
Adds various sounds from /tg/, for tools and welders.

I would've made the power tools use device cells, but it would require another large refactor similar to this one to have each tool check if it can be used before using.

Welder and powertool sprites.  Note the glow for the experimental welder was changed to be bigger after this screenshot.
Welders


Anewbe:
Power cell refactor can be done later, imo.

Anewbe:
Not 110% convinced that having the path be /aaay is a good decision. Other than that, looks good.

PR #3678, opened 2017-08-04
Allows admins to write in blood as a ghost

Cameron653:
Someone suggested this would be useful for events, so might as well make a PR for it.


Cameron653:
Went ahead and tested this ~~Testing code? That's insanity.~~ and made sure everything worked. If you're not an admin it tells you the veil isn't thin enough and that you're not an admin. (Due to the way the check rights proc works. Not sure how to get rid of it.) 

Other than that, a user w/o admin rights was told the veil wasn't thin enough while an admin was able to use the proc as normal. 

Should be ready for merge.

Neerti:
I think the admin check proc has an argument to do the check silently.

Cameron653:
Looking through other examples where check_rights is used, I see this ``` check_rights(R_ADMIN, 0, user) ``` quite a lot. I think this might be the way to do it silently? Going to change the line and recompile and see if it works.

Neerti:
The proc definition is ``/proc/check_rights(rights_required, show_msg=1, var/client/C = usr)``

Protip: If you wanna find a proc/verb definition easily, hit control-h, select 'all files' and type in /proc/your_proc_name_here.

Cameron653:
Alrighty, I'll keep that in mind next time I'm looking for a proc. Usually just look for it with Github's search function.

Anyhow, went ahead and added , 0 to it so it'd silently fail. Tested it and it works as expected.

PR #3684, opened 2017-08-06
Reduces effectiveness of tabling as opening combat move.

Cyantime:
Tabling in its current form is an effectively instant stun with no real counters other then being about four tiles from any tables at all times, which can be extended indefinitely with minimal effort. This degenerates melee combat into "Grab, mash z, click table, push to ground, table again occasionally to reapply stun".

Tabling now requires a fully upgraded aggressive grab.

I was told to send this to here from a downstream fork.

### DIFF
--- a/code/modules/tables/interactions.dm
+++ b/code/modules/tables/interactions.dm
@@ -85,6 +85,8 @@
 			if(occupied)
 				user << "<span class='danger'>There's \a [occupied] in the way.</span>"
 				return
+			if(!user.Adjacent(M))
+				return
 			if (G.state < 2)
 				if(user.a_intent == I_HURT)
 					if (prob(15))	M.Weaken(5)
@@ -106,9 +108,9 @@
 				else
 					user << "<span class='danger'>You need a better grip to do that!</span>"
 					return
-			else if(!M.Check_Shoegrip() && do_mob(user, M, 5+(M.getarmor(BP_TORSO,"melee")/4)))
+			else if(G.state > GRAB_AGGRESSIVE || world.time >= (G.last_action + UPGRADE_COOLDOWN))
 				M.forceMove(get_turf(src))
-				M.Weaken(round(5-(M.getarmor(null, "melee")/20)))
+				M.Weaken(5)
 				visible_message("<span class='danger'>[G.assailant] puts [G.affecting] on \the [src].</span>")
 			qdel(W)
 			return
--- a/code/modules/tables/interactions.dm
+++ b/code/modules/tables/interactions.dm
@@ -106,9 +106,9 @@
 				else
 					user << "<span class='danger'>You need a better grip to do that!</span>"
 					return
-			else
-				G.affecting.loc = src.loc
-				G.affecting.Weaken(5)
+			else if(!M.Check_Shoegrip() && do_mob(user, M, 5+(M.getarmor(BP_TORSO,"melee")/4)))
+				M.forceMove(get_turf(src))
+				M.Weaken(round(5-(M.getarmor(null, "melee")/20)))
 				visible_message("<span class='danger'>[G.assailant] puts [G.affecting] on \the [src].</span>")
 			qdel(W)
 			return
--- a/html/changelogs/Cyantime-Tabling.yml
+++ b/html/changelogs/Cyantime-Tabling.yml
@@ -1,0 +1,36 @@
+################################
+# Example Changelog File
+#
+# Note: This file, and files beginning with ".", and files that don't end in ".yml" will not be read. If you change this file, you will look really dumb.
+#
+# Your changelog will be merged with a master changelog. (New stuff added only, and only on the date entry for the day it was merged.)
+# When it is, any changes listed below will disappear.
+#
+# Valid Prefixes: 
+#   bugfix
+#   wip (For works in progress)
+#   tweak
+#   soundadd
+#   sounddel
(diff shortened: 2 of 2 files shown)


Yoshax:
That seems like too much, to be honest. If you're going to nerf something you shouldn't immediately jump to hitting it in like three different places. Because then said thing becomes useless and never receives any buffs in the future when it is useless.

LorenLuke:
There's about a dozen better ways to do this.

HawkerTheGreat:
Yea, I disapprove 100% with this. Tabling is only difficult to escape when || A) You don't pay attention. (or) B) The MC is having issues. || And unfortunately the MC lag is usually harder to fix then anything else. The reason for combat skills like disarm and grab are to give player an advantage even though they're technically screwed if the other player has better gear. But a lot of these instant mechanics port from bay. 

I would suggest putting a ten second cooldown on tabling as to make it a feature that cannot be spammed constantly - if its reduced, you will simply never see it used. And I can remember quite a few features dying because they were nerfed when they did not need to be.

Cyantime:
@LorenLuke Could you give some examples then? The idea was to make this reasonably escapable if a person is paying attention and not incapacitated, without the only real answer being "have a macro for resist and smash it constantly", and to have someone geared up in armor less vulnerable to being instantly thrown onto a table than a random person.
@HawkerTheGreat I disagree that paying attention can keep you from being tabled. Grabbing, upgrading to aggressive, stepping towards a table (if needed), and clicking the table can be done in less then a second. Moving away during the process does not interrupt it, only resisting to break the grab (which itself is rng) is an actual option.

LorenLuke:
@HawkerTheGreat 
Being thrown into windows and against walls no longer hurts, so I sympathise with this line of thinking.

@Cyantime 
To nerf: 
1) https://github.com/Baystation12/Baystation12/pull/16699
2) Add an adjacency check for the victim such that if you table someone and they're not next to you OR THE TABLE when you click, such that, if they are not, it breaks the grab (rather than being able to table them after they ran four tiles away yet you still have the grab in your hand).
3) Give it an RNG failure chance based on the armor parameters, magboots, or what have you. Because magboots don't keep you from getting knocked over by carp, jumping off a station with them on, or being thrown. To me, they're magnetic enough to keep you from slipping across a surface in zero-g.
4) Make it a short 'do after' rather than instantaneous.

But if you're going to do any of these, pick exactly ONE. The combined effect of all of these means it'd be so useless, I'd rather just bash someone's knees in and wait for them to drop.

Cyantime:
I just threw Bay's fix in, it seemed pretty reasonable.
I also fixed the adjacency issue, because teletabling is pretty awful in any circumstance, and is more of a bugfix than anything.

code/modules/tables/interactions.dm
```diff
@@ -85,6 +85,8 @@
 			if(occupied)
 				user << "<span class='danger'>There's \a [occupied] in the way.</span>"
 				return
+			if(!user.Adjacent(M))
+				return
```
SpadesNeil: This part is a bugfix, not a feature change. Just going to point that out to everyone in protest.

code/modules/tables/interactions.dm
```diff
@@ -106,9 +108,9 @@
 				else
 					user << "<span class='danger'>You need a better grip to do that!</span>"
 					return
-			else
-				G.affecting.loc = src.loc
-				G.affecting.Weaken(5)
+			else if(G.state > GRAB_AGGRESSIVE || world.time >= (G.last_action + UPGRADE_COOLDOWN))
```
SpadesNeil: This prevents spamming it, which was also incredibly broken.

LorenLuke:
@Cyantime 
While I dislike the bay-fix in how long it takes to get to point viable time delay to table (indeed, that makes it almost on par with just forcing to the ground...), it doesn't outright neuter or RNG it. 

Tentative 👍 from me.

Fox-McCloud:
The real problem comes from lack of adjacency checks and the duration.

Without adjacency checks, you're reliant on grab's `process` to de-ference itself (completely) from the mob being grabbed, meaning...sometimes someone can be a few turfs away when you table them.

Once tabled, it's fairly easy to keep someone perma-stunned, since it lasts for a whopping 10 seconds--even if they're spamming resist like no tomorrow, it's still statistically likely you're going to keep them permanently stunned.

This isn't even factoring if if the person has cuffs/cablecuffs on them; if they table you once, that's *it* for you.

On servers like Paradise and Bay, where movement speed is super slow, this problem is exacerbated even more.

Cyantime:
Changelog added.

Anewbe:
What specifically does this PR do, as of this moment, @Cyantime ?

Cyantime:
Right now, this PR causes you to be unable to throw someone on a table without your aggressive grab being upgraded, at the same point that you would be able to increase to neck grab.

HawkerTheGreat:
I still think this is a bad merge, I would prefer this be simply put as a cooldown or a tile check. Besides, a ballistics vest and a sec hard hat isnt going to protect you from a blunt slam to a metal table or wall. You might not get a concussion, but... still gawt suplex'd.

HawkerTheGreat:
Wait, Tabling already requires aggressive, its just possible to quickkey the actions.

Anewbe:
So, based entirely on the code I can see, this is a welcome change. Merging on that basis.

Issue #3406, opened 2017-05-03
The AI can teleport food


PrismaticGynoid:
#### Brief description of the issue
While a food item is being cooked with one of the machines in the kitchen (Deep fryer, grill, oven, likely candy machine and cereal maker too), the AI can click on the machine to remove the food. This moves the food item straight to the AI's core.
#### What you expected to happen
The AI would be unable to interact with the machines, or for the food to be ejected on top of them.
#### What actually happened
The food item in the machine is teleported to the AI's core
#### Steps to reproduce
Be AI, have someone cook something with one of the kitchen machines. Click on it and have the food teleported directly to your core. 
#### Additional info:
 - **Server Revision**: master - 2017-04-29
3c8088015df26236fdb2734f9eda0f97d95ab665 
 - **Anything else you may wish to add**: Cyborgs/robots/drones can also probably use this to teleport food to them.

Atermonera:
I don't know why the attackby() is processing for AIs using cookers, but when the cooker ejects food, it tries to put it into the attacker's hand. If the attacker doesn't have a hand, it puts it at their feet. This can probably be fixed with a pretty simple adjacency check

Neerti:
For some reason this didn't get autoclosed so wew.

PR #3685, opened 2017-08-06
Gripper Fixes and Updates

Mechoid:
Grippers can now interface with smart-fridges, grinders and kitchen machinery. 
Testing was done incrementally as I fixed machine interactions.

New or current computers that need gripper interaction and call ..() in their attackby proc should be functional with them as well. I didn't touch the base machine instead, because it has no strict attackby, and I didn't want to touch it and cause un-forseen consequences.

I can guarantee there are more machines that need gripper interaction checks, that I didn't recall/fix, but I will fix them as I am informed when I get a new list going.

**Fixes: **
-Clerical borgs' grippers can interface with both ID console card slots again. The transition to NanoUI overlooked this.
-Borgs can't teleport food to them from the oven, fryer, grill, etc. in the kitchen anymore, despite the comedy involved.

### DIFF
--- a/code/game/machinery/kitchen/cooking_machines/_cooker.dm
+++ b/code/game/machinery/kitchen/cooking_machines/_cooker.dm
@@ -76,10 +76,21 @@
 	else if(istype(check, /obj/item/weapon/disk/nuclear))
 		user << "Central Command would kill you if you [cook_type] that."
 		return 0
-	else if(!istype(check) && !istype(check, /obj/item/weapon/holder) && !istype(check, /obj/item/organ))
-		user << "<span class='warning'>That's not edible.</span>"
-		return 0
-		
+	else if(!istype(check) && !istype(check, /obj/item/weapon/holder) && !istype(check, /obj/item/organ)) //Gripper check has to go here, else it still just cuts it off. ~Mechoid
+		// Is it a borg using a gripper?
+		if(istype(check, /obj/item/weapon/gripper)) // Grippers. ~Mechoid.
+			var/obj/item/weapon/gripper/B = check	//B, for Borg.
+			if(!B.wrapped)
+				user << "\The [B] is not holding anything."
+				return 0
+			else
+				var/B_held = B.wrapped
+				user << "You use \the [B] to put \the [B_held] into \the [src]."
+			return 0
+		else
+			user << "<span class='warning'>That's not edible.</span>"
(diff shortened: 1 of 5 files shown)


PrismaticGynoid:
[But then, how will the AI be able to steal food?](https://github.com/PolarisSS13/Polaris/issues/3406)

Atermonera:
Grippers call the attack sequence with wrapped when the sequence properly executes, you should be able to just return out of the attackby processing the gripper in such a manner as to not break the sequence, and it'll process wrapped

`/obj/item/weapon/gripper/attack(mob/living/carbon/M as mob, mob/living/carbon/user as mob)
	if(wrapped) 	//The force of the wrapped obj gets set to zero during the attack() and afterattack().
		force_holder = wrapped.force
		wrapped.force = 0.0
		wrapped.attack(M,user)
		M.attackby(wrapped, user)	//attackby reportedly gets procced by being clicked on, at least according to Anewbe.
		if(QDELETED(wrapped) || wrapped.loc != src.loc)
			wrapped = null
		return 1
	return 0`

code/game/machinery/kitchen/cooking_machines/_cooker.dm
```diff
@@ -91,8 +103,8 @@
 		for(var/mob/living/M in cooking_obj.contents)
 			M.apply_damage(rand(30,40), BURN, "chest")
 
-	// Not sure why a food item that passed the previous checks would fail to drop, but safety first.
-	if(!user.unEquip(I))
+	// Not sure why a food item that passed the previous checks would fail to drop, but safety first. (Hint: Borg grippers. That is why. ~Mechoid.)
+	if(!user.unEquip(I) && !istype(user,/mob/living/silicon/robot))
```
Atermonera: Did you test this against #3406 ?

Mechoid: That's for putting food into a machine, and AI's don't have hands or grippers to do that with. Taking food out, however, is a proximity check, so that fixes the referenced one, but I can test it again.

Mechoid:
When I just returned out, it didn't seem to actually work, for some reason. I can change it again and check.

Anewbe:
Tested?

Mechoid:
Tested everything as I went, and then after all was said and done as well.

PR #3688, opened 2017-08-08
Assault rifle tweaks and fixes

SpadesNeil:
- STS-35 no longer has "This one is unmarked" because it literally means nothing useful.
- Assault Carbine (7.62mm) name has been changed to Bullpup Rifle.
- Bullpup Rifle starts with only 10 rounds as opposed to 20, making it slightly less useful than the Z8 marksman rifle.
- Fixed the Bullpup Rifle not using the correct sprites, hopefully.
- Bullpup Rifle's 3 round burst is nerfed to 2 rounds.

**The Bullpup Rifle is no longer an admin abuse gun** unless you give it 20 round mags, which are currently only admin spawnable.


SpadesNeil:
@Anewbe since the Assault Carbine (now Bullpup Rifle) is no longer admin abuse gun, feel free to toss it into some kind of traitor loadout. I have no idea what the cost would be so I leave that to your judgement.

PR #3690, opened 2017-08-08
Makes the surgery drill have the correct icon state

Cameron653:
See: Title. It's "drill" not "surgery_drill" in the surgery.dmi


Neerti:
My bad.

PR #3692, opened 2017-08-09
Better Wheelchairs

PrismaticGynoid:
Wheelchairs can now be collapsed by click-dragging them to you, much like rollerbeds. You can then pick them up. They are, however, too large to stuff into a backpack.

These new collapsible wheelchairs are now available in the loadout, under utility, for 4 points. They're also color-customizable now. No more plain white wheelchairs available only by missing feet.

The loadout wheelchairs will spawn with you buckled into them, just like how it works when you spawn without feet, except that the loadout wheelchairs can be more colors than just white.

You're also now not forced to leave your wheelchair behind when leaving the colony via the gateway/elevator.

[Have some colorful wheelchairs.](https://i.gyazo.com/f5ffc0b02a4f741e0acd0357c97b928d.png)


Anewbe:
How much have you tested this?

PrismaticGynoid:
I tested each change in as many relevant situations as I could think of.

PR #3695, opened 2017-08-10
Allows guns in suit slots to appear on your back

SpadesNeil:
So, fun fact.

If you put an item in your suit belt slot, it uses sprites from **belt_mirror.dmi**

Now with this in mind, I copied and mirrored all the gun sprites from back.dmi and added them to belt_mirror.dmi instead. I did this before on oooooollllllllld vorestation from like 2015. It actually worked very nicely, and looks really cool. It also no longer lets people with armor or hardsuits walk around with concealed shotguns stuffed in their armor.


SpadesNeil:
image

SkowronX:
Gee bill, two shotguns?

Anewbe:
How does this interact with actual back sprites?

SpadesNeil:
Suit slot sprites are always below the normal back sprites. I have yet to find a case where this looks bad. It's actually much cooler looking to have, say, a backpack, overlayed across a shotgun. It's like you tucked it in.

SpadesNeil:
Make sure that if any other back sprites are added that go across the back like this, that they are consistent with what is already in the file.

PR #3706, opened 2017-08-12
Removes hardsuit descriptions that name the station.

SpadesNeil:
It's absurd for this to appear on Syndie suits, or NASA suits, or CentCom suits, or Vox suits, any other suits that don't explicitly belong to the station.

What if you wanna add SolGov suits one day? This will screw that up too.

### DIFF
--- a/code/modules/clothing/spacesuits/spacesuits.dm
+++ b/code/modules/clothing/spacesuits/spacesuits.dm
@@ -62,10 +62,6 @@
 
 	var/list/supporting_limbs //If not-null, automatically splints breaks. Checked when removing the suit.
 
-/obj/item/clothing/suit/space/New()
-	..()
-	desc += " \"[using_map.station_short]\" is written in large block letters on the back."
-
 /obj/item/clothing/suit/space/equipped(mob/M)
 	check_limb_support(M)
 	..()


Anewbe:
Here's an idea: Add a var to the space suits called uses_station_name. If true, then it calls the proc. If false, it skips it.

SpadesNeil:
I disagree with that, as what if you wanted to spawn suits for an event that didn't necessarily belong to the station? Then you have to manually edit the item with VV to remove the description. Plus, nothing else in the server has code or descriptions like this. It's largely just annoying.

SpadesNeil:
Let me put it this way.

Are ERT suits going to have this on them?
Or Merc suits?
Or Breacher suits?
Or SifGuard suits when eventually those are added?

What if I want to use suits we already have for SifGuard suits? That wouldn't make sense if they had the "Northern Star is written in large block letters on the back" in the description when they got spawned in. Do we make a separate subtype for all of the SifGuard versions that has the `uses_station_name` var set to false? Or do we make the admin spawning them manually delete the description?

This is not a good feature. It's more inconvenient than anything. If a coder wants a suit to have a description on it saying it's property of some station even though literally no other item in the entire code has anything remotely similar, then let them do that themselves.

SpadesNeil:
Or what about--

... oh. It got merged. Well I guess I made my point. Cheers.

PR #3708, opened 2017-08-12
Fixes a-help being dis-colored.

killer653:
Apparantly you can't use bold between two different ```<font color>```'s

This fixes the text showing up an incorrect color. Also removes a : because there were two :'s

Tested and can confirm it works.


Atermonera:
<font color='blue'>[get_options_bar(mob, 2, 1, 1)][ai_cl]:</b> [msg]</font>"

It was trying to close the bold tag within the font tag, which is a nono in html

PR #3710, opened 2017-08-14
Weak attacks no longer can damage windows

SpadesNeil:
Attacks with a force less than 5 for normal windows, or 10 for reinforced windows, no longer does any damage to the glass.

The exact number can change, just the goal of this is to stop (especially prisoners) from slapping a pane of glass with a pair of shoes until it shatters. Go ahead. Smack a pane of reinforced glass as many times as you want IRL with your sneaker. You're generally not going to damage it in the course of a day.

We already can't damage windows by punching them for obvious reasons. If your weapon does around the same or less damage than a punch, it shouldn't hurt the window either.

### DIFF
--- a/code/game/objects/structures/window.dm
+++ b/code/game/objects/structures/window.dm
@@ -164,7 +164,7 @@
 		update_verbs()
 		update_nearby_icons()
 		step(src, get_dir(AM, src))
-	if(tforce < 10)
+	if(tforce < 5)
 		return
 	take_damage(tforce)
 
--- a/code/game/objects/structures/window.dm
+++ b/code/game/objects/structures/window.dm
@@ -164,6 +164,8 @@
 		update_verbs()
 		update_nearby_icons()
 		step(src, get_dir(AM, src))
+	if(tforce < 10)
+		return
 	take_damage(tforce)
 
 /obj/structure/window/attack_tk(mob/user as mob)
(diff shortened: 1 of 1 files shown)


Cyantime:
This only applies to thrown objects in it's current state, you want the attackby() section for normal weapon strikes.

SpadesNeil:
I'll fix that shortly, then. I had a feeling I was missing something.

SpadesNeil:
There we go.

SilverTalismen:
This is great! Before it was a real pain in the glass to keep most prisoners.

SpadesNeil:
>This is great! Before it was a real pain in the glass to keep most prisoners.

Go back to your hole and stay there until I get back. :T

Neerti:
@Anewbe Does this still need more staff input?

SpadesNeil:
Keep in mind, 5-10 was just a number I came up with. Change it to whatever seems reasonable. I just don't like prisoners being able to slap the shit out of windows with their shoes.

SpadesNeil:
So what's the word on this?

Woodratt:
There was comments about maybe allowing regular windows still be smashed but that was just talk. It seems to be fine with folks staff end. Sorry that this is taking so long.

chaoko99:
for the record, a drunken Teshari can still smash a window by running into it. Kinda weird.

PR #3715, opened 2017-08-16
Replaces status frame sprite.

Belsima:
images
Old left, new right.

Taken from CEV-Eris and a random folder on my desktop. I mashed the sprites together. Also fixes a minor bug with pens because why not.


Anewbe:
Don't mix and match PRs, please. One PR, one change is sort of the guideline.

Belsima:
It was honestly an accident because I was doing multiple things at once and pushed the pen file by accident, but it's a 1 line bugfix so I figured it wasn't worth undoing.

code/modules/paperwork/pen.dm
```diff
@@ -79,7 +79,6 @@
 
 /obj/item/weapon/pen/reagent
 	flags = OPENCONTAINER
-	slot_flags = SLOT_BELT
```
SpadesNeil: Why'd this get removed? I know it's accidental but still.

Belsima: So that they can go into the ear slot, like regular pens.

SpadesNeil: Oh ok.

PR #3726, opened 2017-08-21
Woodrat breaks everything with floor decals

Woodratt:
Rough port of

https://github.com/VOREStation/VOREStation/commit/a8c9bb786aabedfe715b9ab354a68cd32327e12d

Unsure if I even did it correctly, but seems to function in the brief tests done, no compile errors.

In preparation of porting the massive number of floor decals for the new map.


Anewbe:
Gonna give this a try, can't specifically see anything it breaks.

PR #3732, opened 2017-08-23
Replaces grinder sprites.

Belsima:
images
images

Now it looks like it might actually grind something. Note that the old sprites are not deleted. Taken from the Eris.


Cyantime:
That new sprite looks like a cage or something? It doesn't look like a gibber, and the old one looks like a big sausage grinder and fits perfectly.
Also 3/4 perspective is awful.

Belsima:
Yeah, I wasn't sure on this one. Only decided to toss up a pull because someone suggested porting it.

Atermonera:
I think the issue with the new one is it's lack of a distinct spout. I can see that there're coils for grinding up the meat, but it doesn't look like meat should come out of anywhere

Belsima:
I think it just grinds it out and drops it right there- which also makes more sense given it spits out gibs rather than a tiny bit of concentrated flesh.

Cyantime:
The gibber forcefully throws the meats it makes to the west of it, which is why it has the spout on that side.

PR #3734, opened 2017-08-23
Replaces robot storage pod.

Belsima:
images
images

Now rather than a stolen guest console and a wider suit storage, it's a CEV-Eris guest console and an FTL-13 recolored suit storage.


Cyantime:
Why would you get rid of a unique sprite and replace it with a palette swap of something else?

Belsima:
It's not a unique sprite. They literally took the original and copy + pasted it to widen it.

Atermonera:
Can't look at the .dmi from my phone, but in case you totally removed the old sprites, common (or at least common in days of yore) practice was to rename the old sprites to x_old or something, so that they're still in the code.
If you did just rename the old sprites, then disregard.

@Cyantime, it's mostly just matter of preference. Both sprites involved are minor modifications to an existing sprite, it's just the nature of how they were altered that differs

Belsima:
I'll add the _old ones, smart idea.

Cyantime:
I really feel that the wider version fits better, it looks like a big box that you shove a whole robot into, instead of just a blue suit storage unit.

Belsima:
That's fair enough. If anyone else puts in input, i'll just replace the console.

Belsima:
Actually, I agree tbh. Gonna shut this down.

Issue #3679, opened 2017-08-04
Can't Repair Solar Tracking Computers


Nalarac:
#### Brief description of the issue
If the computer console used for the solar arrays has its glass broken, it can't be restored to its previous condition.
#### What you expected to happen
To be able to repair the computer
#### What actually happened
It ended up as a frame that I couldn't interact with
#### Steps to reproduce
1.) Damage the solar tracking computer (ex. meteor explosion hits it)
2.) Remove broken glass
3.) Try to slot new glass in and notice it doesn't work
4.) Try to dismantle the frame instead and notice it doesn't work
#### Additional info:
 - **Server Revision**: master - 2017-08-04  127ff441f160ba0bab2f6b946cbef29611415a71

PR #3736, opened 2017-08-25
~Solar Computer Fix~

MisterLayne:
This fixes the problem where if a Solar Panel Control Computer were to break, and the player removed the glass, the computer would basically become a physical ghost that you couldn't interact with.

### DIFF
--- a/code/modules/power/solar.dm
+++ b/code/modules/power/solar.dm
@@ -302,7 +302,7 @@ var/list/solars_list = list()
 		M.unset_control()
 	if(connected_tracker)
 		connected_tracker.unset_control()
-	..()
+	return ..()
 
 /obj/machinery/power/solar_control/disconnect_from_network()
 	..()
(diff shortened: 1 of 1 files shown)


Neerti:
A protip, you can use ``#`` and then a few numbers to automatically reference a specific PR or Issue.  If you write, in this case, Fixes #3679 , in the PR description, GitHub will automatically close the issue when this gets merged.  An easy way to find the correct number is to type ``Fixes #`` and the type 'solar' and it should be one of the first options.

SpadesNeil:
`Fixes #Sola` -> Fixes #3736

Neerti:
@MagmaRam Bugfixes never needed changelogs back in the day and I don't see why they should now, since backend changes that players aren't likely to notice don't really need a changelog, and people generally expect that a feature is actually functional when using them, and holding up a bugfix to perpetuate an inefficient, manual thing instead of using a lovely automated system like GitHub Pulse.

So I'm gonna merge this so engineering isn't SoL when a solar computer gets exploded.

PR #3737, opened 2017-08-26
Adds More Ways to Obtain Power Tools

Nalarac:
- Added the hand drill and jaws of life to the protolathe
- Removed titanium from the hand drill and jaws of life
- Syndicate toolbox now comes with power tools


Yoshax:
I don't suppose the sounds on the new tools, like the drill, could be made quieter? Or possibly less abrasive? The drill is very irritating.

MagmaRam:
That'd be something for another PR.

Neerti:
Best way to adjust the volume would be to edit the sound files in an external program since the power tool sounds get played at generally 50 or 75 volume inside hundreds of ``playsound()`` calls.

PR #3740, opened 2017-08-26
Replaces solar panel sprites.

Belsima:
Taken from a russian paradise branch. I don't believe it's the origin, being a nazy pony ERP server without a name, but still.

images


Woodratt:
Does it keep the old sprites listed as old? Or outright removes them?

Belsima:
Removes them. Would you prefer them marked as old? I assumed in the case of the tracker and sp_base at least, it was a direct upgrade.

Woodratt:
Leave them in the dmi marked as old, we may find a use for them in the future.

Belsima:
Alright.

Belsima:
Done.

PR #3746, opened 2017-08-27
Revert "Reshades APC sprites."

Neerti:
The PR that added these got suddenly merged and to my knowledge no discussion was had about doing so, on the staff side.

In my opinion, the sprites for the APC were fine as is, and I'm getting rather annoyed at the newest fad of 'lets change all the sprites for no reason', but the 'emagged' icon_state is way, way too dark and it looks like it was darkened by an automated process.


Belsima:
I could brighten that up if you want- I'm simply going around and finding sprites I prefer over the current ones and porting them. No fad of 'change everything for fun'.

PR #3750, opened 2017-08-28
Rainbow gloves.

MisterLayne:
Those pesky rainbow gloves now actually LOOK like rainbow gloves.


Anewbe:
Is it intentional that this only affects Teshari?

MisterLayne:
No. No that is not intentional >->

PR #3751, opened 2017-08-29
Fixes 'The data cable retracts rapidly into its spool.' times a billion.

Mechoid:
On the tin. Also makes the cable's Destroy() cleaner.
There are no time police.


MisterLayne:
I think I cried when I saw this.. Thank you.

PR #3762, opened 2017-09-03
Brings back the aliens (With hitches).

Mechoid:
Returns xenos as they were before being removed, minus face-huggers until we can get a suitable replacement, for admin event usage.

Something Something Git never pulls when I tell it to pull the first time.


Mechoid:
Dealing with Travis in the morning. Contains unhealthy amounts of \red and \b.

Cameron653:
It's not only the \red and \b but also
```
The command "(! grep -En "<\s*span\s+class\s*=\s*('[^'>]+|[^'>]+')\s*>" **/*.dm)" exited with 0.
1.70s$ awk -f tools/indentation.awk **/*.dm
code/game/objects/effects/alien/aliens.dm :  /obj/effect/alien/egg/proc/GetFacehugger() // Commented out for future edit.
The command "awk -f tools/indentation.awk **/*.dm" exited with 1.
```
Even though it's commented out Travis wants it indented correctly.

Mechoid:
Already dealt with that as well, before moving onto the \red / \bs

Mechoid:
I lied, there were some \blue too.

Anewbe:
Is this complete?

SpadesNeil:
sddefault

PR #3769, opened 2017-09-06
Brain transplants, synthetics, and languages

### LINKED ISSUE #2832: Sign Language and hands
It's possible to use sign language while lacking hands.
Lacking hands would cause some sort of issue with using sign language.
#### What actually happened
Handlessness has no effect on being able to use the language.
#### Steps to reproduce
1.) Have sign language.
2.) Remove hands.
3.) Attempt to use sign language with your arm-nubs.
4.) Be perfectly able to do so.
#### Additional info:

PrismaticGynoid:
Brains now retain languages from the mob they were removed from, rather than forgetting them. Languages are also retained when a brain is transplanted into a new body.

AIs, as well as cyborgs/robots/drones, will load languages from your currently-active preferences when they spawn (just like humanoid mobs do). Brains removed from cyborgs/robots/drones will retain their original languages regardless of the current module.

Brains in MMIs can now speak EAL if they know it, and additionally fixes #2832 - sign language will no longer be usable when lacking both hands. Also fixes silicons being told the wrong language prefix by the known languages window, as it was still showing ':', leading to much confusion among new borg players.

Known issues:
AIs who can use sign language can use it even without a holopad - I'm not sure how to have it check for that. This could be seen as a non-issue, as signs could conceivably be displayed on the AI's screen.

AIs signing over holopad will also have the language verbs overridden by the synth speech verbs (states/queries/declares). I'm pretty sure this has something to do with the already-existing bug that all languages can be understood by everyone when an AI speaks them over holopad.

### DIFF
--- a/code/game/machinery/computer/ai_core.dm
+++ b/code/game/machinery/computer/ai_core.dm
@@ -176,6 +176,8 @@
 					var/mob/living/silicon/ai/A = new /mob/living/silicon/ai ( loc, laws, brain )
 					if(A) //if there's no brain, the mob is deleted and a structure/AIcore is created
 						A.rename_self("ai", 1)
+						for(var/datum/language/L in brain.brainmob.languages)
+							A.add_language(L.name)
 				feedback_inc("cyborg_ais_created",1)
 				qdel(src)
 
--- a/code/game/objects/items/robot/robot_parts.dm
+++ b/code/game/objects/items/robot/robot_parts.dm
@@ -208,6 +208,9 @@
 
 			O.job = "Cyborg"
 
+			for(var/datum/language/L in M.brainmob.languages)
+				O.add_language(L.name)
+
 			O.cell = chest.cell
 			O.cell.loc = O
 			W.loc = O//Should fix cybros run time erroring when blown up. It got deleted before, along with the frame.
--- a/code/modules/mob/language/generic.dm
+++ b/code/modules/mob/language/generic.dm
@@ -121,3 +121,7 @@
 	colour = "say_quote"
 	key = "s"
 	flags = SIGNLANG|NO_STUTTER|NONVERBAL
+
+/datum/language/sign/can_speak_special(var/mob/speaker)
+	var/obj/item/organ/external/hand/hands = locate() in speaker //you can't sign without hands
+	return (hands || !iscarbon(speaker))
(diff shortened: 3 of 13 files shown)


Mechoid:
I believe posis / drone circuits having binary was in part due to the fact they can't get a radio version like the MMI. I'm not a dev, but as a borg player, I'd really appreciate that being kept. Being able to talk to the station synths stopped more than one person just leaving my circuit somewhere. Never really had that problem as an MMI since people seemed more keen on keeping the fleshy bit alive and around than the box or plastic square.

PrismaticGynoid:
You're a security robot, chasing after an intruder who has stolen valuable tech and the spare ID. You corner them in maintenance, and a shootout begins. As you find yourself outmatched, you try to call for help, but realize you can't because your radio and binary communication device were destroyed. The intruder takes this as an opportunity to acquire some more valuable tech - namely, your brain - and takes it with them to their ship. However, as you've now been removed from your body, you suddenly have access to the binary channel again, even though you couldn't use it before without a working binary communication device. You secretly call for help in a way the intruder can't stop or detect, and reveal the location of the intruder's ship, leading to security finding them.

You're a traitor AI, and you've just murdered the engineer before they can pose a threat to you - they just found out that something's up with your laws, and easily have the skills to defeat you if you let them. If you hadn't shut off telecomms before sealing the room and suffocating them, they would have surely revealed your malicious intent. Unfortunately, as they had their suit sensors on and a death alarm installed, they get found pretty quickly - you can't stop them without being revealed, but that's okay, because they aren't supposed to remember how they died anyway. Except the engineer is a drone, and as soon as they get removed from their body by the roboticist, they suddenly have the ability to listen in on your secret commands to your linked cyborgs. They didn't have that before, but now they've alerted security, and nothing goes well after that.

You're a changeling, disguised as a roboticist. The warden, a positronic, just captured your changeling siblings, but not without taking heavy damage, so they come to you for repairs. You use this opportunity to stealthily kill the officer so they won't be a threat, and you can take their ID. However,  since you don't want to just take them out of the round, you take out their brain so they're still able to roleplay. Well, that was a mistake, because they immediately call for help using binary, something they had no access to before, and you have no way of knowing this happened until the full might of security bears down on you the second you step out into the hall.

I've seen all of these happen before. I get that being left alone as a brain can be really unfun, but so is having all of your antagonist plans blown because removing someone from their body somehow gives them access to a secure channel that can be used without anyone knowing, and can't be stopped even with telecomms down. That's why I wanted to remove binary from brains, but I can undo that change if it's really not wanted.

Anewbe:
Will review this soon, sorry for the delay.

Anewbe:
You've tested this thoroughly, yes?

PrismaticGynoid:
Yes, and the only issues remaining are the two mentioned - I honestly have no idea how to fix those.

Anewbe:
God help us all, merging.

PR #3771, opened 2017-09-07
Makes passive vents constructable

Cyantime:
Also fixes the mingle_with_turf proc that they use.


Anewbe:
Have you tested this?

Cyantime:
Yes, of course.

PR #3779, opened 2017-09-08
Must now be on help intent to pry floortiles.

Belsima:
https://github.com/vgstation-coders/vgstation13/pull/12681

Too often are we victims of combat via crowbar destroying the flooring.


Anewbe:
Don't see the need.

Belsima:
It's because if you're in a fight with a crowbar and misclick, you often tear up the floor. I see it very often.

Anewbe:
Yeah, I'm aware, but I don't see the need to remove that, honestly.

Belsima:
It doesn't seem quite right to be bashing someone in the face and then leaning down to pry up a floortile mid-fight.

PR #3785, opened 2017-09-08
Added sounds for changeling absorbtion.

Belsima:
Mildly gross and unsettling, which makes sense given what a changeling is and the flavor text for most of their abilities. From /vg/.


Anewbe:
I can't actually preview sounds, and the account that made this PR has been closed. Closing this PR.

PR #3790, opened 2017-09-08
Replaces bookcase sprites.

Belsima:
images

The left were apparently from RPG maker. The right are from /tg/. Thanks, /tg/.


SpadesNeil:
Oh yea RPG maker would not appreciate us using their assets.

Belsima:
Most of it is licensed purely for use in RPG maker.

PR #3799, opened 2017-09-09
Adds a non-gendered biological option for prometheans.

Belsima:


Anewbe:
Wouldn't it be Neuter?

Belsima:
Neuter sets the examine to "It", doesn't it? That doesn't sound right.

Anewbe:
*shrug

Belsima:
Yeah, neuter is 'it'. I'll wait for a statement from @Mechoid or @Kholdstare34 if you want it to be neuter, but otherwise I'll leave it as plural.

Anewbe:
Querying them.

Mechoid:
Can't speak for Kholdstare, but I have no issue with adding the Plural or Neuter option to Prometheans, if both are wanted. I can see the use of having 'It' instead of 'They', 'He', or 'She', if only based on the status of them.

Belsima:
Added neuter.

PR #3800, opened 2017-09-09
Replaces syringe sprites.

Belsima:
images

From Paradise.


Wickedtemp:
Would the colors of the chems inside the syringe still affect the color of the syringe sprite?

Belsima:
Yes. It would. The current sprites have that blue crap for whatever reason as well. Not sure what they're used for or if they're used.

Anewbe:
I'm actually not a huge fan of this. Seems obnoxious.

Belsima:
What's the difference? They have markers and the draw/inject things are both more obvious and smaller.

Anewbe:
Colors.

Belsima:
The colors much more resemble an actual medical syringe rather than a 60 year old steel one.

Anewbe:
I mean the bright blue and red lettering.

Belsima:
I prefer it, since the current drab grey 'draw/inject' is something I often mix up, leading to a game of swapping between the two to try to extract the right amount. I can revert that, of course, if you want.

Anewbe:
Like I said, I find the lettering to be overly colorful.

Belsima:
So, should I revert it, or make only part of it colored, or what?

Belsima:
images Old big grey words returned.

Wickedtemp:
As someone who only plays Medical and would be dealing with this ALL THE TIME and would be the one primarily affected by this...

I don't mind the coloring of Draw/Inject at all. I'm fine with either, but I'm not going to freak the fuck out over the words being colored. It might even be useful since it's more visible.

chaoko99:
The existing syringesprites use an overlay. Taht's why they're blue.

Anewbe:
Closing this because the account is no more.

PR #3807, opened 2017-09-09
Fixes lit welders not igniting fuel tanks.

Cyantime:


Anewbe:
Tested, including refueling with an unlit welder?

Cyantime:
Of course.

PR #3815, opened 2017-09-11
Makes Prometheans process various Toxins differently. Step One.

Mechoid:
-Prometheans now take less toxin damage from base toxins, and gain nutrition from them. This includes fertilizers.
-Slime-jelly will now have a high chance to act as a general regenerative chemical for Prometheans, with the side effect of color distortion similar to mindbreaker. They do not have the chance of being poisoned by it, since it is part of their base body composition.
-Mindbreaker and Psilocybin now are more potent when used on Prometheans due to the entire surface of their body essentially acting as a sensory organ when required. That's a large amount of data to mess with.
-Soporific and Chloral Hydrate require greater amounts when used on, and no longer put Prometheans to sleep, instead keeping them inhibited (small chance to get up with confusion) or prone for the same duration respectively, with constant vision blurring and occasional deafness to act as an inhibited sensory state in both. This state is extended greatly for chloral hydrate.
-Tiny spacing changes to make reading the blocks easier.

Step one of probably only one of chemical variation-izing Prometheans. Probably.

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Toxins.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Toxins.dm
@@ -361,16 +361,17 @@
 /datum/reagent/slimejelly/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
 	if(alien == IS_DIONA)
 		return
-	if(prob(10) && alien != IS_SLIME) //Partially made of the stuff. Why would it hurt them?
-		M << "<span class='danger'>Your insides are burning!</span>"
-		M.adjustToxLoss(rand(100, 300) * removed)
-	else if(prob(40) || (alien == IS_SLIME && prob(75)))
-		if(alien == IS_SLIME) //Production of the stuff naturally slows as age increases. Added artificially later, it's like a stimpack. Though not all that effective if only one limb has the damage.
+	if(alien == IS_SLIME) //Partially made of the stuff. Why would it hurt them?
+		if(prob(75))
 			M.heal_overall_damage(25 * removed, 25 * removed)
 			M.adjustToxLoss(rand(-30, -10) * removed)
 			M.druggy = max(M.druggy, 10)
 			M.add_chemical_effect(CE_PAINKILLER, 60)
-		else
+	else
+		if(prob(10))
+			M << "<span class='danger'>Your insides are burning!</span>"
+			M.adjustToxLoss(rand(100, 300) * removed)
+		else if(prob(40))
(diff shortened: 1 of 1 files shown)


Anewbe:
Thaaaaat first point is contentious.
Fourth point probably needs discussion too, now that I'm looking at it again...

Wickedtemp:
If Slime Jelly will act as a healing reagent, would it still be locked behind Xenobio or will Chemistry obtain some manner of making small amounts of it? I know there's a recipe for it that uses lithium, sugar and a LOT of phoron to make 5u of Slime jelly, it might not be a bad idea to consider.

Mechoid:
Sopo and chloral already leave them able to see, since they don't have eyes. The only actual in-game change there is for the blurriness, chance to stand up confused for a short moment in sopo, and the ability to hear things sometimes in both. Were there a way to fumble the sentences with asterisks as old-lang did, I would use that instead of occasional deafness/hearing chance.

On the Slime Jelly, I was just told that new-old Xenobio cores don't produce it in the way old-old and old ones did, so I might look into making it 'synthesizable'.

Wickedtemp:
I'd suggest taking a look at Virgo's slime reaction code. It's basically the old xenobio slime abilities. Slime jelly's recipe is 50 Sugar, 50 Lithium, 20 Phoron = 5u Slime Jelly. IIRC most reactions need 10u of slime jelly, so it isn't like this is a cheap thing to do. Without getting extra phoron, you'd be left with enough phoron for maybe 10u of slime jelly each round.

Mechoid:
Slime Jelly, right now, has no medicinal use reaction or otherwise. ..I may look into that for a future thing because I'm a terrible person.

Wickedtemp:
Well, you said in the initial post that it could act as a medicinal chemical for prometheans. Assuming it's useful enough to bother producing in the off-chance it'll be needed, then it'll be used. It's also a fun traitor reagent.

code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Toxins.dm
```diff
@@ -358,11 +361,17 @@
 /datum/reagent/slimejelly/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
 	if(alien == IS_DIONA)
 		return
-	if(prob(10))
+	if(prob(10) && alien != IS_SLIME) //Partially made of the stuff. Why would it hurt them?
```
Anewbe: So rather than doing this, put the IS_SLIME check under the IS_DIONA check and put the normal checks in a else?

PR #3816, opened 2017-09-11
Tweaks instability

Neerti:
Instability should now feel more responsive and less sluggish to go away.  Instead of it being removed at a semi-random and slow rate, it now has a semi-defined half life of about 46 seconds (IE 100 instability will be around 50 in 46 seconds, then 25 in another 46 seconds, and so on).  The radiate pulse now occurs every Life() tick as well, and by extension the light radius update, so you should be able to see your purple halo adjust faster.

This unfortunately necessitated tweaking to the strength of instability events.  Instability events should happen more often to compensate for it going away faster.  It may need to be tweaked further.

Instability Glow (IE radiating) now radiates the instability decayed in the Life() tick instead of a third of the current instability.  This means people have more time to get out of the purple light before they start glowing themselves.  Two glowing people tend to equalize their instability if adjacent for awhile, while a third person can cause runaway instability, as normal.

Glowing now starts at 10 instability instead of 30.

Technomancer cores show the instability delta in their stat panel.

Message for standing in purple made bigger and purple-er.

Fixes bug with certain spells which waited for the effect to go off before qdeling the hand item.

Made the Glow a deeper purple.

Fixes leftover Xenobio tests in example map defines.

### DIFF
--- a/code/game/gamemodes/technomancer/core_obj.dm
+++ b/code/game/gamemodes/technomancer/core_obj.dm
@@ -170,7 +170,8 @@
 	if(core && statpanel("Spell Core"))
 		var/charge_status = "[core.energy]/[core.max_energy] ([round( (core.energy / core.max_energy) * 100)]%) \
 		([round(core.energy_delta)]/s)"
-		var/instability_status = "[src.instability]"
+		var/instability_delta = instability - last_instability
+		var/instability_status = "[src.instability] ([round(instability_delta, 0.1)]/s)"
 		stat("Core charge", charge_status)
 		stat("User instability", instability_status)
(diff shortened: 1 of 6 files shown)


Wickedtemp:
The instability mess was a little irritating to handle. On-screen, I didn't see any change in color aside from the occasional violet sparks on my character. Evacuating from the area was a little more difficult than it should have been in my opinion, everybody there ended up either in critical condition or nearing it, save for the two crew that died. It was kinda just an invisible death aura that impeded escape by knocking you down and putting you in paincrit.

I'm happy to see it adjusted.

Yoshax:
Awesome work :)

Anewbe:
For (super)Science!

PR #3820, opened 2017-09-12
Adds medical tape

Cameron653:
- Adds medical tape for medical to block off areas when treating patients
- Adds medical tape to paramedic and CMO lockers.

This makes it so paramedics can block off an area when they are treating patients, so they aren't bumped into while injecting chemicals or applying medications.

Tested and it just shows up as plain, white tape when put down, as intended.
https://i.imgur.com/gNp90Sb.png

### DIFF
--- a/code/game/objects/items/weapons/policetape.dm
+++ b/code/game/objects/items/weapons/policetape.dm
@@ -59,6 +59,18 @@ var/list/tape_roll_applications = list()
 		hazard_overlays["[WEST]"]	= new/image('icons/effects/warning_stripes.dmi', icon_state = "W")
 	update_icon()
 
+/obj/item/taperoll/medical
+	name = "medical tape"
+	desc = "A roll of medical tape used to block off patients from the public."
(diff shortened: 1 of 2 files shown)


Wickedtemp:
Huh... I like this, but I have to wonder how often it'd be used. Generally, a paramedic's job is to pick up the wounded and get them out of there and back to Medbay as soon as they're stable enough for transit. 

Though... We COULD lay out medical tape within medbay, hopefully to dissuade Sec/Command who want to watch patients get treated :p

MisterLayne:
I don't think medical tape would get used often, security tape for crime scenes is good enough, though it would be nice to use (in agreement with Wickedtemp's comment) in the medbay to keep those pesky people from barging in and getting in the doctors' way.

Cameron653:
Yeah. It's not anything revolutionary, but it's just a nice little thing for medical to say "Hey, we're working here. Don't come in and bump into us."

Issue #3817, opened 2017-09-11
Autoheal is a little bit too much


Wickedtemp:
#### Brief description of the issue
Right now, you can autoheal yourself from 110 damage.
#### What you expected to happen
Ended up taking around 90 burn damage from a mix of fire and acid. You should need medical attention for that.
#### What actually happened
Autoheal took care of it in about a minute.
#### Steps to reproduce
Damage every limb to around 9 brute/burn damage. This adds up to 108 damage. You'll autoheal from this. 
#### Additional info:
 - This is because of autoheal working on each limb independently. If the limb takes 10 damage, the autoheal process stops and has to be restarted by an advanced trauma/burn kit, which then resumes autoheal for that limb and raises the cut-off damage to 50 instead of 10. 

This reduces the effectiveness of fire, since the bulk of the damage is dispersed and you're autohealing as you're being burnt. It also means if someone pumps you full of acid, unless the dose is high enough to kill you outright, you're going to not only survive, but autoheal from all of it. 

As a medical player, this kinda takes action out of my game, and watching someone magically heal themselves all the way from severely damaged to green is a little odd to explain ICly. As a solution for this, I'd suggest either reducing the cut-off point per limb from 10 to 5. Trauma/Burn kits restarting it at 50 seems fine enough to me, I'd personally keep it between 20-60. As an alternative, if 5 damage per limb is too low, it could be modified to have a cut-off of 20 *total* limb damage, instead of 10 damage per limb. I think that makes a little more sense.

Wickedtemp:
Another alternative is to give autoheal a delay. Perhaps maybe 60 seconds or so after damage is received, that way you aren't autohealing fire burns *as* it's burning you.

Atermonera:
Tox damage isn't autohealed, so I think fire is about the only damage source that will regularly cause damage to be distributed over a million tiny burns than in one single wound

Wickedtemp:
It's burn/brute, so a few causes of widespread burn and brute damage would be: Acid injections, atmos fire, having your character itself ignite, brute damage from arithrazine... I'm sure there are others, but those are the ones off the top of my head. I think the best course of action for this is to just add a delay, so if you're taking damage, you aren't autohealing as you're being injured.

Wickedtemp:
Closed, since it's getting a much needed delay.

PR #3825, opened 2017-09-13
Makes Fire Hot Again

Neerti:
Being lit on fire barely did anything, for multiple reasons, all which should be fixed in this PR.
First, the code to do damage based on bodytemperature was literally bugged and only applied the lowest level of damage, no matter what.
Second, wounds made autohealed them fairly fast, sometimes even outhealing the incoming damage.
Third, being on fire raised your temperature rather slowly.
Fourth, the lowest level of 'overheating' damage was rather low.

Changes:
The bodytemperature part of Life() is now unbroken and I accidentally fixed freezing not hurting, so yay.
Added a ten minute delay to wounds being able to autoheal them away.  Applying bandage/salves will skip this.
Temperature climb is now based on fire stacks while on fire, so more stacks means more and longer pain.
Lowest level of overheating was tweaked.
Running into people while on fire will split your fire stacks with them and light them on fire.
Ports /tg/'s on fire sprite.
Different mobs can have different icons for being on fire.
Bonus: Rejuv now removes pain.


Dankman:
how does your change affect chemicals that affect body temperature? prediction: you just also accidentally made temperature chemicals obscenely powerful murder machines

Wickedtemp:
Don't worry about temperature changing chems. All Medical/Science really has in that department is frost oil (and potentially condensed capsaicin? I'm unsure, I'll have to check), which they can't get without a botanist, and leporazine, which stabilizes temperature and nothing else. 

And then fire. But if you're on fire, you have bigger problems to worry about than just raised temperature :p

Wickedtemp:
Question: Does autoheal begin ten minutes after the wound was made, or is it ten minutes after any damage is taken? If they get hit again, does that reset the timer?

Dankman:
Does this make enchiladas deadly?

Neerti:
If I remember correctly chemicals like frost oil and cryotoxin have caps to not go below a specific temperature too.

Dankman:
before merging, could you check? Cause like, changing how the body handles temperature, changes a LOT of stuff, Polaris has fullbody prosthetics, right? Being out in vacuum without a cooler is usually a pretty hard death sentence, but with these buffs to spike damage, it could get really bad, really fast. 

This is a change to a core mechanic that a lot of the game was built around, while leaving everything that was designed with the original functionality in place untouched.

Wickedtemp:
60u Frost Oil injected, temp doesn't go below -22C
60u Condensed Capscaicin injected, no temp change. Ingested, no temp change.

Reagent-wise, it's fine.

code/modules/mob/living/carbon/human/life.dm
```diff
@@ -4,7 +4,7 @@
 #define HUMAN_MAX_OXYLOSS 1 //Defines how much oxyloss humans can get per tick. A tile with no air at all (such as space) applies this value, otherwise it's a percentage of it.
 #define HUMAN_CRIT_MAX_OXYLOSS ( 2.0 / 6) //The amount of damage you'll get when in critical condition. We want this to be a 5 minute deal = 300s. There are 50HP to get through, so (1/6)*last_tick_duration per second. Breaths however only happen every 4 ticks. last_tick_duration = ~2.0 on average
 
-#define HEAT_DAMAGE_LEVEL_1 2 //Amount of damage applied when your body temperature just passes the 360.15k safety point
+#define HEAT_DAMAGE_LEVEL_1 5 //Amount of damage applied when your body temperature just passes the 360.15k safety point
```
Anewbe: It was at 5 at one point in the past and there were issues, people were displeased.

code/modules/mob/living/carbon/human/life.dm
```diff
@@ -1181,7 +1185,7 @@
 
 				// Apply a fire overlay if we're burning.
 				if(on_fire)
-					health_images += image('icons/mob/screen1_health.dmi',"burning")
+					health_images += image('icons/mob/OnFire.dmi',"[get_fire_icon_state()]")
```
Anewbe: Is this the massive /tg/ fire thing? Still not sure how I feel about that.

code/modules/mob/living/living_defense.dm
```diff
@@ -351,6 +351,24 @@
...
+		fire_stacks /= 2
+		L.fire_stacks += fire_stacks
+		if(L.IgniteMob())
+			message_admins("[key_name(src)] bumped into [key_name(L)] and set them on fire.")
+
+	if(L_old_on_fire) //Only ignite us and gain their stacks if they were onfire before we bumped them
+		L.fire_stacks /= 2
+		fire_stacks += L.fire_stacks
+		IgniteMob()
+
```
Anewbe: This concept will probably need to be discussed.

SilverTalismen:
Hi could I see the on fire sprite? :o If you don't mind?

Nalarac:
This would Fix #3817 since that issue was about autoheal which is being changed here.

Anewbe:
Neerti knows what needs to be done with this PR.

Anewbe:
Oh boy, here we go.

PR #3838, opened 2017-09-15
Removes a comment

Cameron653:
Toto, I don't think we're in Vorestation anymore.


Cameron653:
I just looked in the code and there's actually a bunch of these comments. Well, I'll get to removing them eventually.

PR #3839, opened 2017-09-15
Removes reference robots.

Belsima:
Renames the sailor moon references and deletes the sprites that were straight rip-offs of Fallout, given new policies.


Anewbe:
Those policies aren't new, they've just been followed spottily.

Belsima:
Ah, alright. Well, here's step one to respecting these policies, yes?

elgeonmb:
I wouldn't call "relatively common Japanese first names given to robots taken from a show with no robots in it whatsoever" a reference so to speak.

PR #3840, opened 2017-09-15
Removes reference AI hologram.

Belsima:
Given we can now select 'humanf', 'humanm', customized sprites, crew members, and plenty others, might as well delete the Cortana sprite.


Anewbe:
Alright, now you're being spiteful because I told you off. This is closing, do not remake it.

Belsima:
I don't see the issue? This is a direct reference to Halo, and very obviously so in many ways.

PR #3843, opened 2017-09-16
Corrects Borg Light Intensity

Neerti:
Apparently default intensity before the handle_light() change is 1 and not the range repeated, so now it's 1 again.


Anewbe:
If needed we can readjust later.

PR #3845, opened 2017-09-16
Catwalks and Railings

Woodratt:
Added catwalks and railings. Ported from Vorestation who ported them from Eris.

Note, catwalks can be placed on plating (as is seen on the vorestation map Tether) and is done so here as well. However it doesn't seem like it is possible to build said catwalks on plating. Did not bother to adjust this at this time. Something to sort in the future.

Adjusted SC station dmm to use both the railing and catwalks as well as a couple fixes for the floor tile adjustment in a earlier commit.


code/game/objects/structures/lattice.dm
```diff
@@ -67,7 +67,18 @@
 				user << "<span class='notice'>Slicing lattice joints ...</span>"
 			new /obj/item/stack/rods(src.loc)
 			qdel(src)
-
+		return
+	// VOREStation Edit - Added Catwalks
```
Yoshax: :eyes:

code/modules/multiz/movement.dm
```diff
@@ -57,7 +57,10 @@
 		if(!A.CanPass(src, start, 1.5, 0))
 			to_chat(src, "<span class='warning'>\The [A] blocks you.</span>")
 			return 0
-	Move(destination)
+	//VOREStation Edit
```
Yoshax: :eyes:

Cyantime:
As a heads up, ghosts can use the flip/rotate railing verbs, and there's no click delay on attacking them.

code/game/objects/structures/lattice.dm
Woodratt: Funny eyes don't explain to me what is wrong. Most of what I did should be a direct copy of how the code is set up on Vorestation.

code/modules/multiz/movement.dm
Woodratt: See previous comment

code/game/objects/structures/lattice.dm
Anewbe: I think the implication is that we don't need the comment there.

Yoshax: Mostly just a joke. It's a comment added on VORE to show where the code is edited on said server, so that doing syncs with here is easier. It's just slightly amusing in that aspect. It's completely unnecessary but doesn't really do anything.

PR #3854, opened 2017-09-17
Fixes for nettles

PrismaticGynoid:
Nettle plants can now spawn randomly from weeds - it was trying to spawn 'nettles', but the name of the plant is 'nettle', so it never spawned. Now it can.

You now hit people with food if you use it on harm intent. This is mostly so that you can use nettles as a weapon again, but hey, who wouldn't want to hit people over the head with a baguette.

Nettles and other stinging "fruits" can no longer be used to generate infinite chemicals by throwing them. Chemicals that the "fruits" inject are now properly removed when hitting a target.


Anewbe:
Do the non-in-hand plants (like biohazards, etc) ever run out of chemicals? If so, that's an issue.

Other than that, looks fine.

PrismaticGynoid:
Hmm, looks like I actually broke those, working on a fix.

Anewbe:
Thank you.

Anewbe:
Fix tested?

PrismaticGynoid:
Yep, it all seems to be working now.

Anewbe:
Excellent.

PR #3855, opened 2017-09-17
Now requires harm intent to beat people with batons.

Belsima:
This is to prevent security from accidentally beating prisoners and breaking their ribs when they intended to stun them.

port of https://github.com/Baystation12/Baystation12/pull/16976.

### DIFF
--- a/code/game/objects/items/weapons/stunbaton.dm
+++ b/code/game/objects/items/weapons/stunbaton.dm
@@ -165,7 +165,7 @@
 		var/mob/living/carbon/human/H = target
 		affecting = H.get_organ(hit_zone)
 
-	if(user.a_intent == I_HURT || user.a_intent == I_DISARM)
+	if(user.a_intent == I_HURT) //To prevent accidental beatings.
 		. = ..()
 		//whacking someone causes a much poorer electrical contact than deliberately prodding them.
 		agony *= 0.5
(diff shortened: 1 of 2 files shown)


Anewbe:
There's a reason batons work the way they do on Disarm, it's to give them the ability to dislocate limbs, as with every other melee weapon. This is very much intentional.

Yoshax:
Disarm intent baton exists for a reason. With your solution, someone with an out of power baton would be forced to harmbaton instead of disarm, which is reduced damage.

Belsima:
Fair enough. Just was hoping to prevent accidental beatings, which happen surprisingly often. I'll try to see if I can figure any alternative solution. Quoting the Baystation PR, "Secondly, when using an E-gun on stun setting, you need to have it on something other than help. If you then switch to a baton, if it's on disarm, you will begin to beat the shit out of the person. This isn't good game design and sends mixed signals."

Yoshax:
So use the preference that allows you to fire guns on help intent, if that's your issue.

Anewbe:
Yosh beat me to my point.

Belsima:
...I'll be honest; I forgot that existed. Oops. I was testing on another Baycode branch.

Anewbe:
It's best to test code on the codebase it's planned for.

Belsima:
True enough. I was just mirroring the last of my Auroraports as Polaris PRs. Worry not, all future PRs will be tested on Polaris code; only a small batch were tested on Aurora code.

PR #3856, opened 2017-09-17
Adds nicotine to cigarettes.

Belsima:
Straight port from https://github.com/Baystation12/Baystation12/blob/24701f1dfd2cf51381c2f42b27e915006e4b83df/code/game/objects/items/weapons/cigs_lighters.dm.

### DIFF
--- a/code/game/objects/items/weapons/cigs_lighters.dm
+++ b/code/game/objects/items/weapons/cigs_lighters.dm
@@ -342,9 +342,9 @@ CIGARETTE PACKETS ARE IN FANCY.DM
 	weldermes = "<span class='notice'>USER insults NAME by lighting it with FLAME.</span>"
 	ignitermes = "<span class='notice'>USER fiddles with FLAME, and manages to light their NAME with the power of science.</span>"
 
-	New()
-		..()
-		reagents.add_reagent(/datum/reagent/nicotine, 5)
+/obj/item/clothing/mask/smokable/cigarette/cigar/New()
+	..()
+	reagents.add_reagent(/datum/reagent/nicotine, 5))
 
 /obj/item/clothing/mask/smokable/cigarette/cigar/cohiba
 	name = "\improper Cohiba Robusto cigar"
@@ -359,9 +359,9 @@ CIGARETTE PACKETS ARE IN FANCY.DM
 	smoketime = 7200
 	chem_volume = 30
 
-	New()
-		..()
-		reagents.add_reagent(/datum/reagent/nicotine, 10)
+/obj/item/clothing/mask/smokable/cigarette/cigar/havana/New()
+	..()
+	reagents.add_reagent(/datum/reagent/nicotine, 10))
 
 /obj/item/weapon/cigbutt
 	name = "cigarette butt"
--- a/code/game/objects/items/weapons/cigs_lighters.dm
+++ b/code/game/objects/items/weapons/cigs_lighters.dm
@@ -344,7 +344,7 @@ CIGARETTE PACKETS ARE IN FANCY.DM
 
 /obj/item/clothing/mask/smokable/cigarette/cigar/New()
 	..()
-	reagents.add_reagent(/datum/reagent/nicotine, 5))
+	reagents.add_reagent(/datum/reagent/nicotine, 3))
 
 /obj/item/clothing/mask/smokable/cigarette/cigar/cohiba
 	name = "\improper Cohiba Robusto cigar"
@@ -361,7 +361,7 @@ CIGARETTE PACKETS ARE IN FANCY.DM
 
 /obj/item/clothing/mask/smokable/cigarette/cigar/havana/New()
 	..()
-	reagents.add_reagent(/datum/reagent/nicotine, 10))
+	reagents.add_reagent(/datum/reagent/nicotine, 5))
 
 /obj/item/weapon/cigbutt
 	name = "cigarette butt"
--- a/code/game/objects/items/weapons/cigs_lighters.dm
+++ b/code/game/objects/items/weapons/cigs_lighters.dm
@@ -279,6 +279,7 @@ CIGARETTE PACKETS ARE IN FANCY.DM
 	chem_volume = 15
 	max_smoketime = 300
 	smoketime = 300
+	var/nicotine_amt = 2
 	matchmes = "<span class='notice'>USER lights their NAME with their FLAME.</span>"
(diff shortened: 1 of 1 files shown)


code/game/objects/items/weapons/cigs_lighters.dm
```diff
@@ -338,6 +342,10 @@ CIGARETTE PACKETS ARE IN FANCY.DM
 	weldermes = "<span class='notice'>USER insults NAME by lighting it with FLAME.</span>"
 	ignitermes = "<span class='notice'>USER fiddles with FLAME, and manages to light their NAME with the power of science.</span>"
 
+	New()
```
Yoshax: Relative pathing pls no.

code/game/objects/items/weapons/cigs_lighters.dm
```diff
@@ -351,6 +359,10 @@ CIGARETTE PACKETS ARE IN FANCY.DM
 	smoketime = 7200
 	chem_volume = 30
 
+	New()
```
Yoshax: Relative pathing pls no.

code/game/objects/items/weapons/cigs_lighters.dm
Belsima: Would it be better as 
`/obj/item/clothing/mask/smokable/cigarette/cigar/New()
	..()
	reagents.add_reagent(/datum/reagent/nicotine, 5))`
or am I misunderstanding what you're saying?

Anewbe: Yes, the Github's wiki tab has examples of absolute and relative pathing, under the Standards section.

Anewbe: Also, that ..() calls the New() proc from the item above it, pathwise, which is in this case /obj/item/clothing/mask/smokable/cigarette/New(), so if you've added nicotine there, it will also be called in this proc... Which may or may not give you the intended amount of nicotine.

Belsima: I'll be honest, I didn't understand what the non-absolute pathing thing meant the first time I read it there, and I still don't understand what the hell that is or how it relates. I might be a moron.

Anewbe: Relative pathing is when you define a proc [in this case New()] in the object's code, without specifying in the line which object the proc is supposed to function for.

If you have something with the path: /obj/item/thing, 

/obj/item/thing/New()
is absolute pathing. You can tell exactly which item will call this proc, because it's there on the line.

-> New()
is relative pathing. This is less than ideal, because it doesn't directly point you to the item it's supposed to work with. This isn't a huge problem with simple items that might only have one or two procs, but if it gets more complicated it really starts to become an issue. Also it's very easy to lose procs through poor indentation. Overall, we like to avoid using relative pathing.

Belsima: That explains it a lot better, thank you very much.

code/game/objects/items/weapons/cigs_lighters.dm
```diff
@@ -351,6 +359,10 @@ CIGARETTE PACKETS ARE IN FANCY.DM
 	smoketime = 7200
 	chem_volume = 30
 
+/obj/item/clothing/mask/smokable/cigarette/cigar/havana/New()
+	..()
+	reagents.add_reagent(/datum/reagent/nicotine, 10))
```
Anewbe: With the super call [..()], this will have 17 units of nicotine in it.

code/game/objects/items/weapons/cigs_lighters.dm
```diff
@@ -338,6 +342,10 @@ CIGARETTE PACKETS ARE IN FANCY.DM
 	weldermes = "<span class='notice'>USER insults NAME by lighting it with FLAME.</span>"
 	ignitermes = "<span class='notice'>USER fiddles with FLAME, and manages to light their NAME with the power of science.</span>"
 
+/obj/item/clothing/mask/smokable/cigarette/cigar/New()
+	..()
+	reagents.add_reagent(/datum/reagent/nicotine, 5))
```
Anewbe: With the super call [..()] this will have 7 units of nicotine in it.

code/game/objects/items/weapons/cigs_lighters.dm
Belsima: Ah. So, would I just want to remove the ..() and it'll be 10 units, or would I want to just adjust the amount in the add_reagent?

Anewbe: You want to adjust the amount you're adding in that add_reagent call. Removing ..() tends to have unforeseen consequences.

Belsima:
The intent is 2, 5, and 10 respectively.

Belsima:
...ah, shit.

Anewbe:
Travis itself seems to be having an issue. It will hopefully resolve itself in some time.

Anewbe:
Travis is apparently fine again, here's what it has to say:

code/game/objects/items/weapons/cigs_lighters.dm:347:error: ): expected }
code/game/objects/items/weapons/cigs_lighters.dm:346:error: location of top-most unmatched {

code/game/objects/items/weapons/cigs_lighters.dm
```diff
@@ -338,6 +342,10 @@ CIGARETTE PACKETS ARE IN FANCY.DM
 	weldermes = "<span class='notice'>USER insults NAME by lighting it with FLAME.</span>"
 	ignitermes = "<span class='notice'>USER fiddles with FLAME, and manages to light their NAME with the power of science.</span>"
 
+/obj/item/clothing/mask/smokable/cigarette/cigar/New()
+	..()
+	reagents.add_reagent(/datum/reagent/nicotine, 3))
```
Anewbe: Ah, there it is, superfluous close paren.

Anewbe:
Did you test to make sure this works? I'm pretty sure it'll run fine, but I've been wrong before.

Belsima:
It's booting the local server right now. It's based on the nutrient code, so I imagine it'll work.

Anewbe:
Oh, wait, odd little request: Go into Chemistry-Reagents-Toxins.dm and change the flavor of nicotine to "pepper".

Belsima:
Alright.

Belsima:
Tested and it works.

PR #3857, opened 2017-09-17
Slime Update

### LINKED ISSUE #3834: [Slime Issue] Dark Blue Slime Cold Aura
A.) Dark Blue slime's cold_aura needs get_cold_protection >= 1 or it causes temperature decrease. This requires every body part (including hands) to be covered with thermal protection gear. Xenobiologists frequently use galoshes to stop slipping. Therefore, dark blue slimes essentially always cause the cold-related effects. We may want to tweak this a bit.
B.) Dark Blue slime's cold_aura frequently kills offspring that are not also dark_blue. This makes it difficult to actually use those colors. 
A.) Winter gear would protect a bit more efficiently against these slimes.
B.) Children of dark blue slimes are harvest-able.
#### Additional info:

### LINKED ISSUE #3835: [Slime Issue] Amber Slimes feed each other indefinitely
Amber slimes have no cap to the number of slimes, amber or otherwise, that they can cause to grow and split. This gives the Mob controller trouble.
A.) Some sort of cap to the number of slimes that can be fed by amber slimes in the same area.
B.) Some sort of cap on maximum number of slimes in a map.
#### Additional info:

Neerti:
Pink Slime Aura and Mend Synthetic (technomancer) should now fix FBPs again.
Adds procs to get the 'real' number of damage from humans, including limb damage on prosthetic limbs.
Slime following made a bit more robust, slimes (and other simple animals) stop wandering when following.
Slimes following a leader that passes the can_command() check will refrain from attacking things if not attacked first.
Slime kill command implemented, requires the slime to be in your faction or be the slime's friend, and not just obedient.  This can be accomplished with end-game slime agents.
Slimes require three 'free' tiles to split, which should prevent super overcrowed cells and by extension avoid Amber Tide from crashing the mob controller.
Dark Blue Cold Aura now properly adjusts temperature based on protection gradually instead of instantly setting to the lowest value.
Fixes #3834 
Fixes #3835

### DIFF
--- a/code/modules/mob/living/simple_animal/slime/life.dm
+++ b/code/modules/mob/living/simple_animal/slime/life.dm
@@ -97,15 +97,20 @@
 
 // This is to make slime responses feel a bit more natural and not instant.
 /mob/living/simple_animal/slime/proc/delayed_say(var/message, var/mob/target)
-	sleep(rand(1 SECOND, 2 SECONDS))
-	if(target)
-		face_atom(target)
-	say(message)
+	spawn(rand(1 SECOND, 2 SECONDS))
+		if(target)
+			face_atom(target)
+		say(message)
 
 //Commands, reactions, etc
 /mob/living/simple_animal/slime/hear_say(var/message, var/verb = "says", var/datum/language/language = null, var/alt_name = "", var/italics = 0, var/mob/speaker = null, var/sound/speech_sound, var/sound_vol)
 	..()
 	if((findtext(message, num2text(number)) || findtext(message, name) || findtext(message, "slimes"))) // Talking to us
+
+		// First make sure it's not just another slime repeating things.
+		if(istype(speaker, /mob/living/simple_animal/slime))
+			if(!speaker.client)
+				return
 
 		// Say hello back.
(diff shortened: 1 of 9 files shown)


code/modules/mob/living/simple_animal/slime/life.dm
```diff
@@ -142,33 +147,31 @@
...
 
 			if(follow_mob) // We're being asked to stop following someone.
-				if(follow_mob == speaker)
+				if(can_command(speaker) == SLIME_COMMAND_FRIEND || follow_mob == speaker)
 					delayed_say("Yes... I'll stop...", speaker)
 					LoseFollow()
 				else
 					delayed_say("No... I'll keep following \the [follow_mob]...", speaker)
 
-		// Help request
```
Anewbe: This was removed because we couldn't think of a way to make it work?

Neerti: There's no easy and robust way to determine what is 'threatening' to the person asking help.

Anewbe: Ok

PR #3859, opened 2017-09-17
Spawned Skrell have 'hair' again

Anewbe:


code/modules/mob/living/carbon/human/human_species.dm
```diff
@@ -10,7 +10,7 @@
 	delete_inventory()
 
 /mob/living/carbon/human/skrell/New(var/new_loc)
-	h_style = "Skrell Male Tentacles"
+	h_style = "Skrell Short Tentacles"
```
Belsima: Wouldn't the average ones make more sense?

Anewbe: Doesn't really matter.

PR #3866, opened 2017-09-17
Dirt, hair, and underwear

### LINKED ISSUE #3786: Prometheans can eat the same dirt repeatedly
When a promethean stands in place and eats the dirt on the floor, the tile looks like it's cleaned, but the dirt gets restored when the tile is walked on again. This can be exploited by walking back and forth between two dirty tiles to quickly gain massive amounts of nutrition.
When the promethean cleaned the dirt, it would stay gone until the dirt built up again.
#### What actually happened
The dirt overlay gets deleted but the tile is still dirty, and the overlay gets restored completely the moment the tile is walked on again.
#### Steps to reproduce
Be a promethean, walk on tiles until dirt appears. Keep walking between the tiles to eat  and instantly restore the dirt.
#### Additional info:
 - **Server Revision**: master - 2017-09-04
20c4ba8b4a481392f8e81f5633e1951a20e719eb 
 - **Anything else you may wish to add** I'm not sure it makes sense for prometheans be able to eat through shoes, either.

PrismaticGynoid:
Fixes #3786 - prometheans are no longer able to eat the same dirt repeatedly for infinite nutrition. The dirt now gets properly removed from the tile. Also, prometheans are no longer able to eat dirt and other messes off the floor unless their feet are uncovered.

Prometheans are now able to pick their hair in the character setup menu, and they're also able to have underwear (hopefully underwear that won't get dissolved accidentally).

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/prometheans.dm
+++ b/code/modules/mob/living/carbon/human/species/station/prometheans.dm
@@ -21,7 +21,7 @@ var/datum/species/shapeshifter/promethean/prometheans
 	swap_flags =       MONKEY|SLIME|SIMPLE_ANIMAL
 	push_flags =       MONKEY|SLIME|SIMPLE_ANIMAL
 	flags =            NO_SCAN | NO_SLIP | NO_MINOR_CUT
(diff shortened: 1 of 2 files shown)


Wutnold124:
underwhair

Issue #3853, opened 2017-09-17
CryoStorage Escapepod doors


SilverTalismen:
#### Brief description of the issue
When you arm the Cryo escape pod, it bolts open one of the closed doors, so when it launches it immidietly spaces the hallway
#### What you expected to happen
For all the doors to remain closed and perhaps even bolt shut for safety reasons
#### What actually happened
The door here [https://gyazo.com/262bf460e5c71e7c35b987f156aa6b13] will open when you arm the pod and bolt open. [disregard the traitor assistant please]
#### Steps to reproduce
Go to the cryo storage, switch the pod doors to manual and arm it and watch as that door opens and bolts open or remains open and bolts
#### Additional info:
 - **Server Revision**: master - 2017-09-16
 - **Anything else you may wish to add** (Location if it's a mapping issue, etc)

Anewbe:
Should be fixed.

PR #3872, opened 2017-09-18
Yay, My first ever pull request!! [Lighting fixes and airlock fix]

SilverTalismen:
Adds some missing lighting in area's that seemed too dark, and fixes a airlock at the Cryostorage pod that would bolt open when you launch it and vent the haul when it launches.

Not sure why it has so many deletions and stuff, only added some lights and fixed a door.


SilverTalismen:
Should address #3853

Atermonera:
Map changes tend to affect a lot of lines, especially with the map merger being broken

PR #3875, opened 2017-09-18
Adds the Frost Spider. Starts support for more complicated spiderlings.

Anewbe:
Also adds an oodle of spider sprites, c/o Mechoid.


Wickedtemp:
I'm pretty sure there was a cheaply made SciFi Channel movie exactly like this.

PR #3876, opened 2017-09-18
Adds maid uniform selection + bonnet to loadout.

Belsima:
Also condensed scarves into scarf selection because whilst testing they took up way more space than they had any right to.

Bonnet ported from Eros.


Anewbe:
Did you test this? Including the scarves?

Belsima:
Yes.

code/modules/client/preference_setup/loadout/loadout_accessories.dm
```diff
@@ -167,64 +167,16 @@
...
-	display_name = "scarf, striped green"
-	path = /obj/item/clothing/accessory/stripedgreenscarf
-
-/datum/gear/accessory/scarf/stripedblue
-	display_name = "scarf, striped blue"
-	path = /obj/item/clothing/accessory/stripedbluescarf
+/datum/gear/accessory/scarf/New()
+	..()
+	var/list/scarves = list()
+	for(var/scarf in typesof(/obj/item/clothing/accessory/scarf/))
```
Anewbe: If you do this, you need to give the base scarf item a sprite; this call includes the base item in it's list, and it doesn't look like /obj/item/clothing/accessory/scarf overwrites the tie sprite that is the base for accessories.

Belsima: ...huh. I just tested that the selection worked and let you pick different ties- I tested 3 of them, but not the base one. My bad. I'll adjust that.

Anewbe:
You don't need any of the map changes from the Southern Cross files, those should probably be excluded.

Also, a complete feature list should probably go in your First Post, for any PRs you make.

Complete Feature List Here:
- Added a Maid Dress
- Added a Maid Bonnet
- Repathed the Sexy Maid Dress
- Gave the Maid Dress a bio resistance of 5
- Scarves in the loadout are now a selection.

code/modules/client/preference_setup/loadout/loadout_accessories.dm
Anewbe: Yeah, easy to overlook.

Belsima:
Given the 'default' armband is the red one, the new 'default' scarf is a red one.

code/modules/clothing/under/accessories/accessory.dm
```diff
@@ -236,12 +236,9 @@
 
 //Scarves
 
-/obj/item/clothing/accessory/scarf
```
Anewbe: I'm not sure if removing this code actually does anything, fix wise. I believe the game will see the other scarves and think "Oh, there's a base type that doesn't change anything" and spawning or selecting /obj/item/clothing/accessory/scarf will spawn a tie.

Belsima: ...oooh, riiiight. I'll fix that rn.

PR #3879, opened 2017-09-18
Changes Tradeband syllables.

Belsima:
Ported from Paradise. I don't mind less serious syllables for languages, but hearing a supposedly unknown language that's actually an irl language that should be completely unrelated that there's classes for in my school 1 room across from my French class doesn't seem right.


Anewbe:
Instead of removing the old syllables, mix in the new ones? Tradeband sounding imperious is very much a good thing, in my opinion.

Belsima:
That could work, I suppose.

Belsima:
Ah- wait. No. Not really. The old ones were all words hence the space chance. I could get a list of latin syllables and mix it in there rather than latin words, if you want?

Anewbe:
Sure.

Belsima:
Added most of the more common syllables in latin.

code/modules/mob/language/generic.dm
```diff
@@ -26,7 +26,9 @@
 	whisper_verb = "whispers"
 	key = "0"
 	flags = RESTRICTED
-	syllables = list("vol", "zum", "coo","zoo","bi","do","ooz","ite","og","re","si","ite","ish","ar","at","on","ee","east","ma","da", "rim")
+	syllables = list(
+"vol", "zum", "coo","zoo","bi","do","ooz","ite","og","re","si","ite","ish",
+"ar","at","on","ee","east","ma","da", "rim")
```
Anewbe: Why was this changed?

Belsima: Consistent with every other language this way.

PR #3883, opened 2017-09-18
Space Lube dries off eventually

Anewbe:
Takes five times as long to dry as water.


Anewbe:
Self-Merge was approved.

PR #3884, opened 2017-09-18
Condensation can now freeze floors, if you have the scepter.

Anewbe:
Also made it impossible to mop up ice.


SilverTalismen:
Would be kind of neat if Condensation soaked people too, and maybe make you able to freeze them immobile for awhile or something

Anewbe:
It already can soak them, pretty sure.

SilverTalismen:
Would be nice to be able to freeze people temporarily then if they are soaked, although I understand why that may be difficult to code.

Neerti:
Remember to change the ``enhancement_desc`` in the spell datum at the top of the file to tell people looking in the catalog that the scepter has an effect.

Anewbe:
Oh, thank you.

PR #3895, opened 2017-09-20
Adds an endless cable coil to the alien toolbelt

Anewbe:
It's infinite! Worked fine in testing, including use in robotics.

Also cleaned up cable code, just a bit. Should actually be able to examine cable coils  from a distance.


code/modules/power/cable.dm
```diff
@@ -565,15 +565,19 @@ obj/structure/cable/proc/cableColor(var/colorC)
 		w_class = ITEMSIZE_SMALL
 
 /obj/item/stack/cable_coil/examine(mob/user)
-	if(get_dist(src, user) > 1)
-		return
+	var/msg = ""
 
 	if(get_amount() == 1)
-		to_chat(user, "A short piece of power cable.")
+		msg += "A short piece of power cable."
```
Belsima: Just a question- what's the difference between this and to_chat?

Anewbe: In this instance, msg is a string or text bar that gets modified based on how much cable is in the coil, and if you're close enough it adds in the amount of said amount is greater than 2, then passes the msg string to the user through a to_chat().

PR #3897, opened 2017-09-20
Ghosting as an AI now stores the core and frees the slot

Anewbe:


Neerti:
Probably better to override ``ghostize()`` instead?

Anewbe:
Tried that, ran into a ton of issues with recursion.

Anewbe:
Will try the ghostize thing again, may take a little while.

Anewbe:
Closing, will redo at a later date.

PR #3900, opened 2017-09-22
Jumper Cables in Robotics should spawn with cells now

Anewbe:
Was spawning the wrong type, my bad.


Nalarac:
It was actually when I re-added their functionality that they didn't have cells because I didn't want to mess with the map file to change them to the loaded version. So it's my bad.

Anewbe:
Eh, no worries.

PR #3904, opened 2017-09-22
Fixes the Sun/Weather and Ports Vore's Lighting SS

Neerti:
New lighting system is needed to handle updating literally 14k+ tiles whenever the sun moves.  The straight port seems to have had no adverse effects from what I could see.  Don't know if it's more performant but it doesn't seem to be less.

I didn't expect the latter would be needed for the former.

Code is from https://github.com/VOREStation/VOREStation/pull/1763


Anewbe:
\[T]/

PR #3905, opened 2017-09-22
Fixes up some loadout and clothes.

Belsima:
Fannypacks and flats now use colorable options rather than set color seleciton. Webbings and drop pouches use a selection, as do scarves. Added the knife harness to the webbing crate. Removed unused fannypack sprites.


SilverTalismen:
We already have a colorable flats but I would prefer we keep the set color flats

Belsima:
You can set it up to use any color the colored flats have- this just removes the 20 options of flats we have now that clutter up the loadout.

Belsima:
Eh, I'll return it and fix the flats up a better way.

Belsima:
...huh? I think I just fixed one thing after my test. Didn't realize I somehow made it not compile. One min.

Anewbe:
That's an indentation error, iirc those compile but Travis just doesn't like them.

Belsima:
Ahhh. Alright.

Belsima:
Fixed.

code/datums/supplypacks/misc.dm
```diff
@@ -64,7 +64,8 @@
 			/obj/item/clothing/accessory/storage/black_drop_pouches,
 			/obj/item/clothing/accessory/storage/brown_drop_pouches,
 			/obj/item/clothing/accessory/storage/white_drop_pouches,
-			/obj/item/clothing/accessory/storage/webbing
+			/obj/item/clothing/accessory/storage,
+			/obj/item/clothing/accessory/storage/knifeharness
```
Anewbe: This is a bit out-of-scope.

Belsima: My computer is in a cardboard box and will be for weems. Can't change it.

Belsima:
Tried to fix on my phone. Let's see how badly that works.

PR #3911, opened 2017-09-22
Cyber Solutions torso should now be visible.

Anewbe:


Belsima:
My bad, I took it right from the other PR. Assumed it was working in that one.

PR #3912, opened 2017-09-22
Adds a spider spawner grenade, and variant.

Anewbe:
Just because.

For those who don't get the joke: https://www.youtube.com/watch?v=PYtXuBN1Hvc


Atermonera:
I mean we do technically already have this, it's just that obtaining it involves a nurse chitter, a teshari, and abpneumatic cannon

PR #3913, opened 2017-09-22
Hull shields and emergency shields check for outdoor turfs and drops

Screemonster:
This is more of a thing for planetside maps, especially those with funky atmospheric situations. Also includes support for multi-z.

tl;dr, instead of just placing shields on space turfs, the generator now blocks off outdoor areas as well so you can use the shield to keep yourself snug and warm on an arctic hellscape station or whatever else you want to use just as readily as you would one in space.


Anewbe:
So... If I put my shield generator inside my outpost in my arctic hellscape, the shield will actually wrap the outpost, or am I reading this wrong?

Screemonster:
Yep, it'll shield the outpost. It essentially treats outdoor turfs as space turfs for deciding whether or not to place a shield

PR #3914, opened 2017-09-23
Adds the bone clamp, and some alien surgery tools

Anewbe:
Sprites aren't great, but they should do for now.


Wickedtemp:
So, basically, it's a setter and bone gel 2-in-1?

Can it be used to close the ribcage/skull after it's been sawed open during surgery, or is bone-gel still needed for that step?

Anewbe:
Still need bone gel for that, I'm afraid.

PR #3915, opened 2017-09-23
Removes binary from brains.

PrismaticGynoid:
This is being done because of the problem of MMI-type brains suddenly having access to a secure channel they can spy on when removed from a body - something they couldn't access before without a binary communication device or similar. This has been used several times against antags, usually antag AIs.


Anewbe:
Tested?

PrismaticGynoid:
Yep, couldn't find any way of having brains overhear binary with this in.

PR #3916, opened 2017-09-23
Makes all synthetic brains radio-enabled.

PrismaticGynoid:
Mostly because the old radio-enabled MMIs were almost never used. Changes all MMIs and their subtypes (posi and drone brains) to have a radio inside of them, which can only be used when the brain is outside of a body. This radio can be used to talk over the common channel, in case you get left on a table somewhere by an absent-minded roboticist. However, the radio can also be disabled, in case an antag has you and doesn't want to to scream for help.

Since this makes the old radio-enabled MMIs obsolete, they're no longer printable(although it's still defined in the code so I don't have to touch the map file). Also fixes a telecomms runtime with brains talking over the radio.


Anewbe:
Tested?

PrismaticGynoid:
Tested every scenario that I could think of.

PR #3918, opened 2017-09-23
Adjusts liver damage from alcohol

Anewbe:
**Requires Discussion, Do Not Merge**

- Hitting the Alcohol Tox level will cause an uncapped amount of liver damage, because it currently stops at 10, and that's a bit wonky. We probably should have or need to make a proc for that, in the future.
- Hitting the Alcohol Tox level will cause Toxin damage if your liver is bruise, and additional Toxin damage if your liver is broken.
- Alcohol is still pretty survivable. In testing I gave a Skrell an entire bottle of special blend and they didn't die in the 10+ minutes I was running the tests. The other species take somewhere in the range of 19 toxloss and ~25 liver damage.

### DIFF
--- a/code/modules/organs/internal/liver.dm
+++ b/code/modules/organs/internal/liver.dm
@@ -39,10 +39,11 @@
 
 		// Do some reagent processing.
 		if(owner.chem_effects[CE_ALCOHOL_TOXIC])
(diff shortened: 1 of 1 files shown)


Wickedtemp:
From a med player's perspective, since this can be stabilized anyways, I personally would support this. But, I support most damage buffs.

PR #3921, opened 2017-09-23
Merges xenobiologist job with scientist.

Cameron653:
With the new slime update and new xenobio, I took a look at what xenobio had special compared to scientists. To make a long story short: Absolutely nothing except for having access to hydroponics. 
They were exactly identical for all intents and purposes. Bumped scientist job up by 2 to compensate.

- Merges xenobiologist job with scientist
- Adds xenobiologist & xenobotanist alt titles to scientists.
- Bumps up scientist job by 2 to compensate.
- Gives scientists hydroponics access since xenobiologists had them. This can be removed if needed, since science/xenobio should really stick to their own area instead of going to a civillian one, but I digress.

Xenobio/Scientist comparison:
Sci access: access = list(access_robotics, access_tox, access_tox_storage, access_research, access_xenobiology, access_xenoarch)

Bio access: access = list(access_robotics, access_tox, access_tox_storage, access_research, access_xenobiology, access_hydroponics)

Eco modifier of 7 for both, same department (MEDSCI), same spawning clothing, etc.

The only difference was that xenobio had hydroponics access, while scientist had xenoarch access instead. Which isn't that big of a difference that it requires it's entire own job.

### DIFF
--- a/code/game/jobs/job/science.dm
+++ b/code/game/jobs/job/science.dm
@@ -49,52 +49,20 @@
 	department = "Science"
 	department_flag = MEDSCI
 	faction = "Station"
-	total_positions = 5
+	total_positions = 7
 	spawn_positions = 3
 	supervisors = "the research director"
 	selection_color = "#633D63"
 	idtype = /obj/item/weapon/card/id/science
 	economic_modifier = 7
-	access = list(access_robotics, access_tox, access_tox_storage, access_research, access_xenobiology, access_xenoarch)
-	minimal_access = list(access_tox, access_tox_storage, access_research, access_xenoarch)
-	alt_titles = list("Xenoarcheologist", "Anomalist", "Phoron Researcher")
+	access = list(access_robotics, access_tox, access_tox_storage, access_research, access_xenobiology, access_xenoarch, access_hydroponics)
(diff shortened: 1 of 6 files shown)


Cameron653:
Also, since I had to use the github website there's a shitload of commits and I don't know how to squash them.

Screemonster:
Will admit I'm not a huge fan, as there's nothing stopping more scientists from taking that alt-title than there is room to work xenobio, resulting in more xenobiologists than there is room in xenobio and nobody working R&D (and nobody able to latejoin as R&D 'cause all the scientist slots are taken up by the people fighting over who gets to breed slimes). Even if they share access, having a cap on the number of _actual_ xenobiologists makes it a lot easier for them to tell interlopers to go away until the department is less busy.

Magneld:
The jobs were separated after slimes were first added to begin with way back on bay to avoid "6 scientists in xenobio, nothing gets done, everybody hates everybody else."

So no thank you. There are three slots for it already plus the RD and often only one of those are taken.

Cameron653:
Fair points. Going to go ahead and close this since, looking at how extremely small xenobio is, this probably would cause a fair amount of chaos.

PR #3924, opened 2017-09-24
CE now spawns with their belt, spare in the locker

Anewbe:


Belsima:
Do they spawn with a ray scanner like regular engineers? If not, could I bother you to add that?

Anewbe:
Talked about, general response among staff seems to be negative. Will return at a later date to add the T-Ray scanner.

PR #3931, opened 2017-09-24
Makes phoron heal more oxyloss for vox.

### LINKED ISSUE #3808: Phoron is essentially useless for fixing oxy in Vox
Oxy damage in Vox is intended to be fixed with phoron, however it's not at all effective.
Inject phoron. Results.
#### What actually happened
It's honestly about the effectiveness of tricordrazine. Potentially slightly better or slightly worse. This is due to how slowly phoron is metabolized. As a result, I suggest either:

A) Buffing the amount of damage phoron heals from Vox, or
B) Speeding it's metab rate
#### Steps to reproduce
Have a vox that's suffocating.
Inject phoron.
Have a vox that's still suffocating.
#### Additional info:
It's a simple change, so if this get's OK'd, I'll make the PR, but I'd like some input here first.

Cameron653:
Resolves #3808 
Increases it from -9 * removed to -75 * removed, since phoron is a semi-uncommon reagent.

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Toxins.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Toxins.dm
@@ -105,8 +105,8 @@
 
 /datum/reagent/toxin/phoron/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
 	if(alien == IS_VOX)
-		M.adjustOxyLoss(-removed * 9)
-		return
+		M.adjustOxyLoss(-75 * removed) //Half as efficent as dex+ would be.
+		return //You're wasting plasma (a semi-limited chemical) to save someone, so it might as well be strong.
 	..()
 
 /datum/reagent/toxin/phoron/touch_turf(var/turf/simulated/T, var/amount)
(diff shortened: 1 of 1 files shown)


Wickedtemp:
Phoron metabolizes much more slowly than dexalin plus, so if you meant for it's heal-rate to be half of dexalin plus, the amount healed per unit would need to be scaled upward. Metabolizing at 0.05u per tick would give a heal of 3.75 damage each tick. So if you intended a faster healing effect, the damage healed per unit would need to be increased by...quite a bit.

Anewbe:
They're not supposed to be easy to treat, anyway.

Wickedtemp:
@Anewbe - I mentioned it because, judging by the title, it's meant to heal half as fast as Dexalin Plus. With the current value, it'd heal at a tenth of the speed of Dexalin Plus. So, if Cameron intended for it to heal half as quickly as Dexalin Plus, the value needs to be changed. 

If, however, that was not the intention, then they can ignore the value I suggested.

Anewbe:
I have no idea on what they intended the numbers to be, I'm just saying that they're not supposed to be easy to heal, and phoron's fairly common, for an uncommon material.

Cameron653:
Oh, I didn't even think about the fact toxins have .05 metabolism compared to dexp's .2 I'm going to bump it up to 5 oxyloss removal per tick (-100*removed) so it'll be on closer to dexalin, since bumping up the metabolism would cause it to kill people way too fast, and bumping up the number _too_ much would result in vox being able to inject phoron and be able to heal oxyloss way too long. Because at the moment, it heals .45 oxygen per second, which is utterly useless. 

Feel free to discuss some more on how much oxyloss it should remove per tick.

PR #3932, opened 2017-09-24
Allows resting people to get off the ground w/ assistance.

Cameron653:
This allows it so resting mobs/humans can grab onto another person to hoist theirself up off the ground.

### DIFF
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -259,7 +259,7 @@
 			else
 				var/mob/living/carbon/human/hugger = M
 				if(M.resting == 1) //Are they resting on the ground?
-					M.visible_message("<span class='notice'>[M] grabs onto [src] and pulls [get_visible_gender() == MALE ? "himself" : get_visible_gender() == FEMALE ? "herself" : "themselves"] up off the ground!</span>", \
+					M.visible_message("<span class='notice'>[M] grabs onto [src] and pulls [M.get_visible_gender() == MALE ? "himself" : get_visible_gender() == FEMALE ? "herself" : "themselves"] up off the ground!</span>", \
 							"<span class='notice'>You grip onto [src] and pull yourself up off the ground!</span>")
 					M.resting = 0 //Hoist yourself up up off the ground. No para/stunned/weakened removal.
 					if(M.fire_stacks >= (src.fire_stacks + 3)) //Fire checks.
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -259,14 +259,15 @@
 			else
 				var/mob/living/carbon/human/hugger = M
 				if(M.resting == 1) //Are they resting on the ground?
-					M.visible_message("<span class='notice'>[M] grabs onto [src] and pulls [M.get_visible_gender() == MALE ? "himself" : get_visible_gender() == FEMALE ? "herself" : "themselves"] up off the ground!</span>", \
-							"<span class='notice'>You grip onto [src] and pull yourself up off the ground!</span>")
-					M.resting = 0 //Hoist yourself up up off the ground. No para/stunned/weakened removal.
+					M.visible_message("<span class='notice'>[M] grabs onto [src] and pulls \himself </span>", \
+							"<span class='notice'>You grip onto [src] and pull yourself up off the ground!</span>") //AHHH gender checks are hard, but this should work	
 					if(M.fire_stacks >= (src.fire_stacks + 3)) //Fire checks.
 						src.adjust_fire_stacks(1)
 						M.adjust_fire_stacks(-1)
 					if(M.on_fire)
 						src.IgniteMob()
+					sleep(5) //.5 second delay. Makes it a bit stronger than just typing rest.
+						M.resting = 0 //Hoist yourself up up off the ground. No para/stunned/weakened removal.
 				else if(istype(hugger))
 					hugger.species.hug(hugger,src)
 				else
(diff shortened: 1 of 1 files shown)


Belsima:
Maybe use a gender check for himself/herself/themself- theirself doesn't seem right at all. Also maybe a very slight delay for getting up?

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -258,7 +258,16 @@
 									"<span class='notice'>You shake [src] trying to wake [t_him] up!</span>")
 			else
 				var/mob/living/carbon/human/hugger = M
-				if(istype(hugger))
+				if(M.resting == 1) //Are they resting on the ground?
+					M.visible_message("<span class='notice'>[M] grabs onto [src] and pulls theirself up off the ground!", \
+							"<span class='notice'>You grip onto [src] and pull yourself up off the ground!</span>")
+					M.resting = 0 //Hoist yourself up up off the ground. No para/stunned/weakened removal.
+					if(M.fire_stacks >= (src.fire_stacks + 3)) //Fire checks.
+						src.fire_stacks += 1
```
Anewbe: adjust_fire_stacks(1)

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -258,7 +258,16 @@
...
 			else
 				var/mob/living/carbon/human/hugger = M
-				if(istype(hugger))
+				if(M.resting == 1) //Are they resting on the ground?
+					M.visible_message("<span class='notice'>[M] grabs onto [src] and pulls theirself up off the ground!", \
+							"<span class='notice'>You grip onto [src] and pull yourself up off the ground!</span>")
+					M.resting = 0 //Hoist yourself up up off the ground. No para/stunned/weakened removal.
+					if(M.fire_stacks >= (src.fire_stacks + 3)) //Fire checks.
+						src.fire_stacks += 1
+						M.fire_stacks -= 1
```
Anewbe: adjust_fire_stacks(-1)

Cameron653:
Ah, alright. I went ahead and copied the code below it for the firestacks, so I'll update it for both. And this is intended to not have a delay, as it's supposed to be an alternative to typing in rest, which does have a delay.

Cameron653:
There we go. Adjusted it to use adjust_fire_stacks and made it check their shown gender.

Anewbe:
I'm gonna ask that you give it a delay, if you don't mind.

Cameron653:
Alright. Added a delay and fixed the gender check by using \himself since
https://github.com/PolarisSS13/Polaris/blob/master/code/modules/mob/living/carbon/human/human.dm?utf8=%E2%9C%93#L865
something like the above line didn't work.

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -259,14 +259,15 @@
...
-							"<span class='notice'>You grip onto [src] and pull yourself up off the ground!</span>")
-					M.resting = 0 //Hoist yourself up up off the ground. No para/stunned/weakened removal.
+					M.visible_message("<span class='notice'>[M] grabs onto [src] and pulls \himself </span>", \
+							"<span class='notice'>You grip onto [src] and pull yourself up off the ground!</span>") //AHHH gender checks are hard, but this should work	
 					if(M.fire_stacks >= (src.fire_stacks + 3)) //Fire checks.
 						src.adjust_fire_stacks(1)
 						M.adjust_fire_stacks(-1)
 					if(M.on_fire)
 						src.IgniteMob()
+					sleep(5) //.5 second delay. Makes it a bit stronger than just typing rest.
```
Anewbe: Genuinely unsure, but does a do_after not work here?

Cameron653: I'm not sure. I just used sleep as I've seen it used many times before and I know it'd work here. I can see if do_after works.

Cameron653:
Changed the sleep to do_after.

code/modules/mob/living/carbon/carbon.dm
Anewbe: It may not work, I don't know if do_after() works for resting mobs.

Anewbe:
code/modules/mob/living/carbon/carbon.dm:269:error: user: undefined var

Cameron653:
Fixed. Compiling to retest now.

Anewbe:
Also, please compile and test your code before you commit it. Not 100% certain, but that seems like something that the compiler would catch.

Cameron653:
Yeah. I tested it initially before adding all the changes (do after, delay, \himself, etc) so was just waiting for feedback to be finished before giving it one last final test, since the compiler takes ages.

Anewbe:
If you swap the Northern Star Map for the Example map it goes much quicker. I learned this only yesterday.

Cameron653:
Alright. Tested and can confirm everything works: https://i.imgur.com/lSwDwId.gifv

Anewbe:
Just had a thought: Can we add an intent check for the target, so people I push over can't just pick themselves up off of my unless I'm on help intent?

Cameron653:
You can't pick yourself up if you're weakened, paralyzed, or have any other status debuff that prevents you from clicking. 

All this does is set resting to 0, so it wouldn't do anything if they were pushed over as they would be weakened and unable to click.

https://github.com/PolarisSS13/Polaris/blob/e158fcd3cc70e657bddba1a331f6b959441c417a/code/modules/mob/living/carbon/human/human_attackhand.dm#L291

A check for resting would be possible, but lines
https://github.com/Cameron653/Polaris/blob/b41c2fec9a1320c1d63dcd34c8670b587a11e85e/code/modules/mob/living/carbon/carbon.dm#L253-L257

should catch if the person being clicked on is resting, in which case they'd shake them awake.

Anewbe:
Alrighty, thank you for looking into that.

PR #3942, opened 2017-09-25
Moves the Promethean core to the chest. They no longer need heads.

Anewbe:
As requested by the LoreMasters.


Anewbe:
Forgot the .dme

PR #3943, opened 2017-09-25
Simple Animal Tweaks

Neerti:
Makes hivebots a bit smarter, shifts stats around, adds a few new sprites thanks to Mechoid, and adds a few new types.
Adds verbs that can be used as a player hivebot, to get them to follow you.
Spider eyes can now be seen clearly in the darkness, which looks spooky.
Chitter


code/modules/mob/living/simple_animal/aliens/hivebot.dm
```diff
@@ -1,23 +1,28 @@
+// Hivebots are tuned towards how many default lasers are needed to kill them.
+// As such, if laser damage is ever changed, you should change this define.
+#define LASERS_TO_KILL *40
```
Anewbe: Just in theory, couldn't this be used for all simple_mobs?

PR #3945, opened 2017-09-25
Corrects pipe examine info.

Cyantime:
I noticed this and was unreasonably bothered by it.


Anewbe:
Is that how it actually works? I thought the 300 kPa was a hard limit.

Cyantime:
If you make a pressure chamber over a highly pressurized pipe, you can unwrench it fine.

Anewbe:
Nice.

Neerti:
If I recall correctly, the limit is the current atmosphere pressure + 2 atmospheres but I could be wrong.

PR #3950, opened 2017-09-25
Radiation-Mob Interaction tweaks

Anewbe:
- Prometheans no longer hallucinate from seeing the supermatter
- Prometheans no longer take radiation damage/effects
- Species with the RADIATION_GLOWS flag no longer glow like the sun. Specifically, I lessened the radius and intensity of the light.

radiation_glows


Mechoid:
The Lightbulb Head is dead. Long live the Lightbulb Head.

PR #3952, opened 2017-09-25
Makes it possible to get slimejelly

Anewbe:
- Peridaxon + grey slime core = 15u slime jelly

Definitely needs discussion

### DIFF
--- a/code/modules/xenobio/items/extracts.dm
+++ b/code/modules/xenobio/items/extracts.dm
@@ -93,6 +93,13 @@
 		new /obj/item/weapon/reagent_containers/food/snacks/monkeycube(get_turf(holder.my_atom))
 	..()
 
+/datum/chemical_reaction/slime/grey_slimejelly
+	name = "Slime Jelly"
+	id = "m_jelly"
+	result = "slimejelly"
+	required_reagents = list("peridaxon" = 5)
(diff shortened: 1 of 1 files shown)


Wickedtemp:
Would support. Slime jelly doesn't have too many uses for Medical, but that could change in future PR's regarding prometheans. Also right now, it's a decent antag reagent.

Anewbe:
It's in the top two toxins in the game, however I have someone with lots of plans for it, and I want it in. The question is how hard it is to acquire.

Wickedtemp:
Peridaxon isn't at all difficult to produce. Currently, there aren't any medicinal chems that ARE  difficult to produce, with the exception of myelamine and osteodaxon, which have easy methods to mix, it's just the fact they require an exotic reagent that you aren't likely to obtain.

On virgo, the recipe for slime jelly is 50u lithium, 50u sugar, and 20u phoron, which will net you 5u of slime jelly. It's a simple recipe, but the fact it uses so much phoron, (40u if you want to do anything useful with it), it's not made often unless Cargo really comes through with the phoron delivery. 

Given Polaris current population, and how rare it is (in my experience) to get extra phoron from Cargo, I think requiring 10-20u of phoron would be a decent limiting factor.

PR #3956, opened 2017-09-26
Adds Farm PoI

Neerti:
Also fixes various bugs with surface tiles, hopefully.
Adds some documentation to the map file system.
Adds a new 'plane' map, ideal for testing PoIs by loading them manually.
Makes space heaters glow orange.
Adds a new grille type for fancy surface windows.
Adds self planting hydro trays/dirts if a seed is on top of them when the map is loaded.


Anewbe:
Can we get an image or two for the curious?

Neerti:
I don't have a good current image but here's a few old ones.
Older but whole: Out of date picture
Newer but meteors that shouldn't have happened killed my barn: Meteors are mean

PR #3957, opened 2017-09-26
Adds support for closets storing closets, and graves

Anewbe:
**_Feature Creep Intensifies_**

- [x] Prevent Storage Recursion
- [x] Specify which tool can seal the locker, instead of just welder or a binary check
- [ ] Graves seal when closed, blocking vision and air, possibly lower internal temperature(?)
- [ ] Dig graves with shovels (Maybe future PR)
- [ ] Better sprites


code/game/objects/structures/crates_lockers/closets/secure/secure_closets.dm
```diff
@@ -141,7 +141,7 @@
 	else
 		usr << "<span class='warning'>This mob type can't use this verb.</span>"
```
Belsima: Just as a note, missed a << since you fixed them everywhere else you touched.

Anewbe:

PR #3960, opened 2017-09-26
Synth Syringe Fix

MisterLayne:
Before, if you tried to take blood from a synthetic, it would say you took the blood (but there would be no sprite update), and then if you tried to take blood from anyone else with that syringe, it would then tell you that you are already taking blood from them, so.. This fixes that.


Anewbe:
code/modules/reagents/reagent_containers/syringes.dm:89:error: missing ) in macro call

Wickedtemp:
IIRC this used to also come up with prometheans. Annoying stuff!

Anewbe:
I'll have a fix for them soonish.

PR #3966, opened 2017-09-27
Drone circuit + posibrain lorechanges

elgeonmb:
Removes all that mumbo about "you live to serve" and replaces it with a reminder that you're designed to be friendly to human values. I don't really see the trouble with this-- if a new drone or posi wants to be a surly bastard, it can, it's just not getting sleeved in a shiny bipedal body.

Also adds a bunch of indulgent OS names for drone circuits.

(Realtalk this is mostly just me figuring out how PRs work so I can start being something closer to actually productive)

### DIFF
--- a/code/modules/mob/living/carbon/brain/MMI.dm
+++ b/code/modules/mob/living/carbon/brain/MMI.dm
@@ -261,8 +261,8 @@
 	src.brainmob.ckey = candidate.ckey
 	src.name = "[name] ([src.brainmob.name])"
 	src.brainmob << "<b>You are a [src], brought into existence on [station_name()].</b>"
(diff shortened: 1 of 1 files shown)


PrismaticGynoid:
If I may make two suggestions on this:

**One**, think about changing _"designed with human values"_ to _"designed with your creator's values"_. A positronic could be imprinted with propaganda about how skrell are the superior species, after all, or a drone programmed to serve only synthetic masters unless lawed otherwise.
**Two**, _"cyborg chassis"_ should really be _"lawed chassis"_ - we're talking about robots and drones, not cyborgs, and it's the laws that really matter here.

elgeonmb:
And thus we are.

PR #3974, opened 2017-09-28
Adds Directional Shields

Neerti:
Implements shields which block projectiles from outside, but allow projectiles to leave from inside.
Also adds a projector item which makes these, in two variants, rectangle and line.  It was made an item so that down the road making stationary/wearable/etc versions will be easier as they can just contain the item.
Currently there is no way to obtain this without adminspawn.  I have Future Plans(tm) for these.
Gif of shields in action


Anewbe:
Does it run on a charge, or just constantly project?

Neerti:
The basic version implemented has a health system and it self charges.

Anewbe:
But the item itself never runs dry?

Neerti:
Its essentially self recharging.

Magneld:
Does it block mobs also or only shots?

Anewbe:
Only shots. Mobs and throw objects should be able to pass.

PR #3976, opened 2017-09-28
Makes guest passes great again!

MisterLayne:
Not only can you now deactivate the guest passes, but it also tells you when they deactivate it

### DIFF
--- a/code/game/machinery/computer/guestpass.dm
+++ b/code/game/machinery/computer/guestpass.dm
@@ -9,6 +9,7 @@
 
 	var/temp_access = list() //to prevent agent cards stealing access as permanent
 	var/expiration_time = 0
+	var/expired = 0
 	var/reason = "NOT SPECIFIED"
 
 /obj/item/weapon/card/id/guest/GetAccess()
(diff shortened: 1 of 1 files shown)


code/game/machinery/computer/guestpass.dm
```diff
@@ -38,6 +39,43 @@
...
+	update_icon()
+
+/obj/item/weapon/card/id/guest/Destroy()
+	processing_objects.Remove(src)
+	return ..()
+
+
+/obj/item/weapon/card/id/guest/process()
+	if(world.time == expiration_time)
+		expired = 1
```
Anewbe: You don't need these two lines, that same code is done better below.

code/game/machinery/computer/guestpass.dm
```diff
@@ -38,6 +39,39 @@
...
+
+		var/confirm = alert("Do you really want to deactivate this guest pass? (you can't reactivate it)", "Confirm Deactivation", "Yes", "No")
+		if(confirm == "Yes")
+			//rip guest pass </3
+			user.visible_message("<span class='notice'>\The [user] deactivates \the [src].</span>")
+			icon_state = "guest_invalid"
+			expiration_time = world.time
+			expired = 1
+	return
+	..()
```
Anewbe: This should be on the same line as the return, or above it.

PR #3982, opened 2017-09-29
Cake no longer requires egg

Anewbe:
- It kept reacting with other reagents in the microwave. Plus you don't need eggs in cake anyway.

- [ ] Fix the wiki (if this is merged)

- [x] Hide from Big Dairy

### DIFF
--- a/code/modules/food/recipes_microwave.dm
+++ b/code/modules/food/recipes_microwave.dm
@@ -989,7 +989,7 @@ I said no!
 
 // Cakes.
 /datum/recipe/cake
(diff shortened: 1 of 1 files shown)


Neerti:
Could the microwave be given NO_REACT?

Anewbe:
I'll look into that.

Anewbe:
It **has** NO_REACT, issue is with putting the actual egg into the microwave, instead of just yolk. I'll check the wiki and see if this can be made more clear

PR #3991, opened 2017-09-30
Better T-Ray Scanners

MisterLayne:
Name one of the most useless tools in the game? If you guessed the T-Ray, you'd probably be right. It's way easier to just crowbar up the floor than to have to worry about waiting two seconds to only see one tile below you. I thought it would be a good idea to add upgraded TScanners, because, TScanners are shit. That's why.

### DIFF
--- a/code/game/objects/items/devices/t_scanner.dm
+++ b/code/game/objects/items/devices/t_scanner.dm
@@ -132,4 +132,19 @@
 /obj/item/device/t_scanner/dropped(mob/user)
 	set_user_client(null)
 
+/obj/item/device/t_scanner/upgraded
+	name = "Upgraded T-ray Scanner"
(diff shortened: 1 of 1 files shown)


Anewbe:
I have the feeling that this is going to lag to shit, just so you know.

MisterLayne:
It should only SLIGHTLY lag the person who uses it.

Anewbe:
It also affects everyone around them, and local servers handle lag much better than the actual server. Still willing to give it a shot.

MisterLayne:
I hope it doesn't lag too much, because, it'd be nice to be able to see the broken pipes or wires without having to destroy the floor.

Neerti:
T-rays only show to the person holding nowdays iirc.

MisterLayne:
It does, but I'm pretty sure it stored the result of the overlay for later use for someone elseeee using Tray? Not sure, the code was confusing as all hell.

PR #3993, opened 2017-09-30
Xenobiology sliding doors.

SilverTalismen:
Swaps out the non reinforced Xenobiology containment doors with the reinforced kind, for all you squishy xenobiologists.

Shouldn't have any issues.


Anewbe:
Just to double check: You did test that Xenobiologists can get through those, right?

SilverTalismen:
Yes I did, I went through all of the doors and they all worked.
All I did was replace the doors, copy the access code from the old ones into the right place and renamed them

PR #3994, opened 2017-09-30
Changes the leg holster item sprite.

ElectricB1ue:
Taken from baycode, looks at least ten times more tacticool. First of a series of PRs
https://gyazo.com/fab9e97721f0c1c85c8361422a98b171
Picture for the curious


Anewbe:
Why?

ElectricB1ue:
@Anewbe It looks better. The current leg holster item sprite is just the waist holder, which in turn is just the shoulder holster snipped in half

PR #3997, opened 2017-09-30
Changes the carbine inhand sprite.

ElectricB1ue:
Taken from Baystation. Tested and functioning. Shouldn't have broken anything?
https://gyazo.com/31d4cbfcbdc7b500f0a53d4a76ff51a5 Raw sprite
Looks more realistically sized, all around better imo.


Anewbe:
There should be an option to actually put the image in the post, if you save it to your computer.

Neerti:
You can also use ``placeholder_text`` too.
e.g. Carbine

Anewbe:
If these are the ones with the -wielded suffix from Bay, this is not how their supposed to be used.

ElectricB1ue:
It is, if it's necessary I'll copypaste that over but it's fine when it's used with two

Anewbe:
If you want to bring the entirety of Bay's wielded gun sprites, you may. But I will not be merging this PR if all it does is replace our Carbine sprites for the wrong ones.

ElectricB1ue:
I'll work on that tomorrow, then. Gonna close this so it'll all be neater

PR #4003, opened 2017-10-01
Makes the boots of speed slower, and boots.

Anewbe:


mistyLuminescence:
Boots, yes; slower, no, they're a good amount of fast right now.

Anewbe:
They're very much not a good amount of fast right now. The intention was Teshari speed, not light speed.

PR #4010, opened 2017-10-02
Rain and Storms call water_act()

Anewbe:
Should calculate out to the same decrease in fire_stacks. Does other cool stuff ™️. Will do more cool stuff soon ™️.


Anewbe:
Apparently we have another version of this incoming, so I'm closing it.

PR #4014, opened 2017-10-03
Drone permits can now be worn

Anewbe:
- Belt or ear slots. Ear slots because they're fairly rarely used.
- EDIT: Now works as an accessory


SilverTalismen:
Why not make them something you can clip to your uniform like a security holobadge?

Anewbe:
Pretty sure I'd have to repath them for that to work. Might look into it tomorrow.

SilverTalismen:
Cool! ^^
It would be way more useful to be able to just clip them onto my uniform so I don't need to take up space in a bag or something.

Anewbe:
That was easier than expected.

Belsima:
Do they fit in the ID slot?

Anewbe:
No? That's where one's access lives.

Anewbe:
Yeah, that.

Belsima:
Can't hurt to let you store it in the ID slot, can it?

Anewbe:
I'm gonna say yes, it can. One's a governmental thing, the other is a corporate thing, and unless you're an Assistant you won't be able to access anything.

PR #4025, opened 2017-10-05
Prosthetic Organs Tweaks

Anewbe:
- Switches out the current fully prosthetic organ sprites for nicer ones I took from Bay.
- Allows use of prosthetic liver and kidneys in character setup.


Anewbe:
Removed the Lower Body code, can't make it work in a reasonable timeframe, don't feel like trying right now.

PR #3986, opened 2017-09-30
Alt-Title Tweaks

SpadesNeil:
- ~~Cadet added to Assistant alt titles, because no assistant title exists for an aspiring Sec Officer.~~
- ~~Site Manager -> Colony Manager. I just thought it sounded better.~~
- ~~Overseer -> Colony Overseer. Again, just thought it sounded better.~~
- Added ~~"Colony Commander" and~~ "Site Supervisor" to Colony Director's alt titles.
- "Deputy Supervisor", "Deputy Director" added to HoP titles.
- "Administrative Assistant" added to Command Secretary.
- "Barkeeper" added to Bartender.
- "Gardener" is now an alt title. Botanist is the new default title. Gardener really isn't the correct title for what the role does by default, anyway.
- "Logistics Manager" added to Quartermaster.
- ~~"Caretaker",~~ "Maid" added to Janitor.
- "Cleric" added to Chaplain.
- "Head of Research" added to Research Director.
- Added "Cargo Courier" to Cargo Technician.
- ~~"Geologist",~~ "Physicist" added to Scientist.
- "Prosthesis " added to Roboticist.
- ~~"Biomechanical Engineer" -> "Biomechanical Technician" for Roboticist.~~
- "Commander" -> "Security Commander" because "Commander" was confusing for the HoS.
- ~~Added "Security Overseer" to HoS.~~
- Warden finally has alt titles of "Armorer", "Brig Manager"

I'm not really attached to any of these except I insist that "Commander" should be "Security Commander" because "Commander" alone is super confusing and also not correct terminology in the context of a research facility.

PR #4026, opened 2017-10-05
Sensible Job Titles

SpadesNeil:
Basically #3986 except condensed to the ones that were actually good, plus one more.

- Gardener is now Botanist. Gardener has been moved to an alt title.
This is more fitting for what the role actually does. Besides, who calls them the gardener? I call them botanists.
- The alt title Commander is now _Security_ Commander, to avoid confusion. The title of "Commander" alone implies a role equal to Captain or Colony Director, which is not correct.
- ~~Research Director is now Research _Supervisor_ by default which was formerly an alt-title. It's also much more fitting for what the Research head of staff actually does, which is generally just supervise.~~
- ~~_Head of Research_ has been added as an alt title in place of the old alt title. Either of these could be the new default, I don't care which.~~
- ~~Research _Director_ has been removed entirely to stop confusion with Colony Director.~~


SpadesNeil:
Also, can I add Deputy Director back to the HoP's alt titles? Because I feel it fits. @Anewbe @Neerti

Cerebulon:
I'm not a fan of removing RD as a title entirely. Whether they "just supervise" or actively organize research efforts is entirely up to the player (who can pick an alt title to match if they like) and not something built into the role. My RD character(s) have always been more on the active involvement side.

Anewbe:
I uh... Sort of specifically requested that you only change the two titles, being Security Commander and Gardener. This'll need to go back for review, again.

SpadesNeil:
I was under the impression that you disliked the confusion between RD and CD so I sought to fix that.

Anewbe:
I do, but that's not how I'd like to fix it, and I don't have a great need to fix it at this moment.

SpadesNeil:
¯\_(ツ)_/¯

Techhead0:
"Maid" got axed? I was kinda-sorta looking forward to knowing it existed somewhere.

Anewbe:
@SpadesNeil There's a PR in the works that is going to touch these files, probably, and I've specifically been waiting for this PR for a while now, so your PR will probably be merged after that is. Other than that I don't see a reason not to merge this, so hang tight.

Anewbe:
@SpadesNeil Conflicts, sorry. Would you mind terribly giving this a fix?

PR #4029, opened 2017-10-07
Adds Skeletons for background use

Anewbe:


Belsima:
What sprites do the skeletons use?

PR #4048, opened 2017-10-12
Adds a Trait system

Neerti:
Because apparently there's still demand for this (mostly from Anewbe).  I dug this out of code from code I wrote 4 months ago and implemented most of the traits I wrote for this.

The fact that most of these are negatives/neutral/has no effect is intentional, since I want traits to make it easier to express a character concept as opposed to a min-maxing tool like most other games with a trait system (nothing against them, in fact I personally love doing this myself but I don't wanna do it on Polaris).

The mental traits don't have any mechanical effect and just give progressively more distressful messages to you in the chatlog if near something that 'spooks' your character.

Pictures:
Physical traits
Mental traits


Anewbe:
<3

ElectricB1ue:
YES COOL

mistyLuminescence:
Same feedback as on discord, I _love_ the idea, it's thematic as hell and really interesting; only thing I'd change is having the flimsy/frail traits increase incoming damage instead of lowering max HP, and adding everyone's traits to their medical records by default.

SilverTalismen:
What does the cloned trait do exactly?

Anewbe:
Nothing.

elgeonmb:
GOOD SHIT OP

elgeonmb:
also are flimsy and frail mutually exclusive or can I make a Teshari with 20 or 24 health

code/modules/client/preference_setup/traits/trait_defines.dm
```diff
@@ -0,0 +1,212 @@
...
+
+
+/datum/trait/modifier/physical/cloned
+	name = "Cloned"
+	desc = "At some point in your life, you died and were cloned."
+	modifier_type = /datum/modifier/cloned
+
+/datum/trait/modifier/physical/cloned/test_for_invalidity(var/datum/category_item/player_setup_item/traits/setup)
+	if(setup.is_FBP())
+		return "Full Body Prosthetics cannot be cloned."
```
elgeonmb: You can put a cloned brain in an FBP

code/modules/client/preference_setup/traits/trait_defines.dm
```diff
@@ -0,0 +1,212 @@
...
+
+// 'Mental' traits are just those that only sapients can have, for now, and generally involves fears.
+// So far, all of them are just for fluff/don't have mechanical effects.
+/datum/trait/modifier/mental
+	category = "Mental"
+
+/datum/trait/modifier/mental/test_for_invalidity(var/datum/category_item/player_setup_item/traits/setup)
+	if(setup.is_FBP())
+		if(setup.get_FBP_type() == PREF_FBP_SOFTWARE)
+			return "Drone Intelligences cannot feel emotions."
```
elgeonmb: IDK about adding that given those pesky pesky A-classes, but it might be appropriate the whole "can't feel emotions but can fake them pretty well" thing they've got going on.

(Also if we add racism mental traits I don't want to deprive people of their racism bots)

Anewbe: I'm not sure I agree with this point.

code/modules/client/preference_setup/traits/trait_defines.dm
Anewbe: If it eventually has a mechanical effect other than acting as a death counter, we'd rather not have it affect FBPs.

PR #4060, opened 2017-10-13
Adds the Ionic Rapier

Neerti:
Similar to an esword, but lethal towards mechanical threats, and rather weak against pure organics.  It's very strong against simple animal synthetics like hivebots.
Ports /tg/'s esword and double esword sprites, both item and inhands.
Hivebot isSynthetic() now returns true.
Picture of rapier


elgeonmb:
Is this gonna be antag-buyable; I need this to go on racially-fueled rampages.

Wutnold124:
if it ain't buyable it's prolly gonna be a trade-in

PR #4062, opened 2017-10-13
Loadouts should be fixed.

Anewbe:


Anewbe:
Panicked Self-Merge go.

PR #4067, opened 2017-10-14
Port over Chemistry bag from Paradise

Anasari:
Had to make a new PR because I messed up my commit and I can't code.

Port over chemistry bag from Paradise, based on /tg/ chemistry bag. Based on [Paradise #2610](https://github.com/ParadiseSS13/Paradise/pull/2610).

This add in the chemistry bag, an attack_by exception for glass bottle (Which won't work otherwise) & add it to the chemistry wardrobe. It also add it to the biogenerator.

Should make chemistry less miserable. No more putting glass bottles two by two into the chemfridge. Or pills.

Now come with a slowdown and large itemsize!


Anewbe:
You test all of this?

Anasari:
It's been tested. 

The addition to chemistry wardrobe: 

Loading up a bunch of random bottles & pills:

And then dumping it into the chem-fridge:

And the chem-bag:

Anewbe:
Cool, gonna give it the ol' 24 hour sit and chill, but it looks fine as far as I can see.

Wickedtemp:
Yeah, adding a slowdown to it wouldn't be a bad idea. You'd only be walking a few tiles with it, so it wouldn't have any real effect on it's usefulness in Chemistry, while still being a considerable nerf for anyone using it outside of the lab.

If not, it's whatever. Usually, I don't go through a full bottle of any medicine during a round anyway, so it doesn't really matter if I carry 1 or 10.

Anasari:
My experience as a chemist on where I came from is that I seldomly need to carry around the chem-bag myself, as I do not need to carry more than my usual loads of 14 synthflesh patch, 7 perf and 7 pentetic. I'll see if I can add in a slowdown on it later on though. It'd be good to learn!

Anasari:
Changes made. The large size is to prevent the slowdown from being circumvented by simply putting it in a backpack

Issue #4035, opened 2017-10-08
Corpses are invincible shields again!


FenodyreeAv:
#### Brief description of the issue
People in aggro-grab, even dead corpses, block bullets with a very high % chance.

#### What you expected to happen
When a body in an agro-grab falls over, or is behind the one grabbing it, it shouldn't block bullets anymore.

#### What actually happened
Being shot in the back, with the corpse on the other side of you, uses the corpse as a shield anyway.

#### Steps to reproduce
Kill a monkey, become UNSTOPABLE.

#### Additional info:
 - **Server Revision**: Found using the "Show Server Revision" verb under the OOC tab.
 - **Anything else you may wish to add** (Location if it's a mapping issue, etc)

PR #4073, opened 2017-10-15
Dead bodies are no longer perfect shields.

Mechoid:
Using a dead body as a shield is no longer a perfect defensive option, hovering near 60 percent block-rate for humans, and 30 for Teshari and Prometheans, because corpses are a lot of dead-weight for a small creature or jello creature to hold up. Prone-living persons are still counted as a meatshield, due to the fact that the hostage taker cannot force a hostage to stand, allowing the hostage to negate their risk-of-harm by lying down.
Edit: Fixes/Touches on https://github.com/PolarisSS13/Polaris/issues/4035

### DIFF
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -235,9 +235,18 @@
 			//if they have a neck grab on someone, that person gets hit instead
 			var/obj/item/weapon/grab/G = locate() in M
 			if(G && G.state >= GRAB_NECK)
-				visible_message("<span class='danger'>\The [M] uses [G.affecting] as a shield!</span>")
-				if(Bump(G.affecting, forced=1))
-					return //If Bump() returns 0 (keep going) then we continue on to attack M.
+				if(G.affecting.stat == DEAD)
+					var/shield_chance = min(80, (30 * (M.mob_size / 10)))	//Small mobs have a harder time keeping a dead body as a shield than a human-sized one. Unathi would have an easier job, if they are made to be SIZE_LARGE in the future. -Mech
+					if(prob(shield_chance))
+						visible_message("<span class='danger'>\The [M] uses [G.affecting] as a shield!</span>")
+						if(Bump(G.affecting, forced=1))
+							return
+					else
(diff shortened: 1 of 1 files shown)


code/modules/projectiles/projectile.dm
```diff
@@ -235,9 +235,18 @@
 			//if they have a neck grab on someone, that person gets hit instead
 			var/obj/item/weapon/grab/G = locate() in M
 			if(G && G.state >= GRAB_NECK)
-				visible_message("<span class='danger'>\The [M] uses [G.affecting] as a shield!</span>")
-				if(Bump(G.affecting, forced=1))
-					return //If Bump() returns 0 (keep going) then we continue on to attack M.
+				if(G.affecting.stat == DEAD)
+					var/shield_chance = min(80, (30 * (M.mob_size / 10)))	//Small mobs have a harder time keeping a dead body as a shield than a human-sized one. Unathi would have an easier job, if they are made to be SIZE_LARGE in the future. -Mech
```
Anewbe: Uh... Wouldn't the opposite be true, in that a small mob is a less effective shield for a larger mob?

Mechoid: I only touched dead mob shielding, in that a small mob, if holding a dead body, is less able to use it as an effective shield since they can't lift it as easily compared to a larger one. If the probability goes through, it's as effective as a shield as it would be if the hostage was alive. I can make shielding itself check the comparative size, though that seemed out of scope for the referenced issue.

Anewbe: Yeah, that's fair, write that idea on the Misc projects page, if you can/don't mind?

Atermonera:
Prometheans call gib() anyways when they die, so I think that's a bit of a moot point

Mechoid:
For a Promethean holding a dead body as a shield, not being used as a dead shield. The mob who is being grabbed is not the body being checked. ..Unless I'm reading what I wrote horribly incorrectly.

G.affecting is the target mob, in this case the body. M is the grabber.

PR #4077, opened 2017-10-15
Rings

Anewbe:
It only took me three days to port Bay's rings.
- Made some modifications so they'll work with RIGs.


Wutnold124:
thanks dadnewbe

ElectricB1ue:
p cute

PR #4079, opened 2017-10-15
Prosthetic Organ Expansion

Mechoid:
- O_CELL define added, for the prosthetic cell and any future subtypes.
- MMI_Holder and Microbattery organ tags have been changed to use the correct tags.
- Prosthetic organs added to the prosthetic fabricator to match the bioprinter's ability to print organs.


Wutnold124:
neat

PR #4081, opened 2017-10-16
BYOND 512 Compiling Errors Fix

Cameron653:
With 512, they added "filter" and "filters" vars.
Upon attempting to compile the code in 512, you're met with a variety of errors due to this.
This PR goes in and changes everything that has "filter" and "filters" and gives them more appropriate var names, such as changing "filters" to "atmos_filters." and changing the filters path from /obj/machinery/atmospherics/omni/filter to /obj/machinery/atmospherics/omni/atmos_filter, among other things.

Code compiles fine and, as far as I can see, there's no issues. https://i.imgur.com/Yb6CUWS.png

With this, the server is able to be upgraded to BYOND 512 without errors.


Cameron653:
Edit:
As a side note, the reason the map was changed was due to the omni/filter being changed to omni/atmos_filter, the subspace crystals being changed from subspace/filter to subspace/sub_filter, etc.

Additionally, the changeling menu code uses "var filter" but it's in javascript and seems to not be affected by the new BYOND filter and filters changes, so I went ahead and left it.

Anewbe:
You committed a number of backup files, I'm not sure we're supposed to have those.

Cameron653:
Oh. Euugh, let me remove those. BYOND add those in every time I edit a map file whatsoever.

Anewbe:
Yeah, whenever I do map changes, the first thing I do after running the merger tool is revert and delete everything I know I didn't touch.

Cameron653:
Theeere we go. All the backup files removed.

Cameron653:
~~Please merge whenever possible before there's a conflict in one of the 33 files edited~~

Neerti:
Will need to go through all of our current map files but I can do that later in a separate PR.

PR #4088, opened 2017-10-17
These are not Vorestation edits.

SpadesNeil:
Y'all are confusing the shit out of us. We put those there to denote something we changed from default Polaris code.

Since this is all now default Polaris code, you shouldn't have these.


SpadesNeil:
I didn't mean to do that with the merges. <_>

PR #4100, opened 2017-10-18
Changes how IDscan works.

Belsima:
If you disable IDscan, rather than doing the same thing as bolting a door except shittier since they can just cut wires and crowbar it, it allows anyone through, giving it actual UTILITY for, as an example, emergency access.

Ported from Aurora.

### DIFF
--- a/code/game/machinery/doors/door.dm
+++ b/code/game/machinery/doors/door.dm
@@ -442,7 +442,7 @@
 
 /obj/machinery/door/allowed(mob/M)
 	if(!requiresID())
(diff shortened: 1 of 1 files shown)


Anewbe:
I mean... If you have the ability to change IDscan, you have the ability to open the door the current way.

Belsima:
It doesn't seem to confuse them in my experience- they just hack away at it with an axe. Simple enough.

Anyways, it's for those cases where you might wanna help people through somewhere in an emergency but also kinda DON'T wanna stare at the spot for 10 to make sure everyone who wants in gets through when they yell "AI DOOR"?

PR #4110, opened 2017-10-19
Stops MMIs created with a ghosted brain from being put into cyborgs.

Mechoid:
On the Tin. I didn't completely overwrite the check for safety reasons, and it may be usable at a later point.


Anewbe:
Tested, etc?

Mechoid:
Indeed.

PR #4112, opened 2017-10-19
High Capacity Water Tanks

Woodratt:
- Adds in code and sprites for high capacity water tanks. 5 times the volume of a regular one. (None mapped in)
- Ports water and fuel tank sprites from TG.

[Image](https://i.imgur.com/BOFzFMt.png)


Anewbe:
Image is 404ing for me.

Woodratt:
Try again

Anewbe:
Ah, now it works. Thank you.

Belsima:
I very dislike those sprites. Leave this open for a bit and I'll check whatever else I could find, please?

Belsima:
Sprites in my archive: image
Ones in this PR: image
Current ones: image

Anewbe:
Upon seeing the comparison, I'm actually a fan of Woodrat's selection the most, but I'd rather not use the ones Belsima has proposed.

Belsima:
Alright, fair enough.

Belsima:
Other than that, all I can find are this image and this image

Cyantime:
Belsima's last posted sprites are the best ones here.
I can't stand how TG's awful perspective and sprites keep spreading like some sort of disease into other codebases.

Atermonera:
I prefer Woodrat's sprites to the ones from Belsima's archive, and we shouldn't reuse the current ones for a different sized tank

Neerti:
Aren't the perspectives for Belsima's and /tg/'s the same?

code/modules/reagents/reagent_dispenser.dm
```diff
@@ -76,6 +76,14 @@
 		..()
 		reagents.add_reagent("water",1000)
 
+/obj/structure/reagent_dispensers/watertank/high
+	name = "high-capacity water tank"
+	desc = "A highly-pressurized water tank made to hold vast amounts of water.."
+	icon_state = "watertank_high"
+	New()
```
Neerti: Gonna nitpick and ask you not use relative pathing, instead using ``/obj/structure/reagent_dispensers/watertank/high/New()``, since our standards page says so.

code/modules/reagents/reagent_dispenser.dm
```diff
@@ -76,6 +76,13 @@
 		..()
 		reagents.add_reagent("water",1000)
 
+/obj/structure/reagent_dispensers/watertank/high/New()
+	name = "high-capacity water tank"
+	desc = "A highly-pressurized water tank made to hold vast amounts of water.."
+	icon_state = "watertank_high"
+		..()
```
mistyLuminescence: "code/modules/reagents/reagent_dispenser.dm:83:error: : invalid expression"

Might move New() from /watertank/high to before `..()` ?

Cyantime:
I said the last posted ones, these ones 
68747470733a2f2f692e696d6775722e636f6d2f734c664c5a66312e706e67

code/modules/reagents/reagent_dispenser.dm
Anewbe: This ..() should be back one tab, as should the thing below it.
And the add_reagent() call should probably only add 4000, instead of 5000, unless you want the big tanks to have 5000.

Belsima:
I really dislike the idea of using those tank sprites considering they follow a style I was told specifically not to port in the past (and which I personally don't like myself)- I believe these are Ausops sprites, too?

elgeonmb:
gonna agree w/ not using those weird tg sprites plz

Belsima:
Any word on the sprites?

arsserpentarium:
Please, don't forget to modify farmbot code to make this tanks acceptable.

code/modules/reagents/reagent_dispenser.dm
```diff
@@ -76,6 +76,13 @@
 		..()
 		reagents.add_reagent("water",1000)
 
+/obj/structure/reagent_dispensers/watertank/high/New()
+	name = "high-capacity water tank"
+	desc = "A highly-pressurized water tank made to hold vast amounts of water.."
```
Belsima: Two periods. Should only be one.

code/modules/reagents/reagent_dispenser.dm
```diff
@@ -76,6 +76,13 @@
 		..()
 		reagents.add_reagent("water",1000)
 
+/obj/structure/reagent_dispensers/watertank/high/New()
+	name = "high-capacity water tank"
```
Belsima: It doesn't seem to actually be high-capacity, is it? Would it only be high-pressure? If the capacity is in fact larger and you can in fact empty it and refill it to 4000 units rather than 1000, please do correct me.

Anewbe: Water tanks are just big reagent containers. It's seems that you're thinking about air tanks.

Belsima: I'm aware- the default allow you to fill it to 1000 unit and the desc states this one is high-pressure. The name should probably be high-pressure unless you can also empty and refill it to 4k units instead of 1k.

Anewbe: You can't refill water tanks.

Belsima: Huh. I might be confusing it with another codebase, then.

Belsima:
So... what happened regarding the complaints with the sprites?

Anewbe:
Sprite was changed in the last commit.

Belsima:
Ah, my bad. Displays as 1 commit, assumed no change was made.

Belsima:
Could I get a picture of the sprites used?

PR #4127, opened 2017-10-21
Incendiary >= 2 projectile fixes.

Mechoid:
Myriad of Fixes:
- Flamethrower fireballs will not travel in vacuum.
- Flamethrowers, and any other incendiary 2 or greater weapon will not put fuel on dense turfs.
_I don't knnow why it has the first two commits as well, but they don't do anything different._


Anewbe:
Did you create the branch off of your updated local master?

Anewbe:
I might have a different way to do this, actually, I'll try to tell you about it when I'm not waiting for an oil change.

Issue with this is that asteroid floors are (for some reason?) a child of asteroid walls, so now the flamethrowers wont leave a trail on floor rock.

Mechoid:
Was previously an updated local master, though it fell behind and I had to re-update it. Shoooould have a much nicer way of doing the check sequence now?

PR #4132, opened 2017-10-21
Replaces headache-inducing sound.

Belsima:
Ported from Bay- the old sound was that high pitched ambient hum that forced me to toggle ambiance off because it triggers headaches. Every. Time.


Cyantime:
In what world is shipambience.ogg high pitched? It's the most calming sound on the station and I disagree with removing it in any situation.

Belsima:
Actually- wait. Better ambiance ideas incoming.

PR #4134, opened 2017-10-21
Removes duplicate internal organs

Cameron653:
Heart is actually defined here
https://github.com/PolarisSS13/Polaris/blob/master/code/modules/organs/internal/heart.dm
- Removes duplicate internal organs and leaves comments of where to actually find them.
- The file was commented out, but contained old code.


Wutnold124:
god i wish that were me

Cameron653:
...mark this DNM I've been notified that _this is the case for a bunch of organs._ Going to fix.

Cameron653:
Anewbe you got it RIGHT after I finished it ahhh.

Anyways this isn't WIP anymore. Good to go.

Anewbe:
Did you double check that those definitions are actually elsewhere, and did you test that code doesn't expect them to be in this file?

Cameron653:
Yes. I made sure that they were defined elsewhere. They'll all conveniently in the files contained in the 
https://github.com/PolarisSS13/Polaris/tree/master/code/modules/organs/internal
folder.

Anewbe:
o7

Wutnold124:
the hero we need, but not the one we deserve

thanks Cameron653, you cured Multipleorganitis.

Cameron653:
I was trying to figure out how to tear out organs (Kali ma) and saw that the heart was defined twice and led me on this tangent.

Cameron653:
AS IT TURNS OUT, internal/organ_internal is the one that is checked, this one isn't in the .dme

Not actually that big of an issue though as the comments will direct people to the correct place of where to find it, instead of having old, unused code and processes.

PR #4135, opened 2017-10-22
Level Two Infections actually hurt now

Anewbe:


Wickedtemp:
So, going from causing no toxin damage to causing a small amount?

Anewbe:
Yeah, so infections aren't "I'm fine, I'm dead".

Wickedtemp:
Well, you should be having a fuck-ton of pain messages and such, so it should still be evident that they aren't fine. But yeah, I wouldn't say no to more toxin damage on lesser infections. All the more incentive to go to medical *before* they progress to stage three.

PR #4136, opened 2017-10-22
Hopefully fixes ahelps

Anewbe:
Current guess is that I somehow woke up the AFK checker for admins?


elgeonmb:
anewbe why do you keep doing this

Anewbe:
Because Woodrat wants Event Managers, and I want Event Managers, and trying to test AFK code alone is like trying to give yourself a back-rub.

PR #4137, opened 2017-10-22
Cloning machine now auto process meat

Anasari:
As per title. This port in the feature from Paracode (look i'm not gonna check where) where cloningpod sucks in meat automatically (for 50 biomass each.)

Any meat will work, alongside meatpie, meatsteak and meatball due to similar looking typepath. Shouldn't cause an issue though, no one's gonna feed meatpie to a cloning pod. 

Probably.

Mandatory testing screenie:

### DIFF
--- a/code/game/machinery/cloning.dm
+++ b/code/game/machinery/cloning.dm
@@ -172,9 +172,13 @@
 //Grow clones to maturity then kick them out.  FREELOADERS
 /obj/machinery/clonepod/process()
 
+	var/visible_message = 0
 	for(var/obj/item/weapon/reagent_containers/food/snacks/meat in range(1, src))
 		qdel(meat)
 		biomass += 50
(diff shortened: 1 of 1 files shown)


Anewbe:
I'll bring this up to the Staff, but I'm not sure what purpose it really serves?

Wickedtemp:
This would have been extremely useful before synth-meat production was given a nerf. Currently, using an entire bottle of clonexadone will get you 12 pieces of meat, rather than the previous 60.

Anasari:
Reduce the tedium of stuffing meat in one by one into the cloning pod.

Anewbe:
Did you test this?

Anasari:
https://i.imgur.com/mkzmlkI.png

Works as intended.

Anewbe:
Can you bracket that message with `<span class='notice'> </span>`, please? Other than that, looks good.

PR #4142, opened 2017-10-22
Adds sign sprites and replaces some.

Belsima:
image image image 

Ported from /vg/station, Baystation12, /tg/station, and the CEV-Eris.

Medical signs replaced with blue color scheme. Cargo sign updated to be less horrendously ugly with its coloration. External airlock sign updated to be more specific to actual external airlocks and not just any door. Icon states added for many warning signs. Other signs added but not currently mapped in..


Belsima:
...oh, right. I actually DO need to map it. Gimme a bit.

Belsima:
Mapped in and ported image from Paradise.

Belsima:
Oh, what the fuck? I used the map merge tools.

PR #4146, opened 2017-10-23
fixes exterior shields putting shields EVERYWHERE

Screemonster:
My last shield PR screwed up 'cause it didn't check neighbours properly, resulting in literally every "blocked" turf in range getting a hexagon slapped on it regardless of whether there was an adjacent station turf or not. This fixes that.

Also takes open space off the list because it breaks stairs whoops


Screemonster:
image

Here's a screenshot of it actually working this time

PR #4148, opened 2017-10-23
MedHUDs can identify FBPs

mistyLuminescence:
FBPs have a special icon on MedHUDs now (like people infected with virii or cortical borers); thanks to @Neerti for the HUD sprite. [Looks like this](https://cdn.discordapp.com/attachments/176163422751293441/372010121183625216/export2.png).

meet me in the pit travis


MisterLayne:
Don't know why this is controversial. Seems like a pretty good idea to me, honestly. It will make it much easier for medical staff to identify FBPs.

Anewbe:
Not everyone wants it to be immediately obvious that they're an FBP.

SilverTalismen:
But, a single look at medical or security records can also reveal this Anewbe.

Anewbe:
That takes a quarter of a second longer, so not everyone does it.

MisterLayne:
I don't think anyone would care if you knew they were an FBP or not when they're in a situation where they need immediate medical attention.

Anewbe:
People don't just wear the MedHud during emergencies.

mistyLuminescence:
It's controversial IC for those "I-I'm a human I swear" types, but it being controversial OOC confuses me a little. Surely detecting whether something is organic or not would be the one thing a MedHUD could do beyond any doubt? If people want to uphold their "I'm not an FBP honest" gimmick that bad, they can just avoid any medical staff they see.

Belsima:
I, personally, could only support this if there were a system for entirely hiding from medhuds- though I'd want that either way tbh.

Wickedtemp:
YES. YES YES YES. ALL OF MY YES. SO MUCH YES.

When you've been frantically ordered to respond to a medical emergency, you don't have time to check records, you need to *move*. When you arrive at the emergency and see one or several people face-down, with the HUD saying they're all in or near critical condition, you don't have time to read their records, you need to *act*. 

It is slightly irritating to discover that you were called to a medical situation, only to find that the patient in question was, in fact, an FBP. There have been times where Medical's broken into an area in an attempt to rescue what appeared to be a dying patient, who later turned out to be an afk FBP that ended up burnt from hacking a few airlocks or something, and in their rescue attempts, didn't have time to spend reading through their medical records. An icon on the HUD would have prevented the headache.

Plus, there's no real reason why this shouldn't be implemented anyway, other than "it needs to be coded", which is done now. If people wanted to play "Guys, look, I'm a perfectly organic human." FBP, then that's unfortunate because they're listed as an FBP in their records to start with. This change would take information that medical *already has* and simply make it a little more convenient to see, which is useful in emergency situations where Medical doesn't always have time to read through their files.

Curt5:
I struggle to find reasons to argue against this change. It the hectic, life or death seconds, you want to heal the bleeding but alive human before the crumpled and rebuild-able synthetic. Further, there are already so many ways to tell who are synths and who aren't, even from a distance- why not make it simple and easy so doctors can do their job better?

Woodratt:
Not a fan.

Natje0:
This would be great. I don't really see any reason why it wouldn't be done. Sure, it's mess with the "I'm not a synth!" gimmick, but as previously said... It's in the medical notes. This will be great for emergency situations, and for preventing FBP's being brought to medical by paramedics.

MisterLayne:
**oh, look at me, my name's mr. medical hud and I can't tell the difference between a human and FBP**
So far, we've got five STRONG yes, two 'meh' yes, and two 'meh' no..

zekesulastin:
... as much as I agree, MisterLayne, all your tone is doing is reinforcing the no vote from the people with the actual power to merge.

Lance77am:
First off, I'm personally in favor of the idea, but I understand the desire of not needing to give more information out than is really needed to be accessible quickly. I personally **never** look at somebody's medical records unless they're requesting something or I am giving them emergency treatment, or cloning them. Some players shouldn't have a portion of their character potentially invalidated strictly by the fact that someone else is wearing a bit of headgear, and this gets more applicable to say, mercenaries who are dealing with our numerous security synths (Not that they're often wearing a medhud anyway.)

My suggestion would be as follows: Use the icon replacement for people in critical condition only. It'd require a tweak of the current pull, but it would alleviate RP concerns between people, and limit the use of this feature to life-saving situations only. It'd come online when needed for medical, and go offline when not for RP. Because honestly, this addresses the vast majority of situations.

Healthy crowd: No problems, same as now.
Crowd mildly damaged: The same
Crowd of mixed critical/heavily injured: You can note which critical patients are pointless to treat, and which you need to focus on immediately, saving precious seconds. The heavily injured, which come immediately after, would have enough time permitted for the issue of non-instant species identification to be a nonissue. 

If that could happen, I see absolutely no reason to deny this. Otherwise, I **still** support it, as I also support patient confidentiality from an RP perspective and a doctor knowing your medical status on the fly is more useful more often than it will harm you (hopefully never). (Plus combatants don't wear medhuds, generally, and if they do, it still requires communication on their part with their team.)

Wickedtemp:
@zekesulastin That's likely true, but I'd personally hope that "To hell with the PR, someone supporting it was a little snarky." wouldn't be a deciding factor. Something like this should be left up to the decision of debate and discussion, with consideration granted to the playerbase voting for and against it. 

I'd be lying if I said I wasn't a bit irritated at how this has gone in the Discord chat. This was a feature that I'd personally been wanting for months, and it's finally coded and working, the PR is done, and the only opposing views have been the statement of "I don't want it.", usually with no "Because of..." after, and "It'd let Medical distinguish FBP's and Organics"... Which A) Is precisely the point of the PR, and B) Doesn't really hold water, as we already have the means of doing so, in a slightly less convenient manner, by manually checking medical records.

@Lance77am
That could work. Usually, Medical wouldn't be like "Dude, get here and let us help you with that, holy shit." unless the HUD's showing a yellow/orange, around half-empty healthbar or lower. Above that, Medical will generally just make a mental note and keep an eye on the patient.  You could even have this fluffed ICly as the HUD noticing exposed wires/mechanical bits due to the damage.

Razer69:
It's an HUD showing someone's vitals. What logic would apply that makes the thing unable to differentiate between human and robot?

ForFoxSake:
Since I regularly play an FBP in hiding, at first I reeled at the thought of this feature. I even went so far as to think "What if the robotic icon only showed up when the HUD detects that they're damaged?" knowing full well how difficult that would be to implement.

However, having given it a little more thought I feel it may be better to _also_ (in addition to adding the icon) completely remove the health bar from displaying for an FBP, as that would then send a far clearer signal to anyone performing triage to "ignore this person". It also makes more sense to me, than everyone here claiming "How can the hud not distinguish between humans and robots!" while they're ignoring the fact that it detects and displays damage to them in the exact same way.

tl;dr: Keep the new icons, remove health bar for FBPs. Make the medical hud finally make a bit of sense, lore wise.

Razer69:
>while they're ignoring the fact that it detects and displays damage to them in the exact same way.

I'm not sure how a device would detect damage on organs and a metallic structure in the *exact same way*.

Wickedtemp:
@ForFoxSake  @Razer69 
It's already essentially been decided to have the FBP icon display when the individual is injured. Tacking on any more to that wouldn't really be adding much except a longer delay until this is completed and merged.

Atermonera:
To be fair, having it not display an indicator if the patient is synthetic would be easier than only displaying it past a damage level, because the `if(patient.isSynthetic()) holder.icon_state = "hudrobo"` can just be rewritten to not show a hud instead of showing a special hud

mistyLuminescence:
I'm gonna try and get this sorted sometime this week, the hard part is just "detect how much health they have, and if it's below x amount, show this icon" - any suggestions on how to code it, let me know.

Atermonera:
If you sum
`/mob/living/carbon/human/getActualBruteLoss()
/mob/living/carbon/human/getActualFireLoss()
/mob/living/carbon/human/getOxyLoss()
/mob/living/carbon/human/getToxLoss()`
that'll give you the total of the four main damage types. There's getters for brain- and clone-loss but I don't think those're checked by huds anyways
Health as a value is determined by taking the species max health (100 for most, 50 or w/e for tesh) and subtracting the total of those four procs

Wickedtemp:
...They might detect genetic damage. They don't detect brain damage, however. 

Doesn't matter for this one though, since FBP's are synthetic and don't have genetic damage.

Anewbe:
@mistyLuminescence I'll do another review of this tomorrow. If the code looks like it works, I'll give it a trial run, with a set end-date. If you can figure out the fancy VeyMed stuff, that's cool too.

ForFoxSake:
@Atermonera has a fair point, it would be far quicker to have it skip displaying the health bar for fbps, than to add in a long conditional on if they're injured or not.

I'm extremely late in posting this, but I would like to see that suggestion taken up, as it leaves possible room in the future for roboticists to get their own hud that displays damage levels on things they can fix instead.

Atermonera:
The concern wasn't so much with only displaying the health bar if the fbp was injured enough, but rather to display the icon for veymed fbp's (The ones that are closest to mimicking outwards appearance of organics) only if they were damaged enough for it to be generally apparent that they're a robit, as there's a concern for some people who want to hide the fact that they're inorganic, since if it just displays the icon then people playing med can just casually observe that they're inorganic.

Anewbe:
**EXPERIMENTAL END DATE: DECEMBER 3RD 2017**

PR #4163, opened 2017-10-26
Port paper plane from /tg/

Anasari:
As per title. This port over the paper plane from /tg/. They are made by alt-clicking a piece of paper.

This version don't deal any stun damage if it hit you in the eyes (2% chance).

Known issues:
1. Can't get it to change direction of the sprite unlike the original version
2. The hitting eye damage can proc twice. Seems to be a throw code problem.


code/modules/paperwork/paperplane.dm
```diff
@@ -0,0 +1,102 @@
+// Ported from TG. Known issue: Throw hit can possibly double-proc. Seems to be throw code.
+/obj/item/weapon/paperplane
+	name = "paper plane"
+	desc = "Paper, folded in the shape of a plane."
```
Belsima: `Paper folded into the shape of a plane.`

Sounds a bit better, but its up to you.

PR #4166, opened 2017-10-27
Adds New Advanced Shields, Tweaks

Neerti:
Adds several advanced versions of shield machinery, able to be ordered from cargo at a somewhat steep price.  The generators are more efficient.  The capacitors can store and input/output more power.
Active shield generators now glow a bright cyan color.

### DIFF
--- a/code/modules/shieldgen/shield_capacitor.dm
+++ b/code/modules/shieldgen/shield_capacitor.dm
@@ -4,7 +4,7 @@
 
 /obj/machinery/shield_capacitor
 	name = "shield capacitor"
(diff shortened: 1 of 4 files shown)


code/modules/shieldgen/shield_capacitor.dm
```diff
@@ -4,7 +4,7 @@
 
 /obj/machinery/shield_capacitor
 	name = "shield capacitor"
-	desc = "Machine that charges a shield generator."
+	desc = "a machine that charges a shield generator."
```
Anewbe: A machine

PR #4167, opened 2017-10-27
Armor soak should actually work for humans now

Anewbe:


Wutnold124:
...armor soak?

Anewbe:
A secondary armor system that we've had (apparently without it actually doing anything) for a while. After the percentage based reduction, the game checks for armor soak, the reduces the damage taken by that amount.

PR #4168, opened 2017-10-27
Proto/Autolathe lag fix

Neerti:
Makes the two machines concatenate a lot less and not make BYOND's super secret string tree cry, hopefully.


Anewbe:
Tested?

Neerti:
All the UIs appeared to be the same as before the change, and in theory it should be less laggy due to filling BYOND's super secret string tree less.

PR #4170, opened 2017-10-27
weapon.dm fixes

Belsima:
Fixes grammar, and canes no longer fit in your pocket.


GMR25:
I sense a clown in denial.

PR #4175, opened 2017-10-27
Successful defibs are more damaging now

Anewbe:
Up to 10, from 5.
Will bring it up at the Nov 6th Staff Meeting.

### DIFF
--- a/code/game/objects/items/devices/defib.dm
+++ b/code/game/objects/items/devices/defib.dm
@@ -210,7 +210,7 @@
 	var/cooldowntime = (6 SECONDS) // How long in deciseconds until the defib is ready again after use.
 	var/chargetime = (2 SECONDS)
 	var/chargecost = 1000 //units of charge
-	var/burn_damage_amt = 5
+	var/burn_damage_amt = 10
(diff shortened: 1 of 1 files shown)


Wickedtemp:
Usually, burns only occur when a person is defib'd multiple times, or if the defib malfunctions, or if conductive gel isn't utilized properly. Self-adhesive defibs have gel already applied, however paddle-defibs lack it, the gel needs to be applied to the patients chest, or to the paddles, usually in the form of a pad you place on the paddle directly.

I'd support this change *if* Medical is also provided with electroconductive gel that, when used on the patient, or the paddles themselves in the form of disposable gel pads, would minimize the burns done by the defib.

Nalarac:
The purpose of the burns is to stop people from defib spamming a patient.

Wickedtemp:
Then have the burns kick in on the second defib, I guess. I'd be cool with that as well. I just don't want to run into "Oh, hey, we totally would've been able to defib you if the burns from the defib itself hadn't killed you again."

Anewbe:
I have other plans, now.

PR #4176, opened 2017-10-28
Swaps non-mounted recharging guns to batteries

Anewbe:
Should recharge at roughly the same rate as currently.
Long-Term Goal: Get various slowly-recharging cells for RnD to print.

- [x] Still need to double check some things
- [x] Still need to finish discussion


Belsima:
So, recharge-time is just gone?

Anewbe:
No?

Belsima:
Ah- my bad. Misread the code.

code/modules/power/cells/power_cells.dm
```diff
@@ -1,137 +1,92 @@
...
+	charge = 0
+
+/obj/item/weapon/cell/infinite
+	name = "infinite-capacity power cell!"
+	icon_state = "icell"
+	origin_tech =  null
+	maxcharge = 30000 //determines how badly mobs get shocked
+	matter = list(DEFAULT_WALL_MATERIAL = 700, "glass" = 80)
+
+	check_charge()
```
Mechoid: Not swapped to absolute pathing?

Anewbe: Ah, good catch, that was a straight copy-paste from the old file.

PR #4179, opened 2017-10-28
It's time for hair color revolution here too!

Verkister:
-Aye greetings from downstream hell where I managed to solve the locally intense ADD vs MULTIPLY blend mode discourse once and for all with a fresh as hell new option that combines the best of both worlds without the downsides.
-You can now have hairstyles that retain both, shading and highlights in every color how cool is that?
-Got told to port this stuff up here too so you're welcome. Spread the word if ya want. Planning to hit up the other blend stuffs as well some time, but here's a little something for now.

kuva
kuva
Yeah it works just like that in-game too and looks great on every hairstyle and color I briefly managed to test.

Also Scree's handy lion demonstration of this stuff: https://imgur.com/a/cjGPC


Anewbe:
So wait, does this allow multi-color highlights, or does it just make our hairstyles prettier?

Verkister:
Pretty much the latter, combining the modes as multiply mode alone works well for shading but breaks highlights and add mode alone works well for highlights but breaks shading. In this mode pure white still has shading and pure black still has highlights.

Anewbe:
Alright.

Verkister:
Actually I think it could allow multi-color highlights if those were included in the add mode sprites at least to some point heh.

Anewbe:
Something to look into eventually, I guess.

PR #4181, opened 2017-10-28
Adjusts how radiation works, especially for FBPs

Anewbe:
- Fixes the Recharge Station so it actually repairs the odd damage type.
- FBPs can actually take ToxLoss now, but reagents still shouldn't affect them.
- FBPs now take full damage from radiation. **Wear your PPE!**
- Radiation now goes up to 250. If you're above 150 rads you're going to be in trouble. **Wear your PPE!**
- FBPs with moderate tox damage will occasionally lose their ability to walk in a straight line, and at higher levels will occasionally fall over entirely.

- [x] Add Toxloss removal to the Jumper Cables
- [x] Click-Drag to Rechargers
- [x] Adjust Phoron Contamination
- [x] Hunt down stray sources of ToxLoss


Natje0:
I really don't see how it makes sense for robots to get poisoned... Nor do I get why radiation would affect them. 
But, I also don't really know how radiation work in reality.

Dhaeleena:
Wait what why no, why, wait what, WHY
Radiation not killing you is a feature from FBPS and when you have too much radiation it ALREDY makes you hallucinate and walk in random directions and have camera view swaps.

Anewbe:
Gearshy is right, the hallucinations and such are from looking at the SM.
And the radiation not bothering you is a bug, as far as I'm aware. The Robot Analyzer has a number for it, and long ago radiation made them stumble around and fall over. I'm not sure why it was removed.

Atermonera:
From a matsci standpoint, radiation hardened electronics aren't necessary for a lot of stuff and are probably significantly more expensive than not-rad-hardened electronics.

Anewbe:
Software damage, Ater, not hardware.

Anewbe:
Or something like that, I ain't a computer engineer.

Yoshax:
I don't really think this makes sense. I feel like it's a little much. Radiation should maybe have _some_ effect on them, but I kinda feel like all of what's done here is too much.

Anewbe:
What's done here is really very little, in my mind. It's a 10 percent chance to get dizzy for a few ticks, and as long as you don't stand next to the engine for a fairly long time you can just get fixed up by going into a recharger.

That said, how would you do it? I'm all ears.

Yoshax:
I think that, even if it has been a bug, it's been a thing for so long that doing change now is quite odd. I could be missing something, but is the only way for them to get rid of the toxins damage once they have it by going into a charger?

Anewbe:
For the time being, yes.

Yoshax:
It seems quite odd to me with the damage split and ways to heal them. Doesn't seem immediately obvious or sensible, either. Are there not other, perhaps obscure sources of toxin damage that would work on them also because of this?

Atermonera:
Specific cases of chemical ingestion/injection (Acids, radium, uranium), (possibly) floral somatoray, and metal slimes come to mind

Anewbe:
Made it so jumper cables remove all of a revived FBP's toxloss.

Neerti:
FBPs don't process reagents, so poisons shouldn't cause toxloss.  Metal slimes don't inflict extra tox loss iirc and they just have extra health compared to the others.

Atermonera:
They aren't affected by poisons, no, but strong acids would likely cause hardware damage to the affected portion even more so than radiation, mostly because of intensity and the inherent reactivity of the materials used in FBP construction

Atermonera:
Most of my ideas as additional sources of toxloss would probably requiring varying amounts of spaghetti code, given the nature of this

Atermonera:
This should probably get a changelog in addition to the announcement, given the scale of its effects

Cameron653:
I can agree with radiation dealing damage to synths. Sure, it fries their circuits.

But making it so they can take toxin damage from everything, like wearing phoron contaminated clothing? That's... I _really_ don't like the idea of that. 

The 150 radiation change and 250 radiation max is fine, but synths taking damage from _any_ toxin source is just... Eugh. Please no.

Arokha:
I'm not exactly convinced that there won't be strange other sources of toxins around. Like does phoron gas exposure cause toxins damage not via reagents?

Anewbe:
Not in my experience.

Arokha:
Cameron's phoron clothing thing is problematic, though, as it causes toxin damage directly w/o the use of reagents.

Cameron653:
Relevant section of code:
https://github.com/PolarisSS13/Polaris/blob/master/code/modules/mob/living/carbon/human/life.dm#L823-L828

Dankman:
All this does is nerf fbp's for no discernable reason other than "muh realism in spacemans"

Anewbe:
Thanks for the link @Cameron653 , I'll look into that as soon as I get back to this code.

Anewbe:
@Cameron653 https://github.com/PolarisSS13/Polaris/blob/master/code/modules/mob/living/carbon/human/life.dm#L817
So yeah, phoron contamination already doesn't affect FBPs.

Cameron653:
Oh, good to know. 

With that out of the way and there not really being any other sources than rads to give toxloss (and energy darts but I don't imagine that's used much, and if it'd even effect synths.) I can't see too any issues with this.

Mechoid:
Right. To clear some stuff up. Again, this is a bugfix. The only things that cause this should be:
- Radiation; storm, supermatter, uranium, or technomancer's corona
- Phoron pistol, Somatoray on mutation, energy bows, and the genetic paste gun (decloner).
- Instability
- Hardcrit as an FBP.
If others pop up, please tell us, they're more than likely a bug.

PR #4182, opened 2017-10-28
Adjusts simple mobs

Anewbe:
- Lowers the delay between attacks, and makes the min and max wait into vars
- Simple Mobs can now be emp'd, as long as they're synthetic (ie. Hivebots)


code/modules/mob/living/simple_animal/simple_animal.dm
```diff
@@ -1448,6 +1450,19 @@
...
 		adjustFireLoss(agonyDam)
 
+/mob/living/simple_animal/emp_act(severity)
+	if(!isSynthetic())
+		return
+	switch(severity)
+		if(1)
+			adjustFireLoss(rand(15, 25))
+		if(2)
+			adjustFireLoss(10, 18)
```
Neerti: I think you're missing a ``rand()`` here?

Anewbe: Ah, good catch.

PR #4183, opened 2017-10-29
Added a set of atmospheric analysis sensors...

Dunkhan:
Added a set of atmospheric analysis sensors to the integrated electronics module

Consists of 7 new input parts. Pressure, temperature, o2 n2 co2 and phoron levels. The last part is a general atmospheric analyser which has all 6 of those functions in one unit. The multi-purpose part uses triple the complexity and power so it is only worthwhile if you need more than 2 of the output pins.

### DIFF
--- a/code/modules/integrated_electronics/subtypes/input.dm
+++ b/code/modules/integrated_electronics/subtypes/input.dm
@@ -619,3 +619,235 @@
 	activate_pin(2)
 
 
+/obj/item/integrated_circuit/input/atmo_scanner
+	name = "integrated atmospheric analyser"
+	desc = "The same atmospheric analysis module that is integrate into every pda.  \
+	This allows the machine to know the composition, temperature and pressure of the surrounding atmosphere."
+	icon_state = "medscan_adv"
+	complexity = 9
+	inputs = list()
+	outputs = list(
+		"pressure"       = IC_PINTYPE_NUMBER,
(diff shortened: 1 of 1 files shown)


Dunkhan:
Please note: There are no changes to the icons, all parts use an existing icon - the one from the advanced medical sensor. I did not feel that this was important to change but someone will probably want to at some point.

code/modules/integrated_electronics/subtypes/input.dm
```diff
@@ -619,3 +619,235 @@
 	activate_pin(2)
 
 
+/obj/item/integrated_circuit/input/atmo_scanner
+	name = "integrated atmospheric analyser"
+	desc = "The same atmospheric analysis module that is integrate into every pda.  \
```
Belsima: integrated

PDA

also im not sure thats true either

Dunkhan: Could you elaborate? I am not sure what you mean.

Belsima: PDAs only sometimes do; not all of them do. 'every' PDA is incorrect.

Neerti: PDAs still have an atmos sensor function, and its the only portable one that auto-updates iirc.

PR #4185, opened 2017-10-30
Gun changes

Anewbe:
**If I did this right, there will be minimal change noticed player-side.**

Most of this is ported from Bay
- Fire sound can now be attached to projectiles, rather than guns
- Tweaks the way two-handed guns work, doesn't enable any of them yet
- Assorted background tweaks to the way guns work. Seemed fine in testing.
- Can no longer shoot across Z-levels. This seems like a wise balance decision, but probably requires Staff Discussion

This is my own nonsense
- Converts Evasion and Accuracy to use percentage, rather than increments of 15.
- [x] Missed a spot in ported BayCode, will correct.
- [x] Need to correct the gyroget projectile, did it backwards.


Yoshax:
One quick question. How does this affect chameleon guns? They've always taken the sound of the gun they're firing when shot, which was kinda cool for deceit and such. Will they just use the sound of the dinky little bullet thing they fire?

Anewbe:
I'll double check that, good catch.

Belsima:
Just checking if this is still being looked into, given the 3 month gap?

Anewbe:
Yep.

code/game/objects/items.dm
```diff
@@ -1,658 +1,681 @@
-/obj/item
```
Anewbe: This whole file was deleted during one of the master merges, for reasons I do not understand. It **should** be more or less a clone of what it is at present, with two procs added and the definition for /obj/item/device moved to the bottom of the file.

code/modules/projectiles/guns/projectile/automatic.dm
```diff
@@ -35,11 +35,11 @@
 	load_method = MAGAZINE
 	magazine_type = /obj/item/ammo_magazine/m10mm
 	allowed_magazines = list(/obj/item/ammo_magazine/m10mm)
+	projectile_type = /obj/item/projectile/bullet/pistol/medium
 	auto_eject = 1
 	auto_eject_sound = 'sound/weapons/smg_empty_alarm.ogg'
 
-//	requires_two_hands = 1
-	one_handed_penalty = 2
+//	one_handed_penalty = 15
```
Atermonera: 2*15=30?

code/modules/projectiles/guns/projectile/automatic.dm
```diff
@@ -62,13 +62,14 @@
 	load_method = MAGAZINE
 	magazine_type = /obj/item/ammo_magazine/m545
 	allowed_magazines = list(/obj/item/ammo_magazine/m545)
+	projectile_type = /obj/item/projectile/bullet/rifle/a545
 
-	one_handed_penalty = 4
+//	one_handed_penalty = 30
```
Atermonera: 4*15=60?

code/modules/projectiles/guns/projectile/automatic.dm
```diff
@@ -363,12 +367,13 @@
 	load_method = MAGAZINE
 	magazine_type = /obj/item/ammo_magazine/m762
 	allowed_magazines = list(/obj/item/ammo_magazine/m762, /obj/item/ammo_magazine/m762m)
+	projectile_type = /obj/item/projectile/bullet/rifle/a762
 
-	one_handed_penalty = 4
+//	one_handed_penalty = 45
```
Atermonera: 4*15=60?

code/modules/projectiles/guns/projectile/automatic.dm
Anewbe: This is the submachine gun Mercs get, and I don't want it to be unusable with the e-shield.

code/modules/projectiles/guns/projectile/automatic.dm
Anewbe: I felt that 60 was a bit much, honestly. This is an antag-only assault rifle, I feel like they're allowed to cheat a little. Not married to it, though.

code/modules/projectiles/guns/projectile/automatic.dm
Anewbe: 60 seems very, very high for what is essentially a shortened rifle. 45 is still decently punishing, and until any of these are actually activated I'm not sure it really matters, one way or another.

Atermonera:
Right, numbers of questionable maths were intentional, and can be tweaked if/when they're enabled. Code is otw fine

PR #4186, opened 2017-10-30
SC Arrivals Shuttle fix

Woodratt:
- Fixes issues with the SC arrivals shuttles in regards to airlocks
- New Shuttle seats, ported from Vorestation


code/game/objects/structures/stool_bed_chair_nest/bed.dm
```diff
@@ -46,7 +47,8 @@
 	var/cache_key = "[base_icon]-[material.name]"
 	if(isnull(stool_cache[cache_key]))
 		var/image/I = image('icons/obj/furniture.dmi', base_icon)
-		I.color = material.icon_colour
+		if(applies_material_colour) //VOREStation Add - Goes with added var
```
Anewbe: The Downstream doesn't like when we leave their comments in for things we steal.

PR #4188, opened 2017-10-31
Virtual Reality

Atermonera:
Core systems are all implemented.
Nothing is mapped in for this, but mapping it in is really easy. You just need to spawn `/obj/effect/landmark/virtual_reality` in wherever you want VR mobs to spawn, and name them so players can tell them apart, and put as many VR sleepers as you want players to be able to access the given map somewhere normally accessible. The VR area should be unreachable outside of VR, unless you liked Inception.
VR mobs currently spawn as their own species, which is a child of prometheans, with only some of the benefits and abilities. Their only internal organs are currently a brain and eyes.
Exitting VR can be done by using the exit_vr verb, or ghosting. Aghosting functions normally.
If someone tries to eject the sleeper while a player is in VR, it'll prompt the player if they want to leave before being able to remove them.
When you exit VR, there's still a link between the original mob and the VR mob, allowing you to reenter VR and use the same mob, or optionally create a new one. 2/3s of your total brute-, burn-, tox-, and oxy-loss will be applied to the original mob as agony, as a temporary malus (Anewbe has plans to replace this with a modifier)
If you die in VR, you'll be automatically ejected, and the link will be severed between original and VR mobs.
If your original mob dies while you're in VR, you're ejected automatically and stuck into your corpse.
Ghosts cannot currently enter VR, but that's something I'll look into in the future.

Testing!:
[Death in VR](https://puu.sh/xZ0h9/4c9538ec55.mp4)
[Sleeper Ejection and O-mob Death](https://puu.sh/yavdM/7fd8b62a80.mp4)
[Multi-spawn](https://puu.sh/xZ0HO/2f30bc1237.mp4)

Any other things that need to be tested can either be filmed on request, or I can launch a test server to demonstrate. Kinda hard to find everything relevant in my puush history.


code/game/machinery/vr_console.dm
```diff
@@ -0,0 +1,222 @@
+/obj/machinery/vr_sleeper
+	name = "VR Sleeper"
```
Vek-Tsarya: Should this be improper?

Atermonera: VR is an acronym, but the sleeper probably should be, yeah

SilverTalismen:
I would love to see what you can customize your avatar with, and I imagine this would be set up in a area near the holodeck?

Atermonera:
The avatar is basically just a human mob with promethean appearance, shapeshifting, and some of the other abilities. Any customization beyond that is as you're able to do ICly, although you'd probably get chamclothes for proper customization.

It could theoretically be set up anywhere, just needs space enough for however many sleepers.
If there's cameras that the AI can teleport to, it'll be able to go there. Otherwise it won't. Entirely up to the mappers

elgeonmb:
JOYFUL SCREAMING

also yes let's figure out how to let lawed synths in sometime

OhEmGee:
you bet your ass i liked inception

OhEmGee:
you bet your ass i liked inception

But yeah, this is a good feature, and will open tons of new, intresting avenues of RP.

Atermonera:
If the AI gets the capacity to use VR as an area where they can manifest, similar to holopads except with interactivity instead of just observation and speech, then it'd probably need a dedicated verb so that it can switch in and out quickly.
Stationbounds can probably use an extended version of the sleeper, like what I was originally implementing, or a verb a la the AI

elgeonmb:
other desired functionality: emagging a VR sleeper transfers accumulated VR damage as brute, not agony

Anewbe:
Let's go with no, for now.

Belsima:
> If your original mob dies while you're in VR, you're ejected automatically and stuck into your corpse.

smh

Atermonera:
Blame Ice and elge for that, they were the ones to catch it during testing

code/game/machinery/vr_console.dm
```diff
@@ -0,0 +1,222 @@
...
+
+	return 1
+
+/obj/machinery/vr_sleeper/attackby(var/obj/item/I, var/mob/user)
+	add_fingerprint(user)
+	if(default_deconstruction_screwdriver(user, I))
+		return
+	else if(default_deconstruction_crowbar(user, I))
+		return
+
```
Anewbe: Should there not be a ..() at the end of this?

code/game/machinery/vr_console.dm
```diff
@@ -0,0 +1,222 @@
...
+/obj/machinery/vr_sleeper/initialize()
+	update_icon()
+
+/obj/machinery/vr_sleeper/process()
+	if(stat & (NOPOWER|BROKEN))
+		return
+
+/obj/machinery/vr_sleeper/update_icon()
+	icon_state = "syndipod_[occupant ? "1" : "0"]"
+
```
Anewbe: Should there not be a ..() in here somewhere?

code/game/machinery/vr_console.dm
```diff
@@ -0,0 +1,222 @@
...
+	else if(default_deconstruction_crowbar(user, I))
+		return
+
+
+/obj/machinery/vr_sleeper/MouseDrop_T(var/mob/target, var/mob/user)
+	if(user.stat || user.lying || !Adjacent(user) || !target.Adjacent(user)|| !isliving(target))
+		return
+	go_in(target, user)
+
+
```
Anewbe: Should there not be a ..() at the end of this?

code/game/machinery/vr_console.dm
```diff
@@ -0,0 +1,222 @@
...
+	if(usr.incapacitated())
+		return
+	go_in(usr, usr)
+	add_fingerprint(usr)
+
+/obj/machinery/vr_sleeper/relaymove(mob/user as mob)
+	if(user.incapacitated())
+		return 0 //maybe they should be able to get out with cuffs, but whatever
+	go_out()
+
```
Anewbe: Should there not be a ..() at the end of this?

code/game/machinery/vr_console.dm
Atermonera: All of these were copied more or less wholesale from ABS and sleeper code, lack of super's included

Neerti: Don't think you need a super unless you need something in machinery to do something, which considering generally update_icon() does is change the icon and sometimes does lighting stuff, you probably don't need to use a super.

PR #4197, opened 2017-11-01
ATC ships won't go to the current station

### LINKED ISSUE #3089: NPC ships have Northern Star in destination list
Title

NPC ships to have destinations that aren't the currently played station

#### What actually happened
https://gyazo.com/531f213e3f82f4822661a8b54548cd72

#### Steps to reproduce
a) open a server
b) wait around

#### Additional info:

Atermonera:
Commented line will have to be switched when we switch to the SC, but for now, fixes #3089

### DIFF
--- a/code/modules/busy_space/organizations.dm
+++ b/code/modules/busy_space/organizations.dm
@@ -104,7 +104,7 @@
 	// Note that the current station being used will be pruned from this list upon being instantiated
 	destination_names = list(
 		"NSS Exodus in Nyx",
-		"NCS Northern Star in Vir",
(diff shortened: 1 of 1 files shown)


Neerti:
The ``New()`` proc for NT is meant to stop this from occurring.  I guess it broke?

elgeonmb:
the.... third.... PR that's been supposed to fix this issue...

Atermonera:
Yeah, I don't see why the stuff in `New()` isn't working, but this is the simplest, if also most limited, solution

PR #4199, opened 2017-11-01
[DMN WIP] Adds support for two new movement types

Anewbe:
- Adds support for a "Sprint" walk mode and a "Stalk" walk mode. Sprint is faster than running; Stalk is a slow as current walking, but makes no footsteps.
- Swapped m_intent to a number, rather than a string. Higher numbers mean you're moving faster, so Stalk is 1, Walk is 2, Run is 3, and Sprint is 4. Seems to be working fine.
- Tweaks a whole bunch of related stuff, including (hopefully) optimizing footstep processing.

**TODO:**
- [ ] Needs to be discussed by the Staff
- [ ] Some way to limit the use of Sprint, possibly a Stamina tracker?
- [ ] Allow use of the other two walk modes
- [ ] Flip the Defines, Sprint should be 4

### DIFF
--- a/code/__defines/mobs.dm
+++ b/code/__defines/mobs.dm
@@ -225,3 +225,9 @@
 #define SLIME_COMMAND_OBEY		1 // When disciplined.
 #define SLIME_COMMAND_FACTION	2 // When in the same 'faction'.
 #define SLIME_COMMAND_FRIEND	3 // When befriended with a slime friendship agent.
+
+// Movement states
+#define M_SPRINT	1
+#define M_RUN		2
+#define M_WALK		3
+#define M_STALK		4
--- a/code/_onclick/hud/alien_larva.dm
+++ b/code/_onclick/hud/alien_larva.dm
@@ -9,7 +9,7 @@
 	using.name = "mov_intent"
 	using.set_dir(SOUTHWEST)
 	using.icon = 'icons/mob/screen1_alien.dmi'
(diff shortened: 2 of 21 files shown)


Atermonera:
If you can stack modifiers, you could probably use a modifier with slowdown, to simulate getting tired. Apply it every [x] steps while sprinting, try to time it so that it'll reliably build up while constantly sprinting

Anewbe:
That's not how I'd like to do that, but it's not a bad idea.

Eimras:
I recall seeing a sprint somewhere else before. If I remember correctly, it showed a circular progress bar over the intent icon while sprinting, and if you let it fill all the way up, you'd get stuck to walk intent until it drained completely.

code/__defines/mobs.dm
```diff
@@ -224,4 +224,10 @@
...
 #define SLIME_COMMAND_FACTION	2 // When in the same 'faction'.
-#define SLIME_COMMAND_FRIEND	3 // When befriended with a slime friendship agent.
\ No newline at end of file
+#define SLIME_COMMAND_FRIEND	3 // When befriended with a slime friendship agent.
+
+// Movement states
+#define M_SPRINT	1
+#define M_RUN		2
+#define M_WALK		3
+#define M_STALK		4
```
Atermonera: Backwards defines, by PR description

Anewbe:

PR #4205, opened 2017-11-02
Lost Drone Pod Charger

MisterLayne:
After the drone is produced, it turns the pod into a charger.
rektscrub

### DIFF
--- a/code/game/machinery/rechargestation.dm
+++ b/code/game/machinery/rechargestation.dm
@@ -289,3 +289,27 @@
 	if(!usr.incapacitated())
 		return
 	go_in(usr)
+
+/obj/machinery/recharge_station/ghost_pod_recharger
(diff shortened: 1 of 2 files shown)


code/game/machinery/rechargestation.dm
```diff
@@ -289,3 +289,27 @@
 	if(!usr.incapacitated())
 		return
 	go_in(usr)
+
+/obj/machinery/recharge_station/ghost_pod_recharger
+	name = "drone pod"
+	desc = "This is a pod which used to contain a drone.. Or maybe it still does?"
```
Belsima: Should have three dots.

MisterLayne: It's grammatically correct to use three dots? Amazing.

PR #4206, opened 2017-11-03
Minor window fixes

Cyantime:
Replaces some dir == SOUTHWEST checks with the proc for that
Disables rotating fulltile windows, because that just turns them invisible.
Player-made fulltile windows now have the same health as mapped in ones.


Anewbe:
Did you test this?

Cyantime:
Yes.

PR #4211, opened 2017-11-04
Adds a ling ability to slip cuffs

Anewbe:
Lets you bust out of cuffs instantly, but...
- Costs Two Genetic Points
- Uses 40 Chems
- Makes a noise
- Has a 5 Minute Cooldown
- [x] I forgot a fancy icon.


code/game/gamemodes/changeling/powers/escape_restraints.dm
```diff
@@ -0,0 +1,63 @@
...
+	if(world.time < changeling.next_escape)
+		to_chat(src, "<span class='warning'>We are still recovering from our last escape...</span>")
+		return 0
+	if(!(C.handcuffed || C.legcuffed))	// No need to waste chems if there's nothing to break out of
+		to_chat(C, "<span class='warning'>We are are not restrained in a way we can escape...</span>")
+		return 0
+
+	changeling.chem_charges -= 40
+
+	to_chat(C,"<span class='notice'>We slip contort our extremities and slip our cuffs.</span>")
```
MisterLayne: I think you have an extra 'slip' at the beginning.

MisterLayne: I believe you have an extra 'slip' in that sentence that should not be there.

Anewbe: Good catch, will correct.

PR #4215, opened 2017-11-04
Ports Bay's gym equipment

Anewbe:
Get swole.


Belsima:
Could this tie into race or the traits? I don't imagine a weak Teshari doing as well as any Unathi.

Anewbe:
Out of scope.

SpadesNeil:
Huh. @Cameron653 make sure if this gets synced, we override it with our own 'cuz weight loss mechanics. Looks like the same code except that bit.

PR #4216, opened 2017-11-05
gives vatties an econ modifier

elgeonmb:


elgeonmb:
if I were better at code I would find the proc that checks econ mods and give it a default value but I'm not so patchjob

Anewbe:
Yeah, some day I'm gonna rework it into a goddamn species var...

PR #4226, opened 2017-11-06
Fixes mining maint access.

mistyLuminescence:
Cargo Technicians have access to the mining office, and to maintenance, but not to the mining office maintenance door. This PR fixes that.


Anewbe:
Did you test this?
Can Miners get through that door before and after?

mistyLuminescence:
Tested it, and it works correctly, and yes to both.

PR #4228, opened 2017-11-06
Putting wielded fire axes back now works correctly.

mistyLuminescence:
Previously: attempting to put a wielded fire axe back in a fire axe cabinet would fail, and you'd have to unwield it (which wasn't possible without putting something else in your other hand).
Now: attempting to put a wielded fire axe back in a fire axe cabinet will unwield it and put it back in the cabinet.

Also changes some << to to_chat.


Anewbe:
Did you test this?

mistyLuminescence:
Yep, and it works great.

PR #4229, opened 2017-11-06
Adds gold and silver bars to the Vault

mistyLuminescence:
The Vault crate (the one by the freezer) now has:
- 3 gold bars
- 3 silver bars
- Championship belt

Up from:
- 1 gold bar
- Championship belt

Why? Because it's the Vault and that's where you keep your gold and silver bars. There's still no diamond or uranium, so Mining will have to deliver those. (also it gives the vault actual relevance outside of antaggery)

*high-fives R&D*


MisterLayne:
Think it's a good idea, not sure how willing logical command members would be to let RND have stuff from the vault.

mistyLuminescence:
Exactly - it's _tempting goodies_, but only if you're allowed in.

MisterLayne:
I mean.. The vault is bolted by default..

PR #4233, opened 2017-11-06
Halves the size of default defib cells

Anewbe:
Half as large. Can still be removed and upgraded.


code/game/objects/items/devices/defib.dm
```diff
@@ -36,7 +36,7 @@
 	qdel_null(bcell)
 
 /obj/item/device/defib_kit/loaded //starts with highcap cell
```
elgeonmb: no it doesn't

Anewbe:
Closing in favor of #4278

PR #4237, opened 2017-11-07
Ninja cloak is no longer true invisibility

Anewbe:
Alpha = 5
Ling cloak is Alpha = 10


code/modules/clothing/spacesuits/rig/modules/ninja.dm
```diff
@@ -37,8 +37,8 @@
 
 	var/mob/living/carbon/human/H = holder.wearer
 
-	H << "<font color='blue'><b>You are now invisible to normal detection.</b></font>"
-	H.invisibility = INVISIBILITY_LEVEL_TWO
+	to_chat(H, "<font color='blue'><b>You are now invisible to normal detection.</b></font>")
```
mistyLuminescence: Can we change this to "almost invisible", then?

Anewbe: It's good enough as-is.

mistyLuminescence: It's not *true*, though - 5 alpha isn't full invisibility, it's just really, really good camouflage. If you're moving around in a well-lit area, you're still just barely visible to someone who's paying attention.

Belsima:
what the fuck does bikeshed mean

code/modules/clothing/spacesuits/rig/modules/ninja.dm
Belsima: Yeah, it's really not invisible, that's misleading. I can totally see a ninja running around at alpha 5 if I'm already aware they exist.

elgeonmb:
This is a good label

PR #4239, opened 2017-11-07
Fixed my typo with revolvers

Sabess:
Unfortunately missed one typo in testing, which broke a certain optional sprite for the revolver. Oops. Also, I changed the default name of the revolver to ".45 revolver" after realising it could potentially be confused with the antag revolver when being aimed or etc.


code/modules/projectiles/guns/projectile/revolver.dm
```diff
@@ -67,7 +67,7 @@
 		return 1
 
 /obj/item/weapon/gun/projectile/revolver/detective45
-	name = "revolver"
+	name = ".45 revolver"
```
Belsima: Honestly, I prefer the ambiguity, but I have dumb opinions.

Belsima: I like the description idea better.

Anewbe:
In the name is fine, I'm pretty sure you can rename them anyway.

code/modules/projectiles/guns/projectile/revolver.dm
Sabess: the description is already different, and yeah you could just name it "revolver" if you really wanted

PR #4243, opened 2017-11-07
Grid checker is now dense and anchored

mistyLuminescence:
Per request ( @MisterLayne ). It's that big blue box engineers can mess with to fix the grid check random event, apparently. Definitely [looks like it should be dense and anchored](https://cdn.discordapp.com/attachments/176163422751293441/377598902062546944/unknown.png).


MisterLayne:
I'd call this more of a bugfix, honestly. It should be anchored in the first place.

PR #4245, opened 2017-11-08
Ballistic Transportation Pod

MisterLayne:
Adds Ballistic Transportation Pod with one use.

baldie_race


polaris.dme
```diff
@@ -1016,6 +1017,7 @@
 #include "code\game\objects\structures\tank_dispenser.dm"
 #include "code\game\objects\structures\target_stake.dm"
 #include "code\game\objects\structures\transit_tubes.dm"
+#include "code\game\objects\structures\transportpods.dm"
```
MisterLayne: Aware of this. Will fix.

Neerti:
So is this a sort of droppod for mobs?

Anewbe:
My hype should be palpable.

code/game/machinery/transportpod.dm
```diff
@@ -0,0 +1,109 @@
...
+	set name = "Enter Pod"
+	set src in oview(1)
+
+	if(usr.incapacitated()) //just to DOUBLE CHECK the damn sleepy people don't touch the pod
+		return
+
+	go_in(usr)
+
+/obj/machinery/transportpod/proc/doExplosion(var/turf/L) //boom
+	explosion(L, 0, 3, 3)
```
MisterLayne: This is just a placeholder until I'm not too tired to add in an actual cool effect.

Anewbe:
I believe it is, Neerti, but for mobs that already exist, rather than the pods that just spawn in new mobs.

code/game/machinery/transportpod.dm
```diff
@@ -0,0 +1,121 @@
...
+			sleep(20) //Give explosion time so the pod itself doesn't go boom
+			src.forceMove(L)
+			playsound(src, pick('sound/effects/Explosion1.ogg', 'sound/effects/Explosion2.ogg', 'sound/effects/Explosion3.ogg', 'sound/effects/Explosion4.ogg'))
+			in_transit = 0
+			used = 1
+
+
+/obj/machinery/transportpod/examine(mob/user)
+	..(user)
+	user << "The pod is [(used == 1)? "already used." : "ready to be used."]"
```
Anewbe: to_chat please.

code/game/machinery/transportpod.dm
```diff
@@ -0,0 +1,121 @@
...
+	density = 1 //thicc
+	anchored = 1
+	use_power = 0
+
+	var/used = 0
+	var/in_transit = 0
+	var/mob/occupant = null
+
+	var/xc = list(137, 209, 163, 110, 95, 60, 129, 201) // List of x values on the map to go to.
+	var/yc = list(134, 99, 169, 120, 96, 122, 189, 219) // List of y values on the map to go to.
```
Anewbe: So this is just a list of 8 specific locations you can get sent to?

code/game/machinery/transportpod.dm
```diff
@@ -0,0 +1,121 @@
...
+	if(occupant)
+		if(in_transit)
+			var/locNum = rand(0, 7) //pick a random location
+			var/turf/L = locate(xc[locNum], yc[locNum], 1) // Pairs the X and Y to get an actual location.
+			limit_x = xc[locNum]+1
+			limit_y = yc[locNum]+1
+			build()
+			sleep(20) //Give explosion time so the pod itself doesn't go boom
+			src.forceMove(L)
+			playsound(src, pick('sound/effects/Explosion1.ogg', 'sound/effects/Explosion2.ogg', 'sound/effects/Explosion3.ogg', 'sound/effects/Explosion4.ogg'))
```
Anewbe: If this makes an actual explosion at the target location it should already be playing these sounds, iirc.

code/game/machinery/transportpod.dm
```diff
@@ -0,0 +1,121 @@
...
+	name = "Ballistic Transportation Pod"
+	desc = "A fast transit ballistic pod used to get from one place to the next. Batteries not included!"
+	icon = 'icons/obj/structures.dmi'
+	icon_state = "borg_pod_opened"
+
+	density = 1 //thicc
+	anchored = 1
+	use_power = 0
+
+	var/used = 0
```
Anewbe: Rather than have to check this var, wouldn't it just be simpler to move the mob and destroy this...?

MisterLayne: That would be replacing a simple 'used' variable with some code to delete the old one. Just kind of how I thought it would be easier. The used variable also helps with changing the description of the pod.

code/game/machinery/transportpod.dm
MisterLayne: Yes.

code/game/machinery/transportpod.dm
MisterLayne: Doesn't make an actual explosion.

code/game/machinery/transportpod.dm
MisterLayne: I have no clue why is used user <<

code/game/machinery/transportpod.dm
Anewbe: If it's destroyed when it's done doing its thing you don't need to worry about that.

code/game/machinery/transportpod.dm
Anewbe: Why is that?

code/game/machinery/transportpod.dm
Anewbe: Roughly where are these locations?

code/game/machinery/transportpod.dm
MisterLayne: I mean.. You want it to be destroyed after it drops?

code/game/machinery/transportpod.dm
MisterLayne: Uhr.. Cargo area, security, engineering, around primary tool storage (the one near technical storage, I think), Medical..

MisterLayne: Don't remember all the locations, just those general ones.

code/game/machinery/transportpod.dm
MisterLayne: I feel like having it be destructive, at least for what I thought it would be used for, would be kind of.. Rude? I could add an option whether you want it to be destructive or not? Right now it just builds some reinforced squares one tile around the drop point.

code/game/machinery/transportpod.dm
Anewbe: After it drops and spits out the occupant, yes.

MisterLayne:
> Have you tested this? - Anewbe

Yes. Yes I have.

Belsima:
Remember to fix co-ordinates when we update to Southern Cross. Or figure out a way to set rooms, maybe.

PR #4252, opened 2017-11-11
fixes minor typo in admin_verbs.dm

elgeonmb:


Anewbe:
About time someont fixed that.

PR #4256, opened 2017-11-11
Unathi are bigger and have more max blood

Anewbe:
- Increased size by 50%. 
 - Increased blood volume by 50%.


mistyLuminescence:
Pics of how they look when they're 50% bigger?

Neerti:
The sprite is unchanged (for now), its just changing the size flag.

MisterLayne:
> ...(for now)...

PR #4257, opened 2017-11-12
Un-duplicates head marking cache generation.

Verkister:
-Organ icon generation for heads no longer adds the markings list into the cache key twice.
-Actually does exclude heads from the parent caching for limb markings instead of counting on mere comment-format thoughts and prayers.
-Related to similar dismemberment crashes as the earlier organ cache issue fixes.

### DIFF
--- a/code/modules/organs/organ_icon.dm
+++ b/code/modules/organs/organ_icon.dm
@@ -71,7 +71,7 @@ var/global/list/limb_icon_cache = list()
 		overlays |= lip_icon
 		mob_icon.Blend(lip_icon, ICON_OVERLAY)
 
-	//Head markings, duplicated (sadly) below.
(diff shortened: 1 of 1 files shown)


Anewbe:
And you've tested this? It doesn't interfere with normal head icon creation?

Verkister:
Seemed fine on a test run where I had applied a bunch of head markings to the test body that lost its head. I think some of the markings did vanish off the tossed severed head icon, but otherwise everything worked as intended, all of the markings showing up on the head while it was still attached with no crashing or other complications after it came off.

Anewbe:
And you've tested this with the most recent code?

Verkister:
Pretty much as recent as the PR.

PR #4258, opened 2017-11-12
New Drinks

Mewchild:
Adds a handful of new drinks and mixing reagents, including bitters (orangejuice+enzyme), which is a staple of many real world drinks. Bitters have been added to the bar dispenser appropriate cargo crate.


Mewchild:
Disregard for now, didn't test fully before submitting. The power of 10AM coding after an all nighter.

Issue #4251, opened 2017-11-10
show-hidden-verbs don't work reliably


elgeonmb:
#### Brief description of the issue
The AI HUD we've had for like a year now has never actually been backwards compatible; show-hidden-verbs doesn't actually show all hidden verbs in the tabs and in verb autocomplete
#### What you expected to happen
To be able to ignore the HUD and continue typing verbs
#### What actually happened
Having to remember which ten verbs got made into buttons for some reason
#### Steps to reproduce
Use the hidden verbs command
Try using one of these verbs
Quit AI in frustration, become a secofficer main.
#### Additional info:
 - **Server Revision**: Found using the "Show Server Revision" verb under the OOC tab.
 - **Anything else you may wish to add** (Location if it's a mapping issue, etc)

Anewbe:
Inform the Mechoid, this has drawn his ire in the past.

elgeonmb:
@Mechoid save us

elgeonmb:
fixes https://github.com/PolarisSS13/Polaris/pull/893

the relevant pr

elgeonmb:
...also like, if it could be.... on.... by default.... that would be grand....... admin hidden verbs are on by default.....

Atermonera:
Did the original PR actually add any verbs, and if not, would people be opposed to reverting it outright?

elgeonmb:
It didn't add any verbs and I want it dead if possible

Nalarac:
@elgeonmb You can close this now since the fix was merged.

PR #4262, opened 2017-11-14
AI verbs are no longer hidden

Atermonera:
Hiding stuff from players is bad, mmmkay. #4251 
Tested:
AI Commands
AI IM
AI Settings

Didn't rearrange the verbs on the tabs at all.


elgeonmb:
*cries*

PR #4264, opened 2017-11-14
Decapitation now stops defibrillation

Anewbe:
As per the title, decapitating an organic victim will prevent them from being defibbed.
Synthetics should still be fine, if a little worse for wear

### DIFF
--- a/code/game/objects/items/devices/defib.dm
+++ b/code/game/objects/items/devices/defib.dm
@@ -284,7 +284,7 @@
 		return "buzzes, \"Resuscitation failed - Excessive neural degeneration. Further attempts futile.\""
 
 	H.updatehealth()
-	if(H.health + H.getOxyLoss() <= config.health_threshold_dead || (HUSK in H.mutations))
+	if(H.health + H.getOxyLoss() <= config.health_threshold_dead || (HUSK in H.mutations) || (!H.isSynthetic() && !H.can_defib))
 		return "buzzes, \"Resuscitation failed - Severe tissue damage makes recovery of patient impossible via defibrillator. Further attempts futile.\""
 
 	var/bad_vital_organ = check_vital_organs(H)
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -15,6 +15,8 @@
 	var/spit_name = null //String
 	var/last_spit = 0 //Timestamp.
 
+	var/can_defib = 1	//Horrible damage (like beheadings) will prevent defibbing organics.
+
 /mob/living/carbon/human/New(var/new_loc, var/new_species = null)
 
 	if(!dna)
--- a/code/modules/organs/subtypes/standard.dm
+++ b/code/modules/organs/subtypes/standard.dm
@@ -279,6 +279,8 @@
 		spawn(1)
 			owner.update_hair()
 	get_icon()
+	if(vital)	//This is just in case we ever add something that both a) Doesn't need a head to live, and b) Can be defibbed
+		owner.can_defib = 0
 	..()
 
 /obj/item/organ/external/head/take_damage(brute, burn, sharp, edge, used_weapon = null, list/forbidden_limbs = list())


mistyLuminescence:
Will leave a Doctor's Opinion on this later when I'm back from the irl doctor's. TLDR let med have this fun thing that takes a ton of effort and skill, it's awesome.

mistyLuminescence:
Alright, back, time for the Doctor's Opinion ™️ .

Decapitating an organic victim shouldn't _always_ prevent them from being defibrillated. There shouldn't be a "if you were decapitated, you can't be defibbed, no matter how hard you try" clause. You should be able to defibrillate a decapitation victim if you try really, really hard. Why?

Because it's **cool**. The concept of 'you had your head chopped off, but it took us the entirety of medical and a lot of thoughts and prayers to bring you back alive, through the most advanced medical treatments in the galaxy and some of the best equipment around'... is much cooler than 'yeah, we cloned you just like all the other victims with big injuries'. Defibrillating a decapitation patient makes you the Victor Frankenstein of the medical world; it's a technique that might not always be successful, that you have to be both skillful and lucky to perform, and when it works, it's objectively the coolest thing to happen in that particular round. When it fails, go to cloning and give the antag a pat on the back for thwarting medical. 

"But coolness isn't as important as balance!"

Sure, but what's imbalanced about it? There are two situations where someone's going to get decapitated by an antag. Either the antag has all the time in the world, or is in a pretty big rush and has no time at all. In the former, there's absolutely nothing stopping them from wailing on the body with the same weapon they performed the decapitation with. It's not _hard_ to stop a defibrillator if you put your mind to it; you just need to bash the body a bit, raise the damage high enough that it fails. Or turn their sensors off and wait 10 minutes. Or toss them into space to husk them, or disposals to break every bone in their body. Or, you know, just amputate a few more limbs to push the damage even higher. And if you _don't_ have time to do all that? Then why shouldn't Medical be able to save your victim? The amount of effort you put into an attack should be proportional to the amount of effort it takes to bring that victim back to life without cloning. Shoot them with a carbine a few times? Easy. How about a shotgun at point-blank? Hard. A decapitation victim? Even harder - but still doable. How about a decapitation victim that got tossed into disposals? Impossible; this is the point where antags beat medical every time fair and square, and that's fine.

And speaking of balance, doesn't this also enlarge the power gap between synthetics and organics? The part where synthetics don't suffer from pain, broken bones, organ failure, virii, internal bleeding or external bleeding, and have only one (really, really powerful) counter - EMPs? Yes. Yes, it does.

"But it's unrealistic to be able to save someone from having their head cut off."

This is a game set in 2561 where we have talking squids, lightsabers, combat mecha, handheld teleporters, guns that shoot ionic pulses, a big chunk of yellow rock that we shoot with lasers to generate power, sheets of mythical space handwavium that's important enough to justify an entire research colony... I can go on. It's not unreasonable to think that 544 years in the future, the most advanced surgeons would be able to pull a Frankenstein on someone.

"Okay, but, it's too easy. Anyone can do it."

It's not. It's _really_ not. The patient that prompted this PR, last night? Poisoned and then decapitated with a clean cut from an energy sword, and medical was on scene in less than a minute, with the patient back to med in less than two. It took two of the server's best medical players, a pretty big cocktail of basic and advanced chems (bicaridine, tricordrazine, dexalin plus, tramadol, dylovene, carthatoline, alkysine), and one of the fastest surgeries around to bring the patient to **197.2** damage. The defib limit, for anyone wondering, is 200, and the defib causes 5. It was, in no sense of the word, easy, and if the patient had been attacked one more time, even stomped on or dragged instead of grabbed, she wouldn't have been defibrillated.

"But defibrillators are too powerful!"

They're used a lot, but that's because when people die, it's either PvE (in which case defibs should pretty much always be used), or PvP where the killer doesn't finish the job properly (in which case they should've beat up the patient some more if they wanted them to be cloned). We don't need to nerf the cell size or burn damage or Frankenstein capabilities of defibs, we just need to gently remind antags that it's okay to stab your victim sixteen times in the face with a combat knife. 🔪 And if they don't, or can't, take the time to do that? Then there's no reason the best Medical players shouldn't be able to save them.

"Okay, but, it's a bug."

Well, let's make it a feature. 🐛 

"You don't know what you're talking about, though."

I know what I'm talking about. The only times anyone gets saved from decapitation is when they go through medical, and when people go through medical, medical players like myself are the ones who treat them. I almost exclusively play medical and communicate regularly with other medical mains, so I think I can safely say that pulling a Frankenstein on someone doesn't break the medical system. 

-----

TL;DR: Keep it, because:
- It's cool.
- It requires lots of skill and luck, so doesn't happen all the time.
- It's easy to shut down if the antag takes the time to disposal/space/husk/maul/hide the body.
- Removing it, but keeping it for synths, widens the power gap between FBP and organic.
- It's not unrealistic, given the setting and technology level.
- It rewards medical players for putting in effort to save the patient.
- When it's relevant, it doesn't take anything away from the server, but adds a lot.

And remember, Sana loves you.

Curt5:
Not in favor of this fix. A recapitation happens so rarely that removing it will have little impact, I think that the dead player will largely have no preference between being cloned/re-headed, and the idea of it is so cool that I am against removing it.

However, I would be in favor of making re-attaching violently detached limbs more complex. Using bone gel and FixOVein to fix the nerves and broken bones, etc.

mistyLuminescence:
That could work. Add it to the amputation steps rather than the reattachment steps to make it work for only violently-detached limbs, rather than surgically-removed ones.

Maybe instead of saw --> done, have it... saw --> bone gel --> bone setter --> fixovein --> done? This also means if your gimmick is, for some reason, cleanly decapitating people and leaving the bodies for medical to find within 10 minutes, you can steal their fixoveins/bone setters/bone gels and ruin their do-goodery forever.

ElectricB1ue:
Recapitation **IS** incredibly rare. I don't personally believe this to be necessary.

Magneld:
Clarify, do you mean it wont work even if they re-attach the head or it wont work because obviously the head is not there why are you zapping a headless corpse.

mistyLuminescence:
The former; currently, you can defib someone who got decapitated if you reattach the head (and if you're fast enough and they have less than 200 damage after the decapitation + stump amputation). This PR changes that so you can't, even if you reattach the head.

Wickedtemp:
I'm for keeping it. Misty pretty much covered all of the reasons why. 

One issue I see occasionally is the thought of "It wasn't what I wanted, therefore it is a bug and must be removed." To take one of the more well-known examples of this being untrue and harmful, Minecraft. Creepers were originally a mistake in coding. I can't really imagine Minecraft without creepers. 

Sometimes, oversights or even outright bugged code can add to a game. I feel like this is one such case.

Mechoid:
I'm still of the opinion that surgical decapitations should still allow defibrillation, as in the sawing of a head off on an operating-or-other table. Yes, it's a saw, and saws are messy, but it's also a medical saw for doing medical things with, not a chainsaw, you can chop off someone's hand neatly enough that you can immediately re-attach it with no special work, or chop open their skull with no harm at all to the brain inside the bone-jar. Not allowing surgical decaps to at least be defibbed prevents 'Frankensteining', or sticking the wrong head and other bits on the wrong body, possibly from a different species and shocking it to life. If a true 'surgical' removal for heads, I.E. scalpel, hemo, retractor, saw was added, I'd be less frown at it, but right now I'm still kind of irked on it even if I see the point.

Natje0:
I'm against this. I don't see any reason why it should be changed, and Misty states a ton of GOOD reasons why it shouldn't be.

PR #4267, opened 2017-11-15
Makes Icon Scaling Great

Neerti:
Scaling wasn't great before so no Again.
Makes most icons not get blurry when scaled.
Makes all /atom/movable s able to be scaled with the adjust_scale() proc.
Adds ability for species to get scaled automatically.  Currently no species does this, but it is an option in the future.
Adds four traits to adjust your mob scale.  It is purely cosmetic at this time.
Demo
Traits

### DIFF
--- a/code/game/atoms_movable.dm
+++ b/code/game/atoms_movable.dm
@@ -1,6 +1,6 @@
 /atom/movable
 	layer = 3
-	appearance_flags = TILE_BOUND
+	appearance_flags = TILE_BOUND|PIXEL_SCALE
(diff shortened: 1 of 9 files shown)


Andromeda-K22:
something, something mechanical advantages?

Sprite seeming smaller is a little odd.

Neerti:
The smallest sprite is still fairly easy to click on, and projectiles don't care how big or small the sprite is.

SubberTheFabulous:
This looks dorky.
_I love it._

PR #4274, opened 2017-11-16
Adds "Shield Belt"

MisterLayne:
Adds a "Shield Belt". It is a shield that you can wear on your belt slot. Congratulations.

To do:
- [ ] If you're wearing the thing, and you click the action button, the shield sprite doesn't pop up. You actually have to take it off and click it. That makes me sad.
- [ ] Make the shield not-invincible.
- [ ] Figure out why the hell the shield seems to block no damage.


Anewbe:
Uh... Why is this all an armor, rather than a belt?

MisterLayne:
Because belts are part of storage, and storage stores things, so if I made it a subtype of belt, it could hold things.

Anewbe:
Oh, right. I meant to fix that.
This belt probably covers some body parts that it shouldn't, but otherwise I'm not sure what's wrong with it, specifically.

MisterLayne:
Well, I can fix the body part thing with with flags. Maybe the shield just isn't.. Strong?

code/game/objects/items/devices/beltshield.dm
```diff
@@ -0,0 +1,72 @@
...
+	spark_system.start()
+	playsound(user.loc, 'sound/weapons/blade1.ogg', 50, 1)
+	return 0 // This shield does not block all damage, so returning 0 is needed to tell the game to apply the new damage.
+
+/obj/item/clothing/suit/armor/beltshield/attack_self(mob/user)
+	active = !active
+	to_chat(user, "<span class='notice'>You [active ? "" : "de"]activate \the [src].</span>")
+	user.update_action_buttons()
+	update_icon()
+
```
MisterLayne: I believe the reason the icon doesn't update correctly is because there's no 'return 1'. 'Tis time to test that.

PR #4278, opened 2017-11-17
Successful Defibs are more costly, chargewise.

Anewbe:
From 1000 to 1250, and a smaller cell, borg zappers are unaffected.

### DIFF
--- a/code/game/objects/items/devices/defib.dm
+++ b/code/game/objects/items/devices/defib.dm
@@ -36,7 +36,7 @@
 	qdel_null(bcell)
 
 /obj/item/device/defib_kit/loaded //starts with a cell
-	bcell = /obj/item/weapon/cell/high
+	bcell = /obj/item/weapon/cell/apc
 
 
 /obj/item/device/defib_kit/update_icon()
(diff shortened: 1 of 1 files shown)


mistyLuminescence:
But this is even worse than the last one, because we can't take out the cells and upgrade them with the relatively-common high-cap cells... again, pretty sure the defib problem is caused by antags never sticking around after a kill to do anything about it.

Wickedtemp:
This would lead to entire defibs being used up on a single patient, depending on how slow they are with returning to their body after ghosting. Usually, you'd zap 3-4 times. This isn't _defib spam_, because it's kind of the only way to defib someone who's ghosted. The first or second defib lets them know someone's trying to bring them back, and during the second or third, they have a little window to return to their body. 

"Defib Spam" is when you defib someone, they die, you defib them again, they die, rinse and repeat. Which I agree, it should be discouraged. Feel free to tack on burn damage on patients who are _being brought back by defibs repeatedly, not to corpses who keep popping up with MIF's with each attempt_.

mistyLuminescence:
If this nerfs _unsuccessful_ defibs (read: ones where the patient doesn't come back to life), the title's pretty misleading - I was under the impression this would only work on defibs that, well, actually work. Using three defib charges on one body in one go is normal for medical.

Wickedtemp:
At least partially due to defibs not even messaging ghosted players that doctors are attempting to defib them, apparently. 

https://gyazo.com/99483222545f77c130751d42d88fc1bb

mistyLuminescence:
^ #4283 

Consider:
1) defib zaps that don't rez someone use no charge
2) defib zaps that do rez someone use double charge
and 3) we fix 4283 already (working on it!)

mistyLuminescence:
TL;DR version now: APC cell (5k), each charge uses 1250, 4 uses per cell.
I'd support this, even if unsuccessful defib zaps use charge - with 4283 fixed, we won't need to defib-spam MIF patients to get them to come back.

Wickedtemp:
I can get behind these changes.

PR #4281, opened 2017-11-17
Adds a message to the IRC for round start.

Belsima:
If ELLIS ever comes back online, it'll bipply-boop for roundstart rather than just server boot.


Belsima:
Added a bunch of grammar fixes and replaced the odd 'marooned' messages and such.

PR #4292, opened 2017-11-18
Switches us to the Southern Cross for testing

Anewbe:
**END DATE: Some point on 11/18/2017**


Anewbe:
Self-Merging in spite of the Travis Issue, at Staff Request.
It seems to be running fine, and we will be reverting it in a few hours anyway.

PR #4294, opened 2017-11-18
Removes skrell noses

PrismaticGynoid:
Slightly alters the skrell (and skrell hologram) sprites to remove the noses, since skrell do not have noses.
Comparison: 
noselessskrell
http://www.strawpoll.me/14444386


elgeonmb:
👍

MisterLayne:
There's that one pixel difference on the side view sprite. Kind of makes them look like a noseless dinosaur.

Anewbe:
I'm going to say that this is actually a step down, quality wise. The adjusted side sprites just look too flat.

Belsima:
yeah those flatfaces look bad

PrismaticGynoid:
How would you suggest they be changed? All that was done was removing the nose that skrell shouldn't have. It helps them look more alien, in my opinion. A lot of synthetic brands also lack a nose, like WT.

Belsima:
On second thought, I'm liking it a bit more.

Anewbe:
@PrismaticGynoid I honestly don't see the need to remove it. Just say it's part of the way their skull curves, or something.

Atermonera:
I'm with anewbe, even if the pixel is taken to represent a non-existent biological part, it adds curvature. Removing it makes it too flat, like the monitor heads. The reason it's fine for monitor heads is that they're _supposed_ to look flat, they're boxes. Organic stuff is almost never so perfectly flat.
If you were to add shading to it somehow, that might make it better, but it's such a small area I don't think you'll really be able to make up for the lack of curvature from that one pixel. 
:-1:

Cameron653:
https://i.imgur.com/2lNhkzU.png With nose
https://i.imgur.com/TyUZFTe.png Without nose

They look pretty awful without a nose, like their face is a block.

Anewbe:
I talked to the Skrell people in charge and they said they don't want this change.

PR #4297, opened 2017-11-18
Makes Syndie Toolboxes Special Again

Neerti:
The syndie toolboxes have been split into two versions, one with the old contents pre-powertools and one with the current powertools. Both are in the uplink, however only the regular tool version of the syndie toolbox can be found in maint now. This is to make power tools less common, as there are several spots where syndie toolboxes are guaranteed to spawn at,  and it made the CE's tools less special.


Anewbe:
I'm in firm support of this, but we should still bring it up at the Staff Meeting, I know there were comments the last time I brought this idea up.

PR #4302, opened 2017-11-18
Removed Shotgun's ability to use speedloaders.

SpadesNeil:
- Makes no sense from a physics standpoint.
- It's bullshit from a balance stand point.
- Makes shotgun ammo boxes obsolete, which is dumb.

This doesn't get rid of shotgun speedloaders from the autolathe, it just means they are no longer possible to speed-reload your shotgun with. These were added with no ramification for the imbalance issues they cause. Allow me to elaborate:

- A shotgun clip holds 4 rounds, and is a small item. As such, it takes up the same amount of bag space as one box.
- One box can hold eight shotgun shells. Two clips can hold eight shotgun shells. But one box can hold four clips. You can fit **sixteen** shotgun shells into the space of one card board box, which is absurd.
- Clips can go into security belts, but boxes can't. More bullshit.

Seriously, shotguns should not be possible to speedload!

"Okay but Spades I don't want to go back to having to manually load shotgun shells one by one by hand during a crisis." Okay well you don't have to. Literally any cardboard box can be loaded with shells. If you load the box with shells, you can click a shotgun, and it will load the shotgun with shells automatically. This is far more balanced.

### DIFF
--- a/code/modules/projectiles/guns/projectile/shotgun.dm
+++ b/code/modules/projectiles/guns/projectile/shotgun.dm
@@ -10,7 +10,7 @@
 	slot_flags = SLOT_BACK
 	caliber = "12g"
 	origin_tech = list(TECH_COMBAT = 4, TECH_MATERIAL = 2)
-	load_method = SINGLE_CASING|SPEEDLOADER
+	load_method = SINGLE_CASING
 	ammo_type = /obj/item/ammo_casing/a12g/beanbag
 	handle_casings = HOLD_CASINGS
 	fire_sound = 'sound/weapons/shotgun.ogg'


SpadesNeil:
Whatever Travis fucked up was not my fault.

Atermonera:
Also because of active map testing, circa #4292

mistyLuminescence:
Can you still slow-reload using shotgun clips (like you can with the boxes)?

SpadesNeil:
>Can you still slow-reload using shotgun clips (like you can with the boxes)?

I am afraid not. Allowing that would result in a balance issue again.

Anewbe:
I'm not sure I agree with you on that last comment, Spades.

SpadesNeil:
>I'm not sure I agree with you on that last comment, Spades.
>EDIT: Actually, I'm going to say I disagree with you on most of this PR.

Two shotgun clips take up as much bag space as one box. One box can hold four shotgun clips. Two shotgun shells take up only as much space as one clip, which holds four rounds. The math is that clips allow you to carry far too much shotgun ammunition at once.

If we talk about security belts, previously, each of those slots could hold only one shell. Now they can hold four.

>The only thing I can outright say should be changed is the instant reload part.

That is literally all I changed. The only fair solution is to remove the shotgun's ability to speed-reload. If we give the ability to speed-reload to the clip, even if we have it staggered by a cooldown, that's still **up to four times more** ammunition you can now carry that you have easy access to. The excessive amount of ammo should come at the cost of not being immediately accessible.

Anewbe:
Then I'm afraid that this PR will not be merged. We don't see the vast majority of those points as an issue.

SpadesNeil:
You realize that by not merging this, the shotgun remains the most powerful weapon available to the crew by a _wide_ margin, right?

Speed-reloading a _shotgun_ is not something that is available in any other server for very good reason. It breaks the weapon and makes it exceedingly unfair to use. You also formerly lectured me on keeping my PRs separate, so why you suddenly want me to include a new mechanic for clips in the same PR that I am changing only shotguns is something I do not understand in the slightest.

Anewbe:
You realize that merging this wouldn't really affect shotgun power, right? It hits hard as hell in the most powerful damage type and it can be fired about as fast as it can be pumped. It has been one of if not the single best SS13 crew weapon for as long as I can remember, and that's not an insignificant amount of time.

I've lectured you on keeping PRs separate because people tend to cram as many unrelated changes into one Pull as possible. My request is in the same line as the intention of this PR, being a nerf to shotgun ammo clips, and I'm offering it as an alternative to what's represented here.

That said, if you shrink the clip capacity to be equal to the number of shots you can cram into a box, and have sprites to go along with it, that works for me.

SpadesNeil:
My gripe here isn't the shotgun's power or rate of fire. It's that the power and rate of fire are **relentless** when it can be immediately reloaded. It's just a step below having the automatic shotgun which, even when fired semi-auto, shreds _everything_ it encounters. Four shots in one clip, speedloaded, is just madness. I can tear through any threat thrown at me with the kind of firepower clips give me.

>That said, if you shrink the clip capacity to be equal to the number of shots you can cram into a box, and have sprites to go along with it, that works for me.

Alright I'll do that instead. At the very least my ammo capacity is no longer _quadrupled_, and I just get a meager perk of double the speed of reloading. Much more reasonable.

PR #4310, opened 2017-11-20
Adds Hunting Rounds

Neerti:
Adds a system for projectiles to inflict additional damage on specific kinds of simple animals.
Adds special bullets for certain rifles which are very effective against 'animal' SAs such as giant spiders, bears, carp, and such. The bullets do less damage to other targets such as humanoids.
Replaces the southern cross gun locker regular ammo with hunting ammo.

### DIFF
--- a/code/datums/autolathe/arms.dm
+++ b/code/datums/autolathe/arms.dm
@@ -147,9 +147,18 @@
 	name = "rifle magazine (5.45mm practice)"
 	path =/obj/item/ammo_magazine/m545/practice
 
+/datum/category_item/autolathe/arms/rifle_545_hunter
+	name = "rifle magazine (5.45mm hunting)"
+	path =/obj/item/ammo_magazine/m545/hunter
+
 /datum/category_item/autolathe/arms/machinegun_545
 	name = "machinegun box magazine (5.56)"
 	path =/obj/item/ammo_magazine/m545saw
+	hidden = 1
+
+/datum/category_item/autolathe/arms/machinegun_545_hunter
+	name = "machinegun box magazine (5.56 hunting)"
+	path =/obj/item/ammo_magazine/m545saw/hunter
 	hidden = 1
 
 /////// 7.62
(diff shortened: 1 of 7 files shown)


SpadesNeil:
While this is a neat concept, I can't wrap my head around what sense this makes. It's awfully immersion breaking for a "hunting" round to exist with no explanation as to how it works.

Anewbe:
Gameplay is more important than realism.

mistyLuminescence:
Hunting rounds have tiny polyfluorocarbocordrazide injectors in them, a very obscure chemical that acts as a blood thinner, increasing damage taken from bullets - but it only works on unintelligent creatures.

Or they've all got tiny IFF homing systems inside, with safety features that guide it towards less lethal areas on sapient beings. Or... some other space magic. I'unno.

SpadesNeil:
Gameplay is more important than realism, yes, but I'm just saying it should be easy to make up a lore reason why the bullets exist.

Something like what Misty just said.

Not saying this shouldn't be merged. It's just nice to have a lore reason. Otherwise you're stuck asking the question of, "If these exist and they're so readily available, why not use them _all the time_?" I feel that without a good excuse as to why they exist, they detract from the danger of a gun, and thereby, take away from the respect by which a gun should be treated.

SpadesNeil:
tl;dr having a bullet with nerfed damage just so it hurts humans less is very silly

If you're trying to make a weapon intended to murder only mobs, there's so many better ways it can be done.
- Energy weapons that fire bolts of electricity. Fries the nerves of a less sapient victim. (Tasers already kill mobs!)
- Smart guns that won't fire inside of a certain area.
- Distribute only crappy guns (like bolt actions) for sporting purposes.

Anewbe:
As I mentioned before, gameplay is more important than realism. We have a need for this code, it is functional, and therefore it is being merged. However you choose to explain away the effects of these items is your prerogative.

SpadesNeil:
I might add these bullets to R&D later, then.

Also what is the Bikeshed tag for?

Anewbe:
They're already in the autolathe.

The Bikeshed tag is for relatively minor but still somehow controversial changes.

PR #4311, opened 2017-11-20
Mice should no longer move crates in maint.

Belsima:
Ported from VORE.

Also lowered death sound volume.

This probably won't compile. I really need a new motherboard.


Anewbe:
I'm pretty sure there's some sort of "Bump Flag" that you can set on mobs to stop them from pushing things.
If so, that'd be very preferable to this PR, since stopping any function set by Bumped() sounds more like something I'll be  bugfixing in two months.

Belsima:
Removed the crate pushing bit.

PR #4317, opened 2017-11-21
Hacked borgs can no longer be locked down

Anewbe:
NOV 21ST EDIT: I changed how this works, the Scrambled Equipment Module should no longer prevent lockdown.


SilverTalismen:
Finally, hopefully we will see a fix for hacked/Antag borg camera's too, unless its been fixed already.

Anewbe:
Not yet, but sooner or later. I have that one guy who really likes borgs looking into it.

Nalarac:
The emagged items var was originally intended to allow for the scrambled equipment module to give cyborgs the unique items without having any other effects.

Anewbe:
Do they happen to have another var that tracks if they've been hacked or emagged, then?

Mechoid:
That would be 'emag' or 'emagged'. But, then again, it's an upgrade requiring diamond and illegal tech to make.

Anewbe:
Yeah, I don't particularly mind the Scrambled Module doing this.

PR #4319, opened 2017-11-21
Advanced Roller Bed

mistyLuminescence:
Adds the advanced roller bed to R&D, made using 4000 metal, 2000 glass and 2000 phoron (that's 2 sheets, 1 sheet and 1 sheet). It's just like a normal roller bed, except it's lighter (normal-sized) and prettier (blue!).

Why? Because R&D has cool stuff for chemists (bluespace/cryo beakers) and surgeons (IMS, laser scalpels, bone clamp), but nothing for paramedics yet. No, dartguns don't count.

Sprites [here](https://image.prntscr.com/image/2lzU0vY0QT_36UbRIcImFg.png).

also fight me travis


code/game/objects/structures/stool_bed_chair_nest/bed.dm
```diff
@@ -202,10 +202,17 @@
  */
 /obj/structure/bed/roller
 	name = "roller bed"
+	desc = "A portable bed-on-wheels made for transporting medical patients."
 	icon = 'icons/obj/rollerbed.dmi'
 	icon_state = "down"
 	anchored = 0
 	surgery_odds = 75
+	var/bedtype = "regular"
```
Anewbe: Make bedtype a path, rather than a string.

code/game/objects/structures/stool_bed_chair_nest/bed.dm
```diff
@@ -230,12 +240,17 @@
 	icon = 'icons/obj/rollerbed.dmi'
 	icon_state = "folded"
 	slot_flags = SLOT_BACK
-	w_class = ITEMSIZE_LARGE // Can't be put in backpacks. Oh well.
+	w_class = ITEMSIZE_LARGE
+	var/rollertype = "regular"
```
Anewbe: Make rollertype a path, rather than a string.

Nalarac:
I feel like it'd make more sense for this time to use a material like silver instead of phoron.

mistyLuminescence:
Five hours later... they're all paths now, so adding an even better roller bed will be much easier. And phoron's for the, uh, decompression systems. Or.. something like that.

(it's because not enough stuff uses phoron tbh)

PR #4320, opened 2017-11-22
Adds Submap (PoI) Seeding Procedure

Neerti:
Mostly ports /tg/'s method of seeding submaps into specific z-levels. Note that submaps are spawned in before other map generation things, such as tunnels or trees spawning.
PoI spawning has several restrictions, firstly being restricted to a specific area. The cave PoIs only spawn in 'unexplored' sections, so there should be no risk of them blocking the outposts or the premade pathways. PoIs are also forbidden from being too close to the edge of the world, and the entire PoI must be contained in the same area, so it will not eat the outposts or another PoI.
Due to the low number of submaps present, all of the PoIs are guaranteed to spawn. As more are added, this should fix itself.
Fixes a few weather icon related bugs.


Anewbe:
**BUG WAR NOW!!**

Atermonera:
So I heard you like hunting spiders

PR #4321, opened 2017-11-22
Adds Z Coordinate to Suit Sensor Computer

Neerti:
Fixes what I was was the computer not scanning across z-levels but was instead miscommunication on what was bugged, IE that the computer displays suit sensors for everyone on the station but gives no indication as to what deck they are on.
Bonus: Makes the map compile due to a path typo.


Atermonera:
Neerti is that english

Anewbe:
I can read it.

PR #4324, opened 2017-11-22
Adds rotate verb for pianos

BillyBangles:
Ported from Eros at Belsima's request, despite the fact that I'm not even sure we have a piano anywhere on the map. Also includes appropriate directional piano sprites.


Atermonera:
We don't have it mapped in, no (It and the violin were last seen as a steaming pile of laggy garbage that could only possibly sound good if your connection was rock-solid. Everyone else just heard broken sound that was painful to hear).

BillyBangles:
my god don't bring up the violin. *shivers*

But like I said, I ported this at Bel's request. Dunno what she plans to do with them. ¯\_(ツ)_/¯

code/game/objects/structures/musician.dm
```diff
@@ -27,6 +27,25 @@
 		desc = "This is a space piano, like a regular piano, but always in tune! Even if the musician isn't."
 		icon_state = "piano"
 
+/obj/structure/device/piano/verb/rotate()
+	set name = "Rotate Piano"
+	set category = "Object"
+	set src in oview(1)
+
+	if(config.ghost_interaction)
```
Anewbe: This means that if ghost_interaction is on mice, corpses, and restrained people will be able to turn the thing.

My suggestion would be to scrap this if/else, move all of the checks back one indentation, and then add something like 
`
if(istype(usr,/mob/dead/observer) && !config.ghost_interaction))
`
`->return`

Belsima:
@BillyBangles It's due to pain with mapping on downstreams.

BillyBangles:
Fair enough.

PR #4325, opened 2017-11-22
RUST updates

mistyLuminescence:
* RUST updated to work with TEGS. The RUST now dumps heat into the enviroment if the enviroment is cooler than ~5500k. This can be changed by modifiying the USION_MAX_ENVIRO_HEAT define.
* Radiation system changed. Radiation output now scales with reaction temperature. This can be modified by changing the PLASMA_TEMP_RADIATION_DIVISIOR define.
* Instability monitoring system added. This is fairly basic, but should function to alert engineers of when the reactor is becoming unstable for some reason.
* The RUST's light emissions now change color, depending on plasma temperature. Scales from orange - > green - > cyan - > dark blue - > magenta
* Added He3 as a proper chemical, currently just here for ease of use to make fuel rods easier.

All code tested by @Andromeda-K22 / Mazian#0462 on discord who can't into git. I'm just the messenger (and it compiles).

### DIFF
--- a/code/modules/power/fusion/core/_core.dm
+++ b/code/modules/power/fusion/core/_core.dm
@@ -41,7 +41,12 @@ var/list/fusion_cores = list()
 /obj/machinery/power/fusion_core/process()
 	if((stat & BROKEN) || !powernet || !owned_field)
 		Shutdown()
-
+	if(owned_field)
+		spawn(1)
+			owned_field.stability_monitor()
+			owned_field.radiation_scale()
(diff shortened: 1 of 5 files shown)


Andromeda-K22:
i'm going to go ahead and drop in a .dmm for anyone who wants to test these out - this is a RUST version of the north star's engine, and fits snugly in the exact same spot.

https://drive.google.com/open?id=1XfyXQ7yvG5Ns4tSL_RJdU_ka7bwnfzWZ

PR #4329, opened 2017-11-22
Creates synthblood reagent, adds it to merc medic borg hyposprays

Anewbe:
It's twice as effective as real blood, because otherwise in order to refill a dead bleeder you'd need to work on geological time.


mistyLuminescence:
A recipe for this in chemistry would be good, like synthmeat. Maybe something with cryoxadone (because blood is less complex than meat)? Cryox and iron?

Anewbe:
This will not be mixable.

PR #4330, opened 2017-11-23
Mobile Camera Pods

mistyLuminescence:
Adds mobile camera pods (MCPs).

- They're basically spy bugs, but bigger, way more obvious, and tacticooler.
- The box has six of them, and one monitor. Tap them against the monitor to link them, which lets the monitor view them.
- Each MCP has a randomly-generated identification number between 1000 and 9999.
- You can view each MCP through Security camera monitors as well, even when unlinked.
- Throw them into a room before breaching to get some idea of what's inside, or give them to an undercover officer to keep an eye on them, or "hide" (they're not subtle!) them in key locations to monitor on-the-go.

Basically a nerfed spy bug, because throwing a camera grenade into a room before entry is, hypothetically, tacticool as hell.

Also we can't actually _get_ them any way, but beg the admins for a box next time you're Warden!


Wutnold124:
tacticool*

Woodratt:
What happens if you shoot one?

mistyLuminescence:
Nothing, they're items. Currently the best way to deal with them is by throwing them somewhere else; not sure how I'd go about adding a way to shoot them.

Anewbe:
bullet_act()

Woodratt:
Really not keen on invincible cameras that can follow baddies around and relay their location with impunity.

Anewbe:
I agree with Woodrat. This will remain Do Not Merge until the cameras can be easily destroyed or otherwise rendered useless.

mistyLuminescence:
They can't follow them around unless someone's dragging them around, but working on a way to disable them now:

1) Smash them with a force 5+ object
2) Harm-intent click them in your hand
3) Verb to turn it off and on again, delinking it

mistyLuminescence:
All of the above is now possible. Also works a bit smoother.

code/datums/supplypacks/security.dm
```diff
@@ -167,6 +167,16 @@
 	containername = "Security surplus equipment"
 	access = null
 
+/datum/supply_packs/security/camerapods
```
Anewbe: Let's remove this, for now.

mistyLuminescence: Just to check, you know this means they won't be available outside of admins spawning them in?

Anewbe: Yes, that's the idea.

PR #4332, opened 2017-11-23
Adds hair color changing ability for prometheans.

Cyantime:
Part of https://github.com/VOREStation/VOREStation/pull/2515 that was asked for over here.
Adds it to prometheans, and makes them non-monochromatic.


Mechoid:
After prodding Kholdstare, a question: Would it be possible to give this to Prometheans as in the original PR you opened? I would probably do it anyway if the answer is no, just for ease-of-workflow.

Cyantime:
Sure, hold on

PR #4335, opened 2017-11-24
Adds Exploration Equipment

Neerti:
Ports and tweaks the Explorer gear from /tg/. Recolored to remove the mining purple, instead its now a nice blue because winter, I suppose.
Adds a new survival knife that is bootknife ready. Also from /tg/.
Replaces xenoarch GPS with budget /tg/ GPS that is more useful, as it can track other GPSes.
Adds explorer outfit datum for future uses.
Fixes a bug where hoods could be removed from the head with the hooded suits, as otherwise the hood teleports back inside at any distance.
Suit sensors can now see all surface levels if they are on the surface itself.

Pictures:
Explorer

GPS sprites
(Note the duplicate icon_state was fixed after taking this screenshot)


Wutnold124:
yay

Issue #3990, opened 2017-09-30
Skrell female colorable headtail cloth


SilverTalismen:
#### Brief description of the issue
Spawning with the colorable headtail cloth[ female] for skrell just gives you the black headtail cloth
https://gyazo.com/7fc3fb104345cd1a86a481ae4d754fa5
#### What you expected to happen
To get the cloth I wanted.
#### What actually happened
Spawn with just a black pair.
https://gyazo.com/1d4a1c48c2809e2cc0c944939a06fca3
#### Steps to reproduce
Pick out the cloth and a nice color.
spawn in
Realize you only got the black varient.
#### Additional info:
 - **Server Revision**:master - 2017-09-29
 - **Anything else you may wish to add** (Location if it's a mapping issue, etc)

PrismaticGynoid:
The male cloth too, last I checked.

Belsima:
Fix is up.

PR #4111, opened 2017-10-19
adds a buncha crap to loadout

Belsima:
Standardizes some grammar.

Adds: stethoscopes, lockets, medals, goggles, a white cane, a pair of bags, the powdered wig, a colorable cap, rabbit ears (to go with cat ears. im disgusted.), cheap lighters, greatcoats, and cameras.

Reorganizes some loadout items, and made others selections.

Made the wedding dress the actual wedding dress.

i'm sorry for the commits, at&t went to the wrong house we still have no internet

PR #4339, opened 2017-11-25
Loadout additions + fixes.

Belsima:
https://github.com/PolarisSS13/Polaris/pull/4111

This except I also fix headtail cloth.

Closes https://github.com/PolarisSS13/Polaris/issues/3990.

### DIFF
--- a/code/modules/client/preference_setup/loadout/loadout_accessories.dm
+++ b/code/modules/client/preference_setup/loadout/loadout_accessories.dm
@@ -199,7 +199,7 @@
 
 
 /datum/gear/accessory/sweater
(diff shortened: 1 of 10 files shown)


Belsima:
...I don't understand what broke.

Belsima:
Oddly, it compiled on my PC, even though it upset travis.

Atermonera:
The thing that failed was a unit test, which is specific to travis. The thing with comments within a multi-line list needing to end in a comma is also an example of something that travis catches, but the compiler doesn't.

Belsima:
Ah, thanks.

PR #4341, opened 2017-11-25
Reagents use pathing instead of strings

Anewbe:
Heavily inspired by and partially ported from work done on Baystation.
- Reagent commands use the paths of the chemicals instead of strings. This means that you need to use /datum/reagent/dylovene instead of "anti_toxin". It also means that the compiler will catch you if you mess up.
- Swaps the reagent dispenser cartridge sprites for fancier ones that were apparently made long ago by Mechoid.

- [ ] More testing. First tests have been done, but this is gonna break basically anything at random, so I need to spend more time on hunting.


MisterLayne:
It is the pathing revolution.

Anewbe:
Viva.

Wickedtemp:
Finally, no more abnormal looking giant grey squares.

Anewbe:

PR #4354, opened 2017-11-27
Adds sensor preferences!

mistyLuminescence:
Adds an option in the general tab of character setup, under the equipment section (same as PDA style, backpack/satchel/neither, etc) - _sensor preferences_. You can select a mode of suit sensors - off, binary, vitals, tracking - or no preference - and that mode will always be enabled at roundstart for you.

So if you set your sensor preferences to tracking, you'll always start with sensors to tracking.
If you set your sensors to off, you'll always start with sensors disabled.
It defaults to 'no preference', which is the current setting - randomly generated, 25% chance each for off/binary/vitals/tracking.

Bonus, it's different for each character. Do you have a sensible character and a paranoid character? No problem - you could set your sensor prefs to 'tracking' and 'binary' respectively, if you wanted to.

"This doesn't affect me, I don't want it." You don't have to use it! Defaults to 'no preference'.

Tested, and it works, and also Travis fight me.

Edit: pretty sure Travis failed 'cause of SC test
Edit 2: did it again now the test is over

### DIFF
--- a/code/_helpers/global_lists.dm
+++ b/code/_helpers/global_lists.dm
@@ -58,6 +58,7 @@ var/datum/category_collection/underwear/global_underwear = new()
 	//Backpacks
 var/global/list/backbaglist = list("Nothing", "Backpack", "Satchel", "Satchel Alt", "Messenger Bag")
 var/global/list/pdachoicelist = list("Default", "Slim", "Old", "Rugged")
+var/global/list/sensorpreflist = list("Off", "Binary", "Vitals", "Tracking", "No Preference")
 var/global/list/exclude_jobs = list(/datum/job/ai,/datum/job/cyborg)
 
 // Visual nets
--- a/code/modules/client/preference_setup/general/04_equipment.dm
+++ b/code/modules/client/preference_setup/general/04_equipment.dm
@@ -11,12 +11,14 @@
 	S["all_underwear_metadata"] >> pref.all_underwear_metadata
 	S["backbag"]	>> pref.backbag
 	S["pdachoice"]	>> pref.pdachoice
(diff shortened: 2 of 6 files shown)


Wickedtemp:
I like this. I like this a lot.

I always set my characters' sensors to tracking, even if they're antag, so I'd personally use this. And for people who don't like it, well, they don't have to change anything if the default option is 'no preference'.

mistyLuminescence:
update: still ded, probably SC

Neerti:
Not really in favor of this as it seems like that trait proposal that was shot down a few weeks ago that would've set your sensors on full without any thinking required, and this is essentially that.

mistyLuminescence:
This is similar to that, except it's cleaner and has more options (e.g. if you don't want to be tracked, but you still want med to know when you're hurt). It's completely optional, it defaults to no pref/random, but this way it doesn't punish people with memory problems (like yours truly) who genuinely forget to set them a good 50% of the time.

Pros:
- Saves time for people who remember
- Stops punishing those who don't
- Can be configured, opt-in

Cons:
- Probably means more people have sensors on... but only people who'd max them when med goes "sensors pls"
- Might also mean more people have sensors /off/, now they know they have the option of doing that without fiddling with it every round

Anewbe:
This was discussed in the Staff Meeting. There was no solid staff-side support and more than a little staff-side opposition.
It makes antagging more difficult than it already is, and the variance in sensor level that comes from the current system is beneficial, in our opinion.
Closing this PR.

mistyLuminescence:
💀 rip
Reminder that maxing your suit sensors is still a thing you should be doing, folks!

SpadesNeil:
-le yoink-

PR #4360, opened 2017-11-27
Mobs can't push objects that they can't pull.

Cyantime:
Stops mice from rearranging crates in maint, etc.


mistyLuminescence:
This also means construction drones can't move the pipe dispensers/lockers/computer frames etc.

Cyantime:
If they can't pull them and need to fuss around pushing them, they should just be allowed to pull them.

PR #4365, opened 2017-11-28
Adds a Wild Slime type for future surface-things.

Neerti:


ElectricB1ue:
actually this is a feral slime

PR #4374, opened 2017-11-30
Makes the clotting cargo crate more expensive

Anewbe:
As per the last Staff Meeting, cargo's supply of clotting agent is more expensive.

### DIFF
--- a/code/datums/supplypacks/medical.dm
+++ b/code/datums/supplypacks/medical.dm
@@ -82,7 +82,7 @@
 	contains = list(
 			/obj/item/weapon/storage/firstaid/clotting
 			)
-	cost = 40
+	cost = 100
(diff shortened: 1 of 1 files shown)


mistyLuminescence:
100 points is ridiculously excessive for clotting medicine. 40 is pretty high on its own.

Anewbe:
Staffed discussed it, and this was the compromise.

mistyLuminescence:
100 points is a 'compromise' in the same way that "how about I don't kill you, I just stab you in the chest a few times?" is a compromise. If you want to make each individual injector more expensive, why not just have less per crate? 4 clotting injectors and 3 empty injectors, maybe.

Anewbe:
Your other option is having the crate removed entirely. I'm tired of having to fight tooth and nail over every change that affects Medical.

PR #4376, opened 2017-11-30
Adds New Blob

Neerti:
Current blob is quite terrible and buggy. Its pretty slow, unintelligent, and it literally hogs the process controller because it ``sleep()``s constantly.

Semi-ports /tg/'s blob. There are a large variety of different colors, which all have unique properties. I added a few types of my own as well. I elected to have the blob be controlled by a basic algorithm and not a player, so that I can better tune the balance and to reduce salt.

Note that the random event only spawns 'easy' and 'medium' blob types. The others are admin only for now since I don't trust our current event system to make good calls on when a blob should happen. This will change when/if I finish the new event system that makes heavier use off conditionals and activity tracking.

Other features:
Fireaxes can 'cleave' hitting simple mobs next to them. This was added to make fighting certain blobs less tedius.
Adds some helper functions from /tg/.


Anewbe:
Looks solid.
Can I get an image/video of the infesting spores at work?

Anewbe:
Got my explanation, that's all I wanted.

May RNGsus have mercy on us all.

PR #4377, opened 2017-11-30
Can no longer run around while in a roller chair

Anewbe:
'Tis a Bug, with exploitable benefits.


PrismaticGynoid:
This doesn't break wheelchairs, does it?

Anewbe:
Nope, tested those specifically.

PR #4378, opened 2017-12-01
Adds Reach Variable to Weapons

Neerti:
Melee weapons can now potentially attack from farther away.
Obstacles are taken into account, so you cannot hit people through windows, but you can attack over specific things such as tables, or other people.
Currently only the spear can do this, with a range of two tiles, however the attack speed for the spear was reduced, so it may remain a situational weapon.
The intention for this is to make specific weapons feel different to each other besides 'does more damage', and I got ideas on other kinds of weapon adjustments later to make them feel a bit more unique.


Anewbe:
What happens if you slap reach on a weapon that cleaves?

Neerti:
The cleave code checks for adjacency, so it won't trigger on a two tile ranged attack.

Anewbe:
Alright.

PR #4382, opened 2017-12-02
Adds parachutes

Anewbe:
Look here, Anewbe's over-complicating things again.
- Consolidated falling out of the sky into fall_impact().
- Moved backpack sprites into their own file.
- Added some vars. Hovering doesn't do anything yet, because I wanted to roll this out. Softfall makes you not take damage from falling a floor or so. Parachuting means non-human mobs won't take damage when falling out of the sky. Humans have a proc that looks for parachuting items.
- Adds the Parachute. Parachutes are slightly smaller backpacks, with an added feature. If you take a few seconds to pack the parachute, through a verb, it will prevent your horrible grisly splatting.


Neerti:
You seem to have committed your .gitconfig, which is probably a bad idea.

Anewbe:
Oh darn it, standby.

Anewbe:
I believe it has been corrected.

PR #4390, opened 2017-12-03
Makes species bleed out faster or slower

Anewbe:
- Adds a var to let species bleed out faster or slower.
- Skrell bleed out a bit faster.
- Unathi bleed out a bit slower.
- Also makes Unathi have a bit more health, because I was in the file ~~and no one plays them anyway~~.


elgeonmb:
time to play unathi

Anewbe:
Do it, coward.

PR #4391, opened 2017-12-03
Ghosts no longer show up over living mobs

Anewbe:
- Cultists with a lazy visible tag-along are no longer unclickable
- Griefers with a billion spectating specters are no longer mildly more difficult to ban


Cyantime:
Isn't this what 'Toggle Ghost Vison' is for?
https://github.com/PolarisSS13/Polaris/blob/master/code/modules/mob/dead/observer/observer.dm#L722

Atermonera:
That only affects ghosts, afaict. Living people still have trouble targeting cult mobs who've got lazy ghosts floating over them

PR #4392, opened 2017-12-03
Books with actual pages

BillyBangles:
Adds a new object, book/bundle, which works the same way as a paper bundle except you can't take the pages out. You can create one by clipping some papers together and putting the bundle into the book binder. The binder won't take any bundles with photos in it because that won't play nice with the database.

I did this to fix the infamous "PART 1" "PART 2" "PART 3" etc. books from the library system, as well as books that just have a google drive link in them instead of the actual book. What real life book have you read that only has one page anyways?

### DIFF
--- a/code/modules/library/lib_machines.dm
+++ b/code/modules/library/lib_machines.dm
@@ -457,17 +457,37 @@ datum/borrowbook // Datum used to keep track of who has borrowed what when and f
 	density = 1
 
 /obj/machinery/bookbinder/attackby(var/obj/O as obj, var/mob/user as mob)
-	if(istype(O, /obj/item/weapon/paper))
-		user.drop_item()
-		O.loc = src
-		user.visible_message("[user] loads some paper into [src].", "You load some paper into [src].")
-		src.visible_message("[src] begins to hum as it warms up its printing drums.")
-		sleep(rand(200,400))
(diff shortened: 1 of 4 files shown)


Wutnold124:
nice(?)

BillyBangles:
@Anewbe You can merge if you like but you might wanna slap a Do Not Merge on here until I make sure you can upload these to the database.

Anewbe:
We won't merge til you say it's ready, don't worry.

code/modules/library/lib_machines.dm
```diff
@@ -457,17 +457,37 @@ datum/borrowbook // Datum used to keep track of who has borrowed what when and f
...
-		var/obj/item/weapon/book/b = new(src.loc)
-		b.dat = O:info
-		b.name = "Print Job #" + "[rand(100, 999)]"
-		b.icon_state = "book[rand(1,7)]"
-		qdel(O)
+	if(istype(O, /obj/item/weapon/paper) || istype(O, /obj/item/weapon/paper_bundle))
+		if(istype(O, /obj/item/weapon/paper))
+			user.drop_item()
+			O.loc = src
+			user.visible_message("[user] loads a sheet of paper into [src].", "You load a sheet of paper into [src].")
```
Belsima: \the

PR #4393, opened 2017-12-03
Flashes now click when they say they click.

Belsima:


Anewbe:
Make it the gun click instead?

Belsima:
Good idea.

PR #4394, opened 2017-12-03
Maint drones may now ventcrawl.

Belsima:
Port of https://github.com/VOREStation/VOREStation/pull/1016.

### DIFF
--- a/code/modules/mob/living/silicon/robot/drone/drone.dm
+++ b/code/modules/mob/living/silicon/robot/drone/drone.dm
@@ -341,10 +341,4 @@ var/list/mob_hat_cache = list()
 
 /mob/living/silicon/robot/drone
 	can_enter_vent_with = list(
-		/obj/item/device/radio/borg,
-		/obj/machinery/camera,
-		/obj/item/device/camera/siliconcam/drone_camera,
-		/obj/item/weapon/robot_module/drone,
-		/obj/item/weapon/cell,
-		/obj/screen/movable/ability_master,
-		/obj/item/weapon/card/id/synthetic)
+		/obj)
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/silicon/robot/drone/drone.dm
```diff
@@ -337,3 +338,13 @@ var/list/mob_hat_cache = list()
 /mob/living/silicon/robot/drone/construction/updatename()
 	real_name = "construction drone ([rand(100,999)])"
 	name = real_name
+
+/mob/living/silicon/robot/drone
+	can_enter_vent_with = list(
```
Cyantime: You might as well make them able to ventcrawl with any /obj, so you don't need to keep de-equipping your modules to crawl around. You can just re-equip them once you're in the pipe, so it's just an annoyance.

Belsima: They were implemented with them stowed and that's how they were mentioned in the staff meeting, so I'll leave it unless @Anewbe says otherwise.

Anewbe: Don't care how it works, just want Drones with ventcrawl.

Belsima: So, if I just make the can_ventcrawl into /obj that would allow them to ventcrawl with anything, yes? Or no?

Mechoid: /obj or /obj/item would allow it, yes.

Belsima: doned

PR #4395, opened 2017-12-03
Ports check_radiation verb for ghosts

PrismaticGynoid:
From https://github.com/Baystation12/Baystation12/pull/19390. Ghosts now have a check radiation verb they can use to see the radiation level of where they're standing, similar to geiger counters. Tested in every scenario involving radiation I could think of, and it works just fine.


Anewbe:
Looks fine, but have you tested this on our code?

PrismaticGynoid:
In as many situations that I could think of, yes.

Anewbe:
Excellent, thank you.

PR #4397, opened 2017-12-04
Adds blennophobia.

Belsima:
Pushing so that flaws in my code can be pointed out; DNM for now, pending further testing with how things like the hat and species work.


Atermonera:
Your full_fear_up messages are for spiders

Belsima:
how did i miss that

Belsima:
That should be good.

PR #4399, opened 2017-12-04
DJ Outpost PoI

mistyLuminescence:
Adds a DJ outpost to the points of interest list. Once a mildly successful radio station (albeit the only one for many miles around), Sif Free Radio mysteriously went off the air not long ago, their low-quality pirated tunes falling silent. Now, as you wander through the icy wastelands of Sif, the blizzard howling all around you and lashing at your face, you spot a large wooden cabin in the distance. Shelter? Warmth? You approach, seeking both - but will you recover from the blizzard, or will you fall victim to the fate that befell the SFR crew?

[Looks like this](https://image.prntscr.com/image/4CtUH9AhSdO6XWjkQn8B3A.png).


Nalarac:
Are the spiders set to return home so they don't wander away from the PoI?

mistyLuminescence:
Uh... nope. Should they be? How do I do that?

Woodratt:
They'll wander out, they have a return to home variable, can't think of what it's called off the top of my head, but you'll want to set it.

Also did you ensure that APC starts off? Otherwise it'll die due to lack of power and trip an alarm on the engineering sensors.

Magneld:
Is that a combat knife or a tacknife? Either or is fine honestly but the combat one is robust-er. Your choice which one you want it to be though.

Mechoid:
' returns_home ' is the variable you want to set to 1 for going back to their spawn turf.

mistyLuminescence:
APC starts on (will change), combat knife (probably keeping? not sure), and will set spiders to go back to whence they came

Magneld:
It being the combat knife is fine really, was mostly just curious. Pretty sure the explorers will have bootknives anyway so no big deal.

mistyLuminescence:
Still a WIP? Does anything else need changing?

PR #4408, opened 2017-12-05
Reduces adv medical item healing from 5 to 3

Anewbe:
5 Per Wound is a wee bit much. Drops it to 3. Still useful, not as strong.

### DIFF
--- a/code/game/objects/items/stacks/medical.dm
+++ b/code/game/objects/items/stacks/medical.dm
@@ -164,7 +164,7 @@
 	singular_name = "advanced trauma kit"
 	desc = "An advanced trauma kit for severe injuries."
 	icon_state = "traumakit"
-	heal_brute = 5
+	heal_brute = 3
 	origin_tech = list(TECH_BIO = 1)
 
 /obj/item/stack/medical/advanced/bruise_pack/attack(mob/living/carbon/M as mob, mob/user as mob)
@@ -225,7 +225,7 @@
 	singular_name = "advanced burn kit"
 	desc = "An advanced treatment kit for severe burns."
 	icon_state = "burnkit"
-	heal_burn = 5
+	heal_burn = 3
 	origin_tech = list(TECH_BIO = 1)
 
 


Wutnold124:
b-but

my multiples of 5

mistyLuminescence:
Given that the vast majority of injuries are only one wound, not sure this is needed.

Woodratt:
Sounds good Anewbe.

Natje0:
How is 5 a wee much?
I don't see any reason for this nerf.

Ketrai:
I mean, it is an advanced bandage. At this rate you're just going to push medics to carry around Bicardine all the time. Which a lot of them already do. With autoheals almost completely gone, I don't mind that extra 5 points making it slightly quicker to fix up big wounds.

elgeonmb:
Newbe please, don't make me do chemistry. i hate chemistry

MisterLayne:
Think they're called advanced for a reason. If you really wanted to do this change, I'd suggest having a non-advanced version, and adding advanced patches (and regular patches) to be construct-able in the Research Protolathe given enough biological research.

mistyLuminescence:
That _does_ give me an idea... nerf 'advanced' medical kits as per this PR - and rename them to just trauma/burn kits too. Then add an advanced version in Cargo for big points (80/crate?) with a medkit full of hybrid brute/burn kits with bigger autoheal and direct heal.

Gives us 3 tiers:
Tier 1: bandage, ointment. Found in first-aid kits.
Tier 2: trauma kit, burn kit. Found in advanced first-aid kits and medical vendors.
Tier 3: hybrid triage kit. Found in Cargo.

Anewbe:
When the numbers were originally added, we chose 5 damage as an arbitrary start point. We've seen them in action now, and 5 points per wound is more than we thought it was. As such, we're going to try 3. That's how design works.
We don't need three tiers of everything for Medical.

mistyLuminescence:
You're right, we don't need three tiers... how about _four_ tie- okay I'll stop
Honestly, the only reason this is a big deal is because it messes with the defib threshold - 2 less damage per kit applied means marginally less people being brought to like... 194 damage after liberal application of trauma kits to every injury. Those are the times defibs are actually pretty fun.

PR #4409, opened 2017-12-05
Lowers standard laser damage to 30

Anewbe:
300 per carbine is still nothing to laugh at, but now gunfights will take ... Roughly one shot longer?

### DIFF
--- a/code/modules/projectiles/projectile/beams.dm
+++ b/code/modules/projectiles/projectile/beams.dm
@@ -2,7 +2,7 @@
 	name = "laser"
 	icon_state = "laser"
 	pass_flags = PASSTABLE | PASSGLASS | PASSGRILLE
-	damage = 40
+	damage = 30
 	damage_type = BURN
 	check_armour = "laser"
 	eyeblur = 4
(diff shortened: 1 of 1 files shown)


Wutnold124:
only one shot? for shame, anewbe, for shame

REAL laser nerfs just make them really powerful laserpointers that hurt your eyes

Belsima:
Won't ballistics still easily destroy in gunfights, though?

Neerti:
It may be desirable to adjust this too.

Belsima:
I'm just seeing repeated nerfs to lasers; it feels like at some point we may have ballistics nearly objectively outclass them if that continues.

Anewbe:
If we need to adjust ballistics at a later date we will.

Magneld:
With some exceptions this should make the laser carbine about even with most ballistics when rate of fire and ammo capacity is taken into account. Thats the idea anyway we'll see where we end up.

Nalarac:
I have some issues with this change:
1.) The x-ray laser gun will be objectively better (it does 25 damage but you get armor pen and 12 shots so it does 300 damage in the end)
2.) I feel like it's a pretty big nerf to the laser carbine that emagged/unlocked security cyborgs get
3.) (not "necessarily" an issue) It will now take 1 to 2 more shots from the carbine to kill a giant spider

Woodratt:
Maybe try 35 instead?

Anewbe:
Closing in favor of #4415

PR #4422, opened 2017-12-08
Removes the military clothing from the loadout, repaths things

Anewbe:
- Removes the Fleet, Marine, and SifGuard PT uniforms from the loadout. Left the underwear in place.
- Removes the Fleet and Marine Utility uniforms from the loadout. Replaced them with statless versions with the same sprites.
- Removes Service Jackets from the loadout.
- Repaths a bunch of stuff that was still /expeditionary

### DIFF
--- a/code/modules/client/preference_setup/loadout/loadout_uniform.dm
+++ b/code/modules/client/preference_setup/loadout/loadout_uniform.dm
@@ -371,30 +371,17 @@
 	display_name = "sexy maid uniform"
 	path = /obj/item/clothing/under/dress/sexymaid
 
-/datum/gear/uniform/pt
-	display_name = "pt uniform"
-	path = /obj/item/clothing/under/pt
-
-/datum/gear/uniform/pt/expeditionary
-	display_name = "pt uniform, sifguard"
-	path = /obj/item/clothing/under/pt/expeditionary
-
-/datum/gear/uniform/pt/fleet
-	display_name = "pt uniform, fleet"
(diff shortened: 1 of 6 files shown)


Woodratt:
This is probably something that needs a staff discussion tag and talk at the staff meeting.

code/modules/client/preference_setup/loadout/loadout_uniform.dm
```diff
@@ -371,30 +371,17 @@
...
-	display_name = "pt uniform, fleet"
-	path = /obj/item/clothing/under/pt/fleet
-
 /datum/gear/uniform/utility
 	display_name = "utility, black"
 	path = /obj/item/clothing/under/utility
-	cost = 2
 
-/datum/gear/uniform/utility/fleet
-	display_name = "utility, navy"
```
Belsima: I assumed 'navy' was referring to it in the same way as the 'navy' security suit- though, the clarification here can't hurt.

Belsima: though clarifying the 'navy' security suit to 'blue' would also maybe be a good idea

Anewbe: I removed the term "Navy" here because it might be confused with the actual Navy, which is sort of what the Fleet is.

Belsima: fair

PR #4423, opened 2017-12-08
Grind up some more stuff!

mistyLuminescence:
You know grinders? The ones you can grind food items and some materials in?
You can now grind steel, plasteel, snow, sandstone and glass to get:
Steel: iron and carbon (coal)
Plasteel: iron, carbon and platinum
Snow: water
Sandstone: silicon, oxygen
Glass: silicon

Because some of these have more than one reagent, you won't be able to grind them if your beaker doesn't have enough space.
Grinder interaction with other materials and food items is the same as before.

also fight me travis

### DIFF
--- a/code/modules/reagents/Chemistry-Machinery.dm
+++ b/code/modules/reagents/Chemistry-Machinery.dm
@@ -297,11 +297,11 @@
 		/obj/item/stack/material/platinum = list("platinum"),
 		/obj/item/stack/material/mhydrogen = list("hydrogen"),
 		/obj/item/stack/material/steel = list("iron", "carbon"),
-		/obj/item/stack/material/plasteel = list("iron", "iron", "carbon", "carbon", "platinum"), //8 iron, 8 carbon, 4 platinum
+		/obj/item/stack/material/plasteel = list("iron", "iron", "carbon", "carbon", "platinum"), //8 iron, 8 carbon, 4 platinum,
 		/obj/item/stack/material/snow = list("water"),
 		/obj/item/stack/material/sandstone = list("silicon", "oxygen"),
 		/obj/item/stack/material/glass = list("silicon"),
-		/obj/item/stack/material/glass/phoronglass = list("platinum", "silicon", "silicon", "silicon"), //5 platinum, 15 silicon
+		/obj/item/stack/material/glass/phoronglass = list("platinum", "silicon", "silicon", "silicon"), //5 platinum, 15 silicon,
 		)
 
 /obj/machinery/reagentgrinder/New()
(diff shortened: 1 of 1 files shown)


code/modules/reagents/Chemistry-Machinery.dm
```diff
@@ -289,13 +289,18 @@
...
+		/obj/item/stack/material/gold = list("gold"),
+		/obj/item/stack/material/silver = list("silver"),
+		/obj/item/stack/material/platinum = list("platinum"),
+		/obj/item/stack/material/mhydrogen = list("hydrogen"),
+		/obj/item/stack/material/steel = list("iron", "carbon"),
+		/obj/item/stack/material/plasteel = list("iron", "carbon", "platinum"),
+		/obj/item/stack/material/snow = list("water"),
+		/obj/item/stack/material/sandstone = list("silicon", "oxygen"),
+		/obj/item/stack/material/glass = list("silicon"),
 		)
```
MisterLayne: I formally request we actually get crayons added to this list, so we can finally make paint with crayon dust.

mistyLuminescence: crayons ain't a stack tho
maybe if they had crayon juice inside them..... hmm

Anewbe:
If this does get in, which I am **at best** ambivalent about, this isn't how I want it done. A sheet of plasteel isn't an even mix of its components, as you can see in the file I'll link below. Additionally, I think it's safe to assume that every sheet of a material is roughly the same amount of material, so just flat multiplying everything by that constant isn't want I want to see.

[Alloy List](https://github.com/PolarisSS13/Polaris/blob/0bb15b27324a371804f8f625708cbe8bb03b00e0/code/modules/mining/alloys.dm)

mistyLuminescence:
Whoops, only just saw this now. So...

Steel: 10 Carbon, 10 Iron
Plasteel: 8 Carbon, 8 Iron, 4 Platinum
Snow: as-is
Sandstone: as-is
Glass: as-is
Borosilicate glass: 5 Platinum, 15 Silicon

Can do.

mistyLuminescence:
It's the thing where Travis doesn't like comments in lists.

mistyLuminescence:
okay yeah it was the thing where travis doesn't like comments in lists

mistyLuminescence:
True but it's working now and if I touch it again bad things happen to travis

code/modules/reagents/Chemistry-Machinery.dm
```diff
@@ -497,12 +503,18 @@
 		if(sheet_reagents[O.type])
 			var/obj/item/stack/stack = O
 			if(istype(stack))
+				var/list/sheet_components = sheet_reagents[stack.type]
```
Atermonera: Use only `sheet_reagents[stack.type]` or `sheet_components`. Of the four subsequent uses within the scope of this variable, you use the variable for only one of them

mistyLuminescence: 👍

PR #4424, opened 2017-12-08
Crawling

PrismaticGynoid:
Mobs now have the ability to "crawl" to an adjacent turf by click-dragging themselves to it. This would allow people unable to stand to still move around, though very slowly. You could also do this to other movable mobs or objects, if you really wanted to.

Moving this way takes about 2.5 seconds, plus half a second for each point of weakness you have. For a person missing a leg, this would take 5 seconds. It only works if the mob doing this is alive, conscious, unrestrained, not stunned or paralyzed, and in an area with gravity - no crawling through space.

Mobs lying on the ground are now able to buckle themselves to chairs or beds. For people missing legs or feet, this means no more being stuck helpless on the ground forever because Baldie McGreytide clicked on your wheelchair.

Tested in many different situations, and I've patched every bug that popped up.

### DIFF
--- a/code/game/objects/buckling.dm
+++ b/code/game/objects/buckling.dm
@@ -68,7 +68,7 @@
 /atom/movable/proc/user_buckle_mob(mob/living/M, mob/user, var/forced = FALSE, var/silent = FALSE)
 	if(!ticker)
 		user << "<span class='warning'>You can't buckle anyone in before the game starts.</span>"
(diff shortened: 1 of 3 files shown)


mistyLuminescence:
good PR

MisterLayne:
**<iframe width="420" height="315"
src="https://www.youtube.com/embed/XGSy3_Czz8k">
</iframe>**

MisterLayne:
https://www.youtube.com/watch?v=eCiFO7qV54E

PR #4430, opened 2017-12-10
Fixes rollers and mediborgs

mistyLuminescence:
Mediborgs had a hard time with roller beds. They have an easier time with roller beds now.
(... i'm so sorry i was v tired when i made the advroller PR)

Disclaimer: yes this is an actual bugfix, the rollerbed holder item broke
Disclaimer 2: it works for real this time

### DIFF
--- a/code/game/objects/structures/stool_bed_chair_nest/bed.dm
+++ b/code/game/objects/structures/stool_bed_chair_nest/bed.dm
@@ -95,7 +95,7 @@
 		qdel(src)
 	else if(istype(W,/obj/item/stack))
 		if(padding_material)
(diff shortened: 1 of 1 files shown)


Anewbe:
Have you tested this?

mistyLuminescence:
Yep. Turns out there was another advanced roller bug we didn't notice, this fixes that (it's also included in the pill PR but I made a new PR for it when I realised how weird it was).

mistyLuminescence:
Double-triple-quadruple tested it just in case and it definitely fixes the whole 'crisis borgs can't pick up roller beds unless they're already folded' thing.

PR #4442, opened 2017-12-16
Lots of species stuff ported from Baystation

BothNevarBackwards:
A bunch of stuff that lets species have more custom behaviours.

Edit: the footprints are the only thing that has an impact without adding more code.


Anewbe:
Have you tested all of this?

BothNevarBackwards:
Yup, I tested each between commits.

Anewbe:
On our code, or on Baycode?
Because I don't actually remember getting the flash thing to work.

BothNevarBackwards:
I tested it locally. I assume that it has been tested for baycode but all testing I did was with polaris code.

code/game/objects/effects/decals/Cleanable/tracks.dm
```diff
@@ -154,20 +154,49 @@ var/global/list/image/fluidtrack_cache=list()
...
+	drydesc = "They look like dried tracks left by a mammal."
+	coming_state = "paw1"
+	going_state  = "paw2"
+	random_icon_states = null
+	amount = 0
+
+/obj/effect/decal/cleanable/blood/tracks/claw
+	name = "wet tracks"
+	dryname = "dried tracks"
+	desc = "They look like still wet tracks left by a reptile."
```
Belsima: Should probably generalize it to a clawed creature. Non-reptiles can easily have similiar feet e.g. avians.

Neerti:
Can't see any problem with this.

PR #4450, opened 2017-12-18
Allows modkits to be created for any item.

mistyLuminescence:
Port of [Baystation12/Baystation12/19744](https://github.com/Baystation12/Baystation12/pull/19744) - allows modkits to be created for any item. Basically means if you want your grandpa's pulse rifle as a custom item, you can do that now without breaking balance - you get grandpa's pulse rifle modification kit instead.

It works in all the situations I could think of, but if anyone wants to take a look over the code, that'd be great.


mistyLuminescence:
If nobody's noticed anything wrong with this, might be worth giving it a try?

mistyLuminescence:
If anyone has the time to review this, I'd appreciate it.

code/game/objects/items/paintkit.dm
```diff
@@ -18,95 +21,156 @@
...
+	if(istype(W, /obj/item/device/kit))
+		var/obj/item/device/kit/K = W
+		K.customize(src, user)
+		return
+
+	..()
+
 // Root hardsuit kit defines.
 // Icons for modified hardsuits need to be in the proper .dmis because suit cyclers may cock them up.
 /obj/item/device/kit/suit
```
Atermonera: Can you add a subtype or a provision for `/obj/item/clothing/suit/storage/hooded`? Because those store the helmet item internally, so you can't really apply the kit to the hood itself to change it

PR #4451, opened 2017-12-19
Ports hoists, as well as some macros, from Aurorastation.

MarinaGryphon:
Ported from Aurorastation, where I first implemented them. There's still a lot of work to be done to make sure this works properly. It's likely there are a few Aurora-isms in there, like certain macros that Polaris might not have, but I'll fix them ASAP.
Credit to SonicGotNuked for the sprites.

The new macros/procs are used like this:

`span(class, text)` expands to `"<span class='[class]'>[text]</span>"`.
`use_check` is a rather large proc, but it's documented in the code. Essentially, it returns 0 if something can be used by a mob, and if not it returns a value depending on why they can't use it. The flags are in `code/__defines/misc.dm`.


code/_helpers/unsorted.dm
```diff
@@ -1357,3 +1357,48 @@ var/mob/dview/dview_mob = new
...
+// Flags are in `code/__defines/misc.dm`
+/atom/proc/use_check(mob/user, use_flags = 0, show_messages = FALSE)
+	. = 0
+	if (NOT_FLAG(USE_ALLOW_NONLIVING) && !isliving(user))
+		// No message for ghosts.
+		return USE_FAIL_NONLIVING
+
+	if (NOT_FLAG(USE_ALLOW_NON_ADJACENT) && !Adjacent(user))
+		if (show_messages)
+			user << "<span class='notice'>You're too far away from [src] to do that.</span>"
```
Belsima: to_chat

code/_helpers/unsorted.dm
```diff
@@ -1357,3 +1357,48 @@ var/mob/dview/dview_mob = new
...
+		return USE_FAIL_NON_ADJACENT
+
+	if (NOT_FLAG(USE_ALLOW_DEAD) && user.stat == DEAD)
+		if (show_messages)
+			user << "<span class='notice'>How do you expect to do that when you're dead?</span>"
```
Belsima: to_chat

code/_helpers/unsorted.dm
```diff
@@ -1357,3 +1357,48 @@ var/mob/dview/dview_mob = new
...
+		return USE_FAIL_DEAD
+
+	if (NOT_FLAG(USE_ALLOW_INCAPACITATED) && (user.incapacitated()))
+		if (show_messages)
+			user << "<span class='notice'>You cannot do that in your current state.</span>"
```
Belsima: to_chat

code/_helpers/unsorted.dm
```diff
@@ -1357,3 +1357,48 @@ var/mob/dview/dview_mob = new
...
+		return USE_FAIL_INCAPACITATED
+
+	if (NOT_FLAG(USE_ALLOW_NON_ADV_TOOL_USR) && !user.IsAdvancedToolUser())
+		if (show_messages)
+			user << "<span class='notice'>You don't know how to operate [src].</span>"
```
Belsima: to_chat

code/_helpers/unsorted.dm
```diff
@@ -1357,3 +1357,48 @@ var/mob/dview/dview_mob = new
...
+		return USE_FAIL_NON_ADV_TOOL_USR
+
+	if (HAS_FLAG(USE_DISALLOW_SILICONS) && issilicon(user))
+		if (show_messages)
+			user << "<span class='notice'>How do you propose doing that without hands?</span>"
```
Belsima: to_chat

code/_helpers/unsorted.dm
```diff
@@ -1357,3 +1357,48 @@ var/mob/dview/dview_mob = new
...
+		return USE_FAIL_IS_SILICON
+
+	if (HAS_FLAG(USE_FORCE_SRC_IN_USER) && !(src in user))
+		if (show_messages)
+			user << "<span class='notice'>You need to be holding [src] to do that.</span>"
```
Belsima: to_chat

Cyantime:
qdel_null() already exists here.
https://github.com/PolarisSS13/Polaris/blob/master/code/_macros.dm#L66

MarinaGryphon:
Right, QDEL_NULL was just a case issue, then, alright.

code/_helpers/unsorted.dm
MarinaGryphon: I'm pretty sure that [src] inserts \the automatically, as opposed to [src.name] which doesn't. I remember being told that, at least.

code/modules/multiz/hoist.dm
```diff
@@ -0,0 +1,270 @@
...
+	var/obj/structure/hoist/source_hoist
+	can_buckle = 1
+	anchored = 1
+	description_info = "Click and drag someone (or any object) to this to attach them to the clamp. If you are within reach, when you click and drag this to a turf adjacent to you, it will move the attached object there and release it."
+
+/obj/effect/hoist_hook/attack_hand(mob/living/user)
+	return // This has to be overridden so that it works properly.
+
+/obj/effect/hoist_hook/MouseDrop_T(atom/movable/AM,mob/user)
+	if (use_check(user, USE_DISALLOW_SILICONS))
```
Leshana: In this PR at least, `use_check()` it being used only once: here.  
This line could have been replaced with  `if(issilicon(user))`  instead of porting use_check.    

The rest of use_check's functionality is unused.  There are even places only a few lines later where it could have been used but isn't!  
That said, I **do** like the idea of use_check in principle, its kind of cool to consolidate some of these checks in a proc where it can be standardized!  If we actually use it...

MarinaGryphon: It's all used, actually. It checks for certain flags *not* being present, and assumes a default use case.

Anewbe:
In the future, @MarinaGryphon , please keep any and all PRs as simple and focused as possible.

MarinaGryphon:
Yeah, it would've been better to PR the macros first, and then the hoists. Unfortunately I didn't realise that I'd need to port them until I'd already made the PR.

Anewbe:
Roger, just wanted to give you the heads-up.

PR #4452, opened 2017-12-20
You can now remove card decks from containers.

MarinaGryphon:
You draw cards if the deck is in your hand or on a table, but you should be able to remove them from pockets and bags. (Assuming this compiles.)

### DIFF
--- a/code/modules/games/cards.dm
+++ b/code/modules/games/cards.dm
@@ -65,7 +65,10 @@
 	..()
 
 /obj/item/weapon/deck/attack_hand(mob/user as mob)
-	draw_card()
+	if(istype(src.loc, /obj/item/weapon/storage) || src.loc == user.r_store || src.loc == user.l_store) // so objects can be removed from storage containers or pockets
+		..()
+	else // but if they're not, or are in your hands, you can still draw cards.
+		draw_card()
 
 /obj/item/weapon/deck/verb/draw_card()
 
--- a/code/modules/games/cards.dm
+++ b/code/modules/games/cards.dm
@@ -65,7 +65,8 @@
 	..()
 
 /obj/item/weapon/deck/attack_hand(mob/user as mob)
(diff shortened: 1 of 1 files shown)


code/modules/games/cards.dm
```diff
@@ -65,7 +65,11 @@
 	..()
 
 /obj/item/weapon/deck/attack_hand(mob/user as mob)
-	draw_card()
+	var/mob/living/carbon/human/H = user
+	if(istype(src.loc, /obj/item/weapon/storage) || src.loc == H.r_store || src.loc == H.l_store) // so objects can be removed from storage containers or pockets
```
Neerti: I'm fairly sure ``src.loc == H.r_store || src.loc == H.l_store`` is never going to evaluate to true since if an item is in someone's pockets their loc would equal the mob and not the mob's ``l_store``/``r_store``, which are just ref variables to whatever is in their pockets and so would be the cards themselves.

MarinaGryphon: Oh, right, so I need to remove the .loc. Got it. Or I could just check if the loc is the mob?

Cyantime:
You can already remove decks of cards from containers by click+dragging them onto your sprite, the same way as paper bins.

MarinaGryphon:
That's incredibly unintuitive. I can't think of a situation where you'd want to draw cards directly from a deck of cards or a bag, or where it would make sense. It was common knowledge on the server that you couldn't remove cards from pockets or bags, and that you had to ahelp and get an admin to remove them. This still lets you draw from decks on tables and in your hand, which makes sense- it's hard to open a deck of cards in a bag, much less in your pocket. This would overall improve the usability of cards instead of having them get stuck and need to be removed and respawned by admins, because no one knows the technique for removing them from bags or pockets.
Essentially- I don't think that having another method to pick it up is an issue.

PR #4453, opened 2017-12-22
Dropper and bloodpack improvements

mistyLuminescence:
* Improves droppers - rather than splashing half their reagents onto the target, injecting a quarter into the bloodstream, and retaining a quarter in the dropper, they actually... work now. You can also use less than the entire dropper - e.g. one or two drops.
* Improves IV packs - you can label them with a pen now, and their default blood type is their default label. If you're feeling traitorous, label all the O- blood as AB+, and vice versa.


Anewbe:
I'm going to say no to that Saline, full stop.

mistyLuminescence:
Magic space saltwater is dead and we killed it

PR #4456, opened 2017-12-24
Replaced bare numbers in frame construction with defined constants.

Leshana:
This file is much easier to understand with constants instead of bare numbers.  So I created constants for the states of frame construction, types and classes.   Use constants for dir values too.
Also override examine() to show what circuit board is installed in a frame. Otherwise no semantic changes here, just replacing bare literals with constants.

### DIFF
--- a/code/game/machinery/frame.dm
+++ b/code/game/machinery/frame.dm
@@ -7,7 +7,7 @@
 	construction_frame_floor = list()
 	for(var/R in typesof(/datum/frame/frame_types) - /datum/frame/frame_types)
 		var/datum/frame/frame_types/type = new R
-		if(type.frame_style == "wall")
+		if(type.frame_style == FRAME_STYLE_WALL)
 			construction_frame_wall += type
 		else
 			construction_frame_floor += type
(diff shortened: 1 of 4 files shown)


code/game/machinery/frame.dm
```diff
@@ -16,13 +16,17 @@
...
+// Frame Type Datum - Describes the frame structures that can be created from a frame item.
+//////////////////////////////
 /datum/frame/frame_types
 	var/icon/icon_override		// Icon to set on frame object when building. If null icon is unchanged.
 	var/name					// Name assigned to the frame object.
 	var/frame_size = 5			// Sheets of metal required to build.
 	var/frame_class				// Determines construction method.  "machine", "computer", "alarm", or "display"
 	var/circuit					// Type path of the circuit board that comes built in with this frame. Null to require adding a circuit.
-	var/frame_style = "floor"	// "floor" or "wall"
+	var/frame_style = FRAME_STYLE_WALL	// "floor" or "wall"
```
Cyantime: Was this change from floor to wall intentional?

Leshana: No... it was not.
Although it has some great advantages in saved floor space! Imagine how spacious the labs will feel with wall mounted washing machines, wall mounted autolathes, wall mounted mass drivers!

PR #4457, opened 2017-12-24
VR mobs can toggle opaque/translucent appearance

Atermonera:
[Tested](https://puu.sh/yNgjH/33f4283efc.mp4)
Changing body shape works in both opaque and translucent forms.
Changing to human shape while opaque makes the skin stark white, because it doesn't know how to switch from skin color to skin tone.
Changing to opaque form while in human shape defaults the skin tone, I've a note to look into that.


Atermonera:
Conflicts pls, that should've been fixed the first time

PR #4458, opened 2017-12-25
Adds health analyzers to medkits missing them.

mistyLuminescence:
So, previously every 'basic' medkit - regular, oxy, toxin, burn - had a health analyzer... but the advanced and combat ones didn't. Now they do.

This PR adds a health analyzer to the advanced medkit and combat medkit. Boom.

* "Why? Doctors already have health analyzers." Sure, but the medkits didn't. Consistency.
* "Doesn't this add more health analyzers?" Yep. Not like we have them in almost infinite quantities anyway.
* "How about the combat medkit?" If you buy a combat medkit, you get your money's worth. That means a health analyzer.
* "Power creep!" More like consistency creep.

### DIFF
--- a/code/game/objects/items/weapons/storage/firstaid.dm
+++ b/code/game/objects/items/weapons/storage/firstaid.dm
@@ -102,13 +102,14 @@
 /obj/item/weapon/storage/firstaid/adv/New()
 	..()
 	if (empty) return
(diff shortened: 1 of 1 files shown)


Anewbe:
Simply no. This is not necessary, nor an improvement.

PR #4461, opened 2017-12-28
Fix a last-minute typo

Leshana:
Found by @Cyantime


Leshana:
Re-checked for any other typos and found none.
Workaround: Only affects list in UI, so everything still builds fine once the frame is created, so just pretend its a wall frame.

PR #4471, opened 2017-12-30
Makes alcohol have effects when injected.

Cameron653:
- Makes alcohol have the same effect as when drank, just **three times stronger** (This can be raised to be even stronger if desired.)

- This means that the toxin damage is three times higher, hallucinations three times stronger, the dizzyness, slurring, confusion, drowsyness, and eventual paralysis and sleeping is three times stronger/lasts times longer (varies.)

- Makes it so putting alcohol in medibots is actually somewhat viable as a gag.

- It used to just do 2 toxin damage for every unit of alcohol, which was boring. Worst of all, it didn't get your victim/you drunk at all.

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Dispenser.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Dispenser.dm
@@ -80,10 +80,47 @@
 	if(istype(L))
 		L.adjust_fire_stacks(amount / 15)
 
-/datum/reagent/ethanol/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
+/datum/reagent/ethanol/affect_blood(var/mob/living/carbon/M, var/alien, var/removed) //This used to do just toxin. That's boring. Let's make this FUN.
 	if(issmall(M)) removed *= 2
-	M.adjustToxLoss(removed * 2 * toxicity)
-	return
+	M.nutrition += nutriment_factor * removed
+	var/strength_mod = 3 //Alcohol is 3x stronger when injected into the veins.
+	if(alien == IS_SKRELL)
+		strength_mod *= 5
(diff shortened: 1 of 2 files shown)


Magneld:
I didn't even know the drunk effect on injection was removed til you made this so yes please give it back it was hilarious.

Wickedtemp:
If you can ensure that Ethylredoxrazine still works, and check if it works for alcohol in the stomach as well, I would greatly appreciate it!

Also, yis. Would support this.

Cameron653:
Hmm, good catch. Didn't think about Ethylredoxrazine.

Will check it real quick.

Cameron653:
Fixed. Made it so if ethy and alcohol are in your bloodstream, the ethy gets rid of the alcohol but 3x faster since, well, it's directly in the blood with the alcohol.

code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Dispenser.dm
```diff
@@ -80,10 +80,47 @@
 	if(istype(L))
 		L.adjust_fire_stacks(amount / 15)
 
-/datum/reagent/ethanol/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
+/datum/reagent/ethanol/affect_blood(var/mob/living/carbon/M, var/alien, var/removed) //This used to do just toxin. That's boring. Let's make this FUN.
 	if(issmall(M)) removed *= 2
-	M.adjustToxLoss(removed * 2 * toxicity)
-	return
+	M.nutrition += nutriment_factor * removed
```
Anewbe: Let's scrap this line. I think the intention with this being in the ingest proc is to simulate beer and such.

Cameron653: Done

PR #4478, opened 2018-01-01
Makes cloners actually only take meat for biomass when autoprocessing

Cyantime:


Cyantime:
This is more of a bugfix, it's what was intended in https://github.com/PolarisSS13/Polaris/pull/4137 but the var name was 'meat', letting it use any snacks instead of just meat products.

PR #4486, opened 2018-01-01
S&R loadout spacing

Atermonera:
Now it's in the second column, instead of a third. If we add too many more jobs those limits and splits should probably get shuffled, either to keep two long columns or add a third
[PseudoUI](https://puu.sh/ySemj/f2907fd8ac.png)
[Plain](https://puu.sh/ySens/40c9a59ec7.png)


Atermonera:
Fucking... dicks

PR #4493, opened 2018-01-02
Weatherapp Update

### LINKED ISSUE #4500: Weather app has rounding problems
Sometimes it'll display absurdly precise temps, sometimes it won't

Nice, clean values of precision no more than a tenth of a degree or so

#### What actually happened
It gives precision to probably around 20

#### Steps to reproduce
Wait for the planet controller to start teh weather, then check the weather app on a communicator

#### Additional info:
Putting this here so it's clear this is unintended behaviour. When I was working on it I was totally unable to achieve reliable precision, despite rounding on the Byond side, so I suspect that it's a nanoui problem.

Atermonera:
Wind has been removed because that doesn't actually do anything and the dir2text was broken.
The atmospheric scan replicates the PDA function, but won't display non-O2, non-N2 gas pressures if they aren't at all present.
Planetary time is also displayed.

Because I can't get `round()` to play nice and actually had better results without rounding the temperature values, the weather report will sometimes give an absurd level of precision. This is ICly due to the bugs with the sensor equipment unless/until someone can make `round()` play nicer.
I tried replicating the math the atmosscan does and it still didn't work.

[When you're favored](https://puu.sh/ySxiz/936297d0a9.png)
[When you're not](https://puu.sh/ySxnq/d145a2a0a6.png)
  
Fixes #4500


RemieRichards:
You need to ```round()``` to a precision, ```round(num)``` is just ```floor(num)``` (essentially throwing away all decimal values, leaving only integer precision), this also means that ```round(1.5)``` is 1, not ```2``` as you may expect, since you're just flooring it, if you want to round to the nearest integer, use ```round(num, 1)```

tl;dr ```round()``` <i>generally</i> needs 2 arguments.

Atermonera:
Providing two arguments was exactly as effective as providing none

Atermonera:
The issue was, indeed, on the nanoui side. TIL all the helper functions are in the .js folder. This'll now fix #4500

PR #4501, opened 2018-01-04
Prepare Atmospherics Machinery for SSatoms

Leshana:
We want all machines, air, pipe networks etc to initialize quickly and smoothly both at system startup, but especially during dynamic map loading.   To best support the /tg maploader we want atoms to initialize properly.  Atmospherics machinery (both pipes and components) are among the oldest machines needing to be initialized after creation, and have a proc named initialize() with a different purpose than the way the rest of SS13 uses initialize().  We should follow /tg and Bay's example of renaming this proc.

#### Rename /obj/machinery/atmospherics initialize() to atmos_init()
* These days `initialize()` is used to handle general object initialization that is moved outside of New().  The node connection discovery of atmos machinery needs to happen after all that, and so needs to be in a separate proc.
* Make sure to actually call atmos_init during system startup.

#### Add preloader.load() to /atom/New()
* This was present in the Polaris maploader used prior to porting to TG's maploader.   It is used by TG and Baystation, but not by us.
* The effect of not having it is that any custom var overrides in a .dmm file will be applies *after* invoking New(), instead of before.  This caused bad side effects for some types (i.e. atmos machinery).
* Therefore we put the lines back in.

#### Revert calling target.init_dir() before connecting to neighbors
* This change was added in https://github.com/PolarisSS13/Polaris/pull/3775 to counteract `dir` not being set prior to New() for dynamically loaded maps.  However the true root cause of that problem was /atom/New() not calling _preloader.load().  Undoing the change now that /atom/New() is fixed.
* The addition of the init_dir() proc itself however, is useful, because there ARE other times some atmos machinery will want to re-initialize its dir, specifically whenever it is rotated.
* init_dir() must be called in the constructor of all atmospherics machines capable of connecting to another.   Since it has to happen for ALL machines, lets move that call to /obj/machinery/atmospherics/New()

Also did some to_chat() replacement for the sake of code consistency

### DIFF
--- a/code/ATMOSPHERICS/atmospherics.dm
+++ b/code/ATMOSPHERICS/atmospherics.dm
@@ -10,9 +10,6 @@ Pipelines + Other Objects -> Pipe network
 
 */
 /obj/machinery/atmospherics
-
-	auto_init = 0
-
 	anchored = 1
 	idle_power_usage = 0
 	active_power_usage = 0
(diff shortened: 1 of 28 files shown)


Anewbe:
Has this been tested?

Leshana:
Yes, I loaded and walked around, checked pipes and machines.   I also loaded a submap template that had pipes and machines.
image

PR #4502, opened 2018-01-05
Fix choosing unlimited languages on character setup.

Leshana:
* Checks list length after input() returns so you can't cheese it.
* Sanitizes list to proper length so existing characters over the limit will be fixed.


Anewbe:
Tested?

Leshana:
I tried choosing more languages using the click-click-click-click method.  It didn't work anymore.
I also used varedit to add extra languages, and they were erased on join.  So seems to work to me.

PR #4508, opened 2018-01-07
Replaces NT Relation with Economic Status

PrismaticGynoid:
Because starting money is pretty much all it affects, anyway. Instead of just being used to measure NT's opinion of you, you now choose income categories to fall under. This still affects how much money you start with, with the highest category giving a 1.3x modifier, but the lowest has been lowered to 0.5 times. Now you can start with even less money than before.

Currently doesn't change the "this person might be willing to help you" message for traitors, but it would be easy to do something different with that if requested.

econmods

### DIFF
--- a/code/__defines/mobs.dm
+++ b/code/__defines/mobs.dm
@@ -102,14 +102,14 @@
 #define INV_SUIT_DEF_ICON 'icons/mob/suit.dmi'
 #define MAX_SUPPLIED_LAW_NUMBER 50
 
-// NT's alignment towards the character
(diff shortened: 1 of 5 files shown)


Belsima:
>Currently doesn't change the "this person might be willing to help you" message for traitors

Shouldn't it at least be commented out, or are we assuming economic status makes you more likely to trait? Just asking. Fair either way.

PrismaticGynoid:
I wanted to leave that open for staff to decide on.

PR #4511, opened 2018-01-09
SC Map Tweaks

Woodratt:
- Minor improvements and adjustments to the map
- Fixes some of the submap APCs to start off, so they do not die and show up on the alert console


Anewbe:
Something seems to have gone wrong in the DJ PoI, was that touched, or is it unrelated?

Neerti:
The ``planetuse`` type can't be used since PoIs are meant to be independent of which map gets compiled, however ``planetuse`` isn't needed to stop active edges, if you do what I did at https://github.com/PolarisSS13/Polaris/blob/master/maps/southern_cross/turfs/outdoors.dm#L85
Just make a basic 'outdoor plating' similar to ``/turf/simulated/floor/outdoors``, with the atmospheric characteristics defined inside the ``southern_cross/turfs/outdoors.dm`` file but the basic tile being defined elsewhere so PoIs can use them and it will match what the map expects for 'outside' to be.

PR #4513, opened 2018-01-09
Research Machines can use Plastic and Osmium

Mechoid:
Allows Research machines (Protolathe, Circuit Printer, Mech Fab, and Pros Fab) to store and use plastic and osmium sheets. No recipes currently use these.


MisterLayne:
Will any recipes be updated to include either of these two ingredients? I can forsee a lot of stuff using this.

Mechoid:
Plastic, yes, Osmium less so due to platinum being just less rare than diamond, it's more of an in-between tier from silver/gold to diamond.

Anewbe:
Tested?

Mechoid:
Unless my memory is faulty, yep! Even tested it on live.

PR #4515, opened 2018-01-09
Adds a species flag for preventing infections. Added to Prometheans.

Mechoid:
NO_INFECT prevents limbs and other organs from becoming infected by remaining sterile, similar to IS_PLANT without its other functions.

Added NO_INFECT to Prometheans, to fix regenerated wounds still becoming infected.


Nalarac:
Can the player xenomorphs get infections? They could use this if so.

Mechoid:
Two birds one stone, good idea.

PR #4521, opened 2018-01-12
Make Singulo Stable Again

Leshana:
* Prevents emitter beam effects from being pulled by singulo.  Now watch them hit the field gens instead of veering off course.
* Prevent singulo from consuming the PA particle effects.  It already gets fed by impact, don't try to eat them.
* Improve the Setup Singularity debug verb to do more better setup (including PA)
* Fix the energy dissipation values so that singularity does not just keep on growing even when setup properly.

### DIFF
--- a/code/modules/admin/verbs/debug.dm
+++ b/code/modules/admin/verbs/debug.dm
@@ -512,12 +512,25 @@
 		return
 
 	for(var/obj/machinery/power/emitter/E in world)
-		if(E.anchored)
-			E.active = 1
+		if(istype(get_area(E), /area/space))
+			E.anchored = TRUE
+			E.state = 2
+			E.connect_to_network()
+			E.active = TRUE
 
 	for(var/obj/machinery/field_generator/F in world)
-		if(F.anchored)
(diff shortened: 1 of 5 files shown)


Neerti:
From IRC:
```
<GinjaNinja32> @Neerti can you comment that pls i'm at work on my work acct :D
<GinjaNinja32> ...wait what is that a fucking
<GinjaNinja32> istype(src)
<GinjaNinja32> w a t
<GinjaNinja32> >if(!istype(src)) return
<GinjaNinja32> w a t
<GinjaNinja32> ditch S too, just use src
<GinjaNinja32> L274-276 delete, L278 change S to src
<GinjaNinja32> L279 should just be else
```

Leshana:
GinjaNinja32 commented both on the change I made, and on the existing code below it.
* About the code I changed ( singularity.dm L271 ):  I agree! I was unaware of this purpose of `simulated` var. I'll change that.
* About the existing code:  I don't intend this PR to be a general cleanup of singularity code. It is just a fix to observed bugs.  However since I'm in there, I don't mind cleaning up this one proc a bit.

PR #4523, opened 2018-01-12
"Communicators have a flashlight"

Atermonera:
Not pushing directly to master this time, Ater is in fact a donk.
[Tested](https://puu.sh/z0crB/91ea667ab0.mp4)
**Flashlight functions identically to the one in the PDA**


MisterLayne:
Assuming this is the same brightness as the PDA's flashlight?

PR #4524, opened 2018-01-12
Updates changelog

Atermonera:
Went and figured it out, took a bit because I wasn't actually writing them correctly


Anewbe:
Did you run the program to make the thing or did you just write it in manually? One requires black magic, sacrifices, and understanding of Python; the other is apparently the wrong way to do things.

Atermonera:
With the script. Black magic includes changing includes adding '../..' to the path parameters because it needs to actually get to the html directory from tools/genchangelog

PR #4525, opened 2018-01-13
Fixes #4522 - Make maploader correctly read lists of numbers, paths, etc

### LINKED ISSUE #4522: Door Access and Vending Machines fail on loaded Submap Templates
Some items on submap templates loaded at runtime fail to work properly.
* Doors with `req_access` or `req_one_access` set, it will not actually be usable by anyone.
* Vending machines with custom products will display all nulls

#### What actually happened
Example from the "Quarantined Shuttle" submaps' broken vending machines
image

#### Steps to reproduce
1. Load "Quarantined Shuttle" template
2. Restore power to the area (either the true IC way, or var-edit `requires_power = 0` and proccall `power_change()`)
3. Try and use the vending machines.
#### Additional info:
 - **Server Revision**: bb678df01a3616f4bcea02fe204e23c3cb2d12b3

Leshana:
* For non-assoc lists every value was being read as a string.   For vars like "access_req" this is bad - it doesn't work.  Therefore we add support for lists of numbers.
* While I'm here, might as well add support for non-string keys of associative lists too, since that's required by things like vending machines.

### DIFF
--- a/code/modules/maps/tg/reader.dm
+++ b/code/modules/maps/tg/reader.dm
@@ -239,7 +239,7 @@ var/global/use_preloader = FALSE
 
 			if(variables_start)//if there's any variable
 				full_def = copytext(full_def,variables_start+1,length(full_def))//removing the last '}'
-				fields = readlist(full_def, ";")
+				fields = readlist(full_def, ";", TRUE)
 				if(fields.len)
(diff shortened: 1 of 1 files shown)


Leshana:
vending success

PR #4526, opened 2018-01-13
Knife refactor and minor changes regarding.

Mechoid:
Knives become actual knives, hatchets swapped over to be a subtype of machete, which is a subtype of knife that cleaves. Towercap logs can be cut by any sharp or edged weapon.


Anewbe:
Does this remove functionality from any of the knives that used to be hatchet-based?

Mechoid:
Theoretically, only the chopping of tower-caps would be changed, due to them looking for an actual hatchet instead, though they gain all the functionality of normal knives, including ghetto surgery. Actual logs check for sharp or edge I believe, which they still have.

code/modules/surgery/generic.dm
```diff
@@ -303,7 +303,7 @@
 /datum/surgery_step/generic/amputate
 	allowed_tools = list(
 	/obj/item/weapon/surgical/circular_saw = 100, \
-	/obj/item/weapon/material/hatchet = 75
+	/obj/item/weapon/material/knife/machete/hatchet = 75
```
elgeonmb: allow non-hatchet machetes to be used as csaws?

Anewbe: Nah, hatchet makes sense, IMO, regular machete does not.

PR #4530, opened 2018-01-13
Multiple loadout slots

PrismaticGynoid:
Ported from Bay. Each character can have 3 (number can be overridden by config) loadout slots. This way, you can have different outfits for different situations without needing to have a separate character slot or edit your loadout every time.

Tested here, works as intended. The current loadout is set as slot 1, so you don't need to worry about remaking it. This also ports the to_file and from_file macros to make it work, I'm assuming they could later be used elsewhere as well.

[Here's what it looks like.](https://i.gyazo.com/27a6020e386fd9c3752b0d8578e00a36.mp4)

### DIFF
--- a/code/_macros.dm
+++ b/code/_macros.dm
@@ -59,6 +59,8 @@
 #define to_world_log(message) world.log << message
 // TODO - Baystation has this log to crazy places. For now lets just world.log, but maybe look into it later.
 #define log_world(message) world.log << message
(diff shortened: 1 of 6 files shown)


elgeonmb:
This power is too strong

Zero-Bits:
<3
My code has finally begun to creep into other codebases.

PR #4531, opened 2018-01-14
Ports the Paradise rewrite of TG rewrite of the Beam() proc

Leshana:
Ports @Fox-McCloud port of nice Beam datum.  New proc is more efficient and also supports:
- Multiple simultaneous beams from the same source.
- Doesn't block the thread while the beam is on.

This is useful in itself, nobody is saying this is actually a prelude to lightning bolts or anything.


Fox-McCloud:
>This is useful in itself, nobody is saying this is actually a prelude to lightning bolts or anything.

🤔

PR #4542, opened 2018-01-15
Fix uses of TECH_MAGNETS to TECH_MAGNET

Stevenator1:
Stumbled upon these while working on my req_tech and origin_tech parser.

Used the following regex to determine these were the only cases of the problem:
`TECH_(?!(ENGINEERING|MATERIAL|PHORON|DATA|ARCANE|POWER|MAGNET|ILLEGAL|BLUESPACE|BIO|COMBAT)[^A-Z])`


Stevenator1:
I would love to make a unit test for this.

PR #4546, opened 2018-01-15
Return of material scanner overlay

Arokha:
*This PR is a child of: https://github.com/PolarisSS13/Polaris/pull/4545*

*When the other one is merged, the diff will become something like 5 lines.*

This returns the material scanner goggles hud overlay like all the other goggles have. I'm unsure why it was missing one, I remember it used to be light blue, so this readds that so it's similar to mesons/thermals/nvgs/sciencegoggles in that it has a flat color overlay on the screen while worn and on. This one is probably the least annoying color because of how human eyes interpret light-blue (also it shows up under the UI now like all goggle overlays so that's nice).


mistyLuminescence:
i still maintain we should have a toggle for the overlay (and only the overlay, not the goggle functionality) so we can turn it off and still do our jobs when it's too headache-inducing

Arokha:
This one is honestly not that headache inducing since it's just a sort of natural light blue. The mesons, a little more.

The planes update pushes all the overlays under the hud and everything else, though, which I find helps a bit because you have a sort of visually anchoring effect of all the HUD elements.

Arokha:
Rebased on new thingy so all cleaned up now.

PR #4549, opened 2018-01-15
Fixes runtime in seed_packets.dm

Atermonera:
On the tin,
Tested


Atermonera:
P. sure the seed packet runtime usually pops up before the uplink runtime, but anyways it didn't show up in the log

Leshana:
Please do not sleep() in New.    If absolutely necessary, make sure to spawn() first before sleeping or it will block.
But in this case, I think moving that code to initialize()  will solve your issues.

Leshana:
@Atermonera Thanks! Although I should have been more clear.  The plant controller will always have been created before `initialize()` is called, so you actually don't need the sleeping loop at all when the code is in initialize()

PR #4557, opened 2018-01-16
Surface Map Work

Neerti:
Rearranges POI file location to be more logical.
Wilderness and Mining Z-levels are now divided into two halves, to better regulate the 'difficulty curve' of POI spawns. The plan is that the farther from the outpost you go, the harder it gets. This also makes it so you don't get Black Shuttled when looking for your first Wildy POI.
Adds The Wall, seperating the Wilderness from the Plains.
Adds mountain rock around the edges of the map to give a more IC explaination of why you can't go that way besides 'the end of the world is here, literally'.
Adds banners. Red, Blue, Green, NT, Solgov, and Virgov represent.
Falling from the sky should no longer land you in mining.
Spilled fuel from POIs shouldn't spam the chatlog at server startup.
Adds POI capability to outpost level, however POIs are forbidden from spawning close to the outpost.
Adds a few POIs for the 'plains', IE the outpost z-level.
Makes bridges prettier and boat friendly.

Pictures:
Mountains Mountains2
Plains Plains2
Wildy Wildy2
Banners


code/game/turfs/simulated/outdoors/grass.dm
```diff
@@ -13,21 +12,31 @@ var/list/grass_types = list(
...
+
 /turf/simulated/floor/outdoors/grass/sif
 	name = "growth"
 	icon_state = "grass_sif"
 	edge_blending_priority = 4
 	grass_chance = 0
 	var/tree_chance = 2
 
+	grass_types = list(
+		/obj/structure/flora/sif/eyes
```
Anewbe: Does this actually do anything with grass_chance = 0 ?

Neerti: No, was wanting to add more grass types first before activating it.

PR #4558, opened 2018-01-16
Makes janicarts hold more water

Cameron653:
So you don't have to run back to grab some more water every 20 seconds while cleaning up dirt.

### DIFF
--- a/code/game/objects/structures/janicart.dm
+++ b/code/game/objects/structures/janicart.dm
@@ -17,7 +17,7 @@
 
 
 /obj/structure/janitorialcart/New()
-	create_reagents(100)
+	create_reagents(500)
 
 
 /obj/structure/janitorialcart/examine(mob/user)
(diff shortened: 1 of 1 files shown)


Anewbe:
Eh, maybe 250, or 300? 500 is half a water tank, seems a bit excessive.

Cameron653:
Changed to 300

Belsima:
What of the smaller one? Should it be increased too?

Cameron653:
Smaller one? I don't play janitor, but I got a request to increase the janicart. If there's a smaller janicart then please link me to it and I'll also change it.

Belsima:
https://github.com/Cameron653/Polaris/blob/7783afdd0672ec5679d2fc7e75b68f044f4efb9d/code/game/objects/structures/mop_bucket.dm

The mopbucket. Utterly useless as of now.

PR #4560, opened 2018-01-17
Locket desc was too long

Atermonera:
Like, really, that's fine for an extended desc that doesn't show up on the loadout.


Belsima:
Maybe put it as extended desc? Probably not important either way but I like the fluff.

PR #4562, opened 2018-01-17
zShadow qdel fix

Arokha:
The default QDEL hint for mob is DELETE_ME_RIGHT_NOW_PLS because mobs are a scary and mysterious thing and we don't want them sticking around, just in case it somehow destroys the universe.

But, zshadows are not real mobs, we can get qdel-queued instead. Should result in a pretty large performance improvement where multiz shadows are often created and deleted. Also removes ..() in zshadow New, because it's unnecessary and all the functions of zshadow work without it (also it adds the zshadow to lists and stuff).

### DIFF
--- a/code/modules/multiz/zshadow.dm
+++ b/code/modules/multiz/zshadow.dm
@@ -20,14 +20,16 @@
 	if(!istype(L))
 		qdel(src)
 		return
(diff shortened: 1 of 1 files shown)


Leshana:
Tested this.  Good idea and should help save wasted hard-dels just from walking around in stairwells.

PR #4570, opened 2018-01-18
Buffs the mop bucket by the same amount as the janicart.

Belsima:


Neerti:
300 seems a bit much when the janicart has to be dragged where as buckets don't.

Atermonera:
You can pick up the yellow mop buckets?

Neerti:
I'm a dummy and thought this was being applied to the blue buckets.

Belsima:
Yeah, it's still strictly inferior to the janicart, even. I don't think it has any advantage.

PR #4571, opened 2018-01-18
Reinforced Floors are now Explosion Proof

Neerti:
This should make meteors less horrific on the new map.


Andromeda-K22:
*METEORS*.

PR #4573, opened 2018-01-18
Removes HUD items from all borgs, makes records 'just work'

Arokha:
Previously all borgs were given med/sec huds to hold in their grubby ~paws~ ~hands~ claspers so they could read medical records as a workaround because nobody knew how else to do it I guess.

Now you just need to have the appropriate sensor mode. Have security HUD mode on? Shows security records. Medical hud mode? Shows medical records. No mode? No records.

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot_modules/event.dm
+++ b/code/modules/mob/living/silicon/robot/robot_modules/event.dm
@@ -16,7 +16,6 @@
 	src.modules += new /obj/item/borg/combat/shield(src)
 
 	// Med
-	src.modules += new /obj/item/borg/sight/hud/med(src)
 	src.modules += new /obj/item/device/healthanalyzer(src)
 	src.modules += new /obj/item/weapon/reagent_containers/borghypo/lost(src)
 
--- a/code/modules/mob/living/silicon/robot/robot_modules/station.dm
+++ b/code/modules/mob/living/silicon/robot/robot_modules/station.dm
@@ -211,7 +211,6 @@ var/global/list/robot_modules = list(
 
 /obj/item/weapon/robot_module/robot/medical/surgeon/New()
 	..()
-	src.modules += new /obj/item/borg/sight/hud/med(src)
 	src.modules += new /obj/item/device/healthanalyzer(src)
 	src.modules += new /obj/item/weapon/reagent_containers/borghypo/surgeon(src)
 	src.modules += new /obj/item/weapon/surgical/scalpel/cyborg(src)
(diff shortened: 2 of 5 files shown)


Mechoid:
Sensor augmentations aren't the full HuD. This is an issue, if that's what this makes it rely on. Security aug just shows 'Am I wanted or Not', Sec HuD shows the prisoner implant/tracker, job, and if they're wanted. Medical aug shows health if they're on vitals or above. Medical HuD shows if they're ill or injured or robotic regardless.

Arokha:
As far as I'm aware, the sensor augmentations show exactly what they did before, except now they enable records access while on that augmentation, rather than needing to be holding a med/sec hud in the module slot for those to show up.

code/modules/mob/living/silicon/robot/robot_modules/event.dm
```diff
@@ -16,7 +16,6 @@
 	src.modules += new /obj/item/borg/combat/shield(src)
 
 	// Med
-	src.modules += new /obj/item/borg/sight/hud/med(src)
```
Mechoid: No more full Medhud here, augmentation is directly inferior.

code/modules/mob/living/silicon/robot/robot_modules/station.dm
```diff
@@ -211,7 +211,6 @@ var/global/list/robot_modules = list(
 
 /obj/item/weapon/robot_module/robot/medical/surgeon/New()
 	..()
-	src.modules += new /obj/item/borg/sight/hud/med(src)
```
Mechoid: Same as above.

code/modules/mob/living/silicon/robot/robot_modules/station.dm
```diff
@@ -273,7 +272,6 @@ var/global/list/robot_modules = list(
 
 /obj/item/weapon/robot_module/robot/medical/crisis/New()
 	..()
-	src.modules += new /obj/item/borg/sight/hud/med(src)
```
Mechoid: Same as above.

code/modules/mob/living/silicon/robot/robot_modules/station.dm
```diff
@@ -501,7 +499,6 @@ var/global/list/robot_modules = list(
 
 /obj/item/weapon/robot_module/robot/security/general/New()
 	..()
-	src.modules += new /obj/item/borg/sight/hud/sec(src)
```
Mechoid: No more full Sechud here, augmentation is directly inferior.

Arokha:
I compiled and ran Polaris and the sensor augmentation looks exactly like the normal med and sec huds, I'm unsure what you're talking about. Sec huds show the job ID and wanted status, med huds show a health bar and blue cross.

Nalarac:
The medical HUD sensor augmentation works off suit sensors. Sensors off = no reading, binary = blue cross, vitals and tracking = health bar

Arokha:
Doesn't work that way in current master. They just have the normal HUDs in current master. I think it was something changed in a previous PR, probably by me, since Anewbe said "[1:07 AM] Anewbe: Go for it."

So if you're worried that this is removing functionality from borgs: it's not. I can provide screenshots if you want, they just have normal HUDs when using sensor augmentations.

Arokha:
@Atermonera Changelog'd.

Mechoid:
Ahhh, alright. Sorry for the bother then!

PR #4575, opened 2018-01-18
Meteors No Longer Hurt Floors

Neerti:


Neerti:
Not sure this is a bug.

Belsima:
Does that mean they only blow up walls?

Atermonera:
This is affecting the meteor itself, not the explosion it generates. So if a meteor hits deck 3 on SC, it won't leave a trail of broken/destroyed floors from intercept to explosion, it'll just be an explosion.

Anewbe:
Should still punch through walls and stuff, but there won't be a million extra floor holes after a meteor hit.

Leshana:
I think it will make meteors cooler, I support this change.

code/game/gamemodes/meteor/meteors.dm
```diff
@@ -153,16 +153,17 @@
 /obj/effect/meteor/proc/ram_turf(var/turf/T)
 	//first bust whatever is in the turf
 	for(var/atom/A in T)
-		if(A != src)
-			A.ex_act(hitpwr)
+		if(A == src) // Don't hit ourselves.
+			continue
+		if(isturf(A)) // Don't hit floors. We'll deal with walls later.
```
Leshana: I'm fairly sure that turfs do not contain themselves or any other turfs, so this if condition will never be true.

Belsima:
Ah, good stuff then.

PR #4583, opened 2018-01-18
Allows easier prefilling of beakers, bottles, etc

Cyantime:
Rather than sticking it into New() of every subtype, there's a variable for which reagents and how many of each to fill the container with.

### DIFF
--- a/code/modules/reagents/reagent_containers/glass.dm
+++ b/code/modules/reagents/reagent_containers/glass.dm
@@ -19,6 +19,8 @@
 
 	var/label_text = ""
 
+	var/list/prefill = list()	//Reagents to fill the container with on New()
+
 	var/list/can_be_placed_into = list(
 		/obj/machinery/chem_master/,
 		/obj/machinery/chemical_dispenser,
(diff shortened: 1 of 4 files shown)


code/modules/reagents/reagent_containers/glass.dm
```diff
@@ -19,6 +19,8 @@
 
 	var/label_text = ""
 
+	var/list/prefill = list()	//Reagents to fill the container with on New()
```
Leshana: You can optimize memory usage by initializing this to `null`

code/modules/reagents/reagent_containers/glass.dm
```diff
@@ -47,6 +49,10 @@
 
 /obj/item/weapon/reagent_containers/glass/New()
 	..()
+	if(prefill.len)
+		for(var/R in prefill)
+			reagents.add_reagent(R,prefill[R])
+		update_icon()
```
Leshana: And then ...
```dm
	if(LAZYLEN(prefill))
		for(var/R in prefill)
			reagents.add_reagent(R,prefill[R])
		prefill = null
		update_icon()
```

Leshana:
Look at how much code is cleaner now that this change is in place, and without any real performance hit either.

PR #4586, opened 2018-01-19
Fixes crawling when standing. No more accidentally crawling when buckling.

Mechoid:
Can no longer crawl when in a rolling chair to move, but the normal pull-click still works.
I only went partially insane trying to make rolling chairs work like wheelchairs before thinking that's better for its own PR if it is done.


PrismaticGynoid:
Aw, but that would remove my favorite accidental features of 'crawling', being able to slowly walk without using the keyboard, the visually diagonal movement, and for borgs to be able to move things around without dragging or bumping them. Also being able to move someone on snow turf instead of collecting the snow.

Mechoid:
I'm not entirely certain if it was intentional, though after prodding from Anewbe I'll probably bring it up at the staff meeting if I don't undergo amnesia.

Anewbe:
Any last thoughts, Mechoid?

Mechoid:
Humans can't crawl unless laying down, though borgs can maintain their click-drag.

PR #4589, opened 2018-01-19
No longer emphasizes basic words.

MisterLayne:
Names that contain words such as "the, of, by, there" no longer get emphasized when just those words are spoken.
image

### DIFF
--- a/code/modules/mob/hear_say.dm
+++ b/code/modules/mob/hear_say.dm
@@ -87,7 +87,9 @@
 
 // Checks if the mob's own name is included inside message.  Handles both first and last names.
 /mob/proc/check_mentioned(var/message)
+	var/not_included = list("a", "the", "of", "in", "for", "through", "throughout", "therefore", "here", "there", "then", "now", "I", "you", "they", "he", "she", "by")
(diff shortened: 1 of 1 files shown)


code/modules/mob/hear_say.dm
```diff
@@ -87,7 +87,12 @@
 
 // Checks if the mob's own name is included inside message.  Handles both first and last names.
 /mob/proc/check_mentioned(var/message)
+	var/not_included = list("a", "the", "of", "in", "for", "through", "throughout", "therefore", "here", "there", "then", "now", "I", "you", "they", "he", "she", "by")
 	var/list/valid_names = splittext(real_name, " ") // Should output list("John", "Doe") as an example.
+	for(var/check_word in valid_names)
+		for(var/bad_word in not_included)
```
Atermonera: You can use `if(check_word in not_included) valid_names.Remove(check_word)` instead of the for loop and the if loop

Neerti: You could also forego the loop entirely and do ``valid_names -= not_included``, which should make ``valid_names`` not have any common words inside ``not_included``.

PR #4591, opened 2018-01-20
Band-aids three runtimes

Atermonera:
See commit messages. Server restarted before I could really see what caused these, but this should at least stop these from happening.


Atermonera:
Aw ffs I thought I specifically didn't include polaris.dme in the commits

PR #4597, opened 2018-01-20
Fixes necrotic organ repair

Arokha:
I assume in https://github.com/PolarisSS13/Polaris/commit/b1c16857cb that you didn't notice this surgery because it lacked a large comment header like the other surgeries in the file had, and when you were replacing the fix IB section you replaced this section accidentally because they were adjacent.

Without it, you can't fix infections with surgery, which is a little problematic.

### DIFF
--- a/code/modules/surgery/other.dm
+++ b/code/modules/surgery/other.dm
@@ -59,48 +59,48 @@
 // Necrosis Surgery Step 1
 ///////////////////////////////////////////////////////////////
 /datum/surgery_step/fix_dead_tissue        //Debridement
-    priority = 2
-    allowed_tools = list(
-        /obj/item/weapon/surgical/scalpel = 100,        \
-        /obj/item/weapon/material/knife = 75,    \
-        /obj/item/weapon/material/shard = 50,         \
-    )
+	priority = 2
+	allowed_tools = list(
+		/obj/item/weapon/surgical/scalpel = 100,        \
+		/obj/item/weapon/material/knife = 75,    \
+		/obj/item/weapon/material/shard = 50,         \
+	)
 
-    can_infect = 1
-    blood_level = 1
+	can_infect = 1
+	blood_level = 1
 
-    min_duration = 110
-    max_duration = 160
+	min_duration = 110
+	max_duration = 160
 
-    can_use(mob/living/user, mob/living/carbon/human/target, target_zone, obj/item/tool)
-        if(!hasorgans(target))
-            return 0
+/datum/surgery_step/fix_dead_tissue/can_use(mob/living/user, mob/living/carbon/human/target, target_zone, obj/item/tool)
+	if(!hasorgans(target))
+		return 0
 
-        if (target_zone == O_MOUTH || target_zone == O_EYES)
-            return 0
+	if (target_zone == O_MOUTH || target_zone == O_EYES)
+		return 0
 
-        var/obj/item/organ/external/affected = target.get_organ(target_zone)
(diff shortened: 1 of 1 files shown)


code/modules/surgery/other.dm
```diff
@@ -56,9 +56,55 @@
...
+        /obj/item/weapon/material/shard = 50,         \
+    )
+
+    can_infect = 1
+    blood_level = 1
+
+    min_duration = 110
+    max_duration = 160
+
+    can_use(mob/living/user, mob/living/carbon/human/target, target_zone, obj/item/tool)
```
Cyantime: Relative pathing

code/modules/surgery/other.dm
```diff
@@ -56,9 +56,55 @@
...
+            return 0
+
+        if (target_zone == O_MOUTH || target_zone == O_EYES)
+            return 0
+
+        var/obj/item/organ/external/affected = target.get_organ(target_zone)
+
+        return affected && affected.open >= 2 && (affected.status & ORGAN_DEAD)
+
+    begin_step(mob/user, mob/living/carbon/human/target, target_zone, obj/item/tool)
```
Cyantime: Relative pathing

code/modules/surgery/other.dm
```diff
@@ -56,9 +56,55 @@
...
+        var/obj/item/organ/external/affected = target.get_organ(target_zone)
+        user.visible_message("[user] starts cutting away necrotic tissue in [target]'s [affected.name] with \the [tool]." , \
+        "You start cutting away necrotic tissue in [target]'s [affected.name] with \the [tool].")
+        target.custom_pain("The pain in [affected.name] is unbearable!", 100)
+        ..()
+
+    end_step(mob/living/user, mob/living/carbon/human/target, target_zone, obj/item/tool)
```
Cyantime: Relative pathing

code/modules/surgery/other.dm
```diff
@@ -56,9 +56,55 @@
...
+        var/obj/item/organ/external/affected = target.get_organ(target_zone)
+        user.visible_message("<font color='blue'>[user] has cut away necrotic tissue in [target]'s [affected.name] with \the [tool].</font>", \
+            "<font color='blue'>You have cut away necrotic tissue in [target]'s [affected.name] with \the [tool].</font>")
+        affected.open = 3
+
+    fail_step(mob/living/user, mob/living/carbon/human/target, target_zone, obj/item/tool)
```
Cyantime: Relative pathing

Arokha:
Fixed

Anewbe:
That step was removed because, as far as myself and a few others could tell, it was neither working nor necessary to actually treat a necrotic organ, surgically. Additionally, spaceacillin can treat germ_level in any amount, so the surgery, bugged as it was, was deemed unnecessary.
That said, if you've got the step working again, don't see a reason why not to re-add it.

Cyantime:
The step was required to get a non-head/chest limb to open=3, which can't be done in any other way. The peridaxon dropper surgery needs the organ to be at open3 to start.

Anewbe:
The issue was that starting said surgery with an open chest only gave the menu for organ removal, does this PR correct that?

Cyantime:
It isn't needed to open the chest up fully, it's performed before the bones are opened.

Arokha:
Spaceacillin also cannot 'fix' a dead organ, it cannot make it 'un-dead' as it were. This surgery is for AFTER you have corrected their infection, to remove the DEAD flag from their organ otherwise it will continue to cause toxins to them.

PR #4601, opened 2018-01-21
properly names the communicator watch selector

BillyBangles:
whoops. i goofed.


Anewbe:
Tested?

BillyBangles:
Tested, works. The problem was that the "communicator selection" was accidentally named "communicator watch"

PR #4613, opened 2018-01-22
Adds Missing Ceiling Overlay + Fixes Space on Roof

Neerti:
On Multi-Z compatible maps, if a tile is missing a ceiling (IE there is an open space above it), and someone is wearing meson goggles, it will display an overlay to show this. This makes it easier to know where the hole in the ceiling is without having to go to the ceiling. Overlay sprite by Mechoid.
Also fixes space tiles appearing when meteors/explosions/other fun things occur on the Southern Cross map instead of open space.

Picture available in comments.


Arokha:
What about areas that over openspace above them on purpose? Stairwells, atriums, etc. Won't these just be flashing constantly?

code/game/turfs/simulated/floor_icon.dm
```diff
@@ -84,6 +84,11 @@ var/list/flooring_cache = list()
 				continue
 			F.update_icon()
 
+	// Show 'ceilingless' overlay.
+	var/turf/above = GetAbove(src)
+	if(above && isopenspace(above) && !istype(src, /turf/simulated/floor/outdoors)) // This won't apply to outdoor turfs since its assumed they don't have a ceiling anyways.
+		overlays |= image(icon = 'icons/turf/open_space.dmi', icon_state = "no_ceiling", layer = TURF_LAYER + 0.02)
```
Arokha: I'd suggest caching this image so you don't need to make a new one for literally every turf. Just have it stored in a global somewhere. You could also set a `plane` on this image to make it so you can toggle the visibility with a verb or just wearing mesons or whatever.

In the case of planes, I can help you set that up.

Neerti: Will look into seeing if I can put it on a plane for meson vision or something with Aronai's help.

Neerti:
Now it only shows if wearing mesons.
!(Mesons)[https://puu.sh/z7mu6/582dcb04f3.gif]
The overlay taking a moment to show is due to Life() and oldcode that I'd rather not play with at the moment but the meson plane can be useful in the future for things like blueprint plans or other things engineering might like.

Leshana:
Oooh, only with mesons! That may well solve this problem! I have been trying to think of a way to show this information without it being confusing or annoying; this could be it.

code/game/turfs/simulated/floor_icon.dm
```diff
@@ -84,6 +93,11 @@ var/list/flooring_cache = list()
 				continue
 			F.update_icon()
 
+	// Show 'ceilingless' overlay.
+	var/turf/above = GetAbove(src)
+	if(above && isopenspace(above) && !istype(src, /turf/simulated/floor/outdoors)) // This won't apply to outdoor turfs since its assumed they don't have a ceiling anyways.
```
Leshana: `isopenspace(null)` returns false, so does not need to be checked twice.

code/modules/multiz/turf.dm
```diff
@@ -52,6 +52,7 @@
 	below = GetBelow(src)
 	turf_changed_event.register(below, src, /turf/simulated/open/update_icon)
 	levelupdate()
+	below.update_icon() // So the 'ceiling-less' overlay gets added.
```
Leshana: Perhaps instead of adding this  line, you could change line 58 below to be `OS_controller.add_turf(below, TRUE)`  that way the update_icons will be managed by the openspace controller.  This could be especially useful in the case of a 3+ story high openspace area.

Nalarac:
I feel like limiting it to mesons restricts a large majority of the crew from obtaining this information that can be rather crucial at times.

Anewbe:
I think if you're in a space with a breach and you're not engineering, you've got other priorities than figuring out where the hole is... that being "Getting back into somewhere with air." ... so I don't know, I think meson only works...

Atermonera:
The only cases I can think of where someone other than eng needs to see where holes might be, they also need to see what's around that hole (baddies tossing grenades down at sec from a breached top deck). So at least until we get that functionality, if ever, this should suffice

Screemonster:
This _looks_ like it should fix the "deconstructing grass with an RCD makes space on a planet's surface" bug. Will raise an issue if it doesn't.

https://github.com/VOREStation/VOREStation/issues/2565 is the issue at our end.

PR #4617, opened 2018-01-23
Makes Wet Floors More Obvious

Neerti:
Now the 'wet overlay' for floors is visible until it dries instead of only showing itself every so often, so people paying attention can avoid it.
https://puu.sh/z7A1j/b4f6b47b2f.mp4


Belsima:
image

PR #4618, opened 2018-01-23
Makes berry milkshakes makeable

PrismaticGynoid:
Critical bugfix: uncapitalizes a letter which was preventing berry milkshakes from being made.


Anewbe:
**This is why we should move from strings to types...**

Mewchild:
I’m sorry for the typo. A couple small ones seemed to have slipped through on the drinks. Thank you for the fix!

PR #4622, opened 2018-01-23
Adds Material Girders

Mechoid:
Adds material girders.
-Girder decon time is now partially dependant on material. Stronger girders take slightly longer. Steel girders are, actually, one tick faster to decon than before due to this.
-Wall girders are now under the integrity-based construction listing. Material-Name Wall Girder.
-Explosive-resistant girders in walls have a chance to survive as an unanchored girder, if their wall is destroyed.
-Radioactive girders affect the completed wall's radioactivity.
-Stronger materials make Stronger girders, with Steel at 201 health, and Durasteel at around 1000.


elgeonmb:
Does this mean I can finally gun my way through a wall

Atermonera:
Can you make wood girders, are they airtight, and do they burn?

Anewbe:
If the girders are tougher than the walls of the same material, that needs to be adjust down hard.

Mechoid:
You already could gun your way through a wall, physical rounds just to less damage.

Mechoid:
Well, I don't know why steel walls are so wimpy and that their girders are already stronger than the wall itself.

Mechoid:
Oh my lord I hate Git.

Mechoid:
Wood girders and the like do not burn aside from their temperature resistance threshold as per normal material objects.

code/game/objects/structures/girders.dm
```diff
@@ -130,14 +214,16 @@
 
 /obj/structure/girder/proc/take_damage(var/damage)
 	health -= damage
-
 	if(health <= 0)
 		dismantle()
+	else
```
Atermonera: +=

code/game/objects/structures/girders.dm
```diff
@@ -174,36 +260,36 @@
...
+	to_chat(user, "<span class='notice'>You added reinforcement!</span>")
 
 	reinf_material = M
 	reinforce_girder()
 	return 1
 
 /obj/structure/girder/proc/reinforce_girder()
 	cover = reinf_material.hardness
-	health = 500
+	health = health + round(reinf_material.integrity/2)
```
Atermonera: +=

PR #4626, opened 2018-01-24
Make Darksight Great Again

Arokha:
Makes your eyes adjust in the dark. The amount your eyes will adjust is based on a function of how bright the turf is and how far your eyes can adjust, which itself is based on how good your darksight is. Look at the bottom of living/life.dm at handle_darksight.dm for the math used.

(Glowing eyes is a separate PR)

Anewbe points out that this does have immershuns because if you're a Tajaran in maint and some ponce with a flashlight stands next to you, it'll start unadjusting your eyes, and you'll lose darksight, which is... how it'd work in the realisms.


Nalarac:
Since I can not into codes, could you explain how this would work with borgs and FBPs (or prosthetic eyes in general).

Arokha:
Would work exactly the same as the picture, eyes adjust. However it can be overridden on any mob, because of the fact that it's just a `handle_darksight()` proc. If someone wants to make it so FBPs adjust faster or have some other mechanic for silicons, then that's easy.

I just didn't want to do that in the PR because I don't know how balancey things go on your server and figured the across-the-board uselessness of darksight could be fixed with an across-the-board 'your eyes adjust' thing.

Anewbe:
I have poked the merge conflict, as for to get the Darksight change in sooner.

PR #4627, opened 2018-01-24
Species var that enables eyes to glow

Arokha:
By glow I mean be drawn above the darkness plane, not actually generate and illuminate turfs. Just set the var to 1. Covered up by anything that also covers up your facial hair, so opaque helmets and such. If you wear transparent facemasks or anything, your eyes will be on top of them, piercingly glowy.

Adjusting eyes are from https://github.com/PolarisSS13/Polaris/pull/4626


MisterLayne:
Will this be enabled by default for Tajaran?

Arokha:
I didn't enable it for anyone by default. To be honest, I've never played on Polaris. Just whenever I code a new feature I put it here too since we're a downstream. Y'all are welcome to enable it on whichever species you want, though!

Arokha:
@Anewbe @Neerti Did you have any requests for species to have it on? I can just throw the var in if you want, free of charge

Anewbe:
I think just leave it be, for now.

PR #4628, opened 2018-01-24
Mob var for just glowing in general

Arokha:
Allows mobs to glow, and a variable controls that, and their glow color. Could be fun for lava monsters or whatever.

Can add more vars later for how bright it is if you want.


Neerti:
I think there's code to already handle mob glowing in ``handle_light()`` for things like being on fire or borg lights.

Arokha:
Yup! This is in handle_light(). It just adds two vars for mobs to be arbitrary light colors, if you want mobs to just always be glowing, or use some verb to toggle it on/off, etc.

Neerti:
I see, yay.

PR #4631, opened 2018-01-25
Makes the large size modifiers less extreme

Anewbe:
The traits that increase mob size are now down to 5% and 10% larger, from 10% and 20%.

Should make them less blurry.


PrismaticGynoid:
Does 5% have the needlenose effect that 10% does?

Anewbe:
Don't believe so.

PR #4634, opened 2018-01-25
Makes R-UST machines deconstructable, some more fixes

Cyantime:
Makes the R-UST core, fuel injector, fuel compressor, and gyrotron deconstructable.
Makes the R-UST core actually connect to power cables when secured.
Slightly improves the collision checking from #4616


Andromeda-K22:
So THAT'S why it wasn't working.

May want to look into why the computers don't link properly when constructed.

Cyantime:
All of the computers don't start tagged when constructed, you need to multitool them and the machines to assign tags.

Andromeda-K22:
Negative. 
Core control computer does not link properly, and the fuel injection computer does not even present a UI.

Cyantime:
r-ust
I don't know what to tell you, everything works perfectly here.

Mechoid:
I assume this has been tested, from that screenshot? (The machines, currently here, do not actually seem to link / let the console update properly. This means you can't build them in-round, for some reason. I know because I've tried.)

Cyantime:
I built all those machines in current code and they worked fine, I haven't experienced any issues linking or using the computers

PR #4641, opened 2018-01-26
Loadout Transfer stuff thingy.

Woodratt:
- Shifts head, uniform, and suit load out lists to be map dependent. Allows for certain outfits to be available on one map but not another. (Example, pilot uniforms on the SC but not the NS)

- Weapon lockers on the planetside can now be accessed by security.


polaris.dme
```diff
@@ -2392,7 +2389,7 @@
 #include "code\ZAS\Zone.dm"
 #include "interface\interface.dm"
 #include "interface\skin.dmf"
-#include "maps\northern_star\northern_star.dm"
```
Arokha: Intentional to do that in this PR?

Arokha:
Curious as to why you're removing the head/suit/uniform defaults entirely instead of just adding additional map-specific types to them? Seems problematic if every map has to copy-paste these files if they want any items at all.

Woodratt:
Honestly cause I was following what bay was doing, and wasn't certain if having it a 'split' file would work, Suppose I could test that now.

PR #4645, opened 2018-01-27
POI Tweaks

Neerti:
Adds a new bonfire POI, orange juice included.
Makes the two broken turrets in the first room of the UFO not be salvageable. If you want an ayyy gun, you gotta kill a working turret for it.
Tree stumps are no longer dense and can be walked over.
Removed more Syndicate, renamed to mercenary.
Malf drones are now considered synthetic for various purposes such as ionic rapier bonus damage.
Adds ability to have the map generation algorithm skip a POI based on a probability. Currently only the UFO has this set, so it won't show up every round.
Adds new POI specific SMES, comes pre-charged.
Adds full tiled windows to the farm's greenhouse due to farmbot getting confused with glass planes.
Makes several dirt spots indoors where intended.
Changes Rockybase, drones shouldn't wander away from their home now. Enhanced loot somewhat. Made the bots be the same faction as the drones so they won't get killed.


Mechoid:
I assume what can be tested has been?

PR #4646, opened 2018-01-27
Fixes incorrect gib animations.

### LINKED ISSUE #2449: gib() does not respect icons

When a mob is gibbed, it doesn't display the correct gibbing animation. The whole proc is rather old and in need of a little love as it seems to point at mob.dmi with all the other really horrible really old icons.

For simplemobs, when gibbed, to not play an animation if there isn't one defined (as the icon defaults to NULL) or play the appropriate animation if it is. Also for other races when gibbed to show their appropriate icons.
#### What actually happened

Simplemobs without a defined animation very briefly turn into monkeys for their gib animation. Those with an animation defined simply don't show anything, even with the icon defined. Tajara and other nonhuman races suffer from sudden explosive humanism when gibbed.
#### Steps to reproduce

Spawn a mob.
Gib it from the VV panel.

MarinaGryphon:
I *believe* that this should be working properly. It has also been tested, everything checks out (though I couldn't tell if the animations were showing up properly, since I'm not sure what they were intended to look like).
Fixes #2449

### DIFF
--- a/polaris.dme
+++ b/polaris.dme
@@ -2392,7 +2392,7 @@
 #include "code\ZAS\Zone.dm"
 #include "interface\interface.dm"
 #include "interface\skin.dmf"
-#include "maps\southern_cross\southern_cross.dm"
+#include "maps\example\example.dm"
 #include "maps\submaps\_readme.dm"
 #include "maps\submaps\space_submaps\space.dm"
 #include "maps\submaps\surface_submaps\mountains\mountains.dm"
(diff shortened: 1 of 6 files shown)


polaris.dme
```diff
@@ -2392,7 +2392,7 @@
 #include "code\ZAS\Zone.dm"
 #include "interface\interface.dm"
 #include "interface\skin.dmf"
-#include "maps\southern_cross\southern_cross.dm"
```
Arokha: Errrr.

Belsima: someone forgot to uncheck the box

MarinaGryphon: hah oops

PR #4661, opened 2018-01-28
Convert ZAS into StonedMC Subsystems

Leshana:
### Convert the ZAS/Controller into SSair subsystem
* Move the ZAS-related defines into the defines folder so they're available earlier.
* Creates the SSair subsystem which replaces the Setup, Start(), and Tick() procs of air_master.
* For now we leave the old `air_master` global variable, but re-define it to be the same type as `SSair` and set it to be the same instance as SSair at init time.   This is for clarity in code review and testing.  Once verified a subsequent PR will rename air_master to SSair globally.
* Remove the obsolete Setup and Tick procs.
* Adjust admin and debug verbs that dealt with some ZAS internals to work with the new implementation.
* Adds the cool `report_progress()` proc which does both admin_notice and world.log at the same time!

### Convert ZAS "Airflow" into a subsystem plus fixes
* Port of the "Airflow" portions of Yonaguni/EuropaStation#618
* The "airflow" part of ZAS used to be handled by a sleep'd loop.   This has the potential to bunch up and lag.  Switching to a StonedMC managed subsystem improves it.
* Fixed to ensure that zshadow mobs cannot be blown around by the wind no matter how fierce.
* Added a message to mobs informing them when their boots save them from being wind-thwapped.
* Check w_class on non-item objects if they have it defined (might as well since var/w_class is on /obj)
* Tiny optimization of c_airblock


Leshana:
During testing air still flows and stuff still gets blown around.

PR #4663, opened 2018-01-28
Allow admins to see say/emote near their own body while aghosted

Arokha:
image

yw


elgeonmb:
thank you

*brushes away single tear*

Atermonera:
How does this look if you don't use chat tags?

Arokha:
I have never turned off chat tags. It would probably say BODY: though. I see the create_chat_tag thing accepts a string and that string for OOC is `OOC:` and I put `BODY:` for this so I assume it would say `BODY:`

Atermonera:
I don't personally use chat tags, just a matter of preference, so it's understandably a concern when I'm uncertain if particular formatting will look fucky with something developed for chat tags. Appreciate the test

PR #4664, opened 2018-01-28
Synthetics show up on the manifest

Atermonera:
Turns out they weren't showing up because they don't have records. They still don't have records, but now they're manually added to the manifest.

Extra commits because apparently they didn't get stuck into master or something, git is weird sometimes but I just figured this was a growth off of that branch so eh, w/e. They don't actually accomplish anything, just smth to squish squash


Atermonera:
Tested

Atermonera:
And this is why ater goes to bed at a reasonable time instead of thinking, "Hey, let's stay up into the wee morning doing coding. What could possibly go wrong?"

Belsima:
cool and good

PR #4665, opened 2018-01-28
ID computer can now assign planetside jobs, command secretary, IAA

Atermonera:
Also cleans up that template with real indentation and adds some closing {{/if}} tags.
Captain and Custom need to stay separate within the assignment menu, because I don't think there's a particularly clean way to do the custom thing.
Centcom jobs lose a touch of special formatting that I didn't even know they have because who even uses ID computers with centcom access

Edit: IAA and command secretary also now available. Command secretaries have keycard auth access, per instruction from superiors.

### DIFF
--- a/html/changelogs/Atermonera - id_comp.yml
+++ b/html/changelogs/Atermonera - id_comp.yml
@@ -1,0 +1,6 @@
+author: Atermonera
+
+delete-after: True
+changes:
+  - tweak: "ID computer can set command secretary and IAA access"
+  - tweak: "Command secretary has keycard auth. access"
--- a/code/game/jobs/job/captain.dm
+++ b/code/game/jobs/job/captain.dm
@@ -83,7 +83,7 @@ var/datum/announcement/minor/captain_announcement = new(do_newscast = 1)
 	minimal_player_age = 5
 	economic_modifier = 7
 
-	access = list(access_heads)
+	access = list(access_heads, access_keycard_auth)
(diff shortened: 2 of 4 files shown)


Atermonera:
Oh, right, and 
Tested

Belsima:
I've noticed IAA isn't in that list; is that intentional or?

Atermonera:
They aren't listed under any of the lists of positions, and looking at it, command secretary also isn't on there because it's only listed under command_positions, which I explicitly don't use. I'll look into it later

Atermonera:
Testing (Compiled on NS, so planetside stuff doesn't show up):
[Selection](https://puu.sh/zcFbz/9153b44948.png)
[IAA Access](https://puu.sh/zcEX7/8dc9c146f3.png)
[Secretary Access](https://puu.sh/zcFiV/59d9d52dd4.png)

html/changelogs/Atermonera - id_comp.yml
```diff
@@ -0,0 +1,6 @@
+author: Atermonera
+
+delete-after: True
+changes:
+  - tweak: "ID computer can set command secretary and IAA access"
+  - tweak: "Command secretary has keycard auth. access"
```
Anewbe: This is only true if the config for minimum access is on? Otherwise they don't?

Anewbe:
Is this complete, Ater?

Atermonera:
Yes

PR #4667, opened 2018-01-28
RUST additions and updates, Modular guns, and Lead.

Andromeda-K22:
+Adds the new Hydromagnetic Trap (featuring Devsprites)
Currently unbuildable, but links to a EM field with the same ID as the Hydro-magnetic trap itself and generates power on par with radiation collectors. 

+Adds four new 'end-events' to the RUST, in place of the old rupture() proc

- Magnetic Quench. This is the most common form of shutdown, and is similar to the rupture proc, minus explosions.
-Magnetic Resonance Cascade. This one creates a number of electromagnet pulses in a radius from the core.
-Quantum Flux. This one dumps hot oxygen and phoron (hot enough to auto-ignite) in random tiles around the core, and is rare.
-Bluespace Quench Event. This one causes a number of moderately sized explosions around the core in a radius, Rare.

+Modular gun system. Currently implements a research-component based system (plus devsprites) to create modular energy pistols, carbines and cannons. Currently adminspawn only for balancing purposes.

+Lead and radiation_resistance variable. radiation_resistance is now used in place of the weight variable, and most of the materials should remain the same. Plasteel's resistance has been raised to make it not-useless for shielding the engine. Lead is now a material, and is adminspawn/map spawn only, with a radiation resistance of 350.


Andromeda-K22:
...Indentation. Okay. Thanks for nothing, Travis.

Andromeda-K22:
'comments dont matter' they said
they were WRONG
WRONG!

Anewbe:
Keep your pull requests simple and specific, please. Rust updates or Gun things, not both at once.

Cyantime:
Those radiation shielding changes will make anything that isn't a high-temp R-UST pretty meaningless in terms of radiation leakage.
What's the point of having a fourth way of getting power from the R-UST that just seems to be the core output, but stackable? Radiation collectors already work, the core itself generates power from the field temp, it produces plenty of heat for a TEG, what does the hydrotrap add?

Andromeda-K22:
@Anewbe 

I'm still entirely new to using git in any form, so i'm sorry for lumping everything into one big commit. I'm unsure as to how to keep thing separate. If i knew how, i would have. 

@Cyantime 

Not really.

The only material that was buffed was plasteel, which recieved a significantly higher resistance variable **because previously it was about as effective as tinfoil**. Three meters of walls would hardly stop any radiation. Steel is still the same. Boroglass, R-glass, glass, all those materials **are still the same**.

In order of effectiveness from least, to most: 
Steel
Uranium
Plasteel
Durasteel
Gold
Lead
Alien Alloy

It makes no sense - in any, ANY situation, for a space station with an engine that is **known** to produce large amounts of radiation, to have **no meaningful shielding what so ever**.

2. The Hydrotrap
It's less mystical than the radiation collectors using cans of a magical substance that generates power from radiation (of which, does not have anywhere NEAR the energy in it that these collectors are outputting)

It's for compactness. Hydrotraps produce more power, without mystical cans of gas. TEG setups take large amounts of space, and **this have been extensively tested**, will eventually loose efficiency over time without an absurdly large radiator. If someone wishes to construct one, that's fine. But from what i have seen, the desire is to have the engine as non-intrusive as possible.

Atermonera:
Well the first step is to not code on your master branch. There's a couple guides on the github's wiki tab that are very effective at showing how to go about setting up git and using it.

Cyantime:
r-ust
fuel2
I ran a primarily TEG (ignore the collector, was there for testing) R-UST for over 24 hours. It had a reasonably compact cooling loop with two TEGs in parallel, and neither ever dropped below about 2MW. Having large machinery to produce all your power seems in keeping with SS13's general aesthetic.

Two layers of plasteel (in the form of blast doors) and a small airgap will keep a reasonably powered up supermatter's radiation from leaking into the engine monitoring room, with the engine room properly irradiated. With this a single layer of plasteel will block 100% of radiation and make the engine room unacceptable safe.

Andromeda-K22:
'It's too safe' should never be an argument. For all the realistic behavior Polaris encourages, evidently workplace safety standards isn't one of those things.

I'l lower the radiation value, but i stand by the fact that 'it's too safe' isn't a proper arguement against things. We're not goon. This isn't a station full of mad-men. This is station orbiting a planet within it's orbital traffic space.

MarinaGryphon:
Being in an active engine room is probably not a good idea anyway, even if the radiation weren't there. There are radiation suits, anyway- as long as they can handle that amount of radiation,  I think it should be fine as is.

Atermonera:
(The ideal for engines is stability to the point where you can set it up and leave it for a couple hours without really having to check on it, but where it's not quite fire-and-forget, as well as being relatively easy to sabotage, and in most cases of minor sabotage, easy to recover from that, for antagonization)

Anewbe:
Upon review, I'm really, really not a fan of the idea of merging both of these at once. The RUST stuff seems alright, as far as I can tell, but the modular gun stuff is questionable.

Andromeda-K22:
It's a shame i have no idea how to cherry-pick commits, then. /shrug

Anewbe:
The code itself seems fine, and even though I'm not totally alright with adding the modular guns, this is good enough to be merged.
In the future, please be sure not to add multiple, unrelated features in the same PR.

PR #4672, opened 2018-01-28
humans can be older

elgeonmb:
changed human maxage from 110 to 130. Lorewise you can be up to like 150 but if you're older than 130 you probably aren't uhhhh independently mobile


Anewbe:
Why, exactly?

elgeonmb:
It was requested and is a literally one-byte change. No skin off my teeth if you have any objections tbh

PR #4673, opened 2018-01-28
Adds teshari sprite for flower dress

Schnayy:
Apologies if I've made any errors with pushing, it's been quite some time since I've used github. Simply adds teshari sprite and appropriate code (thanks cerebul) for it to be called on. 

https://gyazo.com/5da3f9159442c9059c4eb552403a7b58
https://gyazo.com/5eb1b8844c93c06cd93749777fc338c7


code/modules/clothing/under/miscellaneous.dm
```diff
@@ -388,6 +388,9 @@
 	name = "flower dress"
 	desc = "A beautiful dress with a skirt of flowers."
 	icon_state = "flower_dress"
+	sprite_sheets = list(
```
Anewbe: Is this not inherited from the default under clothing?

Schnayy: I have no clue, is it? Cerebul gave me the code, I'm a clueless babbit.

Cerebulon: You're right it's applied to all clothing/under items already, I assumed it wasn't. You can take this bit of code out schnee

Schnayy: Will do! Let me remember how to do that and I'll update. o7

PR #4556, opened 2018-01-16
What's Yours Is Mined

mistyLuminescence:
TL;DR:
- Mines are now dangerous. Don't step on them.
- Fragmentation mines added. Don't step on them either.
- You can now shoot or explode mines to detonate them from afar (50% chance).
- You can now disarm mines with hacking tools. If you hear beeping, run. Fast.
- It is possible to commit no mistakes and still lose. That is not a weakness; that is life. Stay alive, run fast.
- When you disarm a mine, you can pick it up and move it somewhere else and reactivate it.
- You know mines you disarmed and moved and then rearmed? Don't step on them.
- How to disarm a mine: pulse until it pings, then cut. Run if it beeps. Don't cut blind.
> 

**ENGINEER FIELD MANUAL 5-69**
Take your **screwdriver** and carefully open the panel of the mine. Without opening the panel, the mine cannot be disarmed.
Take your **multitool** and select a wire; note that each mine has a **different wire configuration**, so what works on one might not work on another.
Pulse the first wire with your **multitool** (recommended *f*: 536 MHz). One of three things will happen:

- A beep: this wire is either the detonation wire or a dummy wire (50/50). **Do not cut this wire**. Move on.
- A **series of rapid beeping**: this wire is the timed detonation wire. You have **two seconds to run**.
- A **ping**: this wire is the disarm wire, dummy wire, or trap wire. Cut this wire with your **wirecutters**.

After cutting a pinging wire, one of three things will happen:
- Nothing. This was the dummy wire. Move onto the next wire.
- A **series of rapid beeping**: this wire is the trap wire. You have **two seconds to run**.
- A **click**. The mine is disarmed, and can safely be moved.

Do **not** attempt to disarm a mine without a multitool. Doing so will likely result in loss of life and/or limb, and will forfeit your guaranteed medical care with NanoTrasen.

In the absence of proper equipment, a land mine may be remotely detonated via ballistic or energy impact (including hunting rifles) or secondary explosive charge (e.g. grenade, another mine).

The most common types of mines you will encounter are high explosive and fragmentation.
High explosive mines are marked with 'HE' and a grenade symbol. They explode violently when stepped on or detonated incorrectly. 
Fragmentation mines are marked with 'FRAG' and a grenade symbol. They explode into a shower of shrapnel when stepped on or detonated incorrectly.

[...]

PR #4679, opened 2018-01-30
Adds EMP mines

Atermonera:
They're as strong as EMP grenades, subject to input re: balance. Testing was done with a test of 2, 4, 6, 8, which was just used to ensure it works. EMP grenades have values of 2, 4, 7, 10
Tested

Also refactors mine code to be readable and use inheritance and not be hideous to ponder the basic function of.

Also I have no idea where those extra commits came from and I suspect some further git fuckery may be required on my end.


mistyLuminescence:
Probably not going to play nice with #4556, just a hunch.

Atermonera:
>When you can't actually keep track of whether something's been merged or not.

I'll fix any conflicts once that gets merged, there's important pieces from both. Didn't actually realized it had already done some of the refactoring.

Atermonera:
Fixed conflicts, tidied up the code again.

PR #4681, opened 2018-01-30
Replaces most of the gender macros.

### LINKED ISSUE #4632: Game does not understand gender.
The game seems to not understand the gender of the person.

The game to get 'he' and 'she' right.

#### What actually happened
Apparently people are confused about their gender.

#### Steps to reproduce
I have no clue.

#### Additional info:
 - **Server Revision**: master - 2018-01-22
2692c924487955f91b8f485c1ee9e950fdc9dbc1 

image

MarinaGryphon:
* Fixes #4632.
* Several, but not all, clunky and odd ternaries have been replaced.
* Almost all uses of the gender macros `\he`, `\him`, and `\himself` have been replaced by using the gender datum.

If there are any I missed, please tell me.
Also, this is probably very, VERY high risk.

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -763,7 +763,8 @@
 
 /mob/living/carbon/human/proc/play_xylophone()
 	if(!src.xylophone)
-		visible_message("<font color='red'>\The [src] begins playing \his ribcage like a xylophone. It's quite spooky.</font>","<font color='blue'>You begin to play a spooky refrain on your ribcage.</font>","<font color='red'>You hear a spooky xylophone melody.</font>")
+		var/datum/gender/T = gender_datums[get_visible_gender()]
+		visible_message("<font color='red'>\The [src] begins playing [T.his] ribcage like a xylophone. It's quite spooky.</font>","<font color='blue'>You begin to play a spooky refrain on your ribcage.</font>","<font color='red'>You hear a spooky xylophone melody.</font>")
 		var/song = pick('sound/effects/xylophone1.ogg','sound/effects/xylophone2.ogg','sound/effects/xylophone3.ogg')
 		playsound(loc, song, 50, 1, -1)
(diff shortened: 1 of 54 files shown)


MarinaGryphon:
hey hey it compiles, okay this can be considered LESS terribly risky now! apologies for the late night coding but it was the only time I could stand this tedious work, I'm running a fever and my neck is sore and I'm tired
any further work need sto be done just ask me alright? alright

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -922,10 +923,13 @@
...
-/mob/living/carbon/human/proc/get_visible_gender()
+/mob/living/carbon/human/get_visible_gender()
 	if(wear_suit && wear_suit.flags_inv & HIDEJUMPSUIT && ((head && head.flags_inv & HIDEMASK) || wear_mask))
-		return NEUTER
-	return gender
+		return PLURAL //plural is the gender-neutral default
+	if(species)
+		if(species.ambiguous_genders)
+			return PLURAL // regardless of what you're wearing, your gender can't be figured out
+	return get_gender()
```
PsiOmegaDelta: Sneaking in a comment from another code base, but this could probably be `..()` instead, to rely on however the base `/get_visible_gender()` acquires the appropriate gender.

MarinaGryphon:
The way the base `get_visible_gender()` gets the gender is by calling
`get_gender()`, so that would probably be more trouble than it's worth. One
call to `get_gender()` would become a call to parent and a call to
`get_gender()`.

code/modules/mob/living/carbon/human/human.dm
MarinaGryphon: That would change one proc call to two, unfortunately. Making the base `get_visible_gender()` use the ternary `get_gender()` uses would reduce it to just one proc, but also open the risk of breaking compatibility if get_gender ever uses different logic.

PR #4689, opened 2018-01-31
Sell materials to Cargo!

mistyLuminescence:
Long story short, you can now sell the following materials to Cargo, for the following prices:
Diamond: 20/sheet
Metallic hydrogen: 15/sheet
Uranium: 10/sheet
Platinum: 5/sheet <-- already exists
Phoron: 5/sheet <-- already exists
Osmium: 4/sheet
Gold: 4/sheet
Silver: 3/sheet
Deuterium: 2/sheet
Tritium: 1/sheet
Thalers: 0.02/sheet (50:1) <-- already exists

This means that all those miners out there can still get rewarded for their hard work even when there's nobody in Science. Prices subject to change.

Also, lots of spaces turned into tabs. That's just how things should be done, dammit.

### DIFF
--- a/code/game/supplyshuttle.dm
+++ b/code/game/supplyshuttle.dm
@@ -1,8 +1,8 @@
 //Config stuff
-#define SUPPLY_DOCKZ 2          //Z-level of the Dock.
-#define SUPPLY_STATIONZ 1       //Z-level of the Station.
+#define SUPPLY_DOCKZ 2		  //Z-level of the Dock.
+#define SUPPLY_STATIONZ 1	   //Z-level of the Station.
 #define SUPPLY_STATION_AREATYPE "/area/supply/station" //Type of the supply shuttle area for station
-#define SUPPLY_DOCK_AREATYPE "/area/supply/dock"    //Type of the supply shuttle area for dock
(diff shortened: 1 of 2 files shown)


Cyantime:
Deuterium is a literally infinite resource, you might not want to give any points for it.

mistyLuminescence:
how have I looked through the chem files this many times and only just now noticed this
stand by

mistyLuminescence:
required_reagents = list("water" = 10, "coolant" = 3, "foaming_agent" = 2)
catalysts = list("hydrophoron" = 5)

Also, coolant and kelotane now no longer spam admins when synthesised.
(why was this even a thing?? still applies to nitroglycerin. y'all still get pinged when someone's big-bomb-makin'.)

Nalarac:
I'm not sure about uranium being worth that much to sell. It's fairly easy to get when mining.

mistyLuminescence:
Never actually gone mining, just basing them off the R&D values. Maybe 7 (more than phoron, less than current)?

Anewbe:
This is not going to be merged, at this time.
We don't want more reasons for Research to end up with nothing. If we do end up looking into this in the future, it will be done differently.

PR #4697, opened 2018-02-01
Loadout Text-To-Speech device.

Cameron653:
- Adds a text-to-speech device to the loadout
- Costs 3 loadout points. Feel free to recommend a higher/lower number if this seems to much/little.

Device is based on the circuit TTS device, but is much simpler and can be spawned with at round start. Additionally, it can have the 1st user's name attached to it.

Features of the device:
- Upon first use, the device adds the person's name as a suffix.
- User can type in the message they wish to send. Message will be an audible message, just like the circuit.
- Checks to see if user is incapped/unable to use device. If so, they can't use the device.
- Device can be used via Ctrl-Z, Z with hotkey mode on, or just clicking the device.


Cameron653:
Aaaahhh. Made a bunch of small errors while modifying it for Polaris instead of another station. They're fixed, now.

Arokha:
You don't need to make a new PR to do that. Just reset/recommit or rebase and forcepush to that branch.

Cameron653:
Ah... My client doesn't support rebaseing. Going to have to look up the command line prompt. Give me a bit.

Cameron653:
Heeey, there we go. 6 guides later and I found out how to do it.

Atermonera:
Yeah, if you push any commits to a branch with an open PR, those commits'll automatically be added to the PR and the pr updated

Cameron653:
Anyhow, that's out of the way.

Is the cost for the translator fine? I choose three as it's a pretty niche item, but does have some use.

code/game/objects/items/devices/text_to_speech.dm
```diff
@@ -0,0 +1,28 @@
+/obj/item/weapon/text_to_speech
```
Anewbe: Why is this pathed as a weapon, in the devices folder?

Cameron653: Whoops. I wrote it as a weapon initially and planned to make it a device. Let me fix that real quick.

Cameron653:
...Let me see what went wrong.

code/game/objects/items/devices/text_to_speech.dm
```diff
@@ -0,0 +1,28 @@
...
+	w_class = ITEMSIZE_SMALL
+	var/named
+
+/obj/item/device/text_to_speech/attack_self(mob/user as mob)
+	if(user.incapacitated(INCAPACITATION_ALL)) //Are you in a state to actual use the device?
+		to_chat(user, "You cannot activate the device in your state.")
+		return
+
+	if(!named)
+		user << "You input your name into the device."
```
Arokha: to_chat

code/game/objects/items/devices/text_to_speech.dm
```diff
@@ -0,0 +1,28 @@
...
+		desc = "[initial(desc)] This one is assigned to [user.real_name]."
+		named = 1
+		/* //Another way of naming the device. Gives more freedom, but could lead to issues.
+		device_name = copytext(sanitize(input(user, "What would you like to name your device? You must input a name before the device can be used.", "Name your device", "") as null|text),1,MAX_NAME_LEN)
+		name = "[initial(name)] - [device_name]"
+		named = 1
+		*/
+
+	var/message = sanitize(input(user,"Choose a message to relay to those around you.") as text|null)
+	if(!isnull(message))
```
Arokha: Might want to do just `if(message)`, otherwise you can 'say' an empty string.

code/game/objects/items/devices/text_to_speech.dm
Cameron653: Resolved.

code/game/objects/items/devices/text_to_speech.dm
Cameron653: Done.

PR #4698, opened 2018-02-02
Reorder some of the planetary fall trigger

Cyantime:
Mobs land on the target turf before exploding into bits.
Admins are only messaged for mobs that fall, instead of all atoms.

### DIFF
--- a/code/game/objects/effects/step_triggers.dm
+++ b/code/game/objects/effects/step_triggers.dm
@@ -184,11 +184,11 @@ var/global/list/tele_landmarks = list() // Terrible, but the alternative is loop
 			A.forceMove(T) // Harmlessly move ghosts.
 			return
 
+		A.forceMove(T)
 		if(isliving(A)) // Someday, implement parachutes.  For now, just turbomurder whoever falls.
+			message_admins("\The [A] fell out of the sky.")
 			var/mob/living/L = A
 			L.fall_impact(T, 42, 90, FALSE, TRUE)	//You will not be defibbed from this.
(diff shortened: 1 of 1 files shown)


Anewbe:
I'm pretty sure the explosion is supposed to happen before-hand, just to damage the turf and make the sound. The actual damage is done in the falling proc itself.

Cyantime:
The explosion happens later, but I was having mobs bursting into gibs in a shuttle transit zone because the fall damage happens before they're moved to the final turf.
this

Anewbe:
So does this make it explosion -> move -> damage?

Cyantime:
They move, then the explosion and damage happens. The victim is hit by an ex_act(2), but they're already taking between 462 and 990 brute damage just from the fall. They're irrecoverably dead regardless.

Nalarac:
You say that but I've seen that prometheans and cyborgs can actually survive falling out of a shuttle.

Anewbe:
The proc is currently designed to kill humans, specifically. Prometheans really shouldn't be surviving it, but they're more than a little fucky, and cyborgs only take about 1/5th of that damage, iirc.

Anewbe:
This just in: The "Close and Comment" button is a wee bit too close to the normal button.

Cyantime:
Tested on other mobs, prometheans seem to survive every time, and ex_act(2) does like 15 total damage to them. Borgs take 120 extra damage from the explosion and still survive sometimes, but they're also big chunks of metal falling from space. Simple_animals with meaningful health usually survive the fall, the explosion adds 60 extra damage.

PR #4701, opened 2018-02-02
De-RNGifies budget insulated gloves

mistyLuminescence:
Budget insulated gloves now have a flat 60% electricity resistance (so you take 40% of the original damage), changed from a random % from (0,0.1,0.3,0.5,0.5,0.75,1.35).

In general, they're more reliable now, but they'll never fully protect you like 'real' insulated gloves will.

### DIFF
--- a/code/modules/clothing/gloves/color.dm
+++ b/code/modules/clothing/gloves/color.dm
@@ -5,16 +5,12 @@
 	siemens_coefficient = 0
 	permeability_coefficient = 0.05
 
-/obj/item/clothing/gloves/fyellow                             //Cheap Chinese Crap
-	desc = "These gloves are cheap copies of the coveted gloves, no way this can end badly."
+/obj/item/clothing/gloves/fyellow                             //Cheap Space Chinese Crap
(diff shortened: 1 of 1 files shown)


Belsima:
I'm actually against this, personally; the chance of getting a big reward was always nice, at the cost of a smaller chance of a bigger loss. And it's been demonstrated that RNG is sometimes wanted ingame by the Head Developer, such as in the case of emagging cyborgs.

FenodyreeAv:
It'd keep the old element if it was 0-100%, but minus the chance of immediate death if you had 1.35 gloves and a high output engine.

mistyLuminescence:
honestly having trouble seeing why NT would buy shitty space china gloves if they're so unreliable, having a flat number (or maybe a much smaller range; eg 30% to 50%) seems more fun and less !!fun!!

(also emag RNG is awful too but that's another can of worms)

Belsima:
Emag RNG is awful but I'm gonna use it to support my argument because im two-faced

Cyantime:
This was the entire point of the budget gloves, you might as well just remove them instead of this.

Anewbe:
This will not be merged at this time.
I will be making related changes, but a flat resistance is not desired by the staff.

PR #4703, opened 2018-02-02
Adds generic can_telecomm() proc, applies to Communicators

Arokha:
Adds a generic proc to determine if two atoms should be able to speak to each other over the telecomms network. It's a simple implementation.

It's applied on Communicators now, so you can't use a communicator where you don't have telecomms (sorta defeats the purpose of having telecomms if you can use a thing to make voice calls to/from anywhere).

I intend to apply it to cameras in some way or another (the AI can see borgs *anywhere*, 100% of the time, regardless of z-level, in any area, period. It's a little silly).

### DIFF
--- a/code/game/machinery/telecomms/telecomunications.dm
+++ b/code/game/machinery/telecomms/telecomunications.dm
@@ -314,7 +314,25 @@ var/global/list/obj/machinery/telecomms/telecomms_list = list()
 	circuitboard = "/obj/item/weapon/circuitboard/telecomms/hub"
 	long_range_link = 1
 	netspeed = 40
+	var/list/telecomms_map
 
+/obj/machinery/telecomms/hub/initialize()
+	. = ..()
+	LAZYINITLIST(telecomms_map)
+
+/obj/machinery/telecomms/hub/process()
+	. = ..()
+	telecomms_map.Cut()
+
+	if(!on)
+		return
+
+	for(var/M in links)
+		if(istype(M,/obj/machinery/telecomms/receiver) || istype(M,/obj/machinery/telecomms/relay))
+			var/obj/machinery/telecomms/R = M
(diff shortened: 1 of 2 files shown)


Anewbe:
Communicators already can't be used without Tcomms, if that's the only thing that applying a Tcomms check to them does.

Arokha:
I believe the video can't be. Voice calls can apparently be made to and from any z or area to anyone else in any z or area without issues. So this makes it so that messages and voice doesn't work either.

Anewbe:
I'm not sure I'm understanding you, sorry.
Are you saying that this makes it so you cannot call a communicator on a Z-Level that doesn't have active Tcomms?

Arokha:
Yup, that's right. No point in even having telecomms if there's magical devices that can circumvent it entirely.

Also sort of important if you want anyone to be isolated from the crew, ever. Otherwise people are always able to communicate and there's nothing an admin can do to overcome that.

Anewbe:
Does this account for Multi-Z? If I'm in a three level area on the top floor, and you're on the middle floor, do we both need Tcomms on those levels for it to work, or would just having one Tcomms array for the whole area work?

Arokha:
This doesn't alter anything about how that works.

How telecomms currently works is that it knows nothing about multi-z and you need one relay per floor. Somewhere on SC are relays that aim at each of your multi-z levels and provide them telecomms. They would also provide communicators telecomms.

So, in summary: Can you use radios on that z-level right now, in Southern Cross? Then you can use a communicator there too.

Anewbe:
Alright, thanks for walking me through that.

Arokha:
Deleted this branch when cleaning up because I'm a dummy and it said it didn't have any PRs open. Thanks Github.

PR #4706, opened 2018-02-03
Changes sleeper UI to be centered on the console

Cyantime:
To use the sleeper controls you need to be adjacent to the console, rather than the sleeper itself.

The diff exploded or something and doesn't seem to know what was actually changed.


Anewbe:
Tested and working?

Cyantime:
Yes.

PR #4707, opened 2018-02-03
Gives Virology teeth again

Cerebulon:
Virology got nerfed into the ground to make random events less of a pain in the arse. Since we no longer have randomly spawning viruses, I decided to make the now much rarer viruses a little more effective (If not necessarily dangerous) and give Virology something to actually do.

Additions:
- Adds antibiotic resistance as a trait for viruses. Roundstart viruses can have a % chance between 10 and 80% of resisting antibiotics. Comes into effect when antibiotics (spaceacillin) is already in the body (Previously prevented viruses with 100% success), when antibiotics are used to treat a virus (Previously a flat chance of success) and when the body is developing antibodies (Previously almost certain).
- When antibiotics fail to treat a virus in a host, its resistance increases from 1-10% for future attempts.
- High antibiotic resistance can prevent a host's chance of developing the antibodies required for a cure.
- Allows admin-spawn of "non-infectious" diseases (Direct blood transmission only)
- Adds mappable or admin-spawnable infectious mucus which contains a random virus that will last all round, similar to a virus dish.
- Vomit is now a disease vector and a possible stage 2 symptom.
- Drooling symptom can leave disease-spreading mucus on the floor.
- Internal organ failure symptom (Very slow progression, stage 4)
- Spontaneous combustion symptom (Very rarely triggers, mostly just makes you _feel_ burnt, stage 4)
- "Beneficial" organ rejection preventing symptom (Stage 3)
- Tinnitus symptom (Ringing in your ears, Stage 2)

Tweaks:
- A virus' rate of progression "Verocity" is now named "Aggressiveness" because Verocity isn't a real word.
- Most symptoms now renamed to be descriptive but not literally "[Does X] Syndrome"
- Adds infectious mucus to Quarantined Shuttle PoI.
- Viruses have a much lower chance of curing themselves instead of progressing to the next stage.
- Waiting Syndrome no longer occurs naturally in viruses.

Bugfixes:
- Xenomorphs and supernatural constructs can no longer catch the flu.
- A single beaker of virus food left in the incubator will no longer generate infinite reagent.
- Limb Paralysis symptom can now effect your left leg


Anewbe:
If it's still "Nothing nothing nothing dead" we will not merge this.

Cerebulon:
No, I haven't increased the lethality of anything and even the tier 4 symptoms are essentially just a buildup of minor injuries. Most of what this does it preventing all viruses from being cured after 2 minutes and actually getting to the point where people are coughing and throwing up instead of downing one pill the moment somebody detects a virus in your blood and never worrying about it again.
There's nothing about the current system, or these modifications which are "Nothing nothing nothing dead"
Right now, the system is essentially "Nothing nothing nothing cured completely by accident". This is supposed to make it "Something, something, something, cured"

code/modules/virus2/effect.dm
```diff
@@ -70,7 +70,7 @@
 	chance_maxm = 0
 
 /datum/disease2/effect/gibbingtons
-	name = "Gibbingtons Syndrome"
+	name = "Gibbington's Syndrome"
 	stage = 4
 	badness = 3
 	activate(var/mob/living/carbon/mob,var/multiplier)
```
Atermonera: Can you make all of this use absolute pathing while you're in the file?

Atermonera:
Would you prefer if the waiting syndrome were removed altogether?

code/modules/virus2/effect.dm
Cerebulon: I think I'm totally missing what you're asking here. Which paths aren't absolute?

Arokha: activate() is a proc definition and should be `/datum/disease2/effect/gibbingtons/activate(var/mob/living/carbon/mob,var/multiplier)`

Arokha:
> Would you prefer if the waiting syndrome were removed altogether?

/I/ would, anyway. Or at the least have a very low chance.

Mechoid:
I don't suggest removing waiting syndrome entirely, as it's actually incredibly useful for bus-use; when I, say, want a virus that *just* makes someone cold, for example, I can fill all the other slots with waiting and just have fridge on slot 2 or 3.

Cerebulon:
Could make it bus-only, like with gibbingtons?

code/modules/virus2/effect.dm
Atermonera: And it's like that for just about every proc in the file, I just noticed it while reviewing the diff and figured it'd be easier to make one comment than to mark every single instance of relative pathing

Wickedtemp:
I personally would be in favor of either leaving it. It's useful for playermade viruses, both antag and "beneficial" viruses, and useful for admin-made viruses. 

Low chance to occur naturally would be fine, I'd think.

Cerebulon:
Ok I have no idea what Travis is having issues with here.

Mechoid:
Okay, post-discussion, what specific issues are:
- A virus needs a non-virology-system only way to deal with in-round still, so antibiotic resistance can't become 100% without admin-intervention. Admin-created/edited viruses being 100% uncurable with spacacillin only is a non-issue as it's an event tool. ..Injecting a virus into a person for an event and them just immediately cilin-ing themselves and crushing the event is rather annoying.
- Waiting syndrome needs to only be usable in slots 1 and 4, without admin interventionm to prevent the previously mentioned 'Waiting, Waiting, Waiting, Crippling Effect' gimmicks of traitors or RNG's Wrath

Cerebulon:
Alright, how does base resistance generation being reduced to a maximum of 70% (Currently 80) with a 'hard' maximum of 80% resistance (Currently unlimited, effectively 100%, essentially superbug-able) unless manually set to 100 by the admin virus panel sound?
It got lost in the merge-squash but I already changed Waiting Syndrome to have a "badness" rating equal to gibbingtons and other 'severe' effects which effectively makes it admin-only, is that acceptable or would you rather it be reimplemented as RNG-able but super restricted?

Mechoid:
So it, at max, can RNG to be 70% resistant, but progesses only to 90% naturally. If it's RNG-able it's preferred to be only able to be in slot one or four, if it's impossible to make it lock to that barring admin intervention with the panel / otherwise, incredibly-low chance for appearance RNG-wise may be best

code/modules/virus2/effect.dm
```diff
@@ -70,344 +72,446 @@
...
 	stage = 4
 	maxm = 3
 	badness = 2
-	activate(var/mob/living/carbon/mob,var/multiplier)
-		mob.apply_effect(2*multiplier, IRRADIATE, check_protection = 0)
+
+/datum/disease2/effect/radian/activate(var/mob/living/carbon/mob,var/multiplier)
+	mob.apply_effect(2*multiplier, IRRADIATE, check_protection = 0)
 
 /datum/disease2/effect/deaf
```
Mechoid: Not caused by you, but this seems to be the same datum path as the lesser deafness symptom.

code/modules/virus2/effect.dm
```diff
@@ -70,344 +72,446 @@
...
-	name = "Hallucinational Syndrome"
+	name = "Hallucination"
 	stage = 3
-	activate(var/mob/living/carbon/mob,var/multiplier)
-		mob.hallucination += 25
+
+/datum/disease2/effect/hallucinations/activate(var/mob/living/carbon/mob,var/multiplier)
+	mob.hallucination += 25
 
 /datum/disease2/effect/deaf
```
Mechoid: Said lesser deafness symptom.

PR #4711, opened 2018-02-04
Add shuttle door/environment sensors

Arokha:
You would have to map them in to use them.

The console needs two lists mapped in (or made on a subtype):
`my_doors` is a list of `id_tag = Pretty Name` doors
`my_sensors` is a list of `id_tag = Pretty Name` sensors (`/obj/machinery/shuttle_sensor`)

Shuttle sensors are a new type of environment sensor meant specifically to address the issue of "no sensors in the game let me get air on an ADJACENT turf", so environment sensors **report on the environment 1 step in the dir they are facing**, so dir 2 shuttle sensor gets air on the turf to the south. This allows them to sit on the shuttle walls and not outside the shuttle and get left behind.

image

### DIFF
--- a/code/modules/shuttles/shuttles_web.dm
+++ b/code/modules/shuttles/shuttles_web.dm
@@ -128,6 +128,31 @@
 	icon_state = "flightcomp_center"
 	icon_keyboard = "flight_center_key"
 	icon_screen = "flight_center"
+	var/list/my_doors //Should be list("id_tag" = "Pretty Door Name", ...)
+	var/list/my_sensors //Should be list("id_tag" = "Pretty Sensor Name", ...)
+
+/obj/machinery/computer/shuttle_control/web/initialize()
+	. = ..()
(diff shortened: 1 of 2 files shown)


Cyantime:
Wouldn't it make more sense and be more useful for these to be a subtype of `/obj/machinery/airlock_sensor`, so they can be used for controlling airlocks?

Arokha:
Deleted this branch when cleaning up because I'm a dummy and it said it didn't have any PRs open. Thanks Github.

Arokha:
> Wouldn't it make more sense and be more useful for these to be a subtype of /obj/machinery/airlock_sensor, so they can be used for controlling airlocks?

Not really. I'd have to override essentially every proc on airlock_sensor, as I don't want these to process(), don't want them to behave the same with regards to getting air (one turf over instead of the turf they're on), etc.

PR #4716, opened 2018-02-04
Atom Initializing and Shuttles Subsystem + CALLBACK

Leshana:
### Implement SSatoms
- Moves proc/initialize() from being on /atom/movable, /are and /turf/simulated to being on /atom - Now turfs can initialize too
- Added the SSatoms subsystem which controls initialization of atoms at roundstart and during normal conditions.
- Refactored the way initialize() is called during /New() to utilize SSatoms instead of SScreation
- Removed SScreation, as it was only a stop-gap until SSatoms could be ported.
- Updated the maploader to inform SSatoms when it is loading maps instead of SScreation.initialization.
- Update every initialize() proc to return an initialize hint. Yes, all of them. Hopefully.

### Implement SSshuttles subsystem
- Replaces the shuttle_controller and shuttle process with the shuttles subsystem.  Instead of docking ports being initialized by the game ticker, its part of the StonedMC Master init order.
- The main advantage of this is control over the initialization order, as well as letting Master be aware of CPU we're using up with shuttle processing.
- By being part of the Master init order, we reduce the uncertainty about "are objects initialized yet?" which is nice, since shuttle docks break if machines aren't finished initializing!
- Let objects register to be informed when shuttles have been setup, and switch /obj/shuttle_connector to use it instead of the icky process hack.

### Ports callback datum from TG
- A callback datum system was created, a common framework that handles all of the nuts and bolts so anyone can implement callbacks easily without worrying about how to invoke procs.  Adapted from tgstation/tgstation#22476 by @MrStonedOne 
- Necessary for the callback feature of SSshuttles.

## For Developers
- As usual, you should avoid putting code in atom's /New() procs, instead putting it in `initialize()`
- `initialize()` now gets called with parameters!  First parameter will be `true` during mapload, and `false` for atoms created during gameplay.   For atoms created during gameplay, subsequent parameters will be the same params that were passed to New() (except for the first parameter which was loc)
- `initialize()` **must** return a value.  
  - return INITIALIZE_HINT_LATELOAD to have the `LateInitialize()` proc called after initialize() has been called on all other atoms being maploaded.  (This is a simple way to perform any actions which need to wait until neighbors have been initialized)
  - return INITIALIZE_HINT_QDEL if you want src to be qdel'd.  Do not qdel in initialize(), return this instead.
  - Otherwise just INITIALIZE_HINT_NORMAL


Leshana:
Resolved conflicts due to other PRs being merged.

PR #4719, opened 2018-02-04
Makes containment fields and tesla bolts bypass gloves

Cyantime:
Wearing a pair of rubber gloves on your hands will no longer save you from giant bolts of lightning or running facefirst into a deathfield. The containment field hits the torso, and the telsa bolts select a random body part.


Arokha:
Nooooo my completely nonsensical defense

PR #4726, opened 2018-02-05
Explorer webbing

mistyLuminescence:
Adds a brown webbing vest and a set of brown drop pouches to the explorer lockers. This doesn't allow them to take webbing in the loadout, because I get the sinking feeling something would go wrong if I tried to make that happen.

### DIFF
--- a/maps/southern_cross/structures/closets/misc.dm
+++ b/maps/southern_cross/structures/closets/misc.dm
@@ -47,6 +47,8 @@
 	new	/obj/item/clothing/mask/gas/explorer(src)
 	new /obj/item/clothing/shoes/boots/winter/explorer(src)
 	new /obj/item/clothing/gloves/black(src)
+	new /obj/item/clothing/accessory/storage/brown_vest(src)
+	new /obj/item/clothing/accessory/storage/brown_drop_pouches(src)
 	new /obj/item/device/radio/headset/explorer(src)
 	new /obj/item/device/flashlight(src)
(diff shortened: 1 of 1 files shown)


Woodratt:
Would rather allow it for them to choose in the loadout. Just have not had a chance to look into it.

Burrito-Justice:
Why not use this sprite? I believe it came with the rest of the explorer gear sprites (unless only the sprite was brought in but no item was made for it)
miner_webbing

Atermonera:
This is using something already in the code. If they want to add an additional type of webbing vest, that's fine by me, although other devs may have words re: whether we need another subtype, or which sprite to use, etc

Cyantime:
That sprite is currently used for this armor suit https://github.com/PolarisSS13/Polaris/blob/master/code/modules/clothing/suits/armor.dm#L373

Magneld:
I'd planned to continue screaming at Woodrat until he finally remembered to make webbings and pouches a loadout pick for Explorers and miners, would you all prefer the screaming continue so you can pick what colour/style of webbing you get or is just a single brown vest in the lockers fine.

Cyantime:
Your screaming already yielded results https://github.com/PolarisSS13/Polaris/pull/4771

Atermonera:
As woodrat mentioned previously, we'd prefer to have this just be a loadout option. Since #4771 currently does that, we're closing this in favor of it.

PR #4728, opened 2018-02-06
Attacked mobs will always target attacker

Cyantime:
This should prevent being able to shoot mobs to death with impunity from outside their vision ranges using binoculars or other methods of extending vision.

### DIFF
--- a/code/modules/mob/living/simple_animal/simple_animal.dm
+++ b/code/modules/mob/living/simple_animal/simple_animal.dm
@@ -746,13 +746,13 @@
 	if(stat || M == target_mob) return //Not if we're dead or already hitting them
 	if(M in friends || M.faction == faction) return //I'll overlook it THIS time...
 	ai_log("react_to_attack([M])",1)
-	if(retaliate && set_target(M))
+	if(retaliate && set_target(M, 1))
 		handle_stance(STANCE_ATTACK)
 		return M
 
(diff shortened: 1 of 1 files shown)


Atermonera:
If two separate mobs are attacking one at the same time, will it switch target each time it's attacked?
That is to say, if my buddy stands on the opposite side of an angry spider from me, and we both shoot it, if we shoot fast enough and always hit the spider, will it just be stuck in the middle, unable to reach and murderize either of us because it can't focus on a single target?

Cyantime:
This shouldn't change the behavior in that situation, it only applies if the source of the attack is outside of the mob's view range.

Arokha:
>If two separate mobs are attacking one at the same time, will it switch target each time it's attacked?
That is to say, if my buddy stands on the opposite side of an angry spider from me, and we both shoot it, if we shoot fast enough and always hit the spider, will it just be stuck in the middle, unable to reach and murderize either of us because it can't focus on a single target?

No. There's a thing I wrote into simple_animal code to prevent kiting. They can only switch targets every so often.

Anewbe:
@Cyantime This doesn't interfere with what Arokha mentioned, correct?

Cyantime:
It does not.

Anewbe:
Good enough for me.

PR #4730, opened 2018-02-06
Automatic firemode for guns

Andromeda-K22:
New gun firemode: automatic
Full-auto destruction at the tip of your fingers! If you're an admin. Switch the automatic var to 1 to make any gun automatic.

This code is largely donated by Mport and the SS:CM project. Relevant modifcations are located in automatic.dm, special.dm, and automagical.dm

Currently, only two guns are capable of automatic fire: 
L6 SAW
Suppression Gun


Andromeda-K22:
pls, bless me, travis

Andromeda-K22:
Entirely unsure why that .dme change is there..

Atermonera:
I'm 85% certain we already have this functionality in the form of burst-fire, but have never actually touched gun-code myself. @Anewbe probably has words regarding this

Andromeda-K22:
No.

Burst fire is completely different from this.

Anewbe:
Take it off of the suppression gun. Seeing as we're not a "Combat" server, that definitely doesn't need to be on anything that a random player can touch. There's a good reason that 5-round burst is basically nowhere.
This'll have to be passed by the staff proper, as well.

Andromeda-K22:
The suppression gun can only be accessed by the mercborg and admins - as a note, it only has 100 shots even on automatic, as the example i linked i had *intentionally* edited the cell to allow it to fire for longer than that. 

The automatic mode is also very inaccurate. You're going to hit more walls than people.

Atermonera:
Hitting more walls than people on a station probably means breaches, as a note

code/modules/projectiles/guns/automagical.dm
```diff
@@ -0,0 +1,70 @@
...
+/obj/screen/auto_target
+	name = "targeter"
+	icon = null//We dont want people to see this guy
+	density = 0
+	anchored = 1
+	var/obj/item/weapon/gun/gun
+	var/active = 0//Just tells us that it was clicked on so we should start shooting
+	var/delay_del = 0//Delays the del if we retarget without shooting
+
+	New(loc, var/obj/item/weapon/gun/G)
```
Anewbe: Absolute pathing, please.

code/modules/projectiles/guns/automagical.dm
```diff
@@ -0,0 +1,70 @@
...
+		..()
+		gun = G
+		var/image/I = image('icons/effects/Targeted.dmi', src, "locked")
+		I.override = 1
+		usr << I
+		//autodel() //Currently doesn't really work right.
+		return
+
+	CanPass()//Everything should ignore this guy and just pass by
```
Anewbe: Absolute pathing, please.

code/modules/projectiles/guns/automagical.dm
```diff
@@ -0,0 +1,70 @@
+//True automagic gun modes. Dakka dakka.
```
Anewbe: Since this is really more of an addition to guns, rather than a new type of gun, this should probably be in the default gun file? That'd also probably fix your .dme commit.

Andromeda-K22: It's in it's own file mostly because i'd rather not clutter up gun.dm with helper code / objects that aren't directly related to the gun subtype or the fire() procs.

Anewbe: I may move it later. My reasoning is this: If it can be used by any gun, which seems to be the case, putting it in a separate file, especially one named like that, will baffle future coders trying to look for it.

Andromeda-K22: Feel free to move it later, then. There was previously a block of pseudo code there which was the original code given to me by Mport, which, if you want to see: https://pastebin.com/zf1x7XXD

PR #4733, opened 2018-02-06
Budget Insulated Gloves tweaks

Anewbe:
Discussed in the staff meeting, at least in concept.
- Default gloves, black gloves, etc. have had their shock resistance reduced.
- Budget Insulated Gloves are now almost always better than non-insulated gloves, and they will **never** be worse than having bare hands.
- Proper insulated gloves haven't been touched.


SpadesNeil:
I always thought it was absurd how they could make the shock worse. What, are some fingers lined with copper?

PR #4738, opened 2018-02-06
Hyposprays and autoinjectors have a delay when used on a target that is not in help intent

Anewbe:
As discussed in the Staff Meeting.
- Using a hypospray or autoinjector on someone who is not in help intent will warn them, and take roughly 3 seconds.

### DIFF
--- a/code/modules/reagents/reagent_containers/hypospray.dm
+++ b/code/modules/reagents/reagent_containers/hypospray.dm
@@ -31,7 +31,7 @@
 
 /obj/item/weapon/reagent_containers/hypospray/attack(mob/living/M as mob, mob/user as mob)
 	if(!reagents.total_volume)
-		user << "<span class='warning'>[src] is empty.</span>"
+		to_chat(user, "<span class='warning'>[src] is empty.</span>")
 		return
 	if (!istype(M))
 		return
@@ -40,21 +40,29 @@
 	if(istype(H))
 		var/obj/item/organ/external/affected = H.get_organ(user.zone_sel.selecting)
 		if(!affected)
-			user << "<span class='danger'>\The [H] is missing that limb!</span>"
+			to_chat(user, "<span class='danger'>\The [H] is missing that limb!</span>")
 			return
 		else if(affected.robotic >= ORGAN_ROBOT)
-			user << "<span class='danger'>You cannot inject a robotic limb.</span>"
+			to_chat(user, "<span class='danger'>You cannot inject a robotic limb.</span>")
 			return
 
+		if(!H.stat)
+			if(H != user)
+				if(H.a_intent != I_HELP)
+					to_chat(user, "<span class='notice'>[H] is resisting your attempt to inject them with \the [src].</span>")
+					to_chat(H, "<span class='danger'> [user] is trying to inject you with \the [src]!</span>")
+					if(!do_after(user, 30))
+						return
+
 	user.setClickCooldown(DEFAULT_QUICK_COOLDOWN)
-	user << "<span class='notice'>You inject [M] with [src].</span>"
-	M << "<span class='notice'>You feel a tiny prick!</span>"
+	to_chat(user, "<span class='notice'>You inject [M] with \the [src].</span>")
+	to_chat(M, "<span class='notice'>You feel a tiny prick!</span>")
 
 	if(M.reagents)
 		var/contained = reagentlist()
 		var/trans = reagents.trans_to_mob(M, amount_per_transfer_from_this, CHEM_BLOOD)
 		admin_inject_log(user, M, src, contained, trans)
-		user << "<span class='notice'>[trans] units injected. [reagents.total_volume] units remaining in \the [src].</span>"
+		to_chat(user, "<span class='notice'>[trans] units injected. [reagents.total_volume] units remaining in \the [src].</span>")
 
 	if(!reusable && !used)
 		used = !used
@@ -110,7 +118,7 @@
 			user << "<span class='notice'>\The [src] already has a vial.</span>"
 	else
 		..()
-		
+
 /obj/item/weapon/reagent_containers/hypospray/autoinjector
 	name = "autoinjector"
 	desc = "A rapid and safe way to administer small amounts of drugs by untrained or trained personnel."
--- a/html/changelogs/Anewbe - Hyposprays.yml
+++ b/html/changelogs/Anewbe - Hyposprays.yml
@@ -1,0 +1,36 @@
+################################
+# Example Changelog File
+#
+# Note: This file, and files beginning with ".", and files that don't end in ".yml" will not be read. If you change this file, you will look really dumb.
+#
+# Your changelog will be merged with a master changelog. (New stuff added only, and only on the date entry for the day it was merged.)
+# When it is, any changes listed below will disappear.
+#
+# Valid Prefixes: 
+#   bugfix
+#   wip (For works in progress)
+#   tweak
+#   soundadd
+#   sounddel
+#   rscadd (general adding of nice things)
+#   rscdel (general deleting of nice things)
+#   imageadd
+#   imagedel
+#   maptweak
+#   spellcheck (typo fixes)
+#   experiment
+#################################
+
+# Your name.  
+author: Anewbe
+
+# Optional: Remove this file after generating master changelog.  Useful for PR changelogs that won't get used again.
+delete-after: True
+
+# Any changes you've made.  See valid prefix list above.
+# INDENT WITH TWO SPACES.  NOT TABS.  SPACES.
+# SCREW THIS UP AND IT WON'T WORK.
+# Also, all entries are changed into a single [] after a master changelog generation. Just remove the brackets when you add new entries.
+# Please surround your changes in  double quotes ("), as certain characters otherwise screws up compiling. The quotes will not show up in the changelog.
+changes: 
+  - tweak: "Hyposprays and autoinjectors now have a delay on use when the target is conscious and not in Help Intent."


mistyLuminescence:
Kinda defeats the point of an instant injector, doesn't it?

Belsima:
They should be on help intent if needing help, no?

mistyLuminescence:
You'd think, but this takes away one of the biggest advantages of autoinjectors/hyposprays - the part where it's instant. Add a slight (not 3 seconds) delay against voidsuits/thick armor etc, sure, but you don't need to mess around with finding a vein with these, just jab them anywhere - even when they're not being cooperative.

If we want to discourage combat uses of autoinjectors and hyposprays (and take away med's only reliable way of defending themselves), set the usage time for non-help intents to 0.5 seconds; just enough that if they're already moving, or if they've got good (3 seconds just requires being anything other than a literal snail to avoid) reaction time, they have a good chance of dodging it.

Alternatively, just shoot/aim at them before they're in melee range, or dose yourself up on dylovene first, or get one of your own... endless possibilities that don't involve making the fire axe med's only option.

Belsima:
0.5 seconds sounds good to me, personally.

Also, the fire axe wouldn't be 'med's only option' - syringes on harm, syringe guns, etc. Besides, perhaps medical should use the handy ; key and call security.

mistyLuminescence:
And when the only security officer in syringeproof armor is dead/SSD/the antag?

The two main problems with this are medical can no longer reliably defend themselves like other departments can (engies can make stunprods, cablecuffs, crossbows etc easily; research has a ton of weapons in R&D; security is security), and combat medics become effectively useless. This PR *introduces* these problems, which currently don't exist. Is this an "I ded pls nerf" or does it actually address any existing issues?

html/changelogs/Anewbe - Hyposprays.yml
```diff
@@ -0,0 +1,36 @@
...
+# Optional: Remove this file after generating master changelog.  Useful for PR changelogs that won't get used again.
+delete-after: True
+
+# Any changes you've made.  See valid prefix list above.
+# INDENT WITH TWO SPACES.  NOT TABS.  SPACES.
+# SCREW THIS UP AND IT WON'T WORK.
+# Also, all entries are changed into a single [] after a master changelog generation. Just remove the brackets when you add new entries.
+# Please surround your changes in  double quotes ("), as certain characters otherwise screws up compiling. The quotes will not show up in the changelog.
+changes: 
+  - tweak: "Hyposprays and autoinjectors now have a delay on use when ."
```
mistyLuminescence: "when ."

code/modules/reagents/reagent_containers/hypospray.dm
```diff
@@ -40,21 +40,27 @@
...
 		if(!affected)
-			user << "<span class='danger'>\The [H] is missing that limb!</span>"
+			to_chat(user, "<span class='danger'>\The [H] is missing that limb!</span>")
 			return
 		else if(affected.robotic >= ORGAN_ROBOT)
-			user << "<span class='danger'>You cannot inject a robotic limb.</span>"
+			to_chat(user, "<span class='danger'>You cannot inject a robotic limb.</span>")
 			return
 
+		if(H.a_intent != I_HELP)
```
mistyLuminescence: Does this check if they're conscious?

code/modules/reagents/reagent_containers/hypospray.dm
```diff
@@ -40,21 +40,27 @@
...
+			to_chat(user, "<span class='notice'>[H] is resisting your attempt ot inject them with [src].</span>")
```
mistyLuminescence: "ot"

Wickedtemp:
Does it check if the patient is conscious, cuffed, lying down, in crit, etc? Or are we going to have unconcsious people resisting injectors.

code/modules/reagents/reagent_containers/hypospray.dm
mistyLuminescence: And this is going to go "they're resisting" before they even start resisting, isn't it? Just because someone's not on help intent, that doesn't mean they're actively resisting you.

Wickedtemp:
Update: Official Verdict
Your code is broken and needs to be reworked. I copied it, spawned in as a greyshirt on harm intent and switched bodies to Dr. Nekid Baldman, leaving the greyshirt ghosted/asleep,(https://gyazo.com/969408d9898db7abda03977059849d76) messed around with it, and found that:

- Sleeping people will resist it, https://gyazo.com/4793a9e840a7d622a59fec66ef672750 
- Not only will sleeping people resist it, but so will people who'd otherwise be conscious in hard crit https://gyazo.com/f6ef2ce92577bc0ea963767526db4231
- Not only will sleeping people and people in hard crit resist. I killed them, and they still resisted. Corpses will resist. https://gyazo.com/746f8d543c4d396e3bab93955d71597f
- Not only will sleeping people, unconscious people, and dead people resist... So will YOU. https://gyazo.com/3b34c98fad1bc9ea3944a324447392ec

It might be easier to just identify what reagent is being used to injector powergame and nerf it. It's probably chloral or potassium chlorophoride, give those a delay to their stun. Currently it's like...one tick delay until they're stunned.

Atermonera:
If you must know, the reagent in question is chloral, and the things that other departments can make for personal defense, with the exception of R&D, having access to syringe guns, is that those methods are non-instant. Hypospray with chloral is.

PrismaticGynoid:
If the problem is chloral hydrate being an instant stun, then nerf that - make it take longer to take effect. Don't give nerfs that will affect people using the hypospray for completely legitimate medical purposes. If a person being not on help intent makes it take longer, it should still be much shorter than using a syringe (1 second, 2 if armored?).

mistyLuminescence:
We already know chloral is kinda OP, this has been a thing for a long time. Why not just... y'know. Nerf chloral? Don't introduce more problems without addressing the original one.

Nalarac:
While I don't agree with this PR, I personally wouldn't mind if hypos had some delay (probably not as much as a syringe) when using it against thick materials or voidsuits, since I find it kind of silly that someone can just run up and instantly inject me through my spacesuit (even if it is beneficial).

code/modules/reagents/reagent_containers/hypospray.dm
```diff
@@ -40,21 +40,29 @@
...
 			return
 		else if(affected.robotic >= ORGAN_ROBOT)
-			user << "<span class='danger'>You cannot inject a robotic limb.</span>"
+			to_chat(user, "<span class='danger'>You cannot inject a robotic limb.</span>")
 			return
 
+		if(!H.stat)
+			if(H != user)
+				if(H.a_intent != I_HELP)
+					to_chat(user, "<span class='notice'>[H] is resisting your attempt to inject them with \the [src].</span>")
```
mistyLuminescence: This is still going to go "hey, they're resisting!" as soon as you try and inject them. Possibly untrue (what if they're just not on help intent, and not actively trying to resist?), and possibly spammable on both ends - click someone a bunch and then move away before the 3 seconds are up.

code/modules/reagents/reagent_containers/hypospray.dm
```diff
@@ -40,21 +40,29 @@
...
 		if(!affected)
-			user << "<span class='danger'>\The [H] is missing that limb!</span>"
+			to_chat(user, "<span class='danger'>\The [H] is missing that limb!</span>")
 			return
 		else if(affected.robotic >= ORGAN_ROBOT)
-			user << "<span class='danger'>You cannot inject a robotic limb.</span>"
+			to_chat(user, "<span class='danger'>You cannot inject a robotic limb.</span>")
 			return
 
+		if(!H.stat)
```
mistyLuminescence: Can you still resist being injected if you're lying down, asleep/SSD, or in crit? How about in a straightjacket?

code/modules/reagents/reagent_containers/hypospray.dm
mistyLuminescence: This is still going to go "hey, they're resisting!" as soon as you try and inject them. Possibly untrue (what if they're just not on help intent, and not actively trying to resist?), and possibly spammable on both ends - click someone a bunch and then move away before the 3 seconds are up.

Anewbe: It's utilitarian. Not everyone who's hostile is in Harm intent, and this needs to be more general than "Maybe the target is resisting, maybe not, I don't know, but it's gonna take a minute".

code/modules/reagents/reagent_containers/hypospray.dm
Anewbe: Lying down yes, SSD no, crit no.
If your target is in a straightjacket there is a 100% chance that you'll be able to keep them still long enough to inject.

mistyLuminescence: Fair. If they're in a straightjacket they've probably been resisting anyway.

code/modules/reagents/reagent_containers/hypospray.dm
mistyLuminescence: Utilitarian but arbitrary. Not everyone who's resisting is in non-help intent, and not everyone who's on non-help intent is resisting. Or maybe someone's resisting being injected by one person but not another (e.g. you're a merc, you want to be injected by the antag medicborg but not by the chemist with the deathmix). There's a lot of variables at play here, no pun intended, and this seems like it'll just get in the way more often than not.

mistyLuminescence:
If the intent here is to nerf chloral, why not just nerf chloral? Wickedtemp gave some good ideas on how to do that above^, and it avoids creating new problems such as "combat medics are now officially shit-tier" and "medical has a really hard time defending themselves now" and "we can't tell if someone's on help intent or not just by looking at them", and "chloral is still OP but now actual legitimate medicine gets hurt".

TL;DR why not just nerf chloral

code/modules/reagents/reagent_containers/hypospray.dm
Anewbe: If you've got a better wording, please share. Otherwise it will remain as it is.

Anewbe:
The issue with nerfing chloral is that a) chloral still has legitimate uses as-is, and b) if I do nerf chloral, people will just make another chem combination to abuse the system of auto-hypo.

I will acknowledge that yes, there were issues with the code, thank you for pointing them out. That said, this change was discussed during the bi-weekly Staff Meeting, and was agreed on by the Staff present. I have grown tired of certain individuals making immense posts of complaints literally every time we look at changing anything to do with Medical Code. If this continues to occur, I will start applying GitHub bans.

mistyLuminescence:
Threatening people with bans for making legitimate points won't stop the points from being legitimate, it'll just stop us from being able to point them out. Chloral _does_ have legitimate uses, but if we don't want instastuns to be one of them, we could always just make instastuns not one of them. Do the same to other chems that instastun and suddenly offensive chemistry isn't overpowered.

Hyposprays and autoinjectors also have legitimate uses which will be hurt by this, but unlike nerfing chloral, nerfing hyposprays doesn't address the "you can instastun someone by injecting them with 5u chloral hydrate, and then toolbox them to death" problem.

Belsima:
Could we do both this AND nerf chloral? Chloral is bullshit when stabbed via syringe anyway since it's still an instastun.

mistyLuminescence:
Chloral _is_ bullshit, and it should be nerfed... and nerfing chloral would mean this PR wouldn't be necessary. The problem with this isn't just that it doesn't address chloral being OP, it's that it introduces more problems than it solves.

code/modules/reagents/reagent_containers/hypospray.dm
mistyLuminescence: Set "x is injecting Y with the hypospray" to be visible to both parties, and then "Y resisted!" to be if the timer check fails - or even better, "you couldn't inject Y, stay close to them!" or something.

Wickedtemp:
If you nerfed chloral, while there are more deadly mixes, at least those take a bit of effort. 

Chloral gives you a 60 second stun even if you inject 0.5u. Getting rid of that and the stun for potassium chlorophoride, as far as bullshit stuns go, you don't have many options left. 

And if there are, and it becomes a problem, then cut those as well.

mistyLuminescence:
After a long discussion on Discord with certain folks, I still see no need for this PR. Apparently the problem isn't "chloral is OP" (even though chloral _is_ OP), the problem is "some CMOs have recently been injecting antags with chloral to ggnore them".

And instead of, say, slapping those CMOs - despite what some people who shall remain nameless think, most medstaff don't prep chloral - the solution here is to... nerf the delivery system in general?
Which implies they won't just use the syringe gun instead? Because they'll just use the syringe gun instead.

If we think those CMOs are bad, slap them. They're making things unfun for other people by instastunning every antag in the round.

If we want to keep chloral at its lethality level, add a safety check to hyposprays and autoinjectors - if there's toxins (including chloral, but vital components of all other deathmixes too) loaded, they can't be used unless the safety is disabled (through EMP or emag). Non-antags who want to use deathmixes will have to deliver them some other way.

If we think chloral is too lethal, nerf chloral to not instastun. Keep the sleep effect so it works for popped lung patients, but doesn't work in combat.

If we think being able to inject someone while they're resisting is unrealistic (and shooting lasers at a giant death-crystal to produce power isn't?), add a secondary effect to resisting that stops you being injected, handcuffed or disarmed for 5 seconds, but that can only activate every minute.

But this PR doesn't address the problem; it just creates more problems - the exact opposite of what improvements are supposed to do.

Neerti:
This will be merged despite its unpopularity for a few reasons, which I will list below;

 - This was discussed in the latest public staff meeting, and it was agreed upon that this would be done, so going back on the agreed upon plan afterwards would reflect poorly on us and I imagine it would cause a lot of future issues if we are not able to fully commit to a plan of action that has reached agreement.
 - This change, to my knowledge, only adversely affects the CMO hypospray, and the auto-injectors (as one is a child of the other code-wise).
 - This should not adversely affect those two items too much, for the reasons listed below.
   - Generally people are in help intent by default, otherwise walking down halls would be hell, as such, the delay should be uncommon if the items above are being used legitimately.
   - Three seconds is a fairly short period of time, and code-wise means the difference of one and a half ``Life()`` ticks, so it is very unlikely that the delay will get someone killed.
   - The above is even more unlikely because if someone is in crit, the delay doesn't happen even if they were not on help intent, so this won't affect people using those items on those dying.
   - If the person someone is trying to heal is on harm intent because they are in an active gun fight, and can't wait three seconds, then the two of them need to take a step back and go somewhere safer.
     - Medical people probably shouldn't go running into active gun fights where they could get shot and die.
 - The delivery system is the problem, and not the payload, or more specifically, nerfing Chloral won't fix the problem. Another chem or mix will take its spot as the go-to for screwing someone over, and its difficult to fix that without nerfing a lot of chemistry, which is not something I imagine anyone really wants.

I am confident that this is not going to ruin medical forever. [Compared](https://github.com/PolarisSS13/Polaris/pull/4374) [to](https://github.com/PolarisSS13/Polaris/pull/4264) [other](https://github.com/PolarisSS13/Polaris/pull/3438) [changes](https://github.com/PolarisSS13/Polaris/pull/3984) [to](https://github.com/PolarisSS13/Polaris/pull/4408) [medical](https://github.com/PolarisSS13/Polaris/pull/4269) that have happened in the past, this one is very much a minor change, and the delay shouldn't really matter or even show up that often if you're using the hypo/injectors for their intended purpose as medical tools, and not weapons or combat healing gear.

PR #4741, opened 2018-02-06
Fixes #4669.

### LINKED ISSUE #4669: Shuttles can create ghosts of occupants on the landing pad
If a person (or a ghost) is on the shuttle when it launches a "ghost" sometimes stays behind. These ghosts aren't visible to most people and I don't know why I can see them. These ghosts can't be interacted with, examining turns you in the opposite direction and you cannot throw items at them. Double-clicking them as a ghost teleports you to nullspace. If a person that has a "ghost" uses the shuttle again the "ghosts" disappear.
https://my.mixtape.moe/ksajvq.png
https://my.mixtape.moe/lemgta.png
https://my.mixtape.moe/cuzqdc.png
I didn't expect to have a shuttle-related spiritual experience.
#### What actually happened
I had a shuttle-related spiritual experience.
#### Steps to reproduce
Unfortunately I don't know, as nobody else could see these ghosts.
#### Additional info:
4b6e64111e6521b3d8e9feb3bbac855aac4cc56f 
Only happened on the Sif landing pad One

MarinaGryphon:
Now calls check_shadow() on mobs when they get moved in `area.move_contents_to()`. This should fix #4669 for both elevators and shuttles.

### DIFF
--- a/code/_helpers/unsorted.dm
+++ b/code/_helpers/unsorted.dm
@@ -821,7 +821,9 @@ proc/GaussRandRound(var/sigma,var/roundto)
 					for(var/mob/M in T)
 						if(istype(M, /mob/observer/eye)) continue // If we need to check for more mobs, I'll add a variable
 						M.loc = X
-						M.check_shadow() // Need to check their Z-shadow, which is normally done in forceMove().
+						if(istype(M, /mob/living)
+							var/mob/living/LM = M
+							LM.check_shadow() // Need to check their Z-shadow, which is normally done in forceMove().
 
 					if(shuttlework)
 						var/turf/simulated/shuttle/SS = T
--- a/code/_helpers/unsorted.dm
+++ b/code/_helpers/unsorted.dm
@@ -817,11 +817,11 @@ proc/GaussRandRound(var/sigma,var/roundto)
 					for(var/obj/O in T)
 						O.loc = X
 
-					//Move the mobs unless it's an AI eye or other eye type. Need to check their Z-shadow.
+					//Move the mobs unless it's an AI eye or other eye type.
 					for(var/mob/M in T)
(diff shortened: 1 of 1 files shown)


code/_helpers/unsorted.dm
```diff
@@ -821,6 +821,9 @@ proc/GaussRandRound(var/sigma,var/roundto)
 					for(var/mob/M in T)
 						if(istype(M, /mob/observer/eye)) continue // If we need to check for more mobs, I'll add a variable
 						M.loc = X
```
Neerti: Would it be better to replace this line with ``M.forceMove(X)``, since the comment you made below sounds like it would also fix it, or is there a reason forceMove cant be used?

MarinaGryphon: The comment above it says it shouldn't use forceMove- I immediately went to change it, but decided against it because of the comment.

Arokha: Heeey you found my comment.

It's debatable on whether or not you should use forceMove(). This calls things like entered, crossed, etc on all the things that are moving. Technically things in a whole area that just move DIDN'T move relative to their surroundings, thus I just set loc instead.

Like if you, for example, have ambient elevator music like we do, using forceMove resets the music every time the turbolift moves because you "moved". But in actuality, you're not supposed to have "moved", the whole lift/shuttle/whatever moved with you, so setting loc seems to represent this better.

Arokha:
I wonder if this fixes our "moving upwards and talking in a turbolift causes crazy echoing". Prooobably? Since it'd have the shadows talking.

code/_helpers/unsorted.dm
Anewbe: I can buy that explanation.

PR #4743, opened 2018-02-06
Adds 2 new cave POIs

BillyBangles:
There's a few neat things here.
- A POI in the caves: a mysterious abandoned temple
- A candelabra!
- A variant of candelabras and normal candles that lights itself when it's spawned and never goes out (unless put out by a player)
- A neat red crystal that looks like supermatter but isn't (might need a sprite change if this is a problem)
- A dark ritual.

Screenshots:
https://puu.sh/zhZjV/8146dc460c.png
https://puu.sh/zi9rU/319da8303f.png

(Tested, all works.)


Atermonera:
That diff. Praise be to Map Merger! Praise be to Satan!

Anewbe:
Uh... Yeah, why was the main map touched?

Atermonera:
Looking just at the numbers on the diff, I'd guess that you removed something somewhere and it shifted the index for everything by -1 Unfortunately github is too scared to show the full diff, as bad an idea as that is

BillyBangles:
i do not recall touching the main map. i have no idea what i did to fuck that up.

BillyBangles:
alright that looks like it did it

code/game/objects/items/weapons/candle.dm
```diff
@@ -70,3 +70,42 @@
...
+		src.lit = 1
+		//src.damtype = "fire"
+		for(var/mob/O in viewers(usr, null))
+			O.show_message("<span class='notice'>\The [name] mysteriously lights itself!.</span>", 1)
+		set_light(CANDLE_LUM)
+		processing_objects.Add(src)
+
+/obj/item/weapon/flame/candle/candelabra/everburn
+	wax = 99999
+	light()
```
Anewbe: Uh... This doesn't actually do anything, as far as I can tell? And I'm also mildly surprised it actually compiles with this here. Could be wrong.

BillyBangles: ah, right, forgot to take that bit out, since i made it light itself in new() instead

BillyBangles:
that should do it

PR #4745, opened 2018-02-07
Makes teshari hide work once more.

Cameron653:
- Makes the human.dm variable comments all line up nice and neatly. https://i.imgur.com/Mm5xHLW.png

- Makes a new proc called on update_icon and update_icons, update_icon_special

- update_icon_special checks to see if the creature is hiding. If so, it sets their layer to the "hide" layer (Taken from the old hide proc. Layer is 2.45). Left open to make more edits for further expansion, such as clothing possibly making you able to hide, or maybe another variable that makes you able to show yourself above all layers no matter what. Of course, this is out of the scope of the current PR, but would be possible in the future by using this.

- Adds a humanoid hide proc. Pretty much the same as old proc, but sets a hiding variable to 1

### DIFF
--- a/code/modules/mob/living/carbon/human/human_powers.dm
+++ b/code/modules/mob/living/carbon/human/human_powers.dm
@@ -271,7 +271,7 @@
 	set desc = "Allows to hide beneath tables or certain items. Toggled on or off."
 	set category = "Abilities"
 
-	if(incapacitated(INCAPACITATION_STUNNED)) // No hiding if you're stunned!
+	if(stat == DEAD || paralysis || weakened || stunned) // No hiding if you're stunned!
 		return
 
 	if (!hiding)
--- a/code/modules/mob/living/carbon/human/human_powers.dm
+++ b/code/modules/mob/living/carbon/human/human_powers.dm
@@ -277,8 +277,8 @@
 	if (!hiding)
 		layer = 2.45 //Just above cables with their 2.44
 		hiding = 1
-		src << text("<font color='blue'>You are now hiding.</font>")
+		to_chat(src, "<font color='blue'>You are now hiding.</font>")
 	else
 		layer = MOB_LAYER
 		hiding = 0
-		src << text("<font color='blue'>You have stopped hiding.</font>")
+		to_chat(src, "<font color='blue'>You have stopped hiding.</font>")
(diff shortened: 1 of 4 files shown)


Anewbe:
Tested?

Cameron653:
Indeed. I can take a video when I get home tomorrow. Made sure that you can rest while hiding, but if stunned you get taken out of hiding.

https://i.imgur.com/FB5qXFD.png Hiding resting

https://i.imgur.com/QgbcJxs.png Hiding 

https://i.imgur.com/opeccsX.png Not hiding

Cameron653:
Found a small bug. For some reason when they get KO'd it doesn't get them out of hiding. Looking into fixing it.

Anewbe:
Thanks for the heads up.

Cameron653:
Resolved. Changed to a method that correctly sees if you're stunned or not.

See: https://i.imgur.com/bUFHHDI.gifv
Started at resting
Stood up
Walked
Shot self with stun revolver
Saw was stunned. Moved out of hiding.

Cameron653:
Also, should note, this should be ready to merge, provided there are no more questions/comments/recommended adjustments.

code/modules/mob/living/carbon/human/human_powers.dm
```diff
@@ -265,3 +265,20 @@
...
+	set desc = "Allows to hide beneath tables or certain items. Toggled on or off."
+	set category = "Abilities"
+
+	if(stat == DEAD || paralysis || weakened || stunned) // No hiding if you're stunned!
+		return
+
+	if (!hiding)
+		layer = 2.45 //Just above cables with their 2.44
+		hiding = 1
+		src << text("<font color='blue'>You are now hiding.</font>")
```
Anewbe: Mind swapping this to a to_chat() proc?

code/modules/mob/living/carbon/human/human_powers.dm
```diff
@@ -265,3 +265,20 @@
...
+	else
+		layer = MOB_LAYER
+		hiding = 0
+		src << text("<font color='blue'>You have stopped hiding.</font>")
```
Anewbe: Mind swapping this to a to_chat() proc?

PR #4748, opened 2018-02-07
Comments out the automatic firemode code

Anewbe:
This is more or less a manual revert of #4730, which was merged based on a misunderstanding among the Dev Staff.

Also cleans up the code a bit.

When we're done testing and tweaking this, it may come back.


Neerti:
Specifically the misunderstanding was that people thought 'I want this' meant 'This has been reviewed and ready to go', which resulted in an early merge.

PR #4753, opened 2018-02-07
Ethylredox removes alcohol

### LINKED ISSUE #4752: Ethylredoxrazine is broken.
Ethyl's been broken since day one of Polaris.
Well, I expect it to synth with ethanol to create water, and to neutralize alcohol in the bloodstream and stomach when administered. https://gyazo.com/b81abf227ccc96c091b264778df2a98a
#### What actually happened
It still doesn't remove alcohol from stomach or bloodstream. It also doesn't combine with ethanol to make water.
#### Steps to reproduce
Someone drinks a keg.
"We'll use ethyl!"
Doesn't work. Still have all the booze ever.
"We'll use carbon pills!"
Do so. Realize this is going to take quite a while because carbon removes 1u per tick, which is fine for pretty much everything except things people are likely to binge on, such as coffee and alcohol.
Babysit the drunk patient that you aren't legally allowed to let refuse treatment.
#### Additional info:
 - Been a thing since Polaris was live, pretty sure
 Doesn't synth to make water: https://gyazo.com/35393ea62397daa0e3f00d0c011f14b5
Code that... makes it look like it's supposed to remove the alcohol, but for some reason just isn't: https://gyazo.com/b609bced2de7bc055d49d6d68887d2f9
Not removing alcohol in bloodstream: https://gyazo.com/780b33a5f6a30302de8decac592f5410
Not removing alcohol in stomach: https://gyazo.com/ac78fd61727b64cdfc453759387f2a65

Alternatively, porting Bay's stomach pumps may be better long-term, since that'd cover everything from mushrooms to coffee to alcohol, and simply edit Ethyl's description to fit what it does currently.

Atermonera:
Fixes #4752 
Ethyl does NOT react with alcohol because then as soon as you ingest or inject it, it'll instantly react with whatever alcohol is in that reagent container and you may need more to handle anything in the other container.

Tested, it had a noticeable impact on the rate alcoholic beverages were removed at, sorta hard to give a hard rate without looking at how it's metabolized, but it's notably faster.

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine.dm
@@ -461,11 +461,11 @@
 	if(M.ingested)
 		for(var/datum/reagent/R in M.ingested.reagent_list)
 			if(istype(R, /datum/reagent/ethanol))
-				R.dose = max(R.dose - removed * 5, 0)
+				R.remove_self(removed * 5)
 	if(M.bloodstr)
 		for(var/datum/reagent/R in M.bloodstr.reagent_list)
 			if(istype(R, /datum/reagent/ethanol))
-				R.dose = max(R.dose - removed * 15, 0) 
+				R.remove_self(removed * 15)
 
 /datum/reagent/hyronalin
 	name = "Hyronalin"
--- a/html/changelogs/Atermonera - ethyl_fix.yml
+++ b/html/changelogs/Atermonera - ethyl_fix.yml
@@ -1,0 +1,5 @@
+author: Atermonera
+
+delete-after: True
+changes:
+  - tweak: "Ethylredoxrazine actively removes alcohol from the stomach and bloodstream"


Wickedtemp:
Alright, so testing it ingame:

Stomach booze is removed at 0.5u per tick

Bloodstream is removed at almost five times the rate. 

Given that carbon removes the reagent at 1u per tick in the stomach, I'd suggest increasing the rate for ethanol removal in the stomach juuuust a little.

Wickedtemp:
Update Update: Boosting the ingested ethyl to 15 will mean it removes 1.5 per tick, 7.5u total. Medicines, when ingested, are half as effective, so having it reduced to 5u booze removed will actually mean it's down to 2.5u removed.

Atermonera:
-----Those numbers------>



My head

Wickedtemp:
TLDR: 
Ingested medicine is half as effective so that 5 value for ethyl in the stomach is actually a 2.5 value mechanically. I suggest bumping it up to 10 or 15, so it has a mechanical value of 5 or 7.5

Atermonera:
Ok, so ethylredoxrazine's `affect_blood()` proc, which as far as I'm aware is called when it gets metabolized whether its in blood or stomach, scans M.ingested (Stomach) and M.bloodstr (Blood) for alcoholic reagents, and then removes them.
The 5 and 15 are the ratios of how much ethyl is metabolized to how much booze is removed in each case.

Since compiled code can't be modified at compile time without dark magics of assembly, those should be the same whether its in the blood or the stomach

Arokha:
@Atermonera is correct. The values will not alter the strength at all, only how long it's effective for, the way it is written there.

Arokha:
It's not half as effective exactly, it just deletes half the reagents in your stomach when it transfers them to your bloodstream, like all drugs.

Altering either the ingested or the bloodstr here won't alter that, though. No matter what you set them to, the ingested version will always be half as effective. Considering 10u ethyl injected in this removes **150** ethanol from your bloodstream, that seems fairly reasonable?

Wickedtemp:
I had tested it ingame, setting the value to 5 for ingested actually means 2.5, but oki.

Atermonera:
And according to my current understanding, injecting 10u will remove 50u from the stomach

PR #4756, opened 2018-02-07
Adjusts mining voidsuit resistances

Cyantime:
Increases radiation shielding, because large artifacts being broken create a significant amount of radiation and can't be identified beforehand without science equipment.
Reduces bomb resistance in exchange, because gibtonite doesn't exist here.

### DIFF
--- a/code/modules/clothing/spacesuits/void/station.dm
+++ b/code/modules/clothing/spacesuits/void/station.dm
@@ -79,18 +79,18 @@
 //Mining
 /obj/item/clothing/head/helmet/space/void/mining
 	name = "mining voidsuit helmet"
-	desc = "A special helmet designed for work in a hazardous, low pressure environment. Has reinforced plating."
+	desc = "A special helmet designed for work in a hazardous, low pressure environment. Has reinforced plating and minor radiation shielding."
 	icon_state = "rig0-mining"
 	item_state_slots = list(slot_r_hand_str = "mining_helm", slot_l_hand_str = "mining_helm")
-	armor = list(melee = 50, bullet = 5, laser = 20, energy = 5, bomb = 55, bio = 100, rad = 20)
+	armor = list(melee = 50, bullet = 5, laser = 20, energy = 5, bomb = 25, bio = 100, rad = 50)
 	light_overlay = "helmet_light_dual"
 
 /obj/item/clothing/suit/space/void/mining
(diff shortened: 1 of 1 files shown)


Anewbe:
I'm not entirely sold on this, for two reasons:
1.) Radiation protection doesn't actually do all that much to stop radiation. It helps a bit, but the actual change in this case won't noticeably change anything from the Miner's end. They'll still take enough ToxLoss that most of them will return anyway.
2.) Bomb resistance is generally more useful than rad resistance, because it lessens the chance you get gibbed by explosions. Plus, blast mining is actually doable, if and when the miner scores some bombs (which might be scattered in PoIs).

Cyantime:
The bomb resistance was added when a real threat to miners was digging into a tile of rock that exploded on them, now the roughly equivalent threat is digging into a tile of rock that irradiates them. The hope is that the extra resist allows them to get far enough away from the source before building up significant radiation.
If miners are blast mining they should be in control of how/when the explosions happen, and their bomb resist won't come into play.

An alternative could be adding alden-saraspova detectors to miner lockers, because they do keep complaining about blundering into artifact tiles without any protection or possible way to identify them ahead of time.

Cameron653:
Increased radiation resistance would be nice. Was playing the other day with someone hit an artifact while I was next to them. They got hit w/ enough rads that they would have died if they didn't run to medical ASAP, and so did I. They even got to the point that they started mutating (taking cloneloss).

The artifact line of code is here (https://github.com/PolarisSS13/Polaris/blob/828b5947014963850f9b614694011eb86f2056d3/code/modules/mining/mine_turfs.dm#L491)

I'd say to nerf the radiation the artifact does, because it's just flat out ridiculous.

Anewbe:
Honestly, touching that number seems more sensible to me than touching the armor does. I'll bring that up at the Staff Meeting as well.

Cameron653:
Please do. The number is absolutely insane from what I've seen. I really can't think of too many other radiation sources other than that, and the problem with the artifact radiation is that miners have 0 chance to know about it. It's just a straight up "Fuck you" that makes them have to flee the area & get treatment.

Anewbe:
We'll be adjusting the artifact/rad burst, rather than the armor values. I'll be leaving this open until the PR to lower rads is made.

Cyantime:
Alright.

PR #4757, opened 2018-02-07
Reduces armor values on the vey-med voidsuit

Cyantime:
This is the **only** voidsuit available with 0 slowdown, now it sacrifices armor values in exchange for that. Speed is the most valuable thing in ss13.


mistyLuminescence:
Fair. They're not built for combat, they're built for speed. 👍

Burrito-Justice:
Gotta go fast ❤️

Arokha:
Makes sense to meeee.

PR #4758, opened 2018-02-07
"Planetary" Exploration Airlock Controller

Leshana:
### Ports Cycling Airlocks tweak (BayStation12/Baystation12#17879)
- When cycling into a direction the airlock has already reached the target pressure for, doing so will only toggle the doors.

### Ports "Planetary" Airlock Controller (VOREStation/VOREStation#2976) (Itself a port from BayStation)
- Adds a flag to controllers that instructs the airlock program to always purge between IN/OUT and also to pump air to "outside" instead of tanks when purging to go OUT.
- Tweak the way the airlock fudges target pressures so it more reliably operates when trying to pump down to vacuum.

### Add "shuttle" sensors
- Adds an exterior airlock sensor that reads the air pressure in *front* of it, so it can be mounted on a shuttle but measure air outside the shuttle.

#### Setup for a "cycle_to_external_air" shuttle airlock:
- `X_pump_out_external` - A vent pump on the "outside" of the shuttle, connected by pipes to the pump_out_internal vent.
- `X_pump_out_internal` - A vent pump inside the airlock chamber, connected by pipes to the pump_out_external vent.
- `X_pump` - A vent pump inside the airlock chamber, connected by pipes to air canisters.
- `X_exterior_sensor` - Sensor on the "outside" of the shuttle.  Must set this tag manually on the controller and sensor, its not auto-populated onto controller.
#### Cycling to "outside"
1. Purge the chamber by draining thru X_pump into the canisters.
2. Fill the chamber by siphoning planetary air via X_pump_out_external thru the pipe and out X_pump_out_internal.
#### Cycling to "inside"
1. Purge the chamber by draining thru X_pump_out_internal thru the pipe and out X_pump_out_external
2. Fill the chamber from the canisters thru X_pump

### DIFF
--- a/code/game/machinery/doors/airlock_control.dm
+++ b/code/game/machinery/doors/airlock_control.dm
@@ -218,6 +218,13 @@ obj/machinery/airlock_sensor/airlock_interior
 obj/machinery/airlock_sensor/airlock_exterior
 	command = "cycle_exterior"
 
+// Return the air from the turf in "front" of us (Used in shuttles, so it can be in the shuttle area but sense outside it)
+obj/machinery/airlock_sensor/airlock_exterior/shuttle/return_air()
+	var/turf/T = get_step(src, dir)
+	if(isnull(T))
+		return ..()
+	return T.return_air()
+
 obj/machinery/access_button
 	icon = 'icons/obj/airlock_machines.dmi'
 	icon_state = "access_button_standby"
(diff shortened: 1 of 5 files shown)


Anewbe:
Will this require further mapping changes?

Leshana:
@Anewbe Only where you actually want to use it.   Its backwards compatible so all existing airlocks will work fine.
If you want to use it on an airlock, you set `cycle_to_external_air = 1` on its controller, and setup additional pumps with tags as described above.

PR #4763, opened 2018-02-08
Lowercases the Giant Crab name

Anewbe:
They should have a `the` in front, so the name should be lowercase.


FenodyreeAv:
Disrespect to such a Majestic and Noble creature.

PR #4767, opened 2018-02-09
removes redundant override in human emp_act()

Screemonster:
Removed:
```
/mob/living/carbon/human/emp_act(severity)
    for(var/obj/O in src)
        if(!O)    continue
        O.emp_act(severity)
    ..()
```

Because it calls the parent, which is:
```
/mob/living/emp_act(severity)
    var/list/L = src.get_contents()
    for(var/obj/O in L)
        O.emp_act(severity)
    ..()
```

Fixes bizarre behaviour where calling emp_act() on a human mob would cause it to run emp_act() on all its contents twice.

May affect balance for FBPs if people have been coding around this bug. If this is the case, tweak the numbers in /obj/item/organ/emp_act().

### DIFF
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -197,12 +197,6 @@ emp_act
 		if(.) return
 	return 0
 
-/mob/living/carbon/human/emp_act(severity)
-	for(var/obj/O in src)
-		if(!O)	continue
-		O.emp_act(severity)
-	..()
-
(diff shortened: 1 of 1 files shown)


Neerti:
I swear I fixed this a year or so ago.
In fact I did (https://github.com/PolarisSS13/Polaris/pull/898), seems to have come back somehow.
Hopefully this bug stays gone this time.

Arokha:
It must have been reverted, because if I look in blame, it doesn't even have your change. It only has one from 3 years ago.

Screemonster:
Looks like it snuck back in as part of #1785.

PR #4771, opened 2018-02-09
Clothing adjustments and fix

Woodratt:
- Utility uniforms can now roll up sleeves
- SAR, Shaft Miner, Explorer can now select webbings in the loadout.

### DIFF
--- a/code/modules/clothing/under/solgov.dm
+++ b/code/modules/clothing/under/solgov.dm
@@ -36,6 +36,7 @@
 	desc = "A comfortable turtleneck and black utility trousers."
 	icon_state = "blackutility"
 	worn_state = "blackutility"
+	rolled_sleeves = 0
 	armor = list(melee = 0, bullet = 0, laser = 0,energy = 0, bomb = 0, bio = 0, rad = 0)
 	siemens_coefficient = 0.9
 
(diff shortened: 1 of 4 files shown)


code/modules/clothing/under/solgov.dm
```diff
@@ -36,6 +36,7 @@
 	desc = "A comfortable turtleneck and black utility trousers."
 	icon_state = "blackutility"
 	worn_state = "blackutility"
+	rolled_sleeves = 0
```
Anewbe: Are there sprites for this?

Woodratt: Base black one no. But if it cannot find them it simply states that it can't be rolled. There are sprites for all the others.

Anewbe: It appears that the base black one does in fact have a sprite, so disregard my first comment.

mistyLuminescence:
Thoughts on adding holsters in the loadout selection for explorers (to fit the new phaser guns in)? Not sure if in scope or not.

Woodratt:
Outside of scope, simply because I was not informed that was a request, and it may be one that'll logjam this due to concerns.

PR #4772, opened 2018-02-10
GPS coordinates and POI names are obfuscated

Atermonera:
Tested:
t

The Z level of the current location has also been changed to report z level name instead of number.
POIs with mapped handheld GPS units need to be off by default. If you want explorers to be able to find it with a gps, use `/obj/item/device/gps/internal/poi`

`/obj/item/device/gps/syndie` maintains the current function of GPS units, in addition to their other effects, although they'll also list Z names instead of numbers as well.

Adds a helper proc on datum/map to get the name of a Z level from its number.
Changes the names of planetside Z levels to more accurately reflect what region they contain.

### DIFF
--- a/code/game/objects/items/devices/gps.dm
+++ b/code/game/objects/items/devices/gps.dm
@@ -80,7 +80,7 @@ var/list/GPS_list = list()
 
 	var/turf/curr = get_turf(src)
 	var/area/my_area = get_area(src)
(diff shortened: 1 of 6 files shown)


Nalarac:
The Z's on the station would show Deck 1, 2, and 3 respectively then correct?

Atermonera:
Yes

PR #4777, opened 2018-02-10
Typo, oversight in weather datums

Atermonera:
It's spelled `failure`.

Web editor is suited for fixing stuff like this, but as can be seen in the commit log, takes at least 1 commit for every change to every file, so squash merge is a must.


Atermonera:
Also fixes the weather that acts upon mobs, so that instead of returning from the first mob to fail the condition, it'll continue through the entire iteration

Anewbe:
I have mild worries about the bugfix part of this making lag, but if it happens, can be corrected.

PR #4778, opened 2018-02-10
Adjusts certain simple mobs, adds some things

Anewbe:
1.) Renames (and repaths) the Samak to Savik, as per Lore Staff Discussion.
2.) _Savik_ now hit harder, and go berserk, as per their role as apex predators.
3.) Diyaabs are no longer innately hostile, but rather retaliate (as a pack) when attacked. Their path has been changed, because it seemed like the easiest way to get the correct variables.
4.) Ranged Mercenary mobs no longer drop bullets, so we don't end up with a million casings on a single tile.
5.) Moved the Savik, Diyaab, and Shantak mobs to a new folder, and individual files. Moved the Tindalos and Yithian to a new file.


Atermonera:
A little wary of the space in the directory name, some shells can be really fucky when trying to navigate through those. Since byond is mostly coded in DM it shouldn't be quite as much of a problem

PR #4779, opened 2018-02-10
[WIP] Planetary Lighting Optimization Attempt

Neerti:
Changes how lighting is applied to the Surface.

The current method works by making every tile glow in a specific color and intensity. The issue is that when it is time to change the color/intensity, it can take the light controller a considerable amount of time for it to update every tile (estimated to be around 14,000 tiles), which can take a minute or two for the lighting controller to finish.

New method is to have a special light source for the surface, with different properties to the 'regular' light source that most things use for making light. The special version doesn't care about line of sight or lighting falloff. This special light source is applied to a series of invisible 'sun' objects that get generated as the planetary controller gets created. When the lights need to change, the lighting controller has to change a much smaller group of objects instead of literally every outdoor tile.

The new lighting code is still ugly and full of copypasta, which I would normally clean up but its difficult to do that this time due to lighting making heavy use of compiler macros, which limits my ability to apply OOP to them, and might adversely affect performance for normal lights. I'm also told by a few other people who know lighting better than me (I barely know what I'm doing) that further gains could be made, so I'm placing this here to see what smarter people than me might have to say.

Performance wise, this appears to make lighting updates on the map take around 20 seconds, which unfortunately still makes lights not work until its finished, but at least the wait is much shorter, which now lasts for around a minute or a minute and a half if memory serves.


code/modules/planet/planet.dm
```diff
@@ -50,3 +53,17 @@
 	sun["color"] = new_color
 	needs_work |= PLANET_PROCESS_SUN
 
+/datum/planet/proc/build_suns()
+	var/desired_suns_per_row = 15
```
Anewbe: Might want to give a comment on how to calculate this number, for maps that aren't the same size as ours?

Neerti: Sadly I don't have a way to determine an ideal number. I got to 15 with trial and error and I'm not sure if its worth it to go even higher or not, however the code will adjust for different map sizes already.

Anewbe: Roger.

Neerti:
It's been two months and I was hoping people smarter than me and who knew more about how lighting works would have a better way of accomplishing this, since I'm still not very proud of this code, but the 'lighting stops for two minutes' thing still annoys me.

PR #4783, opened 2018-02-10
Fixes scanning bed in medical.

zaredman:
A constructive test to make sure my map merger & git is fixed by fixing the scanning bed in medical. Works on a localhost map.

I tried orienting it vertical but the opposite way but for some reason the scanner terminal always resets to a left-right orientation even though the DIR var is set to a up-down orientation.


zaredman:
Other one gets the job done by fixing the code issue outright.

PR #4789, opened 2018-02-11
Make ghosts invisible again.

Leshana:
That way you won't be able to see them even with plane anomaly edge cases.


Anewbe:
Merging early with staff permission.

PR #4790, opened 2018-02-11
Lets poi gps units display gps_tag instead of hardcoded string

Atermonera:
Because devstaff are all donks and anewbe merged the first one as I was making this commit, because this came up in discussion.

When mapping, signals should not have specific names, and ideally there'd be several that share a name. Having 15 pois with "Unidentified Signal" is a lot more explorative than having "Sif free radio, crashed ufo, blackshuttledown, doomp, rockybase, shaqden"

Also I don't know where those other changed lines are coming from.


Anewbe:
Merging early with staff permission.

PR #4791, opened 2018-02-11
Makes Tape work on Nerds

Mechoid:
Mouth-taping no longer checks if the target has glasses. For some reason.

### DIFF
--- a/code/game/objects/items/weapons/tape.dm
+++ b/code/game/objects/items/weapons/tape.dm
@@ -87,7 +87,7 @@
 				if(!can_place)
 					return
 
-				if(!H || !src || !H.organs_by_name[BP_HEAD] || !H.has_eyes() || H.glasses || (H.head && (H.head.body_parts_covered & FACE)))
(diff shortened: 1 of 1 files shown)


code/game/objects/items/weapons/tape.dm
```diff
@@ -87,7 +87,7 @@
 				if(!can_place)
 					return
 
-				if(!H || !src || !H.organs_by_name[BP_HEAD] || !H.has_eyes() || H.glasses || (H.head && (H.head.body_parts_covered & FACE)))
+				if(!H || !src || !H.organs_by_name[BP_HEAD] || !H.has_eyes() || (H.head && (H.head.body_parts_covered & FACE)))
```
Arokha: But you still can't tape their mouth if they don't have eyes? >.>;

I get the feeling this if() was copied from blindfold taping.

Mechoid: It was and I didn't notice that cause I was looking for 'glasses'

PR #4792, opened 2018-02-11
Sleeper and Advanced Scanner consoles now find their machines

Anewbe:
Indentation errors are evil.
Also made it so the consoles face their paired machines, for _aesthetics_.

Might be preferable over #4783


zaredman:
GREAT DEBATE 2018

I thought the other way looked nicer. That and most doctors approach it from the east more than west

Anewbe:
We can do both, but this is more long-term.

zaredman:
Oh if this fixes the issue just do yours then. Vertical is nice so you can run through the doors straight south.

Issue #4780, opened 2018-02-10
Free Sif Radio pOI has the wrong ammo type as loot


FenodyreeAv:
#### Brief description of the issue
In the bottom room of the POI there is a rifle that fires 7.62, which has a spare ammo clip of 5.45 on top of it.

#### What you expected to happen
To be able to load the spare ammo into the gun.

#### What actually happened
Fight a spider and realize my bullets don't fit. Die.

#### Steps to reproduce
Go to Sif Free Radio outpost, try to load gun, fail.

mistyLuminescence:
aaaaaa sorry

PR #4795, opened 2018-02-11
Fixes ammo spawn at DJOutpost1.dmm (take two)

zaredman:
If this fails again, I'mma require all sorts of assistance considering it's literally a three-character change in the map file. See #4780 as the issue.

Edit: Failed check. REEEEEEEEE.
Edit2: I don't know if the old-school map merger likes this type of map file? Let's see if this works.


zaredman:
So this passed all checks. Noticing that for some reason the letters on this go from A through Q when the original was A through P. Looks like the addition to the letters was:
```
"C" = (/obj/machinery/door/airlock/glass,/obj/item/device/gps/internal/poi,/turf/simulated/floor/wood,/area/submap/DJOutpost1)
```

Going to dick around a bit more on my side to figure out what the cause of that was.

PR #4797, opened 2018-02-11
Updates changelog

Atermonera:
Oh, and I guess as a note, the command to actually run the script once you've got python installed and the libraries/other steps done is `py ss13_genchangelog.py ../../html/changelog.html ../../html/changelogs`


Anewbe:
Merging before the update.

PR #4799, opened 2018-02-11
Fixes pAI HUDs not working at all.

Verkister:
-For some whatever reason the things had absolutely no function other than toggling a single pAI specific var that had no code whatsoever connected to it.


Arokha:
This appears to make it so there's no way to turn them OFF after. There's only code that sets the vis to TRUE.

PR #4801, opened 2018-02-12
Makes More Machines Movable

PrismaticGynoid:
Adds in the ability to wrench/unwrench a number of machines that previously could not be. Now you're able to move:

- Laptop venders!
- Seed extractors!
- Biogenerators!
- Drying racks!
- Microwaves!
- Cooking machines!
- Gibbers!
- Chemmasters/Condimasters!
- Bioprinters!
- Virology machines!
- Washing machines!

Tested with no apparent problems.

### DIFF
--- a/code/game/machinery/biogenerator.dm
+++ b/code/game/machinery/biogenerator.dm
@@ -46,6 +46,8 @@
 		return
 	if(default_part_replacement(user, O))
 		return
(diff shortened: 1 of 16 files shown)


Arokha:
Yaaaay. No more needing to varedit when players want to rebuild an area. Or. Less, anyway.

PR #4802, opened 2018-02-12
Generalizes stack usage for machine frames.

Cyantime:


Anewbe:
Tested?

Cyantime:
Tested by using plasteel, all stacks should behave the same.

PR #4807, opened 2018-02-12
Tweaks Lead & Radiation Resistance

Neerti:
Changes a few things to the newest material added, Lead, as well as the radiation system that got adjusted when lead was introduced. The aim is to make radiation be more or less what it was before the change, but without making lead useless.

Changes:
- The ```radiation_resistance``` var on material has been reworked to act as an additive bonus to a material's ability to block radiation instead of the only var being checked. This was done to reduce duplicate numbers since usually the var equaled the weight.
- Walls and other things blocking radiation now check a material's ```weight``` once more, however the material's ```radiation_resistance``` is added on top of that, so a material's ability to block radiation is now determined by the sum of the two variables.
- Most instances of ```radiation_resistance``` have been removed from materials because with the above two changes, they were redundant and equaled weight in most cases. Lead is now the only material currently that has additional resistance. This means the engine room should no longer be rad-free with a basic phoron setup and there is a reason to use lead when considering more ambitious engine setups.
- Lead's weight has been reduced so that it isn't the best thing ever for baseball bats.

Bonus:
- Adds Lead Wall and R-Wall types for mappers and badmins.


Andromeda-K22:
The only reason i had for just using radiation resistance was because i was worried about the ghetto weapon squad going nuts with lead baseball bats. 

Then again, the idea of a lead baseball bat is quite amusing.

zaredman:
Would this require the walls around the SM engine room to be changed to lead walls? Or would what we have in place now work just fine?

Neerti:
This returns it to the status quo. Nitrogen is still safe, and Phoron leaks radiation to the monitoring room unless engineering puts more walls up.

Magneld:
Where does one get Lead

PR #4810, opened 2018-02-12
Shapeshifters can properly wear the clothing of other species.

Mechoid:
On the tin, fixes shapeshifters (Prometheans) ability to wear their current outward species appearance's clothing. Tajaran appearance - > Tajaran voidsuits, uniforms, so on.


Anewbe:
Tested?

Mechoid:
Tested indeed.

PR #4805, opened 2018-02-12
Change to first-aid post system

zaredman:
Suggested this be listed as do-not-merge pending staff meeting discussion since I can imagine this ruffling feathers on top of verification that I somehow didn't fuck this one up too. Time will tell. Still ironing out bugs on my end with my git setup. So far results look promising.

The purpose of this request is to try to coax crew from exceeding their general level of knowledge for anything medical (eg: knowing how to use a paddle-style defib and sleeper thinking it is at a 'first aid' level). That, and the Big Boss Medical is very close to just about everything on this new map.

If someone wants to be awesome and add AED-style defibs, that would be sweet. As those don't exist and my only real familiarity (may be a stretch) is with mapping, doing this instead. **This is only for the station level** and does (or should do) the following:

- Removes all first-aid stations with the exception of the one by the bar.

- Re-names the door of the one by the bar to "Triage Station," changes access requirements to those that already have medical access, and adds a spawn for an advanced medkit and a random medkit.

- Places numerous "first aid" wall lockers around the station containing a stretcher, stasis bag, first-aid medkit, and pill bottle of spaceacillin.

- Places numerous "defibrillator" wall lockers around the station, typically beside the above "first aid" wall lockers. These are locked to anyone with medical access.

- Moves pilot gear-up a bit up to where the old first-aid station was.

- Expands maint and re-zones some areas where appropriate.

PR #4815, opened 2018-02-14
Change to first-aid post system (v2)

zaredman:
Had to delete my old repository after a tortoisegit/git incompatability royally porked my master branch.  Still trying to get the hang of the confusing beast that is Git. This is a repost of https://github.com/PolarisSS13/Polaris/pull/4805 with one alteration.

**Please see the comments on #4805 in addition to any that may be posted here.**

Suggested this be listed as do-not-merge pending staff meeting discussion since I can imagine this ruffling feathers on top of verification that I somehow didn't fuck this one up too. Time will tell. Still ironing out bugs on my end with my git setup. So far results look promising.

The purpose of this request is to try to coax crew from exceeding their general level of knowledge for anything medical (eg: knowing how to use a paddle-style defib and sleeper thinking it is at a 'first aid' level). That, and the Big Boss Medical is very close to just about everything on this new map.

If someone wants to be awesome and add AED-style defibs, that would be sweet. As those don't exist and my only real familiarity (may be a stretch) is with mapping, doing this instead. **This is only for the station level** and does (or should do) the following:

- Removes all first-aid stations with the exception of the one by the bar and the one on the center of z1.

- Re-names the door of the one by the bar and z1 to "Triage Station," changes access requirements to those that already have medical access, and adds a spawn for an advanced medkit and a random medkit. Pill locker is placed in the z1 booth.

- Places numerous "first aid" wall lockers around the station containing a stretcher, stasis bag, first-aid medkit, and pill bottle of spaceacillin.

- Places numerous "defibrillator" wall lockers around the station, typically beside the above "first aid" wall lockers. These are locked to anyone with medical access.

- Expands maint and re-zones some areas where appropriate.

- Unlike #4805 this one does NOT make any adjustments to the pilot nook.

Example setup:
example setup


mistyLuminescence:
Possible idea - why not keep the first-aid stations as they are now, but move the wall locker to just outside (all-access), and lock the first-aid station door itself to Medical access, keeping the defib and sleeper inside?

zaredman:
Personally, I feel there are a few too many of them. Unless there's FUBAR breaching, a sleeper-equipped room is typically within, what, 20-30 tiles of any point on the station? A few I ave been chatting with feel it may be too much in regards to balance in addition to the current virtually-all-access and what-really-is-first-aid conundrum.

Odds are this will be hashed out sometime around the upcoming staff meeting. If the majority feels that leaving them intact/changing access req/but still adding lockers is the desired route then this'll be adjusted accordingly.

Andromeda-K22:
As far as i'm aware:
There's one on Deck 1. One, which is already partially medical access. 

There's one by most departments. One for the civ sector, one for engineering, and one for security. I see nothing wrong with this. Deck three lacks a first-aid station. There are in total, four first aid stations, the same as with the Northern Star. I see no issue with this. 

Regardless, this change will just add extra hoops to make it very slightly more difficult to save someone's life in a catastrophe situation. Like Meteors.

Let's not forget that meteors will breach any and all external portions of deck 2. Through the ceiling, probably. There are no consistent ways to access medical via maint, either. So accessing medical may not always be an option. Medical is also not always around and or able to respond, particularly with the numbers we observe during lowpop. It's rare to have **one** doctor. 

This won't fix people exceeding their general knowlege level. It will make it worse. Removing easy access to the defibs (which act like AEDs, and not ACTUAL manual defibrillators, just with paddles), will just result in more people justifying using say, something incredibly complex, like the cloner. 

So, -1 from me on this PR. Either add AEDs first and put them into each department, or leave the first-aid stations as they are.

Wickedtemp:
I'm still in favor of leaving the stations alone for several reasons posted before. The main one being:

- I don't like having to stabilize patients on the floor in the hallway before shuttling them to Medbay. The aid stations give me an alternative. I'd very much like to *keep* that alternative. The gear in them is almost irrelevant to me, personally. I don't care if we get gear lockers or not, I usually have a defib/roller and a stasis bag onhand anyway when it's lowpop and people get hurt. I never used the Northern Star aid stations because their placement seemed odd and usually it'd entail running out of the way of Medbay. These placements are so much better, they're actually being *used* now. I finally have a spot to take patients who either: A) Need to be stabilized a *lot* before transit to Medical. or B) Actually aren't that hurt and I could stabilize and heal on-site- that isn't the *floor in whatever hall I found them in.*

- The sleepers are useful as well if there isn't a Chemist and an EMT or Doctor still has to stabilize someone. They're one of the few 'baby-proofed' machines for Medical where, no matter what you do, unless you literally drain their blood for no reason, you can't fuck it up and hurt the patient. With that said, their actual usage for *healing* rather than *buying the patient time* is negligible. If someone has rapidly rising damage, the sleeper isn't going to *heal* them with only dylo, inaprov, tricord, dexalin and paracetamol. The best you could do is slow the damage and put them into stasis. If the sleeper happens to be enough to heal someone completely, then mechanically, they weren't that injured to start with and likely had injuries that a properly geared doctor could heal with a single syringe and a trauma or burn kit. Also, it's worth noting that if one person gets hurt, they can't heal themselves with the sleeper. They need a buddy to use the sleeper console, so you aren't just going to be able to break into a first aid station and stabilize yourself using the sleeper without any assistance. 

- Also as a few others said. Meteors pretty much swiss-cheese the station now, having the aid stations is a god-send.

- Already covered the 'risk of death' bit. TLDR risk of death remains roughly the same regardless, if a small room with a sleeper is what kept you from dying, you can be stabilized without one with ease by a prepped doctor/emt/etc, who would likely be the one with you if you're in the sleeper anyway. Their primary use would be the stasis levels. 

- The lockers of equipment are irrelevant to me anyway, since we already have that. It doesn't matter if more lockers are added in place of the aid station rooms because we already have enough gear. It isn't really a tradeoff. The rooms themselves are whats useful about the aid stations, in my opinion. 


If you want to discourage non-med from using the first-aid station sleepers, put them behind med access with all the other gear, and have the basic first aid stuff in a locker on the outside, instead of removing the first aid station. 

I'd be in favor of removing the sleepers if Staff disagrees with me and decides they really are *that* much of a problem, if the rooms themselves remain. 

*I just really don't want to go back to stabilizing a patient in the middle of the hallway again all because "Too many sleepers."*

PrismaticGynoid:
I'm fine with removing the sleepers and locking up the defibs, but the first aid stations should still be separate rooms, for the reasons that have been mentioned.

Anewbe:
https://www.dropbox.com/s/crdex5d029ga3xd/Screenshot%202018-02-18%2016.58.06.png?dl=0
Discussed in the Staff Meeting. There were 4 or 5 abstains, but it otherwise leaned affirmative.

zaredman:
For tracking purposes, this git was linked during the staff meeting for all to review. All points brought up were discussed. Abstains tended to abstain due to [Neutral Response](https://www.youtube.com/watch?v=ussCHoQttyQ) and conclusion was to merge as-is.

PR #4816, opened 2018-02-14
Adds hidden alarms, for air alarms, fire alarms, and APCs

Cyantime:
Accidentally copied part of #4766, but in a different way.
The alarms all still trigger, air/fire alarms still shut firedoors in the area, but they aren't displayed on consoles or to the silicons.

Ideally keep an area to either only unhidden or hidden alarm types, things can get sort of weird when mixed.

### DIFF
--- a/code/modules/power/apc.dm
+++ b/code/modules/power/apc.dm
@@ -56,8 +56,8 @@
 /obj/machinery/power/apc/hyper
 	cell_type = /obj/item/weapon/cell/hyper
 
-/obj/machinery/power/apc/hidden
-	hidden = TRUE
+/obj/machinery/power/apc/alarms_hidden
+	alarms_hidden = TRUE
 
 /obj/machinery/power/apc
 	name = "area power controller"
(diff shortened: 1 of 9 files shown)


code/modules/power/apc.dm
```diff
@@ -107,12 +110,12 @@
 	var/failure_timer = 0
 	var/force_update = 0
 	var/updating_icon = 0
-	var/secret = FALSE // If true, it won't show up on the alert computer.
 	var/global/list/status_overlays_lock
 	var/global/list/status_overlays_charging
 	var/global/list/status_overlays_equipment
 	var/global/list/status_overlays_lighting
 	var/global/list/status_overlays_environ
+	var/hidden = 0 //If power alarms from this APC are visible on consoles
```
Anewbe: Nitpick: I'd rather this be FALSE, if at all possible.

MisterLayne: Are we switching from a 0/1 system to a FALSE/TRUE system, finally?

Leshana:
The variable name `hidden` on APCs and air alarms etc  could be confusing, it might easily be confused with having to do with the visibility of the machine itself.  I suggest changing that variable to something more explicit like `alarms_hidden`  or `secret`  or some such.

PR #4818, opened 2018-02-14
Ports Vorestation's emp effects and ion projectiles

Anewbe:
Sort-of-port of https://github.com/VOREStation/VOREStation/pull/3004, but I left in the organ damage because I feel it's an important part of what makes EMP actually bother FBPs.
- EMP will mess with robotic eyes.
- EMP makes robots a bit hungrier.
- EMP gives robots a bit of system instability.
- The way Ion projectiles pulse has been made more flexible.

### DIFF
--- a/code/modules/organs/internal/eyes.dm
+++ b/code/modules/organs/internal/eyes.dm
@@ -93,3 +93,7 @@
 
 /obj/item/organ/internal/eyes/proc/additional_flash_effects(var/intensity)
 	return -1
+
+/obj/item/organ/internal/eyes/emp_act(severity)
+	..()	//Returns if the organ isn't robotic
+	owner.eye_blurry += (4/severity)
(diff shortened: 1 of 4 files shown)


code/modules/organs/internal/eyes.dm
```diff
@@ -93,3 +93,7 @@
 
 /obj/item/organ/internal/eyes/proc/additional_flash_effects(var/intensity)
 	return -1
+
+/obj/item/organ/internal/eyes/emp_act(severity)
+	..()	//Returns if the organ isn't robotic
```
Leshana: No it doesn't.

Leshana: There literally is no return statement here.   The **parent** will return if its not robotic, but the parent returning does not cause this proc to end.  The following line applying eye_blurry will execute whether the organ is robotic or not.

Anewbe: Ah, I see where I goofed. Fix will be up soonish.

PR #4820, opened 2018-02-15
Suit coolers don't magically shut off in mild fires

### LINKED ISSUE #4250: PPE does not protect FBPs from overheating
Wearing relevant protective equipment (firesuits, voidsuits, hardsuits) does not enable FBPs to safely exist in overheated environments the same way it does for organics.

A suit with an active cooler to confer at least part of its usual thermal protection to a synthetic wearer.

#### What actually happened
No appreciable change in maximum safe temperature.

#### Steps to reproduce
- Get an organic and an FBP in a room together.
- Put them in identical voidsuits.
- Start a thermal regulator set to 90C.
- Watch as the FBP cooks alive.

#### Additional info:
 - **Server Revision**: https://github.com/VOREStation/VOREStation/commit/a5844f786ef2a09948e6718a7d418b05d95f9324

Anewbe:
- Kills the bug that makes suit coolers do literally nothing in a fire. They now check if your suit is enough to protect you before doing literally nothing. (Basically, atmos suits are useful to FBPs again/now.) Fixes #4250
- Changes a few checks that were improperly returning `var < 1` as ~~false~~ true when the var was equal to 1. I don't know what's going on there.

### DIFF
--- a/code/game/objects/items/devices/suit_cooling.dm
+++ b/code/game/objects/items/devices/suit_cooling.dm
@@ -46,9 +46,15 @@
 
 	var/mob/living/carbon/human/H = loc
 
-	var/efficiency = 1 - H.get_pressure_weakness()		//you need to have a good seal for effective cooling
-	var/env_temp = get_environment_temperature()		//wont save you from a fire
-	var/temp_adj = min(H.bodytemperature - max(thermostat, env_temp), max_cooling)
+	var/efficiency = 1 - H.get_pressure_weakness()			// You need to have a good seal for effective cooling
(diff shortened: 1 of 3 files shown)


Atermonera:
`x < 1` will only return true if the value is exclusively less than one.
`x <= 1` will return true if the value is inclusively less than one, hence less than or equal to

Cyantime:
Probably fixes #4250

Anewbe:
Disregard the direction/falsity of the symbol in the first post, which will soon include that Issue Report, I know what I meant to say, and fixed what was being bugged.
This is tested and working..

PR #4822, opened 2018-02-15
Player-controlled SA QoL

Arokha:
Gives player-controlled simple_animals a normal HUD, with optional hands (`has_hands` var), the normal walk/run thing, intents, pull, pulltoggle, etc.

Provides a generic proc to attach additional huds should some animals need some sort of ... bear hud. Or whatever.

Lots of changes are really code cleanup. Lots of things defined on 'carbon' had no reason to be on carbon and not on living (all the vars they used were defined on living, probably just defined on carbon due to poor choices or just 'that's how it used to be')


Anewbe:
polaris.dme:1851:error: unable to open "code/modules/mob/living/simple_animal/simple_animal_vr.dm".

code/_onclick/hud/screen_objects.dm
```diff
@@ -486,7 +488,7 @@
 				var/mob/living/silicon/ai/AI = usr
 				AI.view_images()
 		else
-			return 0
+			return attempt_vr(src,"Click_vr",list(location,control,params)) //VOREStation Add - Additional things.
```
Anewbe: Is this supposed to be here?

code/modules/mob/living/simple_animal/simple_animal.dm
```diff
@@ -496,26 +560,13 @@
 	custom_emote(2, act_desc)
 
 /mob/living/simple_animal/bullet_act(var/obj/item/projectile/Proj)
-	if(!Proj)
-		return
 	ai_log("bullet_act() I was shot by: [Proj.firer]",2)
 
-	if(Proj.taser_effect)
```
Anewbe: Were these moved elsewhere, or just scrapped?

code/_onclick/hud/screen_objects.dm
Arokha: Nnnnnope. I'm just going to fix this and that thing I did in that accidental .dme thing and forcepush over this branch. Be riiiight back.

code/modules/mob/living/simple_animal/simple_animal.dm
Arokha: Removed. It was redundant, as stun_effect_act, also defined on simple_animal, gets called when they get hit by a stun projectile anyway. No need to call it here instead of letting the parent take care of it. Only thing we care about is that we got shot, and we might do the react_to_attack thing to respond appropriately.

Anewbe: Right you are.

What about the section just below, with SA_bonus_damage ? Because that seems to be the only spot where that damage is added, and it's sort of important to our whole planetside deal.

Anewbe: I handwave it slightly differently every time, because we abstract away a lot of "reality" issues for the sake of gameplay or story. Today my excuse is tachyons.

And I'm not sure it was ever a concern before now, but yes, it should probably respect whatever armor value the projectile checks for.

Arokha: I'm a little concerned about how this was originally coded. It would do something upwards of 200%+ damage. Even if you had SA_bonus_damage = 0.

Since it does `true_damage = Proj.damage` which, we'll say is 30 for fun. Then `true_damage += Proj.SA_bonus_damage` which could be as low as 0. Then adds `true_damage` bruteloss. ... then calls `..()` which will have them be shot again, using the normal mob/living bullet_act, I'd think?

Anewbe: It seems to be working fine in the live code, as far as I can tell? And there's no specific ..() call in the base code, unless on_hit specifically calls it. I can check and test tomorrow, if this isn't resolved by then.

Arokha: It might 'work fine' but you'd probably doing more damage than you intended:
`simple_animal/bullet_act`: takes into account SA_bonus_damage, does Proj.damage in addition to it, calls ..()
`living/bullet_act`: also does Proj.damage, again, even though you already did it in the first one.

What I'm saying is, with your previous code, you end up with:
* Normal bullet (30 damage), hits Simple Animal, does 30 damage
* SA-strong bullet (30 damage+SA_bonus_damage), hits Simple Animal, does 65 damage
* SA-strong bullet (using the SA_vulnerability var), hits Simple Animal, does 60 damage

Is that what was intended? It seems 'unexpected' that adding something like only 5 SA_bonus_damage to a projectile actually increases the damage of a 30-damage projectile by 216%

Arokha:
Travis pls.

code/modules/mob/living/simple_animal/simple_animal.dm
Neerti: In that examine the SA-strong bullet should be doing 35 damage, so it doing 65 is not intentional.

Arokha: Actually disregard, I am dumb. It is not intentional but that's also not what it was doing. I was looking at the diff wrong. It didn't call parent previously.

PR #4823, opened 2018-02-15
Fix the seedarkness variable

Arokha:
Now essentially works as a "permanent darksight adjustment" variable, removing the slow change.

Fixes @Anewbe's bloodlust.


Anewbe:
Fix is a strong word. It rests ~~for now~~.

PR #4825, opened 2018-02-15
Extend mob 'glow' capabilities

Arokha:
Now you can alter the other set_light variables by setting them on the mob (yay antiphoton effects that follow you)


Andromeda-K22:
I would rather have lag induced by moving mob light sources fixed first, tbh.

Arokha:
I have no idea how to do that, and this is unrelated.

PR #4827, opened 2018-02-15
Port Bay-style attack animations

Arokha:
Does just that. Also adds a few useful helpers that I'm sure @Neerti will use.

Simple animals who claw at people just have a generic slashy attack effect.

image


Cyantime:
And the preference setting to disable these? Because they're rather ugly and the mob movement combined with the inhand sprite is enough to get across the same information.

Arokha:
Bay does not have one. Paradise said they were going to make one, but I'm unsure if they ever did. /tg/ and /vg/ I'm not sure. I suppose I could create one on my own.

Anewbe:
How much slower is it?

Arokha:
Not a ton. I mean instead of finding every client it needs to show this to, it has to find them and also check to see if they have that preference enabled. So whatever cost is_preference_enabled has. I'm just sort of obsessive about efficiency.

Anewbe:
*shrug

PR #4828, opened 2018-02-15
Allow Simple Animals to have armor in a normal sensible way

Arokha:
Replaces their 'resistances' list with a normal 'armor' list. Works exactly the same as armor does on every other item in the game, which is why it's a Good Idea:tm:

### DIFF
--- a/code/modules/mob/living/simple_animal/simple_animal.dm
+++ b/code/modules/mob/living/simple_animal/simple_animal.dm
@@ -125,14 +125,14 @@
 
 	//Damage resistances
 	var/resistance = 0				// Damage reduction for all types
-	var/list/resistances = list(
-								HALLOSS = 0,
-								BRUTE = 1,
-								BURN = 1,
-								TOX = 1,
-								OXY = 0,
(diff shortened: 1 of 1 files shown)


Anewbe:
Any chance I can bug you to implement the armor soak bit on them, too?

Arokha:
I dunno what armor soak is. Explain?

They do have a var that is a flat damage reduction var, if that's what you want. I think it's called `resistance`

Anewbe:
Soak is an armor mechanic that is checked just before normal armor, that reduces the incoming damage by a flat amount, capped at I think 80% and affected by armor penetration.

I'm not sure `resistance` actually works, because I've never been able to get it to function in testing, though I might be doing it wrong.

Arokha:
To be honest, it may not have worked before. I don't know. It should work now, as it's passed whenever we're passing the armor check thing and the armor check thing should take care of it.

You can see it being passed as the `blocked` parameter when we're doing armor checks, so it should function as a flat soak.

Anewbe:
Works for me.

PR #4829, opened 2018-02-15
Lets animals and ghosts look in boxes/backpacks/whatever

Arokha:
In the title! @Cyantime suggested because he seems like he's an OCD sort of fellow.
Just drag them onto yourself like normal. Neither will be able to do anything since they ain't got haaaaands. They can look inside though.

(Well, animals don't have hands until you merge that QoL~)


Nalarac:
Would it be too much to ask that borgs can look in bags as well?

Arokha:
> Would it be too much to ask that borgs can look in bags as well?

Done

PR #4782, opened 2018-02-10
 Handy borg sprites additions and fixes 

Heroman3003:
Added Mr. Handy borg sprites for Janitor, Clerical and Service (regular and hydro versions) modules. Fixed eyes overlay of engineering module not being there. Added medical handy to Crisis module sprite list (they share almost all the sprites except Handy anyway).

https://image.prntscr.com/image/P5cE2xR_SLq0lVWpLBsx4w.png

From left to right:
Engineering (added stripes for engineering aesthetic, fixed eyes overlay)
Surgeon (unchanged)
Crisis (new, same sprite as surgeon)
Janitor (new)
Service (new, hydro version)
Service (new, service version)
Clerical (new)
Research (unchanged)

PR #4831, opened 2018-02-15
Fix misnamed iconstate for engineering mrhandy

Arokha:
Two sets of eyes were named for the science version, leaving the engineering one without an eye overlay.


Cyantime:
This is also done by #4782

Arokha:
True. Merge whichever, or neither. I won't be offended.

Heroman3003:
I wanted to add, that Engineering version's eyes sprite (orange eyes) is actually noticably misaligned compared to borg's animation and basic state. My PR has that fixed, but just putting it out here in case it doesn't get merged.

PR #4835, opened 2018-02-15
Fix botanist hud icon

Arokha:
File missing hudbotanist icon_state. (You should get some icons for Explorers and stuff. A user made them on our server. You should steal them.)


Anewbe:
I could have sworn we killed that bug at least twice. (And I checked out your sprites, I may steal the Pilot one, but Explorer is a bit too `Sciencey Eye of Sauron` for my tastes, based on what our Explorers are supposed to be.)

PR #4839, opened 2018-02-15
Another attempt at leg icon generation fixes.

Verkister:
-Makes leg icon cache code little more robust in the hunt of the increasingly elusive bugson dismemberment crash.
-The old leg layer direction caching code was pretty slippery to grasp and basically the only thing left separating it from the other well behaving limbs at this point.
-The cache blending runs each limb on its own so there is no practical need to (con)fuse the left and right legs together like that.
-Not entirely sure how exactly the old code did the stuff(itjustworks.ogg), but the new one goes straight to the point, picking the side of the processing limb, grabbing its overlay appropriate direction icons onto one temp set and blending it as an overlay set onto base before grabbing the underlay appropriate direction by itself into a separate(free of the overlay set shenanigans) temp set and blending that one as underlay onto the base.


Anewbe:
Tested?

PR #4843, opened 2018-02-16
Department ponchos for all

PrismaticGynoid:
Removes the job restrictions on the department ponchos, making them able to be taken from the loadout by any job, just like the department jackets.


Belsima:
Maybe make it poncho selection?

Magneld:
I'm okay with this.

PR #4844, opened 2018-02-16
Fixes crew member AI holograms.

### LINKED ISSUE #4092: AI Hologram cannot select from crew
In the AI hologram settings, you have the option to have your hologram mimic the appearance of one of the crew members. However, if you select a crewman and then project the hologram, you simply have no hologram.

Have the hologram appear like a crew member.

#### What actually happened
No hologram was projected when the holopad turns on.

#### Steps to reproduce
1: Play AI
2: Go to Hologram Settings
3: Choose a crewmember
4: Project from hologram

#### Additional info:
 - **Server Revision**: master - 2017-10-16
4f86947108f401b0c19b0c4532d201748bc517c9
Has been repeated across several rounds, and with different players.

MisterLayne:
(Partially) fixes AI selecting crew members as holograms. Hasn't been tested with another person, just using spawned mobs. Has a 50% chance to pull the player out of their mob, copy their icon, and then just delete them. Probably requires further testing..

Fixes #4092

### DIFF
--- a/code/modules/mob/living/silicon/ai/ai.dm
+++ b/code/modules/mob/living/silicon/ai/ai.dm
@@ -1,5 +1,6 @@
 #define AI_CHECK_WIRELESS 1
 #define AI_CHECK_RADIO 2
+#define TRACKING_POSSIBLE 0
 
 var/list/ai_list = list()
 var/list/ai_verbs_default = list(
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/silicon/ai/ai.dm
```diff
@@ -528,22 +529,38 @@ var/list/ai_verbs_default = list(
...
 	var/choice = alert("Would you like to select a hologram based on a (visible) crew member, switch to unique avatar, or load your character from your character slot?",,"Crew Member","Unique","My Character")
 
 	switch(choice)
 		if("Crew Member") //A seeable crew member (or a dog)
-			var/list/targets = trackable_mobs()
+			var/list/targets = list()
+			for(var/mob/living/M in mob_list)
+				if(!M)
+					continue
+				if(!M.client) //This *will* keep SSD players from showing up on the list, but if this wasn't here, there would be a runtime if you tried to imitate an SSD mob.
```
Arokha: If you are ignoring clientless mobs, use the player_list instead, which only includes mobs with clients. It will be faster.

code/modules/mob/living/silicon/ai/ai.dm
```diff
@@ -528,22 +529,38 @@ var/list/ai_verbs_default = list(
...
+
+				targets += M
+
 			if(targets.len)
-				input = input("Select a crew member:") as null|anything in targets //The definition of "crew member" is a little loose...
-				//This is torture, I know. If someone knows a better way...
-				if(!input) return
-				var/new_holo = getHologramIcon(getCompoundIcon(targets[input]))
+				received = input("Select a crew member to imitate:") as null|anything in targets
+				var/mob/living/carbon/human/dummy/dummy = received
```
Arokha: Non. Should be `= new()`. This is why your mobs are being deleted randomly. You're making this dummy ref point straight at their currently existing mob and then doing things to it (such as deleting it). It needs to be a new mob.

code/modules/mob/living/silicon/ai/ai.dm
```diff
@@ -528,22 +529,38 @@ var/list/ai_verbs_default = list(
...
+				dummy.client.prefs.dress_preview_mob(dummy)
```
Arokha: The dummy will not have a client. You need to call `target.client.prefs.dress_preview_mob(dummy)`, so you are asking the real mob's client to dress this dummy preview mob you've made in nullspace.

Arokha:
You may want to consider just getting the icon straight from the target. Otherwise you'll impersonate them with their original job gear and whatnot in character setup (including things like being naked...).

MisterLayne:
@Arokha These are all things I originally tried. None of which have worked. It all goes fine up until I try to retrieve the client from the player by using 'received.client.prefs.dress_preview_mob(dummy)'. DreamMaker (for some reason) doesn't want to get the client from received. (I also double checked using the changes you just suggested. Same result):
`code\modules\mob\living\silicon\ai\ai.dm:551:error: received.client.prefs.dress_preview_mob: undefined var`

Arokha:
To be honest, don't bother with any of the preview mob stuff. It has two cases where it will fail:
* Impersonating a player-controlled nonhuman. It will result in "impersonating" whatever character they have loaded instead, which is weird.
* Impersonating a borg. It will do the same as above, and you can't impersonate the borg.

Just do the getholo icon and getcompound icon thing on the target atom itself, and set the result as the hologram.

MisterLayne:
@Arokha Doing that would just be the _original_ code, which is currently broken.

MisterLayne:
I have concluded that crew member holograms are not currently possible, and closing this.

PR #4853, opened 2018-02-17
Power monitoring consoles find all station-related sensors

### LINKED ISSUE #4842: Power monitor can't read all sensors
Power monitoring consoles do not detect all powernets. Instead, they can only see the ones that share a Z-level with the console, as well as the research outpost (xenobio) and telecomms. This also includes the AI's inbuilt power monitor subsystem.
powermonitorgood

#### What actually happened
powermonitorbad
#### Steps to reproduce
Take a look at the power monitoring console on deck 3, or build one on deck 1. Or, be an AI and try to use your power monitor subsystem. Be unable to pick up any additional powernets even with scanning.
#### Additional info:
 - **Server Revision**: Server revision: master - 2018-02-14
76fa2f82bf05075dc151a6332774f916b3a8967f
 - **Anything else you may wish to add**: Why are both telecomms and the xenobio outpost always visible?

Anewbe:
- Tested with the bridge console, it shows every subgrid and the outposts.
- ~~Also deletes a var that gets defined in simple_animal.dm, but is no longer used.~~

Fixes #4842

### DIFF
--- a/code/modules/mob/living/simple_animal/simple_animal.dm
+++ b/code/modules/mob/living/simple_animal/simple_animal.dm
@@ -566,7 +566,6 @@
 
 	//Projectiles with bonus SA damage
 	if(!Proj.nodamage)
-		var/true_damage = Proj.damage
 		if(!Proj.SA_vulnerability || Proj.SA_vulnerability == intelligence_level)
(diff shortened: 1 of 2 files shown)


Atermonera:
Fixes #4842

PR #4863, opened 2018-02-17
Crashed Medical Shuttle PoI

mistyLuminescence:
Adds a crashed medical shuttle PoI full of tears, medical gear and nurse spiders. Screenshots available on request, but omitted from the PR for spoiler reasons.

Thanks to @Schnayy for the nurse spider sprites.


Anewbe:
This honestly borders on a joke PR because of those spiders. I'll let it slide for now, but there's a non-zero change they'll be removed in the nearish future.

PR #4865, opened 2018-02-17
Adds umbrellas

Atermonera:
Tested, works.
Umbrellas will shield from water_act() of rain, and the damage of hail, because those're low-wind
Storms will drop the umbrella and throw it in a random direction, because high-wind
Blizzards don't drop the umbrella because they don't already iterate over mobs, the umbrella doesn't offer any added protection against a blizzard, and why are you taking a high-surface-area piece of taught cloth out in a blizzard?
Sprite color is randomized in new(), but can be manually set with the color var by adminbus, like other colorable things.
Sprites added: Open/closed sprites, and an _a of each for the color blending, as well as inhands
Not mapped because I didn't bother to check the presence of other map prs or ask if anyone else was working on the planetside map at present


Atermonera:
Pics for neerti

PrismaticGynoid:
What about adding them to the loadout?

Atermonera:
Will be brought up at the staff meeting for point valuation. Expect a minimum of 5 points.

Atermonera:
Decided at 3 points at staff meeting, I'll get to that at some point

Anewbe:
Can be added to the loadout in another PR.

PR #4866, opened 2018-02-17
Cryobag oxygen supply increased

Anewbe:
Should last about 5 times as long, so based on what I've been told, you'll get ~15 minutes before it crushes you to death.


Wickedtemp:
Was it tested and timed, or no?

Anewbe:
It literally quintuples the volume of air, I did math based on values I've been passed by people who've actually sat through the testing for me. Either way, more air than it used to be, by a large factor.

PR #4867, opened 2018-02-17
Explorer gasmasks can be equipped through their hoods

Anewbe:


MisterLayne:
Will this mess with anything technical- Like how the hood is affected by weather, or armor, or anything like that?

Anewbe:
Has no effect on the hood proper. Gasmask will be very marginally less useful, but it will be rewritten soon.

code/modules/clothing/masks/gasmask.dm
```diff
@@ -73,7 +73,6 @@
 	icon_state = "explorer"
 	item_state_slots = list(slot_r_hand_str = "gas", slot_l_hand_str = "gas")
 	armor = list(melee = 10, bullet = 5, laser = 5,energy = 5, bomb = 0, bio = 50, rad = 0)
-	body_parts_covered = HEAD|FACE|EYES
```
Arokha: This will prevent this mask from protecting them from their eyes burning on phoron.

Neerti: Does it not inherit it from the base gasmask?

Arokha: Ouu. Good point. Wait, why was this defined to begin with then?

Anewbe: Probably stolen directly from TG?

PR #4873, opened 2018-02-18
Fixes CS sprites (again)

SubberTheFabulous:
Alternative legs are no longer sticks. 
Really small, insignificant changes are my profession.
Before and after, picture:
image


Belsima:
did you double check the sprite states this time

no offense but the other two times we got robots missing parts

SubberTheFabulous:
Yeap, all fine.

PR #4875, opened 2018-02-18
Moves most boxes to use starts_with

Arokha:
Doesn't (shouldn't) change anything about any boxes, other than how they spawn their things at round-start. Stops overriding New() on them, and uses the starts_with list.

There are technical reasons why this is better, not to mention it's probably easier for new coders to see an item's path "= 7" and understand it spawns 7 of them, rather than seeing a for-loop with a 1-to-7 iterator in an overridden New() proc and wondering wth that means.

Also changes the surgery kit's sprite to look like the one below, while I was in there, because it looks exactly like a first-aid kit otherwise.

image


mistyLuminescence:
Love the new surgery kit sprite.

PR #4876, opened 2018-02-19
Simple mobs respect armor pen, sharp, and edge

Anewbe:
- Mobs are no longer forced to always check melee armor. I have **_PLANS_** ™️.
- Simple mobs can now use armor penetration on their melee attacks. I have **_MORE PLANS_** ™️.
- Simple mobs can now be given sharp and edge. I have **_YET MORE PLANS_** ™️.

Long story short, this lets us make more dangerous melee mobs.


Arokha:
Dun dun dunnnnnnnnnn.

You may consider updating attack_generic on other things, like other simple_animals, too. Doors have it as well, windows, etc. Maybe maybe those values do something on them.

Not really important to put in this PR, just something to consider for your ***PLANS***:tm:

PR #4877, opened 2018-02-19
Two organ bugfixes

PrismaticGynoid:
First, allows FBPs with synthetic brains (posis and drones) to actually die from massive brain damage. Before, although the damage would still affect them, robotic vital organs taking 60+ damage would never actually kill the owner.

Second, stops the zombie organs bug from popping up in normal play. You were able to repair all the damage on an organ, but not the ORGAN_DEAD status, leading to "dead" organs that still functioned (and that you could never die from them taking damage). Now, repairing the damage surgically will remove that.

### DIFF
--- a/code/modules/surgery/organs_internal.dm
+++ b/code/modules/surgery/organs_internal.dm
@@ -73,7 +73,7 @@
 		return
 	var/is_organ_damaged = 0
 	for(var/obj/item/organ/I in affected.internal_organs)
-		if(I.damage > 0)
+		if(I && (I.damage > 0 || I.status == ORGAN_DEAD))
 			is_organ_damaged = 1
 			break
 	return ..() && is_organ_damaged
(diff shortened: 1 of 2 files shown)


Anewbe:
Glorious.

code/modules/surgery/organs_internal.dm
```diff
@@ -120,6 +120,8 @@
 					target.sdisabilities &= ~BLIND
 				if(I.organ_tag == O_LUNGS)
 					target.SetLosebreath(0)
+				if (I.status == ORGAN_DEAD)
```
Arokha: There already is a surgery that does this in other.dm. Why add two ways to do the same surgery?

PrismaticGynoid: There isn't, actually. Both of the necrotic organ repair surgeries there only work on external organs, and reworking that to work on internal organs as well still wouldn't solve the 'you can fully repair an organ but still have it be an unkillable zombie organ' problem.

Arokha: Ah, I see. But then shouldn't you have changed line 114? Because you cannot even get to the part where it unsets dead if it has no damage. If you end up with an organ that has 0 damage somehow, but still has the dead flag, 114 will be false and you will not be able to fix it still. Shouldn't it maybe be `if(I && (I.damage > 0 || I.status & DEAD)`?

PrismaticGynoid: Maybe, though the only ways that could happen is being injected with 60+ units of peridaxon (which overdoses at 10u) or through varediting the damage away (which can also be done to the dead status in that case).

Arokha: You saaaay that, and yet you're here fixing a problem that allows exactly that to happen. :p The ability to completely heal dead organs and not remove the dead flag. You might as well just make it future-proof.

Wickedtemp:
<3

PR #4878, opened 2018-02-19
Fixes uncleanable feet.

Verkister:
-Foot washing proc was using a wrong and nonexistent var instead of the one actually assigned for cleaning feet by showers and such.


Anewbe:
You are a champion, that's been on my low-priority list for ages.

Nerezza:
@Anewbe This seems to just rename a local var and mechanically, the proc works the same as it did before (which is to say, it doesn't work). I'm looking through it myself right now to see why showers don't clean feet.

PR #4879, opened 2018-02-19
Adds planetside gun permits

Anewbe:
- Added a planetside gun permit item. Anyone with a registered permit can have a gun on the planet's surface.
- Explorers spawn with a planetside gun permit. One also spawns in each explorer gun cabinet.


zaredman:
Do these permits have names attached to them when people spawn in with them?

Also, what about leaving some with the HOP/Warden for distribution instead of leaving in lockers?

Anewbe:
When they spawn on the Explorer they automatically get named.

And Upper Management didn't say nothin' about HoP or Warden, so *shrug? They're only in the gun cabinets in the first place because I was explicitly told to put them there.

PR #4881, opened 2018-02-19
Removes Stun() proc from flashbangs

Anewbe:
As per the 2/18 Staff Meeting, removes the Stun() from flashbangs. Either you're close enough to be dropped on your ass, or you're too far away to be affected. Will hopefully ease some of the complaints we've historically gotten about them.

Also: **Absolute Pathing**


Neerti:
Should note for those unaware, ``Stun()`` makes you unable to move or act but you still remain standing (its what Beepsky Smash does iirc). This doesn't actually remove the knocking down part, which is ``Weaken()``.

Reason for this removal is that people who were stunned but were standing couldn't do anything, but other people thought they were not stunned and could still be dangerous. With it being knockdown-only, its more obvious that someone is unable to act.

Anewbe:
Thank you.

FenodyreeAv:
Isn't there still Stun() on the visual portion of the grenade?
And doesn't this drop the total stun time from 13 to 3?

code/game/objects/items/weapons/grenades/flashbang.dm
```diff
@@ -5,92 +5,85 @@
...
 //Flashing everyone
-		if(eye_safety < 1)
-			M.flash_eyes()
-			M.Stun(2)
-			M.Weaken(10)
-
-
+	if(eye_safety < 1)
+		M.flash_eyes()
+		M.Stun(2)
```
Neerti: Seems you missed one here.

Anewbe:
Thaaat's not supposed to be there, closing and re-opening when I'm less dumb.

PR #4883, opened 2018-02-19
Makes electrochromatic glass buildable and programmable in game.

Leshana:
- Reinforced glass windows can be upgraded to electrochromatic by hitting them with a cable coil.  The window must not be unpried from its frame to upgrade it (but it can remain anchored).
- Windows can be linked with a tint button by using a multitool on the button to buffer it, then using it on the window to set the ID.  The window must be completely unanchored to edit its id.
- Window tint buttons can have thier ID set *once* with a multtool.
- Inspired by Baystation12/Baystation12#20508
- Also adds a "full tile" polarized window type for use by mappers.

### DIFF
--- a/code/game/objects/structures/window.dm
+++ b/code/game/objects/structures/window.dm
@@ -294,6 +294,21 @@
 			if(is_fulltile())
 				mats.amount = 4
 			qdel(src)
+	else if(iscoil(W) && reinf && state == 0 && !istype(src, /obj/structure/window/reinforced/polarized))
+		var/obj/item/stack/cable_coil/C = W
+		if (C.use(1))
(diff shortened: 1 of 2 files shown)


code/game/objects/structures/window.dm
```diff
@@ -551,6 +566,28 @@
 	desc = "Adjusts its tint with voltage. Might take a few good hits to shatter it."
 	var/id
 
+/obj/structure/window/reinforced/full
```
Anewbe: Is this supposed to be polarized?

Leshana: Yes actually, thanks. I've updated it.

PR #4886, opened 2018-02-19
Fixes some tile overlays on z1

zaredman:
Fixes some tile overlays that were set incorrectly on z1 between Dock 1 and the Medical post.


zaredman:
Did another look-over and don't see anything immediately jumping out at me as more issues.

PR #4888, opened 2018-02-20
Adds various cabin-related things.

BillyBangles:
Makes bonfires able to provide warmth (kinda)
Adds fireplaces, which operate similarly to bonfires.
Fixed an oversight that allowed bonfires to reach an "in-between" state where the fire mysteriously goes out until you add more.
Adds a blue sifwood floor tile
Fixes blue carpets (You can place them again). They're now known as teal carpets, and still come in the cargo carpet crate.
Adds the ability to dig up tree stumps with a shovel or spade.

Changelog and images incoming. Everything tested and working.


BillyBangles:
Showing off the fireplace
Sifwood floors versus normal wood floors
Digging up a tree stump with a shovel

Atermonera:
Right as I was about to add the label too :V

BillyBangles:
well i did say changelog incoming :P

Anewbe:
Minor thing: Can you add a do_after to the stump digging? Just a short one. Stump digging sucks IRL.

code/game/turfs/flooring/flooring.dm
```diff
@@ -275,6 +280,23 @@ var/list/flooring_types
 		'sound/effects/footstep/wood4.ogg',
 		'sound/effects/footstep/wood5.ogg'))
 
+/decl/flooring/sifwood
```
Belsima: Just asking; given all the flags made here, could this be a child of regular wood floor?

BillyBangles: I tried that, and it was having similar problems to the blue carpet, in that you couldn't put it down for some reason. It might not have anything to do with being a child of another type of turf but I made it its own separate thing in an attempt to wrap my head around why it wasn't working.

PR #4893, opened 2018-02-20
Slimes should no longer /list you.

### LINKED ISSUE #4872: Grey baby slimes have weird notification

When slime assaults you.

The grey baby slime (672) has /list Merek Black!

This happened a few times. Don't know what this is.

#### Additional info:

Anewbe:
Should fix #4872 

I probably didn't need to change all of these files, but future-proofing is good.

I'm pretty sure the issue is fixed in slime/combat.dm, where it was passing attacktext instead of pick(attacktext). It seemed to be fixed in testing.


Arokha:
Sneaky secret xenobio mobs not being in s_a path!

PR #4897, opened 2018-02-20
Fix health analyzer names

Arokha:
The one that comes out is called advanced. And there IS an enhanced one (level 3) that this is not.

Now also renames them: [nothing] -> Improved -> Advanced -> Phasic

That seems like a linear progression to me.

### DIFF
--- a/code/game/objects/items/devices/scanners.dm
+++ b/code/game/objects/items/devices/scanners.dm
@@ -266,15 +266,15 @@ REAGENT SCANNER
 		if(0)
 			to_chat(usr, "The scanner will now perform a basic analysis.")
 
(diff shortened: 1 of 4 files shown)


Anewbe:
~~It would be so nice if those weren't so vaguely named.~~

Arokha:
Yeah, I was pondering names that imply obvious improvements over the previous. I personally would have done [nothing] -> improved -> advanced -> phasic maybe. Phasic is MAGICAL and obviously the best. Improved sounds incremental, and advanced sounds 'good'

Belsima:
We really should rename them better.

Arokha:
Done

PR #4898, opened 2018-02-20
Mining Bag / Box Speedification

Mechoid:
Mining ore bags no longer hang the server for 3 seconds, and are now hard-locked to only gathering Ore and its subtypes.

Mining boxes can be emptied by borgs and mining drones as well, as they are able to open normal crates.

In touching the orebag, I somehow broke mining borg auto-pickup. There is now a check in their move for if any borg has an orebag deployed into a module slot, active or not, to maintain old functionality. It also no longer spams the borg's chat when picking up objects.


Nerezza:
We will honor this pull request with an annual holiday.

code/modules/mining/satchel_ore_boxdm.dm
```diff
@@ -62,13 +65,12 @@
...
 	return
 
-
 /obj/structure/ore_box/verb/empty_box()
 	set name = "Empty Ore Box"
 	set category = "Object"
 	set src in view(1)
 
-	if(!istype(usr, /mob/living/carbon/human)) //Only living, intelligent creatures with hands can empty ore boxes.
+	if(!istype(usr, /mob/living/carbon/human) && !istype(usr, /mob/living/silicon/robot)) //Only living, intelligent creatures with gripping aparatti can empty ore boxes.
```
Arokha: Hah

PR #4901, opened 2018-02-21
Adds holoplants, and to cargo

Arokha:
Courtesy of CEV Eris SS13 server (plus a bunch of additional code to make them real machines)


Arokha:
It's blepped.

PR #4903, opened 2018-02-21
Adds halogen counter tool, adds gas scanner to analyzers

Atermonera:
This will conflict with #4897 and it's my fault for being a lazy bum and not finishing this sooner.
Tested:
Gas scanner
Gas scanner

Halogen counter
Halogen counter
Halogen counter uses the EFTPos sprite

Also cleans up a small bit of the associated PDA code
And refactors health analyzer attack() code because jesus christ 61 calls to show_message is too many, just build a message and send it all at once


Atermonera:
And that's what I get for trying to reduce the number of spans by lumping blocks together, fix in a bit

Atermonera:
Apparently 
```if(x) "<span>..."
else "<span>..."
"...</span>"``` angers travis.
Which I suppose makes sense

Anewbe:
Tested and working? Even the med scanner cleanup parts?

Atermonera:
Yes, tested each of the medscanners

PR #4910, opened 2018-02-21
Unobfuscates GPS locations

Atermonera:
After a miscommunication brought them to being rather more useless than was desired.
Tested


Atermonera:
PoI gps units will now show a similar level of information to the current system, where it's raw distance to 10m accuracy, except with more accurate directions thanks to get_adir()

Anewbe:
Tested?

Anewbe:
Got my response elsewhere.

PR #4912, opened 2018-02-21
Small regenerate_icons tweak.

Verkister:
Alright the previous attempt didn't quite work as intended, sorry for the inconvenience.
Anyway I was checking out what sorta things happen to use the legacy support regenerate_icons human proc and found out some of them had connections to circumstances I've heard of causing trouble. Found out all the update calls included in it came with a 0 to prevent each call from doing update_icons on its own before the intended calls at the bottom of the bunch, except for one, UpdateDamageIcon, which also did have the same (var/update_icons=1) skip expectation as the others, causing premature dupe updates in regenerate_icons.


Anewbe:
Here we go again!

PR #4916, opened 2018-02-21
Critters Can Crawl Carrying Client Constructs

Arokha:
When I copied the hud elements from humans to simple animals, I copied a bug with it, which nobody noticed on humans because they can't ventcrawl.

You shouldn't put the hud elements literally inside the mob, especially considering the huds are deleted/created whenever you logout/login. This results in s_a's played by admins having like 999999 of these objects as they aghost over and over.

That will fix the ventcrawl problem, but ventcrawl was written weird to begin with. This is better than it was, but not as good as /tg/'s typecaches. S o m e d a y.

### DIFF
--- a/code/_onclick/hud/human.dm
+++ b/code/_onclick/hud/human.dm
@@ -75,7 +75,7 @@
 		ico = new(ui_style, "black")
 		ico.MapColors(0,0,0,0, 0,0,0,0, 0,0,0,0, 0,0,0,0, -1,-1,-1,-1)
 		ico.DrawBox(rgb(255,255,255,1),1,ico.Height()/2,ico.Width()/2,ico.Height())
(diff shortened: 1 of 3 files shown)


Anewbe:
Tested?

Arokha:
Tested it originally to make sure it didn't break anything more that wasn't already broken, but gave it a more thorough test just now to be sure. Don't see anything it breaks, and it seems to fix the issue.

I'm always a little 'eeeeh' about undoing people's code when I don't know why they did it originally, and there's really no obvious reason it was like this. Hopefully the reason was 'because they didn't know any better' and not 'because of some TOP SECRET reason'.

Anewbe:
~~Knowing Byond, it's going to break Genetics and ZAS.~~

PR #4921, opened 2018-02-22
Drone Foundry PoI Combat Drone Tweaks

mistyLuminescence:
You know the Drone Foundry PoI with all the combat drones? The ones that, when they notice you, fire bursts of lasers at you with pinpoint accuracy until you die horribly?

The ones on the outside are now dead. Some other exploration team wiped them out. You can loot the drones (just like any other lootpile), but don't get too greedy, 'cause the ones inside the facility are still active...

**TAKE 3: IN WHICH THE ONES OUTSIDE THE FACILITY ARE DEAD**

This PR changes the drones on the border of the PoI - the ones outside the facility, that explorers can run into without directly encountering the facility itself - into dead drone lootpiles. A warning, as it were, that "if you explore this area, you're going to get wrecked for sure". To compensate for the lowered number of mobs, some more alive drones have been added to the inside of the facility.


Magneld:
So you ran, alone. to basically one of the final boss type POIs and got your shit smacked.
And you want it nerfed.
No.

ThoughtSynapsis:
Drone base isn't an easy POI, In-fact it's probably the hardest POI currently right next to Neerti's X-com adventures.  Tackling it alone or without perpetration often results in some misery, And as such its situated on the farthest parts of the wilderness.

If it proves too much of a challenge for a group of explorers tooled up or if its an issue with drones on the outline then there will be changes, But a recent PR adjusted there speed meaning you should be able to escape them if you see one.

mistyLuminescence:
@Magneld No, this isn't an "i ded pls nerf" PR, this is based on having fought them both live and in controlled environments, and seen others doing the same. I've had the PR idea for a while, just needed to see it live to experience it.

@ThoughtSynapsis Truth. Even with backup and proper gear, they're still more than capable of killing an explorer. Setting the ones on the border to be weaker would be a good idea - sure, they're slower than before, but they don't have to outrun healthy!you, they have to outrun you when you've been hit with 60 damage of lasers. Maybe keep the ones inside as current tier (i.e. super strong), and set the ones on the border to be threatening-but-not-instagib tier (see: PR), so explorers know what they're getting into but have the option of running away?

ThoughtSynapsis:
As it stands however I haven't seen any indication of the POI being taken on as a team effort. I would be more inclined to increase the border radius of Rockybase with more obvious markers of "Here there be dragons." And work from there.  It's not an easy POI but I have seen individuals beat it! (And don't worry, They didn't roll in there with a mech)

Nalarac:
Yeah, it's fairly easy to just charge the combat drones and melee them down since their melee damage is garbage (sadly not as effective on the mech because it casts FIST!!) I can't say much else because I've only ever assaulted the place as a cyborg.

mistyLuminescence:
Update: only changing the ones on the outside. The ones on the inside are still tanky laser spam hell.

Anewbe:
This honestly looks like an Operator Error induced nerf to me.

You were a Taj FBP, that's top-tier for "I'm gonna leave now" hijinks. Those outside drones are slow enough that a non-VTEC borg (who are notably slow) can escape from them with little trouble. Nalarac can probably testify on that, too. I for one literally got within two tiles of one as a Crisis Borg and managed to float away from it without taking more than one hit.

And, as mentioned elsewhere, if you're anywhere near a drone, and you're alone, you run or you die. You're not supposed to be able to clear the Wilderness by yourself.

Screemonster:
The point isn't that you shouldn't be able to see one and then run away and get backup if you're solo, the point is that if you're solo you shouldn't even be venturing onto that z-level if you're not prepared to get your ass kicked.

mistyLuminescence:
That's not how the rest of the PoIs work - you can run from a crashed UFO or a minefield or some turrets, and come back later or mark the area as unsafe. What makes this any different?

Dankman:
the thing that makes it different is that this PR was posted as a kneejerk reaction to you getting your shit wrecked but hey, you do you.

mistyLuminescence:
You get an F on reading comprehension. As I've already said twice, I've had the idea for this PR for a while now (Jan 28, when someone else commented on the drones being OP, and Jan 29, where I did extensive testing with different weapons and equipment on how much it took to kill them).

Screemonster:
The point is that if you nerf something that is designed to be taken on in groups because you can't survive it solo, in an area that's meant to be taken on by groups, means that when a group does show up there won't be anything to challenge them because the difficulty has been scaled to some chump wanting to be able to safely roam it solo.

Dankman:
did you test with a full crew of people taking on the POI as was intended by its designers? 
or did you test solo.

mistyLuminescence:
As detailed in the opening post, the facility itself still has the original, powerful drones, as well as the big stompy hostile mech. You're not going to be able to survive it solo. You're probably not even going to be able to 1v1 the drones outside solo. This just means explorers can run away after seeing the border drones and going "oh no, this is too hard for us".

pontifexminimus2:
If anything, solos getting wiped out is a feature.

Don't go solo. Fix'd.

mistyLuminescence:
@Dankman Both. Do you have anything helpful to add?
@pontifexminimus2 Solo explorers who don't heed warnings should be wrecked. When those warnings aren't in place, problems arise.

> To reduce the perception that a submap is 'unfair', some steps can be taken to warn the player that the submap they are entering might be the last one they enter for that round.

> In the UFO submap, there is only one door to go inside, and it is divided into several rooms. The first room is safe, and contains two dead turrets, a skeleton, two empty weapon cells, a spent hypo, an unused burn hypo, some captain gear, a deactivated GPS, tools, insulated gloves, and an empty retro laser. This room is very important for the submap's design, and it serves two purposes, one of which is relevant here.

> The room gives a hint towards what dangers the rest of the submap will contain (deadly turrets), and the tools needed to attempt to progress deeper inside. It would feel unfair if the player instantly got gunned down by turrets immediately upon entering, as they had no way to know without metagaming. The first room ensures that observant players will realize that subsequent rooms will be hazardous, and if an unobservant player ignores the warning and dies in the second room, it should feel less unfair.

> As such, if designing a dangerous submap, the danger should be obvious to those entering, and ideally the nature of the danger would also be explained by the surroundings.

Atermonera:
I would rather advocate for deactivated/dead drones on the very outskirts, making the submap a touch larger, than for this. There are viscerators and corpses spaced around the entire edge of the submap, but there are several gaps where you could come across the one of the drones before encountering that. Taking away the rapidfire is actually a very hard nerf to the drones, changes to health aside, even if they're only on the outskirts.

The spacing and nature of the warning signs means that you might miss them, and even if you run into them, you won't really be able to discern the nature of the threat. Ergo, rather than making some of the dangerous part of the PoI, I'd rather make the warning signs more apparent. Less, nerf the bear, and more, there's viscera and bodies strewn about the entry to the bear's cave.

mistyLuminescence:
Swapping the border drones for deactivated/dead drones would work if we can get sprites for them. As they are right now, they just explode when killed, so I'm pretty sure we don't actually have a 'dead drone' sprite.

code/modules/mob/living/simple_animal/aliens/drone.dm
```diff
@@ -280,6 +280,12 @@
 // A slightly easier drone, for POIs.
 // Difference is that it should not be faster than you.
 /mob/living/simple_animal/hostile/malf_drone/lesser
-	desc = "An automated combat drone with an aged apperance."
+	desc = "An automated combat drone with an aged appearance."
```
Belsima: at least remember to fix this regardless of balanace

mistyLuminescence: balanace

Anewbe:
Please separate the two things you've proposed here into separate PRs, so they can be discussed individually.

mistyLuminescence:
me, screaming at travis,
also which two separate things are we talking here

Anewbe:
A drone nerf is one thing. A drone loot pile is another thing.

Belsima:
The nerfed drone is not mapped in.

mistyLuminescence:
It's not a drone nerf any more either, it just adds a subtype of PoI-drone which isn't in the map (could be used as an admin resource, even - it's like a regular combat drone but not OP). The current change is just swapping the exterior drones for lootpile dead drones, and moving them inside instead.

Anewbe:
Quite simply, the HeadStaff have discussed this and we're not comfortable merging this PR. It has changed direction too many times to still be accurately represented by the comments here. Additionally, we have numerous complaints about the contents of the loot pile, in a volume that we are not going to discuss here. If you care to have this change implemented as presented, please make a new PR, specifying exactly what the change is and what the goal of said change is. Additionally, we have other plans for that specific PoI, that we believe will better match the design philosophy of the server.

mistyLuminescence:
I asked for (and implemented) feedback on discord for the contents of the loot pile; what else do you suggest we change about it?

Anewbe:
Close this PR. Separate it into two PRs. It will not be discussed further on this PR.

mistyLuminescence:
camera

PR closed, another PR going up soon-ish.

mistyLuminescence:
Update: another PR _not_ going up soon-ish, but if there's any changes this PR makes that you like the idea of, let me know and I might make another.

PR #4922, opened 2018-02-22
Should actually fix the flashbang thing, hopefully without a billion extra commits

Anewbe:
https://forum.ss13polaris.com/viewtopic.php?f=14&t=477


Atermonera:
Changelog to be written in about 10 minutes, when I run the script

PR #4924, opened 2018-02-22
Eliminate duplicate code in trinary atmos components

Leshana:
## Consolidate trinary_base code
- Consolidate common functions into trinary_base
- The only real difference in atmos_init was which direction each node would connect in.  Delegated that to an overriddable proc, and the rest can be consolidated. Same story for update_underlays()
- Used similar patterns on tvalves which also are trinary even tho they are not subtyes of trinary (for some reason)
## Went even farther in consolidation of trinary atmospherics.
- We can eliminate any overridden methods at all on flipped trinary devices by making a `mirrored` and `tee` variable!  This allows us to do things like flip them without delete/spawning.
- T-Valves are also basically trinary.  Sadly they are not a subtype, so to share code with them I have to put it into an outside proc.  But still worth it.

## Consolidate digital t-valves type paths.
- Because mirroring is now *much* easier to setup than digital remote control, we reverse the order of inheritance to de-duplicate all the radio code from digital t-valves!
- This requires changing the following type paths:
    `/obj/machinery/atmospherics/tvalve/mirrored/digital => /obj/machinery/atmospherics/tvalve/digital/mirrored`
    `/obj/machinery/atmospherics/tvalve/mirrored/digital/bypass => /obj/machinery/atmospherics/tvalve/digital/mirrored/bypass`
- Change made on all maps.

Testing shows they even all still work too:
capture

Port of VOREStation/VOREStation#3093 and VOREStation/VOREStation#3100
*The pipes.  The pipes.  The pipes are coming.  **It begins.** They are coming.*


Anewbe:
Ia ia Atomsia fhtagn!

PR #4929, opened 2018-02-23
Splints no longer perfectly prevent negative effects from broken bones

Anewbe:
- Splinted bodyparts act broken 30% of the time.
- Splinted legs still slow you down like broken ones.
- Hardsuits are mostly unaffected. Power Armor is Power Armor.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_movement.dm
+++ b/code/modules/mob/living/carbon/human/human_movement.dm
@@ -35,7 +35,7 @@
 			var/obj/item/organ/external/E = get_organ(organ_name)
 			if(!E || E.is_stump())
 				tally += 4
-			else if(E.splinted)
+			else if(E.splinted && E.splinted.loc != E)
 				tally += 0.5
 			else if(E.status & ORGAN_BROKEN)
 				tally += 1.5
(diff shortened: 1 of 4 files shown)


mistyLuminescence:
Have you tested how this works with picking up items with broken-but-splinted arms, and walking with broken-but-splinted legs? How long can you hold an item for/walk for on average with this before something goes wrong?

Anewbe:
About 70% longer than if you don't have a splint.

mistyLuminescence:
Pretty sure you drop things near-instantly with a broken hand/arm, and testing broken legs/feet has been pretty weird on my end (sometimes you fall over, sometimes nothing happens at all). Making it only work 30% of the time (read: 'you still drop things near-instantly' and 'broken legs/feet are still weird'), just wanted to check if that was the case during testing or not.

Atermonera:
Changelog to be written in about 10 minutes when I'm running the script

Rand0mname:
I mean it is a broken arm after all. Even with a splint I doubt there is much you could do with it that doesn't involve pain/making it worse.

Screemonster:
How will this handle traumatic_shock? Spikes of pain on 30% of life ticks?

Anewbe:
@screemonster Anywhere the game checks for a broken limb, if you're splinted by anything other than a hardsuit, is going to trigger a little less than a third of the time.

Nalarac:
If you had a broken limb splinted and taking painkillers, would you still have the negative effects? I ask because broken ribs can be temporarily "dealt" with by taking tramadol.

PR #4931, opened 2018-02-23
Adds a cursed sword (and other haunted things)

BillyBangles:
This one's a whopping three (technically 5) parter, so stay tuned.

- [x] Adds the cursed sword item, a mysterious sword roughly on par with the cult sword in the way of damage.
- [x] Adds a framework for admins to let ghosts 'haunt' an object and speak through it.
- [x] Adds a ghost pod for the cursed sword, allowing explorers to find it and have a talking sword companion.
- [x] Adds a POI containing the aforementioned cursed sword ghost pod.


BillyBangles:
Cursed sword example

code/game/objects/items/weapons/melee/misc.dm
```diff
@@ -44,4 +44,57 @@
...
+	if(src.brainmob && src.brainmob.key)
+		switch(src.brainmob.stat)
+			if(CONSCIOUS)
+				if(!src.brainmob.client)	msg += "The blade's bright glint seems dull.\n" //afk
+			if(UNCONSCIOUS)		msg += "<span class='warning'>The blade seems oddly lifeless.</span>\n"
+			if(DEAD)			msg += "<span class='deadsay'>The blade seems completely lifeless.</span>\n"
+	else
+		msg += "<span class='deadsay'>The blade seems completely lifeless, its bright glow now dull.</span>\n"
+	msg += "</span><span class='info'>*---------*</span>"
+	user << msg
```
Arokha: to_chat(user,msg)

code/game/objects/items/weapons/melee/misc.dm
```diff
@@ -44,4 +44,57 @@
...
+	return 0
+
+/obj/item/weapon/melee/cursedblade/New()
+	src.brainmob = new(src)
+	src.brainmob.add_language(LANGUAGE_GALCOM)
+	src.brainmob.loc = src
+	src.brainmob.container = src
+	src.brainmob.stat = 0
+	src.brainmob.silent = 0
+	src.brainmob.name = "cursed sword"
```
Arokha: Where's the framework part?

BillyBangles: In say.dm. For a brain to talk now requires the container its in to have a can_speak variable. Previously it checked if the brain was in an MMI and called it a day.

BillyBangles: The only things that have can_say set by default are MMIs (which includes posibrains and drone chips) and this sword, but admin magic can be used to make other things capable of speech.

Arokha: > Adds a framework to allow ghosts to 'haunt' an object and let them speak through it.

I mean, I interpret this as adding a mechanism by which ghosts can join into objects. Is that not the case? Has to be adminbused in?

Atermonera:
I don't want a changelog for the PoI, and the rest of the stuff doesn't exactly warrant one

code/game/objects/items/weapons/melee/misc.dm
BillyBangles: That's already the ghost_pods. The particular ghost pod for this sword I am currently working on. But yes, the framework I was talking about is mostly for adminbuse. Perhaps that was a bit misleading. I'll reword that.

Neerti:
It might be better to use a child of ``/mob/living/voice`` (which is what communicators ghosts use) to avoid the brainmob baggage.

Also not sure how kosher this is for lore but thats not really my call.

BillyBangles:
The lore excuse is that the sword is made out of the same kind of stone cultists use to _succ_ people's souls and use them for constructs. That's never explicitly stated anywhere but that was kinda the idea in my head.

As for the voice bit, i'll look into that.

Anewbe:
$0.02: All of the everything in PoIs is assumed non-canon outside of the round they're found in, so I'm not quite sure how much of an issue it is. Seems like a slightly more intelligent/undamaged version of those artifacts that parrot what you say near them.

BillyBangles:
I’m done here, wip tag can go.

BillyBangles:
100% forgot about that. Lemme take a peek.

BillyBangles:
There we go. Uses a voice mob now instead of the brainmob. Sword player now gets a neat static overlay as well.

PR #4932, opened 2018-02-23
Small tweak to survival boxes

Arokha:
Synths start with a unique empty type rather than just an unnamed cardboard box, and don't get a tank or the comprehensive supplies (the only useful thing would be the glowstick, but there are a billion of those). That way when synths actually have some sort of emergency supply, it can go in there.

I kind of want to make an emergency cell or something just to give synths some sort of actual emergency supply.


Atermonera:
`Unit 2138 crunches down on an emergency battery pack!`

Arokha:
Yessss.

PR #4935, opened 2018-02-23
Ice Tiles and Ice Puzzles

BillyBangles:
Adds some ice tiles, which have the special property of forcing you to keep sliding until you land on something that isn't ice, or run into something. Uncraftable, but good for puzzles while mapping! Speaking of which, puzzle POIs that utilize this coming shortly.


BillyBangles:
Example of sliding on ice
Fun prank to try on your friends

PR #4936, opened 2018-02-23
Sundry Bugfixes - Feburary

Leshana:
- Fixes VOREStation/VOREStation#3075 - Boxes of Chocolates say they contain donuts.
- Fix runtime when clicking RCD on almost anything. - get_base_turf_by_area() is only for turfs!
- Stop candles from messaging the universe when they light.


Anewbe:
Definitely read the title as _Sultry Bugfixes_ and was mildly concerned. Code looks good, merging.

PR #4941, opened 2018-02-24
Turrets Target through Girders, Airlock assemblies, and other unanchored structures.

Mechoid:
Things using the test projectile (Turrets, Aim intent), will be able to target through girders, airlock assemblies, and unanchored structures like frames or lockers.


mistyLuminescence:
RE turrets: do the turret lasers pass through frames/lockers etc, or are they blocked?

Mechoid:
Lockers, they will dig through just like metal foam. It's a temporary barrier. Frames have an 'Out of Scope, but Adjust If Abused' marker on them. They currently have no health, thus having them fire forever on a frame would fill logs to hell and back.

PR #4942, opened 2018-02-24
Removes translators from the loadout

Anewbe:
Heavily inspired by https://github.com/VOREStation/VOREStation/pull/2972

https://forum.ss13polaris.com/viewtopic.php?f=14&t=489

- Removes the Translators from the loadout. Go talk to RnD, or apply for the language.
- Makes the handheld translator bigger, so it doesn't fit in pockets. You're supposed to be looking at the screen to see it translated.
- Adds the Ear Translator to the RnD console. It's more expensive than the normal one, because it doesn't take up a hand.

### DIFF
--- a/code/game/objects/items/devices/translator.dm
+++ b/code/game/objects/items/devices/translator.dm
@@ -4,7 +4,7 @@
 	desc = "This handy device appears to translate the languages it hears into onscreen text for a user."
 	icon = 'icons/obj/device.dmi'
 	icon_state = "translator"
-	w_class = ITEMSIZE_SMALL
+	w_class = ITEMSIZE_NORMAL
 	origin_tech = list(TECH_DATA = 3, TECH_ENGINEERING = 3)
 	var/mult_icons = 1	//Changes sprite when it translates
 	var/visual = 1		//If you need to see to get the message
--- a/code/modules/client/preference_setup/loadout/loadout_ears.dm
+++ b/code/modules/client/preference_setup/loadout/loadout_ears.dm
@@ -8,8 +8,3 @@
 /datum/gear/ears/headphones
 	display_name = "headphones"
 	path = /obj/item/clothing/ears/earmuffs/headphones
-
-/datum/gear/ears/translator
-	display_name = "universal translator, ear"
-	path = /obj/item/device/universal_translator/ear
-	cost = 8
--- a/code/modules/client/preference_setup/loadout/loadout_utility.dm
+++ b/code/modules/client/preference_setup/loadout/loadout_utility.dm
@@ -120,11 +120,6 @@
 	slot = "implant"
 	exploitable = 1
 
-/datum/gear/utility/translator
-	display_name = "universal translator"
-	path = /obj/item/device/universal_translator
-	cost = 8
-
 /datum/gear/utility/pen
 	display_name = "Fountain Pen"
 	path = /obj/item/weapon/pen/fountain
--- a/code/modules/research/designs.dm
+++ b/code/modules/research/designs.dm
@@ -1790,9 +1790,17 @@ CIRCUITS BELOW
 	name = "handheld translator"
 	id = "translator"
 	req_tech = list(TECH_DATA = 3, TECH_ENGINEERING = 3)
-	materials = list(DEFAULT_WALL_MATERIAL = 2000, "glass" = 2000)
+	materials = list(DEFAULT_WALL_MATERIAL = 3000, "glass" = 3000)
 	build_path = /obj/item/device/universal_translator
-	sort_string = "HABBA"
+	sort_string = "HABQA"
+
+/datum/design/item/ear_translator
+	name = "earpiece translator"
+	id = "ear_translator"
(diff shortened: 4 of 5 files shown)


mistyLuminescence:
Left feedback on the forum thread, but TL;DR: understanding people is better for RP than not understanding people.

Arokha:
> understanding people is better for RP than not understanding people

I'd disagree with that as being the case, in general.

For RP, it can be bad if there are three people in a room, and two of them are constantly talking nonstop in another language, ignoring the third. That is not the fault of the language, though. That's the fault of those two people being assholes. The solution to fix that isn't "make it so the third can understand", it's "make those two people stop being assholes". Do that by doing in-universe stuff about 'common on common' and whatnot, easy enough. Or just the third character complaining to the HoP about how their coworkers are being douchebags, both RPish solutions.

Three people in a room, and an occasional message passed back and forth between two in private, perhaps even ABOUT the third, is great for roleplay. It makes the third have QUESTIONS and DOUBTS and maybe they get suspicious, maybe they don't like what species the two are, and figures they are talking about them behind their back, and stealin' their space jarbs. Or maybe they're just talking about coffee and didn't want to hear the third's opinion, who knows. Regardless, it allows for the character to have additional feelings about those two, just simply based on their use of a language that they don't understand. That in itself is something to act on/feel about.

mistyLuminescence:
In-depth explanation is on the forum thread, but consider: three people in a room, all talking different languages, but all understanding each other perfectly. They're all feeling close enough to use their own languages, and it's not a barrier for communication (barriers ain't fun, on the whole). If you're the sort of person who enjoys the questions and doubts of "what are they talking about? I don't know!", then you could always just not take a translator. Maybe your character can't afford one, or maybe they don't like the idea of it, but either way the option is still open. The existence of translators just allows people who'd rather have communication than confusion, to have that.

Arokha:
That sounds like the languages don't add or remove anything by existing in that situation. It sounds like the exact same thing can happen, without languages.

The problem is that people DO take the translators, and they make languages pointless for those that want to speak to people in private, using languages, because they are ubiquitous.

It can only go one way or the other, not both:
* Languages are pointless aside from being a funny text color and a word on character setup, because potentially everyone has translators.
- or -
* Languages have mechanical uses/benefits because translators are rare, and it allows you to speak in private.

I, for one, prefer languages to be an actually useful mechanical thing.

mistyLuminescence:
It _could_ happen without languages, but then it lacks the element of closeness/intimacy/just plain "I no speak Galcom good". Little things, but little things that mean a lot in the context of a HRP server. There are still ways to speak in private that translators can't catch onto - communicators, PDAs, even whispering; translators didn't kill the privacy star. I don't mind so much if there's no significant mechanical benefit to speaking a different language, even assuming everyone has a translator (8 points is pricey); it's a roleplay tool, and that's how it's used.

Arokha:
I never said that translators somehow removed all privacy. They removed one means of getting it, for very little in return in terms of roleplay. Your one example sounds interesting, but makes zero sense to me.

It's intimate because ... someone is speaking in their native language and you're listening via a metal box? That sounds like the opposite of intimate.

How much more intimate would it be if you COULDN'T understand them, and then decided to LEARN their language so you could talk to them in their native language? Do you not think that is more powerful?

Very-Soft:
Translators have a potentially significant mechanical benefit. This mechanical benefit is great enough, that a significant portion of the population of Vorestation took a translator, because why wouldn't you? It lets you understand almost anyone, all the time, and you get it for free to start with! People noticed this behavior, and so made them more expensive, because being able to understand almost anyone is powerful.  But the expense of them in loadout points was not enough to discouraging most people from taking them, really.

So the headset versions were removed from the loadout, and the handheld version was made not pocketable.

In their present state on Vorestation, they are not seen nearly so often, but can still be relatively easily acquired from RnD. Which is honestly a small price to pay. 

Being able to understand nearly everyone, at really no cost (because let's face it, 8/15 loadout points is basically nothing, you can still take a bunch of other things with that.) does nothing to enhance RP. It negates most of the RP that can come with language being a thing. If everyone's speaking strange languages, but they all have translators... then what's the point of there being all those languages?

Belsima:
Vorestation also has perhaps 10-15 more languages than Polaris, and thus I don't think can be safely compared, as it'd be harder to share a language with others there anyway.

Arokha:
I think he meant after I removed them from loadout.

Screemonster:
> So here's what we came up with - rather than a translator turning everything you hear into [specific language], it turns everything you say into [specific language]. So you speak Tradeband, but it comes out as Gutter, for example (and everyone nearby who understands either, will hear you). You can still have your intimate conversations where you speak your own language (by setting the translator to turn it into a language the other person understands), and you can still have your secret conversations (by taking off the translator). Boom, problem solved.

Can't reply on forums 'cause not registered yet, but...

**"THE BENE GESSERIT WITCH MUST LEAVE."**

Belsima:
I very much like the idea of a one-to-one translator.

Arokha:
Fancy schmancy! I like it.

PR #4943, opened 2018-02-24
Robolimbs require repair at 30 composite damage

Anewbe:
- 30 brute and burn is now the cutoff for external robolimb repair, rather than 30 brute or 30 burn. This seems like an oversight, honestly, but I'm not 100% certain.
- Sparking now actually shows that you need repair.


Atermonera:
Discussed at staff meeting, if we need to tweak the number we can tweak it

PR #4945, opened 2018-02-24
Generic aura animation proc, works on any icon

Arokha:
Generic proc that allows you to animate an aura around something. Useful to get rid of all those hardcoded 'the teleport animation is the outline of a bald human facing south' things.

I know your server doesn't have resizey things but it's still a good demo.


Anewbe:
_ponders the possibility of making gib work in a similar manner..._

PR #4948, opened 2018-02-24
Adds TG-style floor tile swapping

Nerezza:
Hold the tool for pulling a floor tile in your offhand and use a stack of floor tiles (ie, carpet) on the target floor to (almost) immediately swap the tiles.

This respects the different tools and their effects so replacing wooden tiles should be done with a screwdriver, not a crowbar.

Also adds is_holding_item_of_type() to cyborgs/drons.

### DIFF
--- a/code/game/turfs/simulated/floor_attackby.dm
+++ b/code/game/turfs/simulated/floor_attackby.dm
@@ -4,39 +4,14 @@
 		return 0
 
 	if(flooring)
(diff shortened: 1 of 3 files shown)


Atermonera:
Tested?

Nerezza:
Myself, yeah. That's why I made a flooring fixes PR to fix unrelated bugs I'd found o3od

PR #4949, opened 2018-02-24
Flooring fixes

### LINKED ISSUE #4947: Floors tiles can't be placed unless the target flooring already exists on the map somewhere
First, I'm 100% sure this is why blue (now teal) carpet doesn't work, however this bug extends to ALL floor types. Second, I recall symptoms of this on Baystation from like 3 or 4 years ago so this has likely looooong been an issue.

Basically, flooring types are defined in code/game/turfs/flooring/flooring.dm but the flooring_types list used in tile placement (in code/game/turfs/simulated/floor_attackby.dm) only populates with floors that exist in-round already.

This means that a hackish solution to floor tiles not being placeable is to toggle build mode and manually place one of the target floor type somewhere (ie, put just one teal carpet square in central and teal carpet begins working everywhere in our current build) or mapping in a room full of floor tiles we want constructible.

The real solution would be for the flooring_types list to populate with all of /decl/flooring at server start, but I have no idea how to do that myself

Nerezza:
Firstly it fixes a long-standing bugs with constructable flooring not actually populating on server startup, but instead through a proc that's only called on existing floor tiles. Fixes #4947

Second it changes sif wood tiles into a child of wood tiles now that the reason for carpet colors not working is known & fixed.

Credit to `@jade` on Discord for pointing me in the right direction for the first fix.

Edit: Oh and third it removes three 'fifty' stacks that were incorrectly defined (and duplicates), which would create tile stacks that were non-functional.

### DIFF
--- a/html/changelogs/Nerezza - Floor-fixes.yml
+++ b/html/changelogs/Nerezza - Floor-fixes.yml
@@ -1,0 +1,37 @@
+################################
+# Example Changelog File
+#
+# Note: This file, and files beginning with ".", and files that don't end in ".yml" will not be read. If you change this file, you will look really dumb.
+#
+# Your changelog will be merged with a master changelog. (New stuff added only, and only on the date entry for the day it was merged.)
+# When it is, any changes listed below will disappear.
+#
+# Valid Prefixes: 
+#   bugfix
+#   wip (For works in progress)
+#   tweak
+#   soundadd
+#   sounddel
+#   rscadd (general adding of nice things)
+#   rscdel (general deleting of nice things)
+#   imageadd
+#   imagedel
+#   maptweak
+#   spellcheck (typo fixes)
+#   experiment
+#################################
+
+# Your name.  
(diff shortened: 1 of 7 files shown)


html/changelogs/Nerezza - Floor-fixes.yml
```diff
@@ -0,0 +1,37 @@
...
+#   rscdel (general deleting of nice things)
+#   imageadd
+#   imagedel
+#   maptweak
+#   spellcheck (typo fixes)
+#   experiment
+#################################
+
+# Your name.  
+author: N3X15
```
Neerti: Think you forgot to change this.

Nerezza: That's weird, it's changed locally o3o

Nerezza: Saved it (it was unsaved) and amended c.c;

PR #4950, opened 2018-02-24
Should be able to attach accessories by clicking the target item again

Anewbe:
- Current workaround is to hit tab to switch to hotkeys and just press E.
- Removal of functionality was unintentional.
- This was tested. Hopefully better than the last one.

### DIFF
--- a/code/modules/clothing/clothing_accessories.dm
+++ b/code/modules/clothing/clothing_accessories.dm
@@ -1,11 +1,9 @@
 /obj/item/clothing/proc/can_attach_accessory(obj/item/clothing/accessory/A)
-	var/obj/item/clothing/accessory/attach = A
-	if(src.valid_accessory_slots && (attach.slot in src.valid_accessory_slots))
-		if(accessories.len && restricted_accessory_slots && (attach.slot in restricted_accessory_slots))
+	if(src.valid_accessory_slots && (A.slot in src.valid_accessory_slots))
+		if(accessories.len && restricted_accessory_slots && (A.slot in restricted_accessory_slots))
 			for(var/obj/item/clothing/accessory/AC in accessories)
-				if (AC.slot == attach.slot)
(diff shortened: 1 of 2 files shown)


code/modules/clothing/clothing_accessories.dm
```diff
@@ -1,21 +1,20 @@
 /obj/item/clothing/proc/can_attach_accessory(obj/item/clothing/accessory/A)
-	if(valid_accessory_slots && istype(A) && (A.slot in valid_accessory_slots))
-		if(accessories.len && restricted_accessory_slots && (A.slot in restricted_accessory_slots))
+	var/obj/item/clothing/accessory/attach = A
```
Leshana: Why are we assigning `A` to `attach` when it is the same type, and neither var is assigned to again?  Couldn't we just use `A` throughout the proc with the same effect?

Anewbe: For some reason it seemed to go horribly wrong without this define. I'm basically just a monkey, here, I'm not sure why it was failing.

Anewbe:
_screams internally._
_screams externally._
_screams eternally._

code/modules/clothing/clothing_accessories.dm
```diff
@@ -1,21 +1,19 @@
...
-		return 1
-	return 0
-
-	if(accessories.len && restricted_accessory_slots && (A.slot in restricted_accessory_slots))
-		for(var/obj/item/clothing/accessory/AC in accessories)
-			if (AC.slot == A.slot)
-				return 0
-	return 1
+					return FALSE
+			return TRUE
```
Atermonera: Unnecessary line, since it's in the `if` statement, if it makes it past the `for` loop, it'll run into the `return TRUE` on the line immediately below when it gets out of this `if`

PR #4952, opened 2018-02-25
Minor robolimb sprite tweaks

PrismaticGynoid:
Tweaks a number of robolimbs in minor ways, mostly to remove clipping issues where certain parts would be visible sticking through clothing (but a couple other minor tweaks too).
Unbranded: Adjusts displaced feet (were sticking out past shoes), removes a pixel on each arm and a few on the chest (was sticking through shirts)
Bishop: Adjusts displaced left foot (was sticking out past shoes)
Cyber Solutions: adjusts displaced left-east leg sprite(was sticking through pants)
Grayson: Adds the missing left-east and right-west hand sprites (because... every other brand has them)
Morpheus: Removes the feet from the leg sprites, adjusts displaced left-east and right-west legs and feet (were clipping through shoes)
Zeng-Hu: Edits the face shape to bring it more in line with the human shape, and tweaks face texture so it doesn't look like it has a 3 o'clock shadow on darker skin tones

Comparison (some changes hard to see): 
synthlimbcomparison


SubberTheFabulous:
...I was pretty sure, that I fixed the clipping leg in CS. Huh.

PR #4954, opened 2018-02-25
Adjusts PoIs

Anewbe:
- Adjusts the loot and spiders in a large number of PoIs, mostly replaced with randomizers. In some places ammo counts were lowered.
- Removes the Medical Nurse Spider. That was a terrible joke, and they were somehow less dangerous than the regular variant.
- Adjusts the fabricator tags on all of the PoI drone spawners I could find to "Unknown". With any luck, this'll let drones spawn on station again.
- The PoI with the angry drones has been made more threatening, visually. If you ignore all of the dead things and oil stains, that's on you.


code/modules/mob/living/simple_animal/animals/giant_spider.dm
```diff
@@ -78,18 +78,6 @@ Nurse Family
 	var/atom/cocoon_target
 	var/egg_inject_chance = 5
 
-/mob/living/simple_animal/hostile/giant_spider/nurse/medical
-	desc = "Furry and beige, it makes you shudder to look at it. This one has brilliant green eyes and a tiny nurse hat."
```
Belsima: these dudes being less dangerous was intentional and i'd love you forever if we kept at least one; they're a nice unique feature that you get to see at least on occasion, but i feel having even two is overuse

mistyLuminescence:
1) Can you make a comprehensive list of PoI changes this PR makes, rather than just going "it changes things" and leaving us to check what, exactly, has been changed?
2) Just because you don't find something funny, that doesn't make it a terrible joke.
3) Having checked the proposed changes to the Crashed Medical Shuttle PoI, aside from having removed the nurse spiders, you've increased the lethality of the PoI while simultaneously gutting it of virtually everything Medical would actually care about the PoI for. Although I understand you have a personal vendetta against the Medical department, I'm not sure removing all the stuff people stationside would actually care about is the best idea. Originally, we had some medical supplies for explorers, and a good amount of medical gear that stationside Med would appreciate. Now, we have, uh... really, nothing that makes the PoI actually worth tangling with mutant spiders for. It's gone from "actually worth going out of the way to find" to "if you find it, avoid it because it's not worth the hassle". If you're going to make such comprehensive alterations to it, might I suggest discussing with explorers and medical players first whether they'd actually enjoy such a change?

maps/submaps/surface_submaps/plains/plains.dm
```diff
@@ -110,10 +110,10 @@
 	mappath = 'maps/submaps/surface_submaps/plains/Field1.dmm'
 	cost = 20
 
-/datum/map_template/surface/plains/Theifc
-	name = "Theif Cave"
-	desc = "A theives stash"
-	mappath = 'maps/submaps/surface_submaps/plains/Theifc.dmm'
+/datum/map_template/surface/plains/Thiefc
+	name = "Thief Cave"
+	desc = "A thieve's stash"
```
mistyLuminescence: Thieves', for more than one, or thief's, for just one.

maps/submaps/surface_submaps/plains/plains_areas.dm
```diff
@@ -37,8 +37,8 @@
 /area/submap/Field1
 	name = "Field 1"
 
-/area/submap/Theifc
-	name = "Theif Cave"
+/area/submap/Thiefc
+	name = "Thieve's Cave"
```
mistyLuminescence: And here too. Thieves' or thief's.

Atermonera:
The bikeshed tag is appropriate here. The changes are trivial, and the disagreements also quite frankly petty given the animosity between you two. 

Changes to PoIs, their addition, and possibly their removal, are all very minor compared to, say, the change to robolimb repair thresholds, which will affect far more people than the removal of some of the loot of a PoI to bring it in line with the threat level it poses to explorers. The purpose of the changes to the Med shuttle PoI are indeed balance; the threat posed by the spiders is extraordinarily minor compared to the loot, namely a bluespace beaker and a _combat_ medkit, among other things.

Whether you think Anewbe is personally nerfing everything medical related because of some personal vendetta against medical, you, or anyone else, he's a dev. That means that balancing mechanics is his purview. Sometimes, things need to be rebalanced. If we don't nerf things when necessary, that leads to inevitable power creep, which is something to avoid from a design standpoint.

Loot on the medical shuttle with this PR:
- Fully-loaded defib's cell has been removed
- Most of a surgery set
- 5 or so random medicine
- Stasis bag
- Random medkit

A complete list of the removed items to the medical shuttle are as follows:
- 3 0-slowdown voidsuits, hidden under webs
- A combat medkit hidden under one of three random medkit spawners (You were told to pick either this OR the defib)
- A working NanoMed Plus
- About a dozen Random Medicine Spawners
- One of the fancy Health Analyzers hidden under a box
- A bluespace beaker hidden under a box of vials
- A syringe of Osteodaxon (Bone-fixing med) hidden in a crate
- Four ERT-level chem dispenser cartridges (Containing Tricord, Tramadol, Bicard, and Dermaline)
- Two medical computers (Records laptop and crew monitor). which don't really make sense on a non-NT ship

The threat currently posed by this shuttle is, again, very minor compared to PoI's with a similar volume of loot. It's basically just Nurse Spiders, which are VERY weak, and then you take ten minutes cutting open all the cocoons. Honestly, upon hearing about this list of stuff, I was reminded of Kelenius, who _often_ snuck unwanted features in with their code which required extremely close examination of any code they submitted.

If this ends up being too harsh, again, balancing stuff is our job as devs. It's very hard to exactly find the right place to balance something, so it'll get buffed or nerfed a bit when it becomes an issue until we find somewhere around where it should be, based on risk and reward.

Nerezza:
@Atermonera When you put it like that, it makes sense and I agree with the changes.

The difficulty of the mobs present should definitely dictate the value of the loot present.

mistyLuminescence:
Should've only been two voidsuits, unless I miscounted, and it was either a combat medkit or a combat defib; combat medkit and regular defib was fine, I was told. I'm not sure having things on the same tile as other things counts as "hiding" them, especially since in such a cramped PoI not every object can have its own tile. At least the cocoons part is true, nurse spiders have an unfortunate tendency to cover literally everything in cocoons.

The problem with it now is that it doesn't justify _actually visiting_; the PoI is in the mountains, so requires going out of one's way to find and mine to. Nothing in the PoI is actually useful to medical, when the intended purpose was "explorers find it, loot it and bring the stuff to med". It's now challenging enough to justify having good loot, but... without any good loot. It doesn't further explorer interaction with the station any more (they're not going to waste their time bringing back basic gear medical has access to already), and it doesn't even justify visiting for explorers - a defib, stasis bag, one random medkit and a half-dozen random medical items isn't even worth tangling with fourteen spiders (including two carriers) for. If the previous version of this PoI had too much reward for not enough risk, this is too much risk for not enough reward.

Atermonera:
The reasoning behind the use of 'hidden' is that they aren't on the topmost layer, so in the map editor, finding stuff like that requires clicking on every. single. tile. to find what's actually there. It's very easy to hide a laser cannon under a pizza-box (May not actually be, just an approximate example off the top of my head, you hopefully get the point of hiding powerful stuff under things with large sprites).

Carriers are a bit more threatening, I don't know their specific mechanics off the top of my head (Two shots with phase pistol to kill at least), but Nurses die in one shot from phase weapons, having 40 health, and are really only a source of concern if they get something to inject with eggs (Which doesn't exactly include explorers killing them in one hit).

Belsima:
I also want to note I think the med shuttle loot should be lowered, I just love the medical nurse

mistyLuminescence:
Yeah, the medical nurse is great. If we've still got the sprites, Mechoid (or anyone with buttons who likes spiders) could probably edit a regular nurse spider to have the hat, but it's not the same any more.

Fair point about the map editor, though- I placed the important stuff first, then filled in the rest after. We should probably write that down somewhere - "when mapping, have the biggest stuff on the bottom layer, and the smallest stuff on the top layer, so nothing is obscured". And carriers are the ones that explode into a bunch of tiny spiders when killed, I think.

Atermonera:
Oh and looking for stuff in the actual map files, like what git shows in diffs, is nigh-impossible unless you know EXACTLY what you're looking for, because that info is _very_ densely packed

Nerezza:
Maybe we should require a list of the most valuable loot available from new PoIs in their PRs?

mistyLuminescence:
Probably a good idea. What concerns me most is that the crashed med shuttle is in an isolated, hard-to-reach area (mountains means digging through rock to find it), has significant risk (over a dozen powerful spiders), and has nothing even remotely useful for either explorers or medical to justify it. At least previously it was worth taking a detour for.

Woodratt:
Having watched people clear the POI in the past, It's not that hard. You go forward, lure one spider to you kill it. Go forward, lure one spider to you, kill it.

Nurse Spiders are not strong spiders at all. 

As it stands this has also led to a certain unwritten down rule that we're going to start implementing in the future, which will be that devs are to be notified what potential loot a POI will have. That way we avoid these sorts of suprises.

mistyLuminescence:
Yeah, it wasn't that hard to begin with, but if we're going to make it hard, it should at least be worth it. Risk vs reward and all that.

code/modules/mob/living/simple_animal/animals/giant_spider.dm
Arokha: Leave 'em in the code. Good for adminbuse!

maps/submaps/surface_submaps/mountains/CrashedMedShuttle1.dmm
```diff
@@ -79,10 +79,10 @@
 "bA" = (/obj/structure/girder/displaced,/turf/template_noop,/area/submap/CrashedMedShuttle)
 "bB" = (/obj/structure/shuttle/engine/propulsion{tag = "icon-propulsion_r (WEST)"; icon_state = "propulsion_r"; dir = 8},/turf/simulated/floor/outdoors/rocks,/area/submap/CrashedMedShuttle)
 "bC" = (/obj/structure/shuttle/engine/heater{icon_state = "heater"; dir = 8},/obj/structure/window/reinforced{dir = 4},/turf/simulated/mineral/floor/ignore_mapgen,/area/submap/CrashedMedShuttle)
-"bD" = (/obj/structure/table/standard,/turf/simulated/shuttle/floor/purple,/area/submap/CrashedMedShuttle)
-"bE" = (/obj/structure/table/standard,/obj/item/weapon/storage/fancy/vials,/turf/simulated/shuttle/floor/purple,/area/submap/CrashedMedShuttle)
+"bD" = (/obj/structure/table/standard,/obj/item/weapon/storage/firstaid/adv,/turf/simulated/shuttle/floor/purple,/area/submap/CrashedMedShuttle)
```
mistyLuminescence: Better, but still not actually worth going out of your way to follow the signal for, and definitely not worth the risk posed by the dozen+ (fourteen, last I counted?) Actual Dangerous Spiders. Plus it removes the vials, so... not really that much better, but.

maps/submaps/surface_submaps/wilderness/spider1.dmm
```diff
@@ -3,18 +3,19 @@
 "c" = (/turf/template_noop,/area/submap/spider1)
 "d" = (/obj/structure/flora/tree/pine,/turf/template_noop,/area/submap/spider1)
 "e" = (/obj/effect/spider/eggcluster/small/frost,/turf/template_noop,/area/submap/spider1)
-"f" = (/mob/living/simple_animal/hostile/giant_spider/frost,/turf/template_noop,/area/submap/spider1)
-"g" = (/obj/structure/flora/tree/sif,/turf/template_noop,/area/submap/spider1)
-"h" = (/obj/effect/spider/cocoon,/mob/living/simple_animal/hostile/giant_spider/frost,/turf/template_noop,/area/submap/spider1)
+"f" = (/mob/living/simple_animal/hostile/giant_spider/frost{returns_home = 1},/turf/template_noop,/area/submap/spider1)
+"g" = (/mob/living/simple_animal/hostile/giant_spider/nurse,/turf/template_noop,/area/submap/spider1)
```
Atermonera: Should that spider be set to return home like the frost spiders are?

Anewbe: It was specifically set not to. It's a single nurse, the hope is that it'll wander around and make webs, so the PoI isn't two spiders and a bunch of web surrounded by a million miles of nothing else.

And, again, single nurse. Not really threatening in any way.

maps/submaps/surface_submaps/plains/plains.dm
```diff
@@ -111,8 +111,8 @@
 	cost = 20
 
 /datum/map_template/surface/plains/Thiefc
-	name = "Thief Cave"
-	desc = "A thieve's stash"
+	name = "Thieves Cave"
+	desc = "A thieves stash"
```
mistyLuminescence: thieves', with the apostrophe at the end, possessive of multiple thieves

maps/submaps/surface_submaps/plains/plains_areas.dm
```diff
@@ -38,7 +38,7 @@
 	name = "Field 1"
 
 /area/submap/Thiefc
-	name = "Thieve's Cave"
+	name = "Thieves Cave"
```
mistyLuminescence: thieves', apostrophe at the end

mistyLuminescence:
The issues I raised about crashed med shuttle all still apply. Also, thieves'. Worth a forum thread?

PR #4955, opened 2018-02-25
Fixes a bug in Promethean water ingestion

Anewbe:
- Copy paste is bad, kids.
- Credit to @Screemonster for pointing it out.


Screemonster:
The other thing I noticed in that thing - I saw it used fire_stacks.

... are fire_stacks still subtracted if they're wearing a sealed suit? Wouldn't want to see them putting on a completely airtight (and presumably watertight) outfit and yet somehow still melt like the wicked witch of the west if they go out in the rain.

Anewbe:
It is, for now. My next big project will fix that.

PR #4958, opened 2018-02-25
DNA injectors cause cloneloss.

Belsima:
Port from https://github.com/VOREStation/VOREStation/pull/921. Being hulk's gotta have some kinda downside, eh?

### DIFF
--- a/code/game/objects/items/weapons/dna_injector.dm
+++ b/code/game/objects/items/weapons/dna_injector.dm
@@ -67,6 +67,7 @@
 	if(istype(M,/mob/living))
 		var/mob/living/L = M
 		L.apply_effect(rand(5,20), IRRADIATE, check_protection = 0)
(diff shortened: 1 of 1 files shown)


Belsima:
@Anewbe May want to bring up to Higher Powers, if this matters.

Arokha:
This was mostly to prevent brute-forcing genetics because people were building the old genetics equipment, I didn't want to REMOVE that from the game, but also didn't want to make a secret genetics lab always able to make hulks

Belsima:
Yeah, I worry the same, so I liked the adjustment.

Atermonera:
If the user doesn't have any cloneloss, it'll apply 2.
After that it'll double the existing cloneloss.
Intended?

Belsima:
@Atermonera Yes, intended behavior. Meant to make getting all mutations stack Badly.

PR #4960, opened 2018-02-25
Makes tape appliable to hazard shutters just like airlocks

Nerezza:
Clicking hazard shutters with tape will now tape them up like airlocks instead of trying to open them.


Anewbe:
Tested?

Nerezza:
By myself, yes. It works exactly as one might except with no bugs added o3od

PR #4961, opened 2018-02-25
Adds new languages

elgeonmb:
- Adds Terminus, a fairly uncommon human-only language invented by Diet Racist xenoarchs
- Adds High Skrellian, a Skrell language spoken by the Wise and True Skrell on the other side of Loop 2-- and you can go to those wonderful Far Kingdoms too, comrade, if only you work hard for the glory of the City-State! Whitelisted and probably won't be open to applications until we have at least some Far Kingdoms lore.
- Adds Minbus, an adminbus language that sounds like Russian.
- Also the local xenochat is now ,u instead of ,4 because saycode is voodoo magic.

Terminus syllables from Bay12's Space. Minbus syllables from Bay12's Terran.

### DIFF
--- a/code/modules/mob/language/generic.dm
+++ b/code/modules/mob/language/generic.dm
@@ -42,7 +42,7 @@
 // Galactic common languages (systemwide accepted standards).
 /datum/language/trader
 	name = LANGUAGE_TRADEBAND
-	desc = "Maintained by the various trading cartels in major systems, this elegant, structured language is used for bartering and bargaining."
+	desc = "Spoken by the humans of the upper-class Sagittarius Heights, Tradeband was designed to be pleasing both to humans and to their Skrellian trading partners."
 	speech_verb = "enunciates"
 	colour = "say_quote"
 	key = "2"
@@ -53,10 +53,21 @@
 "nt","ti","us","it","en","at","tu","te","ri","es","et","ra","ta","an","ni","li","on","or","se",
 "am","ae","ia","di","ue","em","ar","ui","st","si","de","ci","iu","ne","pe","co","os","ur","ru")
 
+/datum/language/terminus
+	name = LANGUAGE_TERMINUS
+	desc = "A soft language spoken by the people of the sparsely populated, socially-concious Precursors' Crypt region."
+	speech_verb = "mentions"
+	exclaim_verb = "insinuates"
+	colour = "terminus"
+	key = "4"
+	syllables = list ("die", "en", "skei", "van", "son", "der", "aar", "ch", "op", "ruk", "aa", "be", "ne", "het",
+ 	"ek", "ras", "ver", "zan", "das", "waa", "geb", "vol", "lu", "min", "breh", "rus", "stv", "ee", "goe", "sk",
+ 	"la", "ver", "we", "ge", "luk", "an", "ar", "at", "es", "et", "bel", "du", "jaa", "ch", "kk", "gh", "ll", "uu", "wat")
+
 // Criminal language.
 /datum/language/gutter
 	name = LANGUAGE_GUTTER
-	desc = "There is no true language named Gutter. 'Gutter' is a catchall term for a collection of unofficial SolCom dialects that has somehow managed to spread across the stars."
+	desc = "Gutterspeak originated as a Thieves' Cant of sorts during the early colonization era. The language eventually spread from the cartels and triads to the disenfranchised people of the Bowl."
 	speech_verb = "growls"
 	colour = "rough"
 	key = "3"
--- a/code/modules/mob/language/generic.dm
+++ b/code/modules/mob/language/generic.dm
@@ -60,6 +60,7 @@
 	exclaim_verb = "insinuates"
 	colour = "terminus"
 	key = "4"
+	flags = WHITELISTED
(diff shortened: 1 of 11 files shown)


elgeonmb:
Oh shit I totally forgot: also modernizes Gutterspeak and Tradeband lore

code/modules/mob/language/generic.dm
```diff
@@ -53,10 +53,22 @@
 "nt","ti","us","it","en","at","tu","te","ri","es","et","ra","ta","an","ni","li","on","or","se",
 "am","ae","ia","di","ue","em","ar","ui","st","si","de","ci","iu","ne","pe","co","os","ur","ru")
 
+/datum/language/terminus
+	name = LANGUAGE_TERMINUS
+	desc = "A soft language spoken by the people of the sparsely populated, socially-concious Precursors' Crypt region."
```
Atermonera: conscious

code/modules/mob/language/generic.dm
```diff
@@ -53,10 +53,22 @@
...
+	flags = WHITELISTED
+	syllables = list ("die", "en", "skei", "van", "son", "der", "aar", "ch", "op", "ruk", "aa", "be", "ne", "het",
+ 	"ek", "ras", "ver", "zan", "das", "waa", "geb", "vol", "lu", "min", "breh", "rus", "stv", "ee", "goe", "sk",
+ 	"la", "ver", "we", "ge", "luk", "an", "ar", "at", "es", "et", "bel", "du", "jaa", "ch", "kk", "gh", "ll", "uu", "wat")
+
 // Criminal language.
 /datum/language/gutter
 	name = LANGUAGE_GUTTER
-	desc = "There is no true language named Gutter. 'Gutter' is a catchall term for a collection of unofficial SolCom dialects that has somehow managed to spread across the stars."
+	desc = "Gutterspeak originated as a Thieves' Cant of sorts during the early colonization era. The language eventually spread from the cartels and triads to the disenfranchised people of the Bowl."
```
Anewbe: It's just `Gutter`.

code/modules/mob/language/generic.dm
```diff
@@ -42,7 +42,7 @@
 // Galactic common languages (systemwide accepted standards).
 /datum/language/trader
 	name = LANGUAGE_TRADEBAND
-	desc = "Maintained by the various trading cartels in major systems, this elegant, structured language is used for bartering and bargaining."
+	desc = "Spoken by the humans of the upper-class Sagittarius Heights, Tradeband was designed to be pleasing both to humans and to their Skrellian trading partners."
```
Atermonera: `to humans and their Skrellian`

code/modules/mob/language/outsider.dm
```diff
@@ -115,3 +115,17 @@
 		return 1
 
 	return 0
+
+//for your antag purposes. machine-translatable, if you don't want that use Cult or one of the animal langs
```
Anewbe: Machine-translatable? Not sure that's a feature?

code/modules/mob/language/outsider.dm
```diff
@@ -115,3 +115,17 @@
...
+/datum/language/minbus
+	name = LANGUAGE_MINBUS
+	desc = "The Powers That Be have seen it fit to grace you with a special language that sounds like Russian for some reason."
+	speech_verb = "says"
+	ask_verb = "asks"
+	exclaim_verb = "shouts"
+	colour = "deadsay"
+	key = "r"
+	flags = RESTRICTED | WHITELISTED
```
Anewbe: Not 100% sure what the WHITELISTED flag does, but I'm not sure it should be here.

code/modules/mob/language/station.dm
```diff
@@ -85,6 +85,20 @@
 	flags = WHITELISTED
 	syllables = list("qr","qrr","xuq","qil","quum","xuqm","vol","xrim","zaoo","qu-uu","qix","qoo","zix")
 
+/datum/language/skrellfar
```
Anewbe: Doesn't look like Skrell can actually get this, is that intentional?

code/modules/mob/language/outsider.dm
elgeonmb: has machine_can_understand = 1 like most languages, so the translator still works on it.

code/modules/mob/language/outsider.dm
elgeonmb: yeah WHITELISTED is probably unnecessary but it also doesn't hurt anyone

code/modules/mob/language/station.dm
elgeonmb: yeah. whitelisted for Skrell too, possibly subject to change, but my Plan with the Far Kingdoms does involve putting a lot more trust into players to portray it correctly than I expect from current Near Kingdom Skrell.

code/modules/mob/language/outsider.dm
Anewbe: Yeah, if this is admin use only I'm not certain that's desirable.

Anewbe: I don't really care, I don't see myself using it, though.

PR #4963, opened 2018-02-26
Adds ice puzzle POIs

BillyBangles:
Introduces Ice Puzzles to the Caves. Technically they're all the same POI, but the actual puzzles within are different, for some variety. Won't be as easy to memorize the solution when there's several different variations of the same puzzle. Puts the new slippery ice to good use, I'd say. They're not too difficult, so expect some low to mid tier loot.

Also adds an alternative to dungeon wall that looks more like rock, for that cheese-proof cave feel. The edges are kinda broken though so if someone could suggest a fix for that I'd appreciate it. No bugs other than decorative, though. And finally, adds ice crystals, which are just an alternative to the normal crystals. The sprites used are placeholder for now, pending nicer ones.

No pictures, this time. No spoilers. <3


Atermonera:
In keeping with the new and not-entirely-encoded dev policy concerning PoIs, please list all loot and all challenges to getting that loot (Aside, from obviously, the ice puzzles, although some indication of how difficult they are would be appreciated)

BillyBangles:
The loot comes in the form of four bone loot piles which could potentially drop any of the items associated with them, one random ammo drop, and one random medkit drop.

The challenge is the puzzles, like you said. No mobs here. They're not really all that difficult. If you throw yourself at it long enough you'll figure it out within a decent amount of time, especially if you're good at these sorts of puzzles. I figured a medkit and a very small chance to get something shiny from the bone piles was some reasonable low-tier loot.

BillyBangles:
For reference, I showed Ice Cave 1A to anewbe and he worked it out in his head within like 2 minutes. In game it will take a little bit longer but that's because A. You can't see the whole puzzle on one screen, and B. The ice stuns you for about 2 seconds every time you slide.

PR #4969, opened 2018-02-26
Alternate Chaplain Items

Belsima:
Adds Kirpan and Prayer Beads as alternatives to the Null Rod, ported from Paradise and Aurora. If you have other suggestions, please tell- this is up so I can receive Dev opinion on the current state, and I would like it to be DNM.


Belsima:
Holding for bigger chaplain plans. Better to roll together.

PR #4981, opened 2018-02-28
Adds 'Intercom Convo' verb

Arokha:
Allows an admin to 'schedule' up to 20 messages to play on the radio.
The syntax is printed to your chat box for helping you out if you've never used it or forget how it works, but you just do repeats of:
```
Sender Name|Message Body
Delay Time
```

So, for example:
```
Blooper|Gosh I love not being a fox.
5
Bleeper|Wow you suck, foxes are the best!!!
5
Foxes|We are the best, aren't we
```
Says those 3 messages, with 5 second gaps between them:
image

While I was in there, I fixed:
* The global announcer is now able to talk on any channel, which fixes anything that previously was trying to send messages on specific channels (morgue trays is one example)
* Some dumb relative pathing in radio.dm

### DIFF
--- a/code/game/objects/items/devices/radio/intercom.dm
+++ b/code/game/objects/items/devices/radio/intercom.dm
@@ -49,6 +49,12 @@
 	name = "entertainment intercom"
 	frequency = ENT_FREQ
 
+/obj/item/device/radio/intercom/omni
+	name = "global announcer"
+/obj/item/device/radio/intercom/omni/initialize()
+	channels = radiochannels.Copy()
+	return ..()
+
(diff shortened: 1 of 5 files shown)


mistyLuminescence:
Suggestion: a way to change the speech verb so it's not always 'states'.
Suggestion 2: support for intercom emotes over radio (e.g. audible coughs).

Anewbe:
Out of scope.

Arokha:
The autosay proc just doesn't support those. I do like the suggestions but ye, not in this PR.

PR #4984, opened 2018-02-28
De-removes recapitation defibrillation

mistyLuminescence:
**WIP**

TL;DR version: A while ago, if someone was decapitated, it'd sometimes be possible to take their head and reattach it to their body, and then defibrillate them. This was removed in #4264 ; this PR is effectively a revert of that PR. To be discussed in [this forum thread](https://forum.ss13polaris.com/viewtopic.php?f=14&t=500).

- [X] Get recapitation defibrillation working
- [ ] Add a complex (many steps) surgery
- [ ] Other tweaks (for both patient and doc)

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -15,7 +15,6 @@
 	var/spit_name = null 				//String
 	var/last_spit = 0 					//Timestamp.
 
-	var/can_defib = 1					//Horrible damage (like beheadings) will prevent defibbing organics.
 	var/hiding = 0						// If the mob is hiding or not. Makes them appear under tables and the like.
 	var/active_regen = FALSE //Used for the regenerate proc in human_powers.dm
 	var/active_regen_delay = 300
--- a/code/game/objects/items/devices/defib.dm
+++ b/code/game/objects/items/devices/defib.dm
@@ -289,7 +289,7 @@
 		if(H.health + H.getOxyLoss() + H.getToxLoss() <= config.health_threshold_dead)
 			return "buzzes, \"Resuscitation failed - Severe damage detected. Begin manual repair before further attempts futile.\""
 
-	else if(H.health + H.getOxyLoss() <= config.health_threshold_dead || (HUSK in H.mutations) || !H.can_defib)
+	else if(H.health + H.getOxyLoss() <= config.health_threshold_dead || (HUSK in H.mutations))
 		return "buzzes, \"Resuscitation failed - Severe tissue damage makes recovery of patient impossible via defibrillator. Further attempts futile.\""
 
 	var/bad_vital_organ = check_vital_organs(H)
--- a/code/modules/organs/organ.dm
+++ b/code/modules/organs/organ.dm
@@ -89,7 +89,6 @@ var/list/organ_cache = list()
 	processing_objects -= src
 	if(owner && vital)
 		owner.death()
-		owner.can_defib = 0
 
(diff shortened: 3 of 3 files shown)


Wickedtemp:
Anewbe said to open up a PR if we disagreed with it, so... That's what this is.

Curt5:
In favor of re-adding recapitations. It's one of those times where 'It's not a bug, it's a feature' is actually true.

Mechoid:
I'm going to point out, removing the can_defib is not advisable to do this, if you're going to be touching this. I'll be posting my own thoughts in the thread in a moment.

mistyLuminescence:
There's probably a more elegant way of fixing it, yeah. Maybe setting can_defib to 1 when the head gets reattached? You still can't defib someone without a head, at least, because they're missing a vital organ (brain). Thoughts on how to implement it would definitely be appreciated.

Mechoid:
I've posted in the thread, and I'm not personally comfortable discussing my own proposal for making a re-capitation defibbable, due to and more than what I've noted in the thread.

Atermonera:
Controversial because this will probably take a lengthy discussion, regardless of outcome.
[Forum Thread](https://forum.ss13polaris.com/viewtopic.php?f=14&t=500)

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -15,7 +15,6 @@
 	var/spit_name = null 				//String
 	var/last_spit = 0 					//Timestamp.
 
-	var/can_defib = 1					//Horrible damage (like beheadings) will prevent defibbing organics.
```
Belsima: wow removing this is an absolutely terrible idea please do not

Arokha:
It looks like a lot of the comments in favor here and in the thread are 'because this thing is hard to do, I want to be able to do it and get a reward', which is a bit nonsensical. You could say the same about virtually anything not possible in the game if it's hard (really, just if 'you think' it's hard), and if that's your single factor arguing in favor of it, then it's probably not actually worth doing.

Another thing I noticed is that people seem to be confusing the OOC difficulty and IC difficulty in doing this in their arguments. I see several people saying 'no, this isn't that easy' but are referring specifically to the OOC game mechanical difficulty of it and likelihood of being able to accomplish it, while others are saying 'too easy' referring to the in-universe seeming ease of attaching a limb requiring virtually no effort.

That aside, I do agree with Mechoid's comments in the thread. The tools available in surgery are insufficient to even vaguely represent the in-universe difficulty one would encounter when reattaching limbs. Maybe it should require the IMS to do, plus drugs, plus more steps or something, who knows.

mistyLuminescence:
> TL;DR: Keep it, because:
* It's cool.
* It requires lots of skill and luck, so doesn't happen all the time.
* It's easy to shut down if the antag takes the time to disposal/space/husk/maul/hide the body.
* Removing it, but keeping it for synths, widens the power gap between FBP and organic.
* It's not unrealistic, given the setting and technology level.
* It rewards medical players for putting in effort to save the patient.
* When it's relevant, it doesn't take anything away from the server, but adds a lot.

From the original PR. It's not so much as "we should be able to do it because it's hard", it's "we should be able to do it because it's awesome and rewarding, but it should be hard so it doesn't happen all the time" (it already was hard before it was removed). Adding a complex surgery would go a long way towards making this more thematic, more fun, and more awesome.

mistyLuminescence:
There's something even better coming Soon ™️ - stay tuned and keep an eye on the forum thread. You'll know it when you see it. Keeping the forum thread open 'cause the topic's still going, it just won't be implemented with this PR specifically.

PR #4986, opened 2018-02-28
Hail is less omnicidal

Anewbe:
Brought to my attention in: https://forum.ss13polaris.com/viewtopic.php?f=14&t=495. Significantly tones down hail.

- Hail no longer affects simple mobs or borgs. Humans should still find shelter.
- Hail has a 10% chance to turn into more hail, down from 40%.


MisterLayne:
Does this mean that explorers will actually get to kill the spiders, now?

Anewbe:
Yep.

PR #4987, opened 2018-03-01
VeyMed limbs (and organs) are now more vulnerable to EMP

Anewbe:
VeyMed is supposed to be a very, very complex design. It'd have to be to fool those ~~MedHuds~~ eyeballs, which ~~I'm pretty sure~~ I know it it can do? So makes sense a pulse that damages complicated electronics would especially affect a VeyMed.

- Mechanically, the amount of damage you take from EMP is now multiplied by the robotic var of the affected organ. The average limb or internal organ = 2, assisted organs = 1, and VeyMed (or more accurately, lifelike limbs) = 3.
- EMP damage has been lessened to compensate for this change. Normal limbs and organs should take roughly the same amount of damage, depending on RNG.


Anewbe:
Even VeyMed? I could have sworn there was a PR that changed that.

Anewbe:
You saw nothing.

Andromeda-K22:
Why is there not a forum thread regarding this?
This'll just make people stop choosing VeyMed limbs.

Atermonera:
That is intentional, they're supposed to be extraordinarily expensive and delicate, and we currently have a surplus in use among people across the various economic classes, even though it should be exclusive to Wealthy (Or very maybe but probably not above-average)

Andromeda-K22:
Then, instead of nerfing limbs so nobody will use them, why not...just,
make it so people have the Wealthy economic status (or above) can use them?

This should still go to a  forum debate. Regardless.

Belsima:
People will use them plenty; the upsides of looking exactly like a human often outweigh mechanics from all I can tell.

Nerezza:
Brand fluff on the wiki would probably help a bit. Explain which prosthetics are expensive, which are cheap, and what each brand is geared towards.

Right now people may not be aware that VeyMed's supposed to be super expensive.

PrismaticGynoid:
I went to sleep being entirely against this, and woke up being entirely for it. It makes sense that Vey-Med limbs, which probably have advanced sensors built in to mimic human touch, would be more affected by an electromagnetic pulse (and they probably don't have any shielding from that, since it's not exactly a common thing for most people to encounter). It would actually give them a mechanical trade-off for not looking obviously robotic, and I think would help reduce the number of people who use it in certain roles. Also I just like the idea of assisted organs taking less damage from EMPs, which this seems to do.

PR #4988, opened 2018-03-01
Removes some alt-titles with insufficiently clear roles

Anewbe:
As part of an ongoing effort to write better guides for our jobs and alt-titles, we are clearing out alt-titles that don't have clear differences from their default roles, are rarely used, are inaccurate to the current station, or imply things that should not be implied. This is the first of a few PRs that will whittle down the alt-titles list into something that is clear and concise in what is expected of the role.

### DIFF
--- a/code/game/jobs/job/civilian.dm
+++ b/code/game/jobs/job/civilian.dm
@@ -49,7 +49,7 @@
 	minimal_access = list(access_hydroponics)
 
 	outfit_type = /decl/hierarchy/outfit/job/service/gardener
-	alt_titles = list("Hydroponicist", "Gardener")
+	alt_titles = list("Gardener")
 
 //Cargo
 /datum/job/qm
@@ -105,7 +105,7 @@
 	minimal_access = list(access_mining, access_mining_station, access_mailsorting)
 
 	outfit_type = /decl/hierarchy/outfit/job/cargo/mining
-	alt_titles = list("Drill Technician","Prospector")
+	alt_titles = list("Drill Technician")
 
 //Service
 /datum/job/janitor
@@ -123,7 +123,7 @@
 	minimal_access = list(access_janitor, access_maint_tunnels)
 
 	outfit_type = /decl/hierarchy/outfit/job/service/janitor
-	alt_titles = list("Custodian", "Sanitation Technician")
+	alt_titles = list("Custodian")
 
 //More or less assistants
 /datum/job/librarian
@@ -141,7 +141,7 @@
 	minimal_access = list(access_library)
 
 	outfit_type = /decl/hierarchy/outfit/job/librarian
-	alt_titles = list("Journalist", "Professor", "Historian", "Writer")
+	alt_titles = list("Journalist", "Writer")
 
 //var/global/lawyer = 0//Checks for another lawyer //This changed clothes on 2nd lawyer, both IA get the same dreds.
 /datum/job/lawyer
--- a/code/game/jobs/job/captain.dm
+++ b/code/game/jobs/job/captain.dm
@@ -87,4 +87,3 @@ var/datum/announcement/minor/captain_announcement = new(do_newscast = 1)
 	minimal_access = list(access_heads, access_keycard_auth)
 
 	outfit_type = /decl/hierarchy/outfit/job/secretary
-	alt_titles = list("Command Liaison", "Bridge Secretary")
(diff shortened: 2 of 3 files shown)


code/game/jobs/job/civilian.dm
```diff
@@ -141,7 +141,7 @@
 	minimal_access = list(access_library)
 
 	outfit_type = /decl/hierarchy/outfit/job/librarian
-	alt_titles = list("Journalist", "Professor", "Historian", "Writer")
```
Belsima: I find Historian to be a rather nice role; I actually have a character where I do the obvious and... write about History, work with anomalists, etc. 'Professor' doesn't make sense, obviously but I find Historian to be rather pleasant. I do see an argument against it at the same time, so removing it seems fair, but I also see little benefit from removing it; there doesn't seem to be much gain for what would disappoint me and a couple others I know.

Belsima: Also, smh no bikeshed label?

MisterLayne: @Anewbe I agree. I think Historian would be nice to keep. It's not really an unclear role. It provides a title for any character who knows extensively about the History of the universe.

Cerebulon: It may be a word with a clear meaning, but it isn't a job on the station that has any specific role. (Primarily) A librarian tends to the library and loans books, a journalist reports the news and a writer writes books.
Being an expert on something academic doesn't give you something specific to do. Anybody, in any job can be an expert on something like that. Pick the job/title which most closely matches what your character actually does on-station.
Also I have seen it used maybe once as a one-off.

Atermonera: Liberal arts degrees are lumped together because, well, they're liberal arts degrees. There's much less difference between them than between, say, mechanical engineering and chemical engineering, even though the latter two are both part of the 'engineering' field.
To be fair, as an engineering student I may be biased in that regard, but my impression is that while the content of different liberal arts degrees may differ, the general technical skills - analytical reading and whatever else they do - are much more similar between degrees than they are for engineers. ChemE's and MechE's don't need nearly as much knowledge of electronic stuff as EE's (Electrical engineers)

MisterLayne: @Cerebulon That's a fair point. Wish we could give the librarian something to actually do so they would be played more, honestly.

Belsima: github ate my comment i think its not displaying here

Gist of it was that Historians ended up being where you go to ask lore questions, except IC- which was an interaction I greatly enjoy. Having that cut out would remove a lot of the fun I have playing in the Library, which is being the person to go for information like that.

MarinaGryphon: I would also be disappointed to see the historian role removed. Professor, perhaps, but on a research station, especially in an area as far out (relative to Sol) as Sivian orbit, it's important to keep history in mind. This includes knowing the lore and history of other species, which is important to keep in mind. They would be able to advise the crew on what they believe is the best course of action, and provide a historical context for events that happen on the station. It's also important to remember that history is being made as we speak, with things like the Almach Association, the creation of Prometheans, and so on.

MarinaGryphon: Is there any response to these concerns? I'd hate to see this merged just because it was inactive for 24 hours, which I believe happened to a few other more controversial/bikeshed PRs. Historians definitely have a place on the station.

Belsima: Yeah, historians as the 'IC lore person' is actually a rather important role at times, especially as the lore develops and we currently try to integrate it further into the game, this'd be a step back.

Atermonera: I'll kick your concerns staff-side. I think they were addressed a bit in internal discussion, but nowhere near to the volume of opinion you folks are giving

Belsima: I want to just clarify; Writer and Historian have much of the same job. The question is one of perception. As a historian, others tend to perceive you as the way to go to learn of lore, leading to interactions _very_ different from playing a writer.

Atermonera: Right, so, staff consensus remains, if you want to play a librarian/writer with a deeper knowledge of history than most folks, you're free to establish yourself as such ICly.
Furthermore, nobody's seen a historian in months, so as far as we're aware, it's not actively in use. (And starting to play one now to try and drive a point would seem like you're just trying to make evidence in your favor, tbh)

Atermonera:
Discussed at the staff meeting, consensus is go-ahead on axing everything in this pr.

PR #4989, opened 2018-03-01
Makes drills report faults over supply channel

Screemonster:
Drills now have an inbuilt radio that reports over the supply channel when they have an error, so miners are alerted to go and fix them. Previously it was a visible message only, so the miners would have to be physically present at the moment they fail to get the error message.

In order that miners know which drill just failed, they can use a multitool to give it a 3-digit ID number which will be appended onto the name.

Covers a missing line where the "resources depleted" error message wasn't being called in circumstances where it _found_ minerals when it was started, but they were subsequently depleted.

Fixes a runtime in the while loop in process() - the resource_field list was having items removed before pick() was called on it, which threw an error when the last item on the list was removed.

### DIFF
--- a/code/modules/mining/drilling/drill.dm
+++ b/code/modules/mining/drilling/drill.dm
@@ -61,7 +61,7 @@
 	if(!active) return
 
 	if(!anchored || !use_cell_power())
-		system_error("system configuration or charge error")
+		system_error("System configuration or charge error.")
 		return
 
 	if(need_update_field)
@@ -99,7 +99,7 @@
 		for(var/metal in ore_types)
 
 			if(contents.len >= capacity)
-				system_error("insufficient storage space")
+				system_error("Insufficient storage space.")
 				active = 0
 				need_player_check = 1
 				update_icon()
@@ -135,7 +135,7 @@
 		active = 0
 		need_player_check = 1
 		update_icon()
-		system_error("resources depleted")
+		system_error("Resources depleted.")
 
 /obj/machinery/mining/drill/attack_ai(var/mob/user as mob)
 	return src.attack_hand(user)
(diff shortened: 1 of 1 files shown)


code/modules/mining/drilling/drill.dm
```diff
@@ -131,12 +135,19 @@
...
 /obj/machinery/mining/drill/attack_ai(var/mob/user as mob)
 	return src.attack_hand(user)
 
 /obj/machinery/mining/drill/attackby(obj/item/O as obj, mob/user as mob)
 	if(!active)
+		if(istype(O, /obj/item/device/multitool))
+			var/newtag = text2num(sanitizeSafe(input(user, "Enter new ID number or leave empty to cancel.", "Assign ID number") as text, 4))
+			if(newtag)
+				name = "[initial(name)] #[newtag]"
+				user << "<span class='notice'>You changed the drill ID to: [newtag]</span>"
```
Belsima: to_chat(user, "<span class='notice'>You changed the drill ID to: [newtag]</span>")

code/modules/mining/drilling/drill.dm
```diff
@@ -234,6 +245,7 @@
 
 	if(error)
 		src.visible_message("<span class='notice'>\The [src] flashes a '[error]' warning.</span>")
+		faultreporter.autosay(error, src.name, "Supply")
```
Belsima: Could you give an example of the message this gives?

Screemonster: image

Belsima: that looks super off but i dont have solutions so i guess i wont complain, its an upgrade either way

Atermonera: Fix the error messages to be properly capitalized?

Mechoid:
May I suggest giving mining modules for borgs a multitool, so they can give the ID as well?

Screemonster:
Touching borgs is out of the scope of this PR, since multitools can do a lot more than just mess with drills. "should minerborgs be able to reprogram the telecoms relay and open abandoned crates" is a separate discussion.

code/modules/mining/drilling/drill.dm
Screemonster: Done.

PR #4990, opened 2018-03-01
Adds a Frankenstein Surgical Procedure

Mechoid:
Recapitation surgical procedure, takes around four to five minutes to complete, and requires patience during the long-duration steps, and post-operation treatment.

Recapitation and surgical reattachment of the brain to the body will give the 'surgically reattached brain' modifier, or, as it is lovingly named codewise, 'Franken Sickness'. Once the modifier expires, the sickness will be replaced with a permanent neurological-only modifier, which maintains the inaccuracy and pain levels.
If you do somehow die, and you want to be defibbed instead of cloned afterward, you will have to have the surgery done *again*, as trauma will result in failure of the connection, adding yet another stack of the modifiers if you undergo it.

The speed-buffs on Borgs, and Alien tools, will make this much 'quicker' but the time limit is rather short in either case, depending on outside factors.

May be part of a larger 'Package', will know Soon TM


Andromeda-K22:
disagree with the five minute long surgery time - unless it resets the defib counter? this is basically worthless.

Nalarac:
So if you simply want to reattach the head so you can proceed with cloning, you do or do not need to do this surgery operation?

Anewbe:
From what I can see @Nalarac, that part hasn't changed.

Belsima:
What advantages would this have over cloning?

Nalarac:
it would have the advantages of being defibbed over being cloned.

Atermonera:
The effects of the modifier are listed in `code/modules/mob/_modifiers/cloning.dm`, which as of time of writing is the first file of the diff.
Each aspect of the modifier is commented appropriately to indicate in-game effects.

The modifier appears to be generally weaker than cloning sickness, which makes sense, because there's a limited window in which the procedure will actually be useful, because defibs.
That said, if you for whatever reason get your head transplanted OUTSIDE of a decap in the process of becoming dead, you'll still get the modifier because, well, your head got transplanted, what do you expect? It's gonna be 100% fine?

Belsima:
I mean- why do this instead of cloning? Only roleplay reasons? I'm quite happy with that, too, just asking.

PrismaticGynoid:
Some people have DNCs on their records, or have the cloning incompatibility trait.

mistyLuminescence:
Roleplay reasons, and cloning sickness is a lot more punishing mechanically than frankensickness. I'll write up my final thoughts when the PR's finished, but so far I like what I see very much.

mistyLuminescence:
> Like one who, on a lonely road,
> Doth walk in fear and dread,
> And, having once turned round, walks on,
> And turns no more his head;
> Because he knows a frightful fiend
> Doth close behind him tread.
> [Coleridge’s ‘Ancient Mariner.’]

Okay, final thoughts: this is incredible and I love it. It's interesting, challenging, _awesome_, and it's only going to happen twice a year but I love it anyway. 👍

Anewbe:
Did you test all of this?

Mechoid:
Tested it as much as I could think of.

PR #4996, opened 2018-03-02
Fixes drone borgs spawning with dropped IDs

Verkister:
Someone had moved the thing back to the outcommented older code part without updating the old code there.


Anewbe:
Did you test this?

Verkister:
Yeah worked as supposed. The line was basically copied from the pre-change equip_survival_gear() version. The older drop prone code did equip the ID onto the uniform accessory slot automatically but the newer version just made it spawn in the backpack, which I think was the normal thing since the first move anyway.

PR #5005, opened 2018-03-04
Mob Spawners can now override certain vars.

Mechoid:
Random Mob Spawners can now overwrite certain vars on the mobs they spawn.

### DIFF
--- a/code/game/objects/random/random.dm
+++ b/code/game/objects/random/random.dm
@@ -1598,9 +1598,12 @@ var/list/multi_point_spawns
 				prob(45);/mob/living/simple_animal/hostile/shantak)
 
 /obj/random/mob/spider
-	name = "Random Spider"
+	name = "Random Spider" //Spiders should patrol where they spawn.
 	desc = "This is a random boring spider."
 	icon_state = "guard"
+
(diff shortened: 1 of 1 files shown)


Anewbe:
Spiders should probably return home too.

Arokha:
Stealin' muh ideas! How dare

code/game/objects/random/random.dm
```diff
@@ -1540,11 +1550,33 @@ var/list/multi_point_spawns
 				prob(3);/mob/living/simple_animal/parrot,
 				prob(1);/mob/living/simple_animal/giant_crab)
 
+/obj/random/mob/spawn_item() //These should only ever have simple mobs.
+	var/build_path = item_to_spawn()
+
+	var/mob/living/simple_animal/M = new build_path(src.loc)
+	M.ai_inactive = 1 //Don't fight eachother while we're still setting up!
+	if(overwrite_faction)
```
Anewbe: I feel like it'd be simpler to define overwrite_faction as null, then after this check have M.faction set overwrite_faction. That way a mapper only needs to set one var to set up the randomizer.

PR #5008, opened 2018-03-04
we now know who made gateways

elgeonmb:
this is old lore and should feel bad about itself


Belsima:
its been five hours. travis is still working. is she okay.

PR #5015, opened 2018-03-05
Makes slightly smaller objects capable of bashing open busted APCs

Nerezza:
Currently broken APCs have to be busted open by a weapon with a force of 5+ and a weight class of Normal or higher. Basically, fire extinguishers.

This PR changes the weight class (not the force) requirement to Small or higher, so that tools like wrenches can bust open broken APCs.

This is partly to alleviate the tediousness of replacing broken APCs, and partly because what's acceptable for bashing open a broken APC can be difficult to discern and allowing a wider variety of 'weapons' for the task is likely to make it easier on fresher engineers. As a bonus the failure line now suggests that the chosen weapon is the issue, though as it is now that isn't always the case.

Full disclosure: This hasn't been tested, but the nature of the changed code allows me to guaruntee it's bug-free o3od

### DIFF
--- a/code/modules/power/apc.dm
+++ b/code/modules/power/apc.dm
@@ -674,7 +674,7 @@
 				istype(W, /obj/item/weapon/wirecutters) || istype(W, /obj/item/device/assembly/signaler)))
 				return src.attack_hand(user)
 			//Placeholder until someone can do take_damage() for APCs or something.
-			to_chat(user,"<span class='notice'>The [src.name] looks too sturdy to bash open with the [W.name].</span>")
+			to_chat(user,"<span class='notice'>The [src.name] looks too sturdy to bash open with \the [W.name].</span>")
 
 // attack with hand - remove cell (if cover open) or interact with the APC
 
--- a/code/modules/power/apc.dm
+++ b/code/modules/power/apc.dm
@@ -656,7 +656,7 @@
 		if ((stat & BROKEN) \
 				&& !opened \
 				&& W.force >= 5 \
-				&& W.w_class >= ITEMSIZE_NORMAL )
+				&& W.w_class >= ITEMSIZE_SMALL )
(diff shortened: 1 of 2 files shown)


Nerezza:
This bashing applies to APCs broken by explosions/meteors, by the way. Aside from those (and the ultra-rare case of a gridcheck event occurring without the gridchecker machine present n the grid) you shouldn't really have APCs in this broken, bashable state.

This is pretty much just a change to a "reconstruct a ---ed up area of the station" step that can often be frustrating, since almost literally everything else you're working on doesn't require you to hunt down some form of heavy weapon to dismantle.

code/modules/power/apc.dm
```diff
@@ -674,7 +674,7 @@
 				istype(W, /obj/item/weapon/wirecutters) || istype(W, /obj/item/device/assembly/signaler)))
 				return src.attack_hand(user)
 			//Placeholder until someone can do take_damage() for APCs or something.
-			to_chat(user,"<span class='notice'>The [src.name] looks too sturdy to bash open.</span>")
+			to_chat(user,"<span class='notice'>The [src.name] looks too sturdy to bash open with the [W.name].</span>")
```
Anewbe: \the , does grammar things.

Anewbe:
And you really, really should be testing this.

Nerezza:
I just tested it, it works

Nerezza:
Also I went ahead and checked everything that calls set_broken() on APCs, the proc that has to be called before this APC bashing can be done. The callers are:

1. Explosions
1. Gridcheck
1. Trying to modify an active SMES (DON'T try to modify an active SMES, trust me I just looked at what it does and it isn't pretty)
1. Malf AIs have a 1% chance to bust APCs with an electrical pulse, and if they drain the power from an APC completely while using Machine Overload

PR #5018, opened 2018-03-06
Syringe adjustments

Anewbe:
Port and minor tweak of: https://github.com/VOREStation/VOREStation/pull/3174

- Injecting someone with a syringe now attempts to empty the syringe in 5 unit increments, without having to click a billion times.


Wickedtemp:
Seems to work well enough on virgo. To be honest there wasn't a whole lot of difference between it and how it was before from a gameplay perspective.

PR #5023, opened 2018-03-06
Attaches Tajaran tails to 3 more station rigsuits

Nerezza:
Adds cat tails to RD, EVA, and Industrial suits:
rigtailsexample
(Tiny tweaks since this preview was put together and unsealed variations are unpictured, but there)

I also did Unathi sprites too, but hated how they came out so I'll be redoing them (later this week or next week) before submitting those.

As the preview suggests, these were loaded into the example map and tested fine o3od


Nerezza:
Just a final adjustment. After a couple days I realized the bends in north-facing sprites for the EVA and Industrial suits looked weird so I smoothed those out. All good now!

PR #5025, opened 2018-03-07
Fixes for gripper interaction with frame parts and mecha equipment

Heroman3003:
- Added new proc to grippers: drop_item_nm. Functinally same as drop_item, except its not a verb and does not display a message to the user.

- Added a drop_item proc functionality to silicon/robot. If robot has active module and its a gripper, calls drop_item_nm on that gripper.

- Rewritten frame parts to use user instead of usr.

- If frame parts are used from a gripper, they're dropped before deletion.

This fixes weird interactions with exosuit parts (installing same part onto a mech several times while also keeping it in a gripper).
Also fixes bug where sometimes frame parts would get stuck in a gripper and could produce several new bugged frames before disappearing after a while.
Could potentially make it easier to interact with objects calling drop_item with grippers in the future.


Heroman3003:
Added changelog

Heroman3003:
Okay, any clarifications on how to add in-game changelog, as it appears there is no explanation for that anywhere.

Atermonera:
https://github.com/PolarisSS13/Polaris/blob/master/html/changelogs/example.yml

Copy that file, rename it to `[username] - [general categorization of changes].yml`
Fill out the author and changes section with appropriate information, push new file to the pr

code/game/machinery/wall_frames.dm
```diff
@@ -29,71 +29,75 @@
...
 
 	if(!(ndir in cardinal))
 		return
 
-	var/turf/loc = get_turf(usr)
+	var/turf/loc = get_turf(user)
 	var/area/A = loc.loc
 	if(!istype(loc, /turf/simulated/floor))
-		usr << "<span class='danger'>\The frame cannot be placed on this spot.</span>"
+		user << "<span class='danger'>\The frame cannot be placed on this spot.</span>"
```
Atermonera: Please update these to `to_chat(user, [message])`

PR #5027, opened 2018-03-07
Couple borg UI fixes.

Verkister:
-Fixes missing default module UI icons.
-Fixes the UI module icons breaking on relog. (oh please how could you possibly even start a proc with ..() when the parent is nothing but a single return???)


Atermonera:
If it's just a single return it doesn't really matter where it gets put.
That said, if at some point the super where to actually do something, the order of child-parent proc execution may become incredibly important. I don't have an example of this off the top of my head, but that's probably why it was put at the top to begin with.

Verkister:
Afaik a single return at the very beginning will stop the proc before it even happens and moving the thing to the bottom sure fixed the bug. The order is always from top to bottom with this stuff and ..() basically means where the entire parent proc goes.

Atermonera:
Nnnnope, it'll just return out of that specific function call, same as any other return.
The ..() is calling a function of the same name on the parent type, to my understanding.

I've definitely seen supers used at the start of a large volume of functions that work just fine

Atermonera:
And, I guess as a note, it does this because calling a function is a lot easier than telling the compiler to remember where each function is and then copy that in place of any supers, OR to scoop out a function's memory and replace it with another function (as the exec() family does in C)

Verkister:
Well this one definitely was not one of those that work just fine. Tested and confirmed broken on top, fixed at bottom in this particular case.

PR #5028, opened 2018-03-07
Ward-Takahashi names + Northern Star

Belsima:
Ward-Takahashi now have stellar objects for names, and ships now no longer visit the station we're on.


Woodratt:
We're not on the Northern Star? I'm not sure why ships going to the Northern Star is bad.

Belsima:
Now they do visit the star, not the cross @woodratt

PR #5032, opened 2018-03-08
Adds Tajaran Akhani language.

Cerebulon:
Adds the second major language of the Tajaran, Akhani. Spoken by the other major playable ethnic group.
(Details developed with, and approved by Serithi.)

Also fixes plural form of Tajaran (From Tajara to Tajaran) in a dozen or more places.


mistyLuminescence:
Is Akhani given to Tajaran automatically, or do they select it using one of their language slots?

elgeonmb:
Good PR

Cerebulon:
Misty, it's available to all Tajaran but not selected as default.

Atermonera:
The color is this, by the by

elgeonmb:
I think some of these grammar fixes are changing the adjective form rather than the singular? I'm not actually sure what the adjective form is supposed to be, though.

Cerebulon:
Pretty sure it's also Tajaran. From the wiki:
One Tajaran, Two Tajarans/Tajaran, The Tajaran People.

elgeonmb:
👌

Leshana:
You renamed the species!? Like not just text, but the actual key name used as keys in lists and everything.

Cerebulon:
Yes because there's a shit ton of code which directly references that species ID and displays it. Which is janky as fuck but overhauling the entire thing to separate ID and Name could potentially break a ton of shit and this... Shouldn't. It doesn't break whitelists I'm sure of that much and save files/configs which saved this should be the only thing which are effected. Which is fixed by clicking Tajaran again and saving it and that should be entirely it.

PR #5036, opened 2018-03-08
Adds emergency bolt overrides to airlock code

Nerezza:
Airlocks now have an emergency_override var that gives the option to manually alter the bolt state when the airlock's unpowered by cranking the override with a wrench for 8 seconds (for the default wrench).

By default only external airlocks have this enabled so that people don't have to break a window to bypass inaccessible external airlocks anymore. However, map makers can set the variable to 1 on other airlocks to provide the same function (good for Virology airlocks and probably some Science airlocks).

All airlocks that start with emergency_override set to 1 will have their description altered in ~~New()~~ initialize() to indicate the presence of an emergency override. If the panel is open, detailed examine will inform the user that a wrench can crank the emergency override.

This does not add a way to add/remove an emergency override to/from doors in-round aside from var edits, to prevent people from adding a crank to the vault and cranking it open.

Fully tested. The examines work, the cranking works if area power is disabled and if door power is disabled, and editing a normal airlock in the map editor to have an emergency override 100% works as intended.

### DIFF
--- a/code/game/machinery/doors/airlock.dm
+++ b/code/game/machinery/doors/airlock.dm
@@ -11,6 +11,7 @@
 	var/main_power_lost_until = 0	 	//World time when main power is restored.
 	var/backup_power_lost_until = -1	//World time when backup power is restored.
 	var/has_beeped = 0					//If 1, will not beep on failed closing attempt. Resets when door closes.
+	var/emergency_override = 0				//If 1, the airlock's bolts have an emergency override crank.
 	var/spawnPowerRestoreRunning = 0
 	var/welded = null
 	var/locked = 0
@@ -37,6 +38,11 @@
 	var/denied_sound = 'sound/machines/deniedbeep.ogg'
 	var/bolt_up_sound = 'sound/machines/boltsup.ogg'
 	var/bolt_down_sound = 'sound/machines/boltsdown.ogg'
+
+/obj/machinery/door/airlock/New()
+	.=..()
+	if(emergency_override)
+		desc += " This one is equipped with an emergency bolt override under its panel."
 
 /obj/machinery/door/airlock/attack_generic(var/mob/user, var/damage)
 	if(stat & (BROKEN|NOPOWER))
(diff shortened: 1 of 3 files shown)


code/game/machinery/doors/airlock.dm
```diff
@@ -38,6 +39,11 @@
 	var/bolt_up_sound = 'sound/machines/boltsup.ogg'
 	var/bolt_down_sound = 'sound/machines/boltsdown.ogg'
 
+/obj/machinery/door/airlock/New()
+	.=..()
```
Arokha: New() doesn't have a return value, may not want to confuse people by putting this here, or, better yet, put this in initialize() so it doesn't slow down map loads.

Nerezza: Fix't and re-tested

Nerezza:
A little extra for staff consideration:

External airlocks and a few other airlocks (such as Virology) are constantly bolted and controlled by their respective airlock controllers for safety reasons.

This becomes detrimental when the power goes out however, as currently there's no way to override bolts for unpowered doors without an emag. Take escape pod hatches for example, if those are unpowered then that would ICly be a VERY BAD thing.

Brainstorming over the problem made me realize airlock override cranks are a scifi trope for exterior airlocks, so I decided to stick together a minimalistic implementation that could be configured per-airlock in the map editor as needed.

There ARE external airlocks that probably shouldn't get this new mechanic of course, for example the ones in Drone Fabrication and the Bomb Testing Range ones since those lead into very sensitive areas, but that can be fixed simply by setting their emergency_override var to 0.

Belsima:
Can you choose this to be on/off when you actually make an airlock?

Anewbe:
This was discussed during the staff meeting, and it was decided that this doesn't fix the actual problem, being the doors don't open with power. Unless you're an engineer, this change won't benefit you if you join (on the shuttle) when the power's out. Additionally, based on how the arrivals shuttle moves, it'd also probably space you the moment you step out into the airlock proper. We're aware of the problem, but this isn't the way we'd like to solve it. Sorry for the delay on this response, I've been out of sorts.

PR #5037, opened 2018-03-09
Rewrite human/update_icons()

Arokha:
What if we had a language that supported polymorphism? That'd be cool right? So we didn't need to overload the function of "draw handcuffs on me" to also "drop my items" and maybe that helmet should draw it's own light, not the human, and maybe there should be some consistency to how we obtain the icons that appear on the mob?

Rewrites update_icons() on /mob/living/carbon/human to be stateful. No longer does it update overlays_standing and then redraw all the layers. If you want to update gloves, you just update gloves. If you put on a backpack, it doesn't also reapply your shoes for fun.

Items create their own worn icons, you can override things about how they're drawn on the items, things are as they should be. God's in his heaven and all is right with the world.

I also added this, just cus. Only animates slow like this when you're not falling. If you fell over due to slips or something, you just go thud. Resting does this. Also after slips, when you get up, it's animated. Basically getting up is always animated. Falling may or may not depending on whether it's on purpose or not.

### /obj/item
* make_worn_icon() - Returns an image that can be applied to the mob. You should pass it many things, such as the body type you want, a default icon to pick from, etc. Read the comments.
* get_worn_icon_file() - Returns the icon (file or memory object) that should be used when drawing this item onto a mob.
* get_worn_icon_state() - Returns the icon_state that should be used when drawing this item onto the mob.
* get_worn_layer() - Returns the layer that this object should be rendered on (should be a FLOAT_LAYER layer, aka negative numbers. Right now it's -100 + define from update_icons())
* apply_blood() - Should the object be bloody in the sprite? Do it here. Has a generic implementation for several things like helmets, gloves, and boots (things that did it before)
* apply_custom() - Stub, can be used to do whatever to the icon before it's converted to an image.
* apply_accessories() - Draws all the accessories onto the object (all clothing appears to support this).
* apply_addblends() - Applys that 'addblends' thing that someone added. I dunno what it does, but this does it.

### /mob/living/carbon/human
* update_icons() - Now runtimes whenever called, stop calling it. If you want to update a slot, update a slot.
* update_icons_layers/huds - Same as above, don't call it. HUDs are also statefully drawn, updating each individual HUD as needed by itself.
* update_icons_body() - This one actually does stuff: important to call whenever the body has changed (lost/gained limbs, for example).
* regenerate_icons() - Calls every individual slot update icon proc, plus update_icons_body(). Should be used sparingly, like after changing a mob's entire species (or spawning a new one)

## Priorities
When designing your items and setting the variables, keep the following in mind, as it's now identical for every item (unless you override some of the above procs, but why would you go and do that?). These are checked, in this order, to see if they are set. If they do, they are used.
### Icon File
1. icon_override var
2. sprite_sheets[species_name]
3. item_icons[slot_name]
4. default_worn_icon var
5. default_icon var (passed into get_worn_icon_file and make_worn_icon)
6. Give up, return nothing.
### Icon State
1. item_state_slots[slot_name]
2. item_state var
3. icon_state var


Nerezza:
This is kind of a roundabout manner to fix bare feet not cleaning in showers, but I suppose it'll do o3od

I kid of course (but it really does fix that problem from what I'm reading). I was poking around the update_icons_layers code just a couple days ago going "Man, there HAS to be a better way to do all this." and here it is! I'll probably read through this entirely tomorrow to see what sort of new layer-torturing methods I can employ in the near future. *nod*

Arokha:
Pls dun torture layers, they have enough trouble as it is. And yes it does fix feet cleaning. There's a proc called update_bloodied() on humans that reboops their hand and foot blood if necessary.

code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -137,40 +103,23 @@ Please contact me on #coderbus IRC. ~Carn x
...
 
-	//5: Do any species specific layering updates, such as when hiding.
-	update_icon_special(ma_compiled, FALSE)
+	//Do any species specific layering updates, such as when hiding.
+	update_icon_special()
 
-	//6: Set appearance once
-	appearance = ma_compiled
+/mob/living/carbon/human/update_icons_layers()
+	crash_with("CANARY: Old human update_icons was called.")
```
Nerezza: Looks like you meant to put update_icons_layers here

Arokha: Can fix, but I'd rather wait for more review comments so I can do it in a big fix commit.

Nerezza: That's exactly the thing I expected you to say, dun worry ;>

Belsima:
How does the added animation handle fainting or collapsing?

Arokha:
It only does the .3s animation if you're resting or sleeping. Otherwise it's .1s which, in testing, is ... basically imperceptible to the human eye (or maybe byond just skips it due to tickrate, who knows).

For example, if you're on the arrivals shuttle and not buckled, you go THUD when it arrives, then animate standing back up.

code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -1,133 +1,99 @@
...
+#define HEAD_LAYER				23		//Head-slot item
+#define HANDCUFF_LAYER			24		//Handcuffs, if the human is handcuffed, in a secret inv slot
+#define LEGCUFF_LAYER			25		//Same as handcuffs, for legcuffs
+#define L_HAND_LAYER			26		//Left-hand item
+#define R_HAND_LAYER			27		//Right-hand item
+#define MODIFIER_EFFECTS_LAYER	28		//Effects drawn by modifiers
+#define FIRE_LAYER				29		//'Mob on fire' overlay layer
+#define WATER_LAYER				30		//'Mob submerged' overlay layer
+#define TARGETED_LAYER			31		//'Aimed at' overlay layer
+#define TOTAL_LAYERS			32//<---- KEEP THIS UPDATED, should always equal the highest number here, used to initialize a list.
```
Anewbe: Highest number here, or highest number+1?

Arokha: Should be highest number. Won't hurt anything for being LONGER, only SHORTER though, so this wouldn't actually cause an issue, but I'll fix it now with the other comment above if you're ready to merge.

Arokha:
It's tested as well as I can test with a few other people. We walked around, tested wearing clothes, suits, hats, shoes, etc. Tried spacesuits, tried blood washing, etc.

I'm sure someone will find an item that didn't comply with this sort of arrangement, it'd just need item_state set on it probably, or something similar. It's basically impossible to test holding/wearing every item in the game, but we tested a lot.

I'm sure soooooomeone will find an item that uses a weird sprite or something, but I fixed every issue we found in testing.

Anewbe:
Excellent. If you're looking to do those two tweaks, I'll hold off on the merge, but other than that I didn't find any glaring issues.

Arokha:
Dun did it

KasparoVy:
Fantastic refactor

Arokha:
Thonks. It only required unlimited suffering.

PR #5045, opened 2018-03-10
Another possible crash bug breakthrough.

Verkister:
So we had an incident that brought definite results pointing to a certain circumstance with an observer managing to crash everyone with an impressively high rate of success. This observer ghost was a teshari wearing a straightjacket. Some comparisons to other less definite cases were made, pointing towards characters with species specific tails arriving in gear that hides tails. Some code diving later I came to a conclusion that the code part responsible for generating cache for tail icons was locked behind the suit coverage check, suggesting that arriving onscreen for the first time while wearing one of those suits would prevent the tail part of the cache from generating at all.

Basically this thing calls the cache part before doing the suit coverage check so wearing a tail hiding suit should no longer block the code from ever reaching the tail cache generation.


Nerezza:
This code shouldn't be touched while https://github.com/PolarisSS13/Polaris/pull/5037 is open.

However, you might be onto something. I've noted that Teshari seem related to the crash bug, and it's possible code related to their icons will do _not_ account for the fact that their tails are _part of their bodies_ and not separate sprites.

I'd say there's a high possibility that clients are trying to get non-existent Teshari tail sprites.

KasparoVy:
That label sent my sides into orbit

Arokha:
To  be honest, I do think it's Teshari tails. Their one unique thing is that they have two images instead of one (because there's some 'tail skin' variable on species, the name escapes me), and they are the only species that has this, and obviously not every tail crashes everyone.

With update_icons rewrite, I notice teshari tails are now the wrong color sometimes, but, I also don't crash anymore in 512.1415. Played a full 5.5 hour round with no crashing and saw several teshari.

Verkister:
I still think this should get tested out anyway as I cannot see anything good resulting from worn gear completely blocking parts of initial icon cache generation.

Leshana:
As far as I read it, the thing this PR does is:  In the situation when someone with a species that has a tail is wearing tail-hiding clothing, it would call `get_tail_icon()` and throw away the result as opposed to before when it didn't bother calling it at all.   

The only side effect of get_tail_icon() is caching, but that cache is only accessed from within get_tail_icon(), so I doubt there is a desired outcome from this change at this point. Instead upgrade to 512.1417/9

Arokha:
512.1417 does fix the crash bug:

image

Let's not make further "fixes" to something that's not our problem, anyway.

Anewbe:
Yeah, gonna close this as per the last comment. If this is a Byond bug us flailing at it probably won't make things better.

PR #5050, opened 2018-03-11
Ports antibiotics overhaul

Anewbe:
Ports https://github.com/VOREStation/VOREStation/pull/3186

### DIFF
--- a/code/modules/organs/organ.dm
+++ b/code/modules/organs/organ.dm
@@ -148,25 +148,25 @@ var/list/organ_cache = list()
 		germ_level = 0
 		return 0
 
-	var/antibiotics = owner.reagents.get_reagent_amount("spaceacillin")
+	var/antibiotics = owner.chem_effects[CE_ANTIBIOTIC] || 0
 
 	var/infection_damage = 0
 
-	if((status & ORGAN_DEAD) && antibiotics < 30) //Sepsis from 'dead' organs
+	if((status & ORGAN_DEAD) && antibiotics < ANTIBIO_OD) //Sepsis from 'dead' organs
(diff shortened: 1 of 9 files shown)


code/modules/organs/organ.dm
```diff
@@ -148,25 +148,25 @@ var/list/organ_cache = list()
 		germ_level = 0
 		return 0
 
-	var/antibiotics = owner.reagents.get_reagent_amount("spaceacillin")
+	var/antibiotics = owner.chem_effects[CE_ANTIBIOTIC] || 0
```
Atermonera: `X||0 == X`

code/modules/organs/organ.dm
```diff
@@ -234,19 +234,19 @@ var/list/organ_cache = list()
 
 //Germs
 /obj/item/organ/proc/handle_antibiotics()
-	var/antibiotics = owner.reagents.get_reagent_amount("spaceacillin")
+	var/antibiotics = owner.chem_effects[CE_ANTIBIOTIC] || 0
```
Atermonera: See above

code/modules/organs/organ.dm
Anewbe: I'm pretty sure this is a form of null check?

Leshana: Yes, in fact, strictly speaking  X||0 does not equal X if X was null

Arokha: No, X||0 when X is null will be 0 because false, rather than null, which is important in some cases like using the result in a string and you don't want it to be null and leave an empty string, but want the zero to appear instead.

Leshana:
Horray! It passed compile and unit tests.   Aronai says the change looks reasonable.

PR #5052, opened 2018-03-11
Ports ghost pAI alert

Anewbe:
https://github.com/VOREStation/VOREStation/pull/2089


PrismaticGynoid:
Must we buff Trine?
(just kidding this is good feature)

PR #5053, opened 2018-03-12
Should speed up ZAS

Anewbe:
- I'm not sure what sort of horrible unforeseen side effects this may have on lag, or other things, but it _seems_ to help ZAS pick up the pace?


Leshana:
Its probably a good idea to have #defines for all subsystem priorities in subsystems.dm at some point, like there is for init order.

PR #5056, opened 2018-03-12
I h*ckin' did an ambient music.

Serithi:
As far as i can tell i implemented it properly, testing it out has been finicky as all hell but it played in the mines like i expected. I figure if i did a goof, the Reddit method of giving the wrong answer first so someone replies with the correct one will work.


Arokha:
I like this much more than the eventually-very-grating DF song :v

PR #5059, opened 2018-03-13
Fixes and tweaks for slime behavior.

Heroman3003:
- Fixed dark purple slime being exactly same color as purple.

- Fixed rainbow slimes born from mutations not having unity applied properly.

- Fixed light pink slimes born from mutations not having maximum discipline and obedience.

___

- Changes to loyal slime behavior:

- Now will properly attack and eat monkeys even if faction is same.

- Will not lose their loyal status over generations anymore.

- Will now still attack other-colored non-unified slimes when loyal and not unified itself rather than acting as if unified.

___

- Changed the attack command to act more logically when order is given to "slimes" in general.


Neerti:
Did this receive input @Anewbe ?

Anewbe:
`Will not lose their loyal status over generations anymore.`
That was intentional, from what conversations I've had. Other than that the concept seems fine.

Sorry about the long delay.

Heroman3003:
Hm, okay, but in original version, the faction _is_ inherited. Why does that happen if loyalty isn't supposed to carry over? Especially considering that interactions with other colors of slime are basically all exceptions from normal faction behavior.

Neerti:
I might've misunderstood what was meant by 'loyalty' not being carried over.

For some reason my dumb brain thought it was ``obedience`` being lowered by one per split, which is intentional.
A slime being your best pal and splitting and no longer being your best pal is not intentional.

Heroman3003:
Yes, thats exactly what I meant. Obedience and discipline carry over same way they do now (except for lightpinks, that simply ignore inheiritance and have both at max as they're supposed)

PR #5060, opened 2018-03-13
AFK Kicker should affect ghosts and people in the lobby

Anewbe:


code/controllers/Processes/inactivity.dm
```diff
@@ -7,10 +7,10 @@
...
 			var/client/C = last_object
 			if(C.is_afk(config.kick_inactive MINUTES))
-				if(!istype(C.mob, /mob/observer/dead) && !istype(C.mob, /mob/new_player))
-					to_chat(C,"<span class='warning'>You have been inactive for more than [config.kick_inactive] minute\s and have been disconnected.</span>")
-					var/information
+				to_chat(C,"<span class='warning'>You have been inactive for more than [config.kick_inactive] minute\s and have been disconnected.</span>")
+				var/information
 
+				if(C.mob)
 					if(ishuman(C.mob))
```
Arokha: don't need `if(C.mob)` given that it's fine to pass null to ishuman, and it will just be false

PR #5065, opened 2018-03-13
[PORT] Revamp logging for consistency

Leshana:
Ports VOREStation/VOREStation#3204  and the subsequent PRs that enhanced it:
VOREStation/VOREStation#3211 - Puts topic logging into debug log to get it out of the way.
VOREStation/VOREStation#3219 - Makes logging about engine stuff consistent too.
VOREStation/VOREStation#3223 - Fixes issue from final production testing.

### summary
Previously, logging would occur using shift time on everything, even runtimes. Why this was the case, I have no idea. I don't understand why it would do that. Now we log with ISO 8601 timestamps at the start of every line, and we log in a consistent manner easily parsable by 3rd party log inspection tools.

When you want to log attack stuff, use `add_attack_logs(attacker,defender,"What happened",admin_notify = TRUE)`

Everything else is things like `log_say(message,speaker)`

### Developer Notes
With this change, logging out put is improved in a few ways
- Each round now has a separate new logfile instead of appending to previous round on the same day!
- Some logging is logged into separate files, like runtimes and init and qdel stuff.
- Logging format is far more standardized.  This allows you to use tools like splunk to parse logs and actually read them.
- Using attack logs is far cleaner.  No longer must you use 5 lines of code every time you want to do attack logging.

### DIFF
--- a/code/_helpers/logging.dm
+++ b/code/_helpers/logging.dm
@@ -140,7 +140,7 @@
 
 //more or less a logging utility
 //Always return "Something/(Something)", even if it's an error message.
-/proc/key_name(var/whom, var/include_link = null, var/highlight_special_characters = 1)
+/proc/key_name(var/whom, var/include_link = FALSE, var/include_name = TRUE, var/highlight_special_characters = TRUE)
 	var/mob/M
 	var/client/C
 	var/key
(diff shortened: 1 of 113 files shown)


code/_helpers/logging.dm
```diff
@@ -27,7 +27,7 @@
 
 /proc/log_debug(text)
 	if (config.log_debug)
-		diary << "\[[time_stamp()]]DEBUG: [text][log_end]"
+		debug_log << "\[[time_stamp()]]DEBUG: [text][log_end]"
```
Atermonera: Should we be switching other debug-logging things to use this as well, if/when this is merged?

Leshana: Its a nice idea, but this PR will work fine even before any of that is done.

PR #5067, opened 2018-03-14
[PORT] Cleans up Humanoid Hiding

KasparoVy:
Ports the humanoid hiding cleanup from https://github.com/VOREStation/VOREStation/pull/3234
- Refactors humanoid hiding to use `status_flags` and a define rather than a snowflake variable.
- Swaps a magic number for a define.
- Amends the humanoid hiding eligibility check.


KasparoVy:
Sorry, pending an edit to match the logic of hiding in `living_powers.dm` with `human_powers.dm`.

PR #5069, opened 2018-03-15
Ported Grabber and Thrower integrated circuits

Heroman3003:
Grabber: can pick up and store up to ten items.
Thrower: can throw items in a direction relative to the location of machine. Can use items from grabber's storage or items lying near machine.


Heroman3003:
closed for fork cleanup

PR #5070, opened 2018-03-16
Updates changelog

Atermonera:
Big update because this past week has been very busy for me (Deadweek, all dead, no hope).

@Cerebulon please rename the changelog file from `Example - Copy.yml` so I don't accidentally delete it preemptively in the future b/c most of the time when I see that file it's the default with all the comment pruned out, which generates an unnecessary changelog. (It also doesn't tell me who the author is or what general category the work within corresponds to).
@MisterLayne it's 
> changes:
>   - rscadd: [...]

The colon after `rscadd` is important, the script findings errors is a nuisance to deal with, because it usually requires reverting all files that the script had already read and then redoing any proofreading edits I made.


Cerebulon:
Whoops, did that in a hurry.

PR #5071, opened 2018-03-16
Makes coffee milkshakes STOP KILLING PEOPLE 

Cameron653:
You see, coffeemilkshake had an overdose proc of 45.

However, it wasn't a child of coffee. This is due to milkshakes having special effects on Unathi and stuff like that.

This meant that having > 45u of coffeemilkshake would OD you. Sure! It'll just make you jittery then, right?

Nope. It _**FUCKING KILLS YOU**_ because it defaults to the /reagent OD since /drink doesn't have a OD set.

- This PR makes it future toxic drinks can have an OD and call back to the reagent OD instead via having the overdose proc for /drinks calls ..(). This can also be edited in the future.

- Gave coffeemilkshake a specific OD proc.

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Food-Drinks.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Food-Drinks.dm
@@ -450,8 +450,8 @@
 	if(adj_temp < 0 && M.bodytemperature > 310)
 		M.bodytemperature = min(310, M.bodytemperature - (adj_temp * TEMPERATURE_DAMAGE_COEFFICIENT))
 
-/datum/reagent/drink/overdose(var/mob/living/carbon/M, var/alien) //Added this is a failsafe, as coffee milkshakes were FUCKING KILLING PEOPLE
-	M.make_jittery(5)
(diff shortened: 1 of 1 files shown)


code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Food-Drinks.dm
```diff
@@ -450,6 +450,9 @@
 	if(adj_temp < 0 && M.bodytemperature > 310)
 		M.bodytemperature = min(310, M.bodytemperature - (adj_temp * TEMPERATURE_DAMAGE_COEFFICIENT))
 
+/datum/reagent/drink/overdose(var/mob/living/carbon/M, var/alien) //Added this is a failsafe, as coffee milkshakes were FUCKING KILLING PEOPLE
+	M.make_jittery(5)
```
Arokha: I mean, doesn't this mean that drinking more than 30u of milk make you jittery?

Cameron653:
Changed to make it so future toxic drinks can have an OD and call back to the reagent OD instead.

Gave coffeemilkshake a specific OD.

PR #5072, opened 2018-03-17
Arranges lots of things into planes

Arokha:
How do you feel about that? Isn't it planetastic?

This is one of those PRs with a billion single-line changes.

Also made the cryopod show the color of the drugs in it, for fun, while I was in there.


Arokha:
Oh. Uh. Your Travis isn't new enough to have vis_contents. Hooold on.

Arokha:
Ok your SERVER is still on 511 so this won't work. Um. Lemme know when you update server I guess.

Neerti:
Server is 512 now.

Arokha:
Howso? I tested it for a while and didn't notice any issues, but it might be that I'm using a different version of byond than you. What is broken?

MarinaGryphon:
Visible_message now always uses the blind message. Toggle_darkness doesn't work. Chem machines are on the wrong layer. There are lots of very strange lighting artifacts, especially when travelling in disposals, etc. I'm not sure how many are related to this, but I believe visible_message may be, since it uses `MOB_CAN_SEE_PLANE(M, plane)`- though I'm not sure how recent this is.

Arokha:
MOB_CAN_SEE_PLANE is older than this. Do you mean toggle_darkness as a ghost? That's super weird, since that plane wasn't even changed. 🤔 

Chem machines are on the wrong layer how? I'll take a gander. But what do you mean?

MarinaGryphon:
Apparently it's only in the chem lab- it might just be acting up because we just updated. If it happens next round, I'll look deeper into it. Specifically, machines in the chem lab such as chem dispensers and grinders are under the table. Soft drink dispensers, booze dispensers, etc. are on top of their tables as normal. (This could be an issue with the chem lab, chemistry-only machines, the table, etc- however, since the grinder is on top of the table in the kitchen and under it in the chem lab, it's either a mapping issue, post-update server weirdness, or the table.)
The issues with darkness, though, seem to be from this-  I believe this is the first update since it was pushed, and these issues just started happening.

PR #5074, opened 2018-03-18
Makes fingerprint logs flow smoother

Cameron653:
Makes the admin logs give the important info (name/ckey) before giving the less important info (wearing gloves/has no fingerprints).


Cameron653:
Someone touched the file in a more efficient way & did the same thing I did here, so just going to close this.

PR #5075, opened 2018-03-19
Adds snow footsteps

BillyBangles:
Walking on snow no longer sounds like [generic footstep noise]
I also slightly nerfed ice tiles while I was at it, since it was in the same dm file.


BillyBangles:
Planning on adding some grass sounds too, as soon as I can record some nice sounds. (the grass in front of my house just doesn't _sound_ right, it's all dead right now)

MisterLayne:
Would suggest renaming the PR title to fit more of what you actually did.

BillyBangles:
The ice nerf was literally a one digit change, didn't strike me as important enough. Much larger scale changes will happen later when I poke stun code.

PR #5078, opened 2018-03-19
Adds permanent markers

BillyBangles:
Adds markers in place of crayons. Don't worry, you can still order a box of crayons from cargo in case you get hungry. This is recommended, as trying to eat a marker will OD you in one bite.
This also changes the chemistry recipe for the various colors of paint to use marker ink instead of crayon dust.

Marker sprites and marker box sprite


BillyBangles:
Okay I straight up downloaded the current version of southern_cross-1 from here and didn't touch anything except for that crayon box. As far as I can tell map merger worked how it's supposed to. What am I doing wrong here?

BillyBangles:
nevermind, fuck it, restarting

PR #5079, opened 2018-03-20
Adds permanent markers

BillyBangles:
Take two. I'm good at git I swear.
Adds markers in place of crayons. Don't worry, you can still order a box of crayons from cargo in case you get hungry. This is recommended, as trying to eat a marker will OD you in one bite.
This also changes the chemistry recipe for the various colors of paint to use marker ink instead of crayon dust.

Marker sprites and marker box sprite


Anewbe:
```
code/game/machinery/vending.dm:839:error: /obj/item/weapon/storage/rollingpapers: undefined type path
code/game/machinery/vending.dm:852:error: /obj/item/weapon/storage/rollingpapers: undefined type path
```

code/game/objects/items/crayons.dm
```diff
@@ -99,7 +99,93 @@
...
+	colourName = "rainbow"
+	uses = 0
+
+/obj/item/weapon/pen/crayon/marker/rainbow/attack_self(mob/living/user as mob)
+	colour = input(user, "Please select the main colour.", "Marker colour") as color
+	shadeColour = input(user, "Please select the shade colour.", "Marker colour") as color
+	return
+
+/obj/item/weapon/pen/crayon/marker/attack(mob/M as mob, mob/user as mob)
+	if(M == user)
```
MisterLayne: Maybe we could have an intent for this, possibly? I really don't want to end up accidentally clicking on myself and eating a marker.

BillyBangles: There are very many items you can harm yourself with on accident regardless of intent. Just be careful. As a pro-tip, you don't have to click the tile you're on to draw on it, you can draw on adjacent tiles.

Atermonera: Add drawing faces on people in sharpie when?

BillyBangles: Now that I can get behind.

Atermonera: img

code/game/objects/items/crayons.dm
```diff
@@ -46,11 +46,11 @@
 	if(colour != "#FFFFFF" && shadeColour != "#000000")
 		colour = "#FFFFFF"
 		shadeColour = "#000000"
-		user << "You will now draw in white and black with this crayon."
+		to_chat(usr,"You will now draw in white and black with this crayon.")
```
Leshana: I believe that the previous code's choice in using `user` was the correct one.  I think changing it to `usr`, although unlikely to cause any obvious problems right away, isn't good.  We should continue to use `user` so that this proc will work correctly when invoked on behalf of someone.  This ought to be fixed when convenient.

PR #5080, opened 2018-03-20
Fixes the drone borg ID thing AGAIN.

Verkister:
Ya didn't even check the validity of this comment ya put back in too, did ya? "//Drone Permit moved to equip_survival_gear()"


Arokha:
This will prevent anyone who spawns w/o a backpack (which is just a choice in character setup) from getting the permit entirely. You may want to use the backbag checks that other spawning stuff uses to at least attempt to put it in a hand or just fix the reason for this spawning altogether.

Verkister:
figured out proper fix

Arokha:
Nice

PR #5085, opened 2018-03-21
Port Shuttle Pilot Helmet Animation

Leshana:
Bonk helmet on shuttle computer to link. No, it doesn't work if you're off the shuttle, no remote info transfer.

Animation of it doing the thing (16mb)
https://cdn.discordapp.com/attachments/173831852602294272/408898940252192769/pilot_helmet.gif

Rightclick on the helmet to change the color of the animation.

Port of VOREStation/VOREStation#2940
Also ports prior bugfixes of VOREStation/VOREStation#2921 and VOREStation/VOREStation#3037


Leshana:
Stand by.  This works with southern cross compiled, but not without it. Testing didn't pick that up of course, but Travis did.  Working on best solution.

Atermonera:
Pilot helmets are defined for us in
https://github.com/PolarisSS13/Polaris/blob/master/maps/southern_cross/items/clothing/sc_head.dm#L3

Which, as you may guess, is only included when the SC is ticked. You'll probably need to move their definition, etc., to not be in a map file.

PR #5087, opened 2018-03-21
Cache dips new mannequins for fresh observers

Verkister:
Found another slightly different crash bugson by simply observing without having done any icon generation relevant things prior. (such as opening char setup for refreshed previews etc.) Immediately crashed myself first, and then literally everyone else I passed by.
Found out recently updated observer join copies the ghost appearance directly from a new mannequin deleted right afterwards, and there was no room for body icon caching inbetween or something that way. The mannequin mob preparations only had the icon updates for worn gear before deletion.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_species.dm
+++ b/code/modules/mob/living/carbon/human/human_species.dm
@@ -9,6 +9,7 @@
 	living_mob_list -= src
 	dead_mob_list -= src
 	delete_inventory()
+	update_icons_body()
 
 /mob/living/carbon/human/skrell/New(var/new_loc)
 	h_style = "Skrell Short Tentacles"


code/modules/mob/living/carbon/human/human_species.dm
```diff
@@ -9,6 +9,7 @@
 	living_mob_list -= src
 	dead_mob_list -= src
 	delete_inventory()
+	update_icons_body()
```
Arokha: This is already done, though, in human/New().

human/New() -> set_species() -> regenerate_icons() -> update_icons_body()

Arokha:
I'll mention I haven't had a single crash on 512.1415, whereas I'd crash 4-5 times every hour previously.

code/modules/mob/living/carbon/human/human_species.dm
Verkister: Is it guaranteed to spawn without species set at all as the human/New() seems to only set species if there is none.

Verkister:
I'm also on 1415 currently an still managed to crash literally everyone who was around that round today, including myself as the first one.

code/modules/mob/living/carbon/human/human_species.dm
Arokha: Yes, it is. Species is null by default on humans. You would have to have a subtype of /mob/living/carbon/human with a New() that digs into all_species, sets species to a reference of the datum, etc.

No such thing exists, and mannequins definitely have set_species() called on them. How else would they appear as your species?

Verkister: Fair enough. Gonna have to dig into the stuff a bit deeper.

Arokha: I mean, it's a byond bug. No matter what you do, you may not be able to fix this. Assuming you can is folly.

You should make sure you're using the latest beta, and collect crash dumps to give to Lummox for troubleshooting. And make test cases for him if you can.

PR #5092, opened 2018-03-22
Removes some problematic HIDEJUMPSUIT flags

Anewbe:


code/modules/clothing/suits/jobs.dm
```diff
@@ -45,7 +45,7 @@
 	desc = "A long, flowing white robe. It looks comfortable, but not very warm."
 	icon_state = "whiteout_robe"
 	item_state_slots = list(slot_r_hand_str = "suit_white", slot_l_hand_str = "suit_white")
-	flags_inv = HIDEJUMPSUIT|HIDETIE|HIDEHOLSTER
```
mistyLuminescence: Not a big deal, but doesn't this _actually_ hide your jumpsuit as far as sprites go? Last I checked it did, anyway.

PR #5093, opened 2018-03-22
Cherrypicks missing typing indicator commit from Virgo

Neerti:
Should fix any lingering typing bugs now that we have the complete implementation.


Leshana:
It appears this is a port of VOREStation/VOREStation#3266

PR #5098, opened 2018-03-24
Changes to objects random

Woodratt:
- Fixed the icon_state for 'structure/plushie/carp' and the random first aid kit spawner
- Added a random chance tool spawn for power tools (most of the time it should still just be regular tools)
- Adjusted the spawn rate of medkits, combat medkits should be more rare
- Cash split into its own loot item
- Plushies split into large and small plushies
- All the extra plushie spawns added to the random plushie spawn
- Eightball and conch shell added to toy spawns
- Added spawn points for the large plushies, cash, and the power tools


code/game/objects/random/misc.dm
```diff
@@ -236,6 +248,20 @@
 				prob(2);/obj/item/weapon/reagent_containers/syringe/drugs,
 				prob(1);/obj/item/weapon/reagent_containers/syringe/steroid)
 
+/obj/random/cash
+	name = "random currency"
+	desc = "LOADSAMONEY!"
```
Belsima: its spelled LODSAMONE sorry

PR #5099, opened 2018-03-24
Removes some unnecessary hoops from head icon updates.

Verkister:
-Now head organ's get_icon() only applies hair/beard overlays on dismembered heads. Intact heads get to use the proper mob layer system as they should without the heft of needless duping underneath.
-Also apparently happens to fix markings covering up eyes.

### DIFF
--- a/code/modules/organs/organ_icon.dm
+++ b/code/modules/organs/organ_icon.dm
@@ -61,7 +61,8 @@ var/global/list/limb_icon_cache = list()
 		var/datum/sprite_accessory/marking/mark_style = markings[M]["datum"]
 		var/icon/mark_s = new/icon("icon" = mark_style.icon, "icon_state" = "[mark_style.icon_state]-[organ_tag]")
 		mark_s.Blend(markings[M]["color"], ICON_ADD)
-		overlays |= mark_s //So when it's not on your body, it has icons
+		if(severed)
+			overlays |= mark_s //So when it's not on your body, it has icons
 		mob_icon.Blend(mark_s, ICON_OVERLAY) //So when it's on your body, it has icons
 		icon_cache_key += "[M][markings[M]["color"]]"
 
@@ -75,11 +76,13 @@ var/global/list/limb_icon_cache = list()
 			else
 				eyes_icon.Blend(rgb(128,0,0), ICON_ADD)
 			mob_icon.Blend(eyes_icon, ICON_OVERLAY)
-			overlays |= eyes_icon
+			if(severed)
+				overlays |= eyes_icon
 
 	if(owner.lip_style && (species && (species.appearance_flags & HAS_LIPS)))
 		var/icon/lip_icon = new/icon('icons/mob/human_face.dmi', "lips_[owner.lip_style]_s")
-		overlays |= lip_icon
+		if(severed)
+			overlays |= lip_icon
 		mob_icon.Blend(lip_icon, ICON_OVERLAY)
 
 	if(severed) //Properly applied on intact mobs at update_icons_body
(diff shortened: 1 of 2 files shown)


Arokha:
Was this tested? When severed, it won't have an owner anymore, so I'm not sure how it could get that info. Also nothing in removed() calls this proc so it would likely not update the icon anyway.

code/modules/organs/organ_icon.dm
```diff
@@ -80,23 +80,24 @@ var/global/list/limb_icon_cache = list()
 		mob_icon.Blend(mark_s, ICON_OVERLAY) //So when it's on your body, it has icons
```
Arokha: Not going to do the same up here?

Verkister: Markings don't have a separate mob layer afaik

Arokha: They don't, they're blended into mob_icon right there, but you can see on markings and lips and stuff, it's adding them to overlays on the head, which is for the same purpose (showing up on the severed head)

Verkister:
Worked like a charm in testing.

code/modules/organs/organ_icon.dm
Verkister: Isn't the mob_icon here just the icon for the limb though? Each called separately from here to update_icons_body() to be assembled into the full body. Like the icon caching builds the mob from pre-painted parts made by this thing. The eyes and lips are applied onto the full mob tho and could be given the same treatment

Arokha: mob_icon is what update_icons_body() applies onto the mob's sprite itself. Anything added to overlays here is totally invisible because this object is inside the body and thus not visible to anyone until it's dropped and becomes it's own atom. So markings, lips, eyes, for their overlays thing, can be deferred until it's dropped.

Same for every limb, really, if you wanted to make this super-complete and do it on all the limbs and update the larger /removed.

Verkister: yeah noticed now

code/modules/organs/organ_icon.dm
```diff
@@ -80,23 +65,40 @@ var/global/list/limb_icon_cache = list()
...
+					eyes_icon.Blend(rgb(eyes.eye_colour[1], eyes.eye_colour[2], eyes.eye_colour[3]), ICON_ADD)
+				else
+					eyes_icon.Blend(rgb(128,0,0), ICON_ADD)
+				mob_icon.Blend(eyes_icon, ICON_OVERLAY)
+				overlays |= eyes_icon
+
+		if(owner.lip_style && (species && (species.appearance_flags & HAS_LIPS)))
+			var/icon/lip_icon = new/icon('icons/mob/human_face.dmi', "lips_[owner.lip_style]_s")
+			overlays |= lip_icon
+			mob_icon.Blend(lip_icon, ICON_OVERLAY)
```
Arokha: Wait, this part is important though. Only the overlays part can be skipped. The mob_icon blends have to be done otherwise they won't have eyes on their body.

Verkister: Tested another sort of approach and founf out the markings are important too. Wouldn't show up in preview at all when skipped.

Arokha: Just do
```
if(severed)
    overlays |= thing
```
Is what I was intending to suggest. The mob_icon stuff needs to stay where it was.

Verkister:
Closing this for now until the codebases are up-to-date enough with each other for more reliable testing.
(setup preview either flawless or broken and I can't find the code difference causing it)

PR #5104, opened 2018-03-24
Revert "Fixes cameras making completely black photos"

Anewbe:
Not 100% certain this will fix the issue, but if it does, it's preferable to the alternative. Will test and correct ASAP.


Neerti:
Upon checking out remotely, this fixes the character preview issue, but the ghosts being naked issue appears to be a separate issue unfortunately.

PR #5105, opened 2018-03-24
Nanoui now supports use of multiple templates

Atermonera:
This shouldn't change any existing NanoUI, but in the future I'll likely be refactoring that stuff so everything uses the same template for the same function. Looking at you, PAI records templates that happen to be almost exactly what I'm looking for for records display.
The order in which templates are added matters: those added first will be displayed first. I didn't test changing a template that was initially put above a different one, but I suspect that it will not re-order them.
There isn't currently a way to remove templates.
You no longer have to use the key "main", any will do so long as you're consistent.

This was tested and developed using code I'm working on for communicators, that stuff isn't included here. The first status bar at the top is taken from one file, the rest of the main template is on another, which is exchanged when it becomes `test text` with a link home. The External Device menu will be utilized for cartridge functions, which I'm mostly doing this for.
It was also tested on PDAs, which retain their old function, to ensure it doesn't break anything.

`Add_template()` on its own doesn't do much, because it explicitly states that the ui has to be opened afterwards.
`Append_template()` will add the template or change an existing key to a new template, re-open the ui, and update it with the data. The third step is important, otherwise it won't finish updating until it auto-updates, if at all.

[Testing with multiple templates](https://puu.sh/zONWN/a88a240a02.mp4)
[Testing with a single template](https://puu.sh/zOP54/6c2bd9a2fc.mp4)

### DIFF
--- a/code/modules/nano/nanoui.dm
+++ b/code/modules/nano/nanoui.dm
@@ -519,3 +519,8 @@ nanoui is used to open and update nano browser uis
   */
 /datum/nanoui/proc/update(var/force_open = 0)
 	src_object.ui_interact(user, ui_key, src, force_open, master_ui, state)
+
+/datum/nanoui/proc/append_template(var/key, var/filename)
+	add_template(key, filename)
+	open()
+	update(1)
--- a/nano/js/nano_state.js
+++ b/nano/js/nano_state.js
@@ -7,8 +7,7 @@ function NanoStateClass() {
 		return;
 	}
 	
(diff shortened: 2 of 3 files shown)


Atermonera:
And then we inflate the diff because that space-indentation where everything else is tab-indented, even if it's just webcode, is bad.

Leshana:
Note: This PR broke all nanouis which use the nano map.  I'll work on fixing it when time is available.

PR #5112, opened 2018-03-26
prettified lorecodex, added a month's worth of news

elgeonmb:
lore articles are now listed reverse-chronologically. "--" no longer used in article titles. Adds literally a month of news why the hell do I never remember to push it


elgeonmb:
okay looks like I'm activating my development environment after all

PR #5125, opened 2018-03-29
Normal brute kits disinfect wounds

Anewbe:
Could've sworn this was already a thing, I blame ridiculous non-standardized medical stack code.


Magneld:
Conflicted. I can accept the brute packs coming with some sort of disinfectant worked into the bandages but seems a bit redundant given ointments already do the same job and are part of the white kit already. Though given that MDs will soon be joining with white kits as default I can still see the need. I'll go with whatever is decided but I'm kinda squinty about it.

Issue #2051, opened 2016-07-06
Wallets Block HUD Interaction


Nalarac:
#### Brief description of the issue

Wearing a wallet over the ID slot will prevent interacting with the individual's records through the HUD.
#### What you expected to happen

Being able to interact with the records.
#### What actually happened

Shows up as "none"
Medical HUD with ID
http://prntscr.com/bppxvo
Medical HUD with wallet (ID inside)
http://prntscr.com/bppxxn
#### Steps to reproduce

1.) Wear wallet (with ID inside)
#### Additional info:
- **Server Revision**: master - 2016-07-06

Nalarac:
Still an issue on: master - 2017-12-08 11757132352255fa88f0d06597edcb4efa816fa7

Issue #4808, opened 2018-02-12
Newscaster does not recognize Wallets


MerekBlack:
#### Brief description of the issue

Newscaster doesn't recognize you if ID in wallet.

#### What you expected to happen

Recognition.

#### What actually happened

Not that.

#### Additional info:
 - **Server Revision**: Found using the "Show Server Revision" verb under the OOC tab.
 - **Anything else you may wish to add** (Location if it's a mapping issue, etc)

Atermonera:
Related to #2051 , it's the same issue in different spots. So if someone fixes one, please fix the other too.

PR #5130, opened 2018-03-30
Fixes wallet-HUD interaction issues.

MarinaGryphon:
Newscasters and HUDs should now properly identify people who have wallets.

### DIFF
--- a/code/game/machinery/newscaster.dm
+++ b/code/game/machinery/newscaster.dm
@@ -936,17 +936,11 @@ obj/item/weapon/newspaper/attackby(obj/item/weapon/W as obj, mob/user as mob)
 	if(istype(user,/mob/living/carbon/human))                       //User is a human
 		var/mob/living/carbon/human/human_user = user
 		if(human_user.wear_id)                                      //Newscaster scans you
-			if(istype(human_user.wear_id, /obj/item/device/pda))	//autorecognition, woo!
-				var/obj/item/device/pda/P = human_user.wear_id
-				if(P.id)
-					scanned_user = GetNameAndAssignmentFromId(P.id)
-				else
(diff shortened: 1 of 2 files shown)


Anewbe:
Tested and working? Does this work for PDAs and Communicators as well?

MarinaGryphon:
If Get_ID works, it should. I'll test it in a moment.

code/game/machinery/newscaster.dm
```diff
@@ -936,17 +936,11 @@ obj/item/weapon/newspaper/attackby(obj/item/weapon/W as obj, mob/user as mob)
...
-			if(istype(human_user.wear_id, /obj/item/device/pda))	//autorecognition, woo!
-				var/obj/item/device/pda/P = human_user.wear_id
-				if(P.id)
-					scanned_user = GetNameAndAssignmentFromId(P.id)
-				else
-					scanned_user = "Unknown"
-			else if(istype(human_user.wear_id, /obj/item/weapon/card/id))
-				var/obj/item/weapon/card/id/ID = human_user.wear_id
-				scanned_user = GetNameAndAssignmentFromId(ID)
+			var/obj/item/weapon/card/id/I = human_user.wear_id.GetID()
```
Leshana: I recommend changing this to simply `human_user.GetID()` so it will also support IDs held in active hand.

MarinaGryphon:
Fixes #2051 and #4808 (wasn't sure which was still open, so I put both).

MarinaGryphon:
Tested, works properly with PDAs, wallets, communicators, etc.

TheFurryFeline:
Looks like the involved Issues are still open, the numbers being what Marina stated above.

PR #5136, opened 2018-04-01
Amend example MOTD to reference Polaris code

Lorwp:
Title. This server threatening issue has finally been resolved.


Woodratt:
Technically that's in the example stuff, but ... eh!

Lorwp:
On the actual MOTD server-side it's still referring to the Northern star, but that's for someone with box access to fix.

PR #5138, opened 2018-04-01
Small fixes and ports discovered during merge.

Leshana:
A series of small couple-line fixes from vorestation or discovered during syncing polaris to vorestation.
- Port of VOREStation/VOREStation#2905: Replace messy string compare with nice type compare for assistant job.
- Port of VOREStation/VOREStation#2957: Move admin planes to more useful places for better shenanigans.
- Port of VOREStation/VOREStation#3120: Only spam ghosts with chat messages from things with real clients.   That way ghosts don't get spammed by simple animals.
- Fix small forgotten lines of text accidentally left out of VS syncs.
- Fix typo'd plane in lighting.dm and remove redundant definitions.

### DIFF
--- a/code/modules/mob/living/simple_animal/animals/cat.dm
+++ b/code/modules/mob/living/simple_animal/animals/cat.dm
@@ -156,6 +156,7 @@
 /mob/living/simple_animal/cat/fluff/Runtime
 	name = "Runtime"
 	desc = "Her fur has the look and feel of velvet, and her tail quivers occasionally."
+	tt_desc = "Felis medicalis"
 	gender = FEMALE
 	icon_state = "cat"
 	item_state = "cat"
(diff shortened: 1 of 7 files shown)


code/modules/mob/living/simple_animal/animals/cat.dm
```diff
@@ -156,6 +156,7 @@
 /mob/living/simple_animal/cat/fluff/Runtime
 	name = "Runtime"
 	desc = "Her fur has the look and feel of velvet, and her tail quivers occasionally."
+	tt_desc = "Felis medicalis"
```
Belsima: every single one of these names bothers me personally and i still think anything you can directly see outside of examines should be only in-character

Issue #5140, opened 2018-04-02
Graphical glitches in vents


Belsima:
#### Brief description of the issue

When moving through vents, especially quickly, parts of the screen will go black and the entirety of the pipeline will appear to be some gray thing I forgot the name of. During these occurrences, severe lagspikes also strike. It makes navigating vents nearly impossible.

![image](https://i.imgur.com/FvVHf37.png)

#### What you expected to happen

A seamless, smooth experience,

#### What actually happened

I can't see and I jump 6 tiles at a time due to those lagspikes.

#### Steps to reproduce

1. Spawn as mouse or other animal with ventcrawling verb.
2. Alt-click on vent connected to pipeline.
3. Scurry down the pipe.
4. Observe.

#### Additional info:
 - **Server Revision**: Found using the "Show Server Revision" verb under the OOC tab.
 - **Anything else you may wish to add**: This also sometimes occurs without the grey objects.

Leshana:
CNR.  Are you using BYOND version 512.1418?  Because that version broke sprites.   Works fine in 512.1419

PR #5147, opened 2018-04-03
Switch cables and pipes to use define constants instead of literals for layers.

Leshana:
Also fix a few things to be on the correct plane.  Basically the under-floor stuff is supposed to be on PLATING_PLANE.
Fixed wires to be on top of pipes like they used to be, otherwise they'd be obscured too often.

image

Portable atmospherics go on object layer so they are above stationary machinery, given that they move around. This also fixes reported issue with canisters going under meters.
image


MisterLayne:
The engineering gods are praising you for your gift to them.

elgeonmb:
will this by any chance fix https://github.com/PolarisSS13/Polaris/issues/5140 ?

Atermonera:
That looks more like an icon-fetching issue than a layering issue, so I'd be surprised if it did.

PR #5153, opened 2018-04-05
Adds Tile-based Attacks

Neerti:
Adds ability to attack mobs by clicking their tile.
When someone clicks a tile with a weapon while off help intent, and if a mob is occupying that tile, the mob will be attacked as if they were clicked directly.
If more than one mob is on a tile, one is chosen randomly.
You cannot hit yourself by clicking your own tile.
Weapons with cleaving abilities will attempt a cleave on the tile clicked on, making it very easy to hit (simple) mobs with those weapons if near you.

Other changes.
Cleave proc can accept any atom now and not just a mob.
Also cleans up weapons deciding how they can cleave somewhat.
Adds argument for attack animations to exclude the weapon icon regardless of player prefs.

### DIFF
--- a/code/game/objects/items/weapons/material/knives.dm
+++ b/code/game/objects/items/weapons/material/knives.dm
@@ -99,14 +99,8 @@
 	icon_state = "machete"
 	force_divisor = 0.3 // 18 when hardness 60 (steel)
 	attack_verb = list("slashed", "chopped", "gouged", "ripped", "cut")
-	var/should_cleave = TRUE //Now hatchets inherit from the machete, and thus knives. Tables turned.
+	can_cleave = TRUE //Now hatchets inherit from the machete, and thus knives. Tables turned.
 	slot_flags = SLOT_BELT
-
-// This cannot go into afterattack since some mobs delete themselves upon dying.
-/obj/item/weapon/material/knife/machete/pre_attack(var/mob/living/target, var/mob/living/user)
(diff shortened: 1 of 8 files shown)


Belsima:
Unless I misread, isn't this a big Teshari nerf?

Atermonera:
And any other mobs with small sprites, but I would consider that an improvement to balance, because trying to not be hit by virtue of having just a small sprite feels very cheap to me.

Neerti:
I don't recall 'having a small sprite' being a factor in the Tesh design, compared to things like 'is very fast' or 'gibs if breathed on'. The current Tesh lore maintainer also tells me they don't recall that as well.

IMO if we had this earlier we could've saved a lot of headaches involving salty security and a certain custom borg sprite, and possibly other things.

Belsima:
'twas far back, in the original Zuhayr thread, IIRC.

Atermonera:
Also, does this only affect melee attacks, or does it also affect ranged? Because I know guns do some special pathfinding and still have a good chance to hit you if you stand between them and where the dude is actually shooting

Neerti:
This only affects melee weapons.

Currently if you fire a gun at a tile it'll hit the first thing it bumps into, which would be a mob on the tile you clicked if the projectile manages to make it that far and the projectile isn't told to keep going (e.g. mob is lying down and isn't specifically the target).

PR #5158, opened 2018-04-09
Polar-izes shipping locations

elgeonmb:
may this be the last fucking time I have to type out all those systems I swear to god


Belsima:
im very happy we removed the lore-violating tau ceti, but even happier we added tau ceti

elgeonmb:
Oh for the sake of transparency: also removes the Almach Rim from the list of valid destinations pending travel there not being illegal as balls. May or may not add the location back to specific organization's lists.

PR #5161, opened 2018-04-10
Refactors human/examine.dm

Atermonera:
_Obstinate grumbling_
[tested](https://hastebin.com/tureqivedo.vbs) but based on the medhud refactor there may be some small issues lurking around. The changelog is present to encourage people to actually report when they think something is broken because with the medhuds they just accepted the broken behaviour was intended when I really did a `<brr>` instead of a `<br>`.
There's a few changes to the logic that I feel make more sense as well, such as actually hiding concealed holsters, and hiding all holsters if they're supposed to be hidden.
Switched it to using bitflags, because that's kinda the entire point of them. There's room for another 3 or 4 flags on `skip_gear`, which is intended for worn things, and it currently has flags for backpack and belt even though we don't currently hide them. Room for another 6 or 7 flags on `skip_body`, which is intended for body parts.

Also fixes a runtime somewhere in there because using `usr` is bad and whoever didn't actually update it to use `user` when adding that parameter should feel bad.


code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -1,22 +1,31 @@
-/mob/living/carbon/human/examine(mob/user)
+// Clothing flags, organized in roughly top-bottom
+#define EXAMINE_SKIPHELMET			0x0001
+#define EXAMINE_SKIPEARS			0x0002
+#define EXAMINE_SKIPEYEWEAR			0x0004
+#define EXAMINE_SKIPMASK			0x0008
+#define EXAMINE_SKIPJUMPSUIT		0x0010
+#define EXAMINE_SKIPTIE				0x0020
```
Arokha: Maybe these should be moved to a defines/examine.dm. They might be useful elsewhere or on other mob types.

code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -26,94 +35,96 @@
...
 		if(wear_suit.flags_inv & HIDEJUMPSUIT)
-			skiparms |= 1
-			skiplegs |= 1
-			skipchest |= 1
-			skipgroin |= 1
-			skipjumpsuit |= 1
-			skiptie |= 1
-			skipholster |= 1
+			skip_body |= EXAMINE_SKIPARMS | EXAMINE_SKIPLEGS | EXAMINE_SKIPBODY | EXAMINE_SKIPGROIN
+			skip_gear |= EXAMINE_SKIPJUMPSUIT | EXAMINE_SKIPTIE | EXAMINE_SKIPHOLSTER
```
Arokha: What if instead you made EXAMINE_SKIPJUMPSUIT just the sum of all these bits rather than a single bit? Is there ever a time where you want to skip the jumpsuit but also not skip the tie and holster?

code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -26,94 +35,96 @@
...
+
 		if(wear_suit.flags_inv & HIDEJUMPSUIT)
-			skiparms |= 1
-			skiplegs |= 1
-			skipchest |= 1
-			skipgroin |= 1
-			skipjumpsuit |= 1
-			skiptie |= 1
-			skipholster |= 1
+			skip_body |= EXAMINE_SKIPARMS | EXAMINE_SKIPLEGS | EXAMINE_SKIPBODY | EXAMINE_SKIPGROIN
```
Arokha: Maybe something similar here, if possible?

code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -26,94 +35,96 @@
...
-			skipshoes |= 1
-			skipfeet |= 1
+			skip_gear |= EXAMINE_SKIPSHOES
+			skip_body |= EXAMINE_SKIPFEET
+
 		if(wear_suit.flags_inv & HIDEGLOVES)
-			skipgloves |= 1
-			skiphands |= 1
+			skip_gear |= EXAMINE_SKIPGLOVES
+			skip_body |= EXAMINE_SKIPHANDS
```
Arokha: Could do the same here.

code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -26,94 +35,96 @@
...
+
 		else if(wear_suit.flags_inv & HIDEHOLSTER)
-			skipholster |= 1
+			skip_gear |= EXAMINE_SKIPHOLSTER
+
 		if(wear_suit.flags_inv & HIDESHOES)
-			skipshoes |= 1
-			skipfeet |= 1
+			skip_gear |= EXAMINE_SKIPSHOES
+			skip_body |= EXAMINE_SKIPFEET
```
Arokha: And here.

Atermonera:
It's preferable to make them individual toggles, in case there are currently or ever need to be cases where it could affect one thing but not the others.
In general, I find it more sensible to explicitly toggle each specific flag than to implicitly toggle one by toggling two others.
And as for not making an extra define with the sum of the bits, that's akin to implicitly toggling, because coders still probably have to check what the define actually does, instead of it being immediately apparent that you're using these flags.
And, I keep the body and gear separate because they have inherently different applications. One determines what gear to hide, the other determines what wounds and flavor text, etc, to hide. They may be similar in function, but they are not identical, so to try to equate them will lead to cases where the behaviour becomes undesirable.

PR #5162, opened 2018-04-11
Optimizes Powercell Update Icons

Neerti:
Hopefully it will prevent this monstrosity.
why

Local testing:
Before change before
After change after


Anewbe:
Early merging because lag.

PR #5164, opened 2018-04-11
Fixed legcuffs not being placeable on handcuffed people

Heroman3003:
And being placeable several times on non-handcuffed.


Heroman3003:
Added a fix for legcuffs icon not being applied correctly.

PR #5170, opened 2018-04-13
Upgrades Planetary Weather to vis_contents

Neerti:
Makes the weather on Sif use vis_contents instead of overlays, which should fix all weather-related icon issues, and may or may not be faster.
Weather updates instantly now.
Fixes indoor PoIs having outdoor weather.


Arokha:
Travis failed for no apparent reason. Might just go to details and click restart job on that last one.

Leshana:
@Neerti Okay, I just checked this branch out on my local, rebased on `master` to fix conflicts and get it all sync'd up, and then ran the unit tests. No active edges, all unit tests passed.   So I think it was just drunk or something.

Atermonera:
Yeah, we've restarted the build multiple times on separate days, trying to get it to work.

PR #5172, opened 2018-04-14
Add several new events

ReddeyfishVor:
These are some of the events we've added have have been using on CHOMPStation.
* window break, causing a single window to fail, and all connected nearby windows to take heavy damage
* electrified door, which causes a door to be damaged and electrified
* escaped slimes, which distributes slimes over the station, with the recommendation to return them to xenobiology
* wormholes, which is the R-UST failure turned into event form, covering the map with portals for a short period of time. If you don't move, you don't get teleported to a random spot nearby the station.
Also includes 
* a refactor of a common function to randomly select an area on the station.
* a minor change to wallrot to report its position after a long duration.

### DIFF
--- a/code/_helpers/events.dm
+++ b/code/_helpers/events.dm
@@ -1,0 +1,15 @@
+/proc/get_station_areas(var/list/area/excluded_areas)
+	var/list/area/grand_list_of_areas = list()
+	// Assemble areas that all exists (See DM reference if you are confused about loop labels)
+	looping_station_areas:
+		for(var/parentpath in global.the_station_areas)
+			// Check its not excluded
+			for(var/excluded_path in excluded_areas)
+				if(ispath(parentpath, excluded_path))
+					continue looping_station_areas
+			// Otherwise add it and all subtypes that exist on the map to our grand list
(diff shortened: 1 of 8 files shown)


elgeonmb:
Lorewise I'd rather not wormholes but as ever I bow to the gods of Gameplay in this matter.

ReddeyfishVor:
Apologies for the failed travis build, chose to not include the dme so that these events would remain unused in your codebase, but forgot that I refactored one of the used events and would require a change to the dme.

You don't have to use these events (such as wormhole), but they are available if you want to. As is, your current enabled events will not change, requiring a modification to code/modules/events/event_container.vr

Atermonera:
Escaped slimes also don't particularly make sense on a map where xenobio is a 2 minute shuttle ride away from the main station.

Atermonera:
Close and comment looked SUSPICIOUSLY LIKE CANCEL COMMENT

elgeonmb:
No objection to having them in the code for bus shenanigens

Anewbe:
Have these been tested?

Leshana:
@Reddeyfish The window break one is a cool idea.

PR #5176, opened 2018-04-15
Better-quality replacements for some gunshot sounds

Serithi:
Sniper and shotgun sounds were kinda ehhh, sound quality wasn't the best (sniper even had some audio clipping) and had a ton of outdoor reverb. Likewise the generic gunshot sound was literally from the original Doom, a sound from 1993 can be kind of out of place in [$CURRENT_YEAR].


Atermonera:
Does it count as a sprite change if they're under the same license?

Belsima:
no because these arent sprites

PR #5182, opened 2018-04-18
Meteor events have fewer waves with less dangerous meteors

Anewbe:
- Meteor events now have a more stable number of waves. Moderate events are now between 3 and 4 waves (from 2, 4, or 6) and major events are now between 3 and 5 waves (from 3, 6, or 9) .
- Meteor events in general throw weaker rocks.

### DIFF
--- a/code/modules/events/meteors.dm
+++ b/code/modules/events/meteors.dm
@@ -6,7 +6,7 @@
 	var/start_side
 
 /datum/event/meteor_wave/setup()
-	waves = severity * rand(1,3)
+	waves = 2 + rand(1, severity)		//EVENT_LEVEL_MAJOR is 3-5 waves
 	start_side = pick(cardinal)
 	endWhen = worst_case_end()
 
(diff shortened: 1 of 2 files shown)


code/modules/events/meteors.dm
```diff
@@ -37,10 +37,10 @@
 			command_announcement.Announce("The colony has cleared the meteor shower", "Meteor Alert")
 
 /datum/event/meteor_wave/proc/get_meteors()
-	switch(severity)
-		if(EVENT_LEVEL_MAJOR)
+	if(EVENT_LEVEL_MAJOR)
+		if(prob(10))
 			return meteors_catastrophic
-		if(EVENT_LEVEL_MODERATE)
```
Atermonera: There's no longer an if statement that specifically processes `EVENT_LEVEL_MODERATE`, such that it will return `meteors_normal`, instead of `meteors_threatening`

Fix'd be:
>if(EVENT_LEVEL_MODERATE || (EVENT_LEVEL_MAJOR && prob(90)))
> > return meteors_threatening
> else if(EVENT_LEVEL_MAJOR) // 1 in 10 EVENT_LEVEL_MAJOR will be catastrophic
> > return meteors_catastrophic
> else return meteors_normal

Anewbe: I'm not sure what you mean here, honestly?
This event only shows up in two severity levels: Major and Moderate. Minor has its own event, Space Dust. Unless I'm going brain-dead again, if this pops up at Major severity, it rolls for `meteors_catastrophic`, and if said roll fails it drops down to `meteors_threatening`. If this event pops up at Moderate, it's always `meteors_normal`.

Atermonera: kk, that clears my confusion

PR #5185, opened 2018-04-18
Fixes borg synth sheet naming when depleted to 1

Verkister:
-The thing changing the sheet stack names to singular/plural no longer breaks the naming for borg sheet synthesisers.


Verkister:
Yeah

PR #5186, opened 2018-04-18
Remove the 'tag' var from every map

Arokha:
It should not be used. It interferes with locate(), and any atom that has `tag` set cannot be garbage collected at all, as the `tag` counts as a reference, therefore it will never reach 0 references.

Older versions of DreamMaker would arbitrarily add the 'tag' var when you did 'Generate instaces from direction' and 'Generate instances from icon-states', however this was removed in later versions. These are all likely from said older versions.

Additionally includes a Travis test to ensure no more tag usage occurs.


Arokha:
image
Tested with a test commit to make sure it caught as expected, and it appears to.

PR #5188, opened 2018-04-19
Fixes mob wall-bashing.

Mechoid:
Weak mobs can no longer smash against walls.

### DIFF
--- a/code/game/turfs/simulated/wall_attacks.dm
+++ b/code/game/turfs/simulated/wall_attacks.dm
@@ -57,12 +57,19 @@
 			source.thermal_conductivity = initial(source.thermal_conductivity)
 
 /turf/simulated/wall/proc/fail_smash(var/mob/user)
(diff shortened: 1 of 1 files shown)


code/game/turfs/simulated/wall_attacks.dm
```diff
@@ -57,12 +57,19 @@
...
 /turf/simulated/wall/proc/fail_smash(var/mob/user)
-	to_chat(user, "<span class='danger'>You smash against the wall!</span>")
+	var/damage_lower = 25
+	var/damage_upper = 75
 	if(isanimal(user))
 		var/mob/living/simple_animal/S = user
 		playsound(src, S.attack_sound, 75, 1)
+		if(!(S.melee_damage_upper >= 10))
+			to_chat(user, "<span class='notice'>You bounce against the wall.</span>")
+			return 0
```
Anewbe: `return FALSE` if you don't mind?

PR #5194, opened 2018-04-19
Species economic modifier is now a species var

Anewbe:
Seems to work as before, aka Tested and Working.
- `species.dm` had a million spaces in it, that's why the diff's huge.
- The economic values were copied straight from the current list. Since Vatborn have the same modifier as baseline humans, they should be able to inherit from the human species.


elgeonmb:
praise

PR #5195, opened 2018-04-20
Sundry Fixes - Emitters, Bedsheets, Conveyors

Leshana:
- Fixes edge case allowing emitters to fire without being on a cable knot.
- Fixes VOREStation/VOREStation#3031 - Conveyor belts in cargo shuttle not obeying switches.
- Fixes VOREStation/VOREStation#3475 - Cannot cut up bedsheets (while holding them)
Port of VOREStation/VOREStation#3485


Neerti:
A minor nitpick, the two numbers referenced in the PR description appear to have been copypasted from the Virgo repo, since here they're pointing to PRs. Since they're PRs I don't think it'll actually matter when this gets merged but if they were pointing to issues, it might've unintentionally closed those.

Leshana:
That is a good point. I have edited the description to point at the real issues.

PR #5205, opened 2018-04-22
Replaces Ice turf sprite for a prettier one

SubberTheFabulous:
New vs Old:


Schnayy:
Can we see how it looks tiled?

SubberTheFabulous:
Compare:
Old

New

PR #5207, opened 2018-04-24
Dufflebag of Holding

Novacat:
Adds a dufflebag of holding, essentially a bag of holding reskin that functions exactly identical


Belsima:
image
The sprite seems lackluster, still appearing mostly a normal dufflebag, as opposed to 
image

Novacat:
I am not too happy with the inventory icon either, but my spriting skill is non existent, I would happily use a replacement.

PR #5209, opened 2018-04-25
 Ports the medical item sounds from Bay

Arokha:
A port of Baystation12/Baystation12#21268, which adds sounds to the healing item stacks, taperolls, and hyposprays


Belsima:
Could these sounds be put somewhere for us to hear? When I see 'gauze' and 'duct tape' sound I expect one of those annoying FPS medical sounds that is overly loud and out of place.

Belsima:
Clicking download brings me to an .ogg file on github without a play button or a way to download.

Arokha:
Click 'View', and then there is a download button

Belsima:
The download button brings me to the .ogg file without a play button or a way to save it.

Arokha:
I don't have that problem, sorry. In fact for me it takes me to a page where I can play, or if I save the page, it saves the ogg file. Maybe try right-clicking on the download or 'view raw' in the middle and doing 'save target as' or 'save link as' or whatever your browser has.

Anewbe:
Sounds're good.

PR #5210, opened 2018-04-26
Fixes an exploit with circuit printers

Arokha:
Allows you to print literally anything.


arsserpentarium:
Hahahaha. You forgot about tools too!

PR #5215, opened 2018-04-27
Cleans up Event messages

Lorwp:
Add `[station_name()]` when practical to event notices

### DIFF
--- a/code/modules/events/dust.dm
+++ b/code/modules/events/dust.dm
@@ -3,13 +3,13 @@
 	endWhen		= 30
 
 /datum/event/dust/announce()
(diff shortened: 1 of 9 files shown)


code/modules/events/dust.dm
```diff
@@ -3,13 +3,13 @@
 	endWhen		= 30
 
 /datum/event/dust/announce()
-	command_announcement.Announce("Debris resulting from activity on another nearby asteroid is approaching your colony.", "Dust Alert")
+	command_announcement.Announce("Debris resulting from activity on another nearby asteroid is approaching [station_name()]", "Dust Alert")
```
Atermonera: `the [station_name()]`

PR #5219, opened 2018-04-27
NanoUI multiple templates example.

Leshana:
## Background
PR #5105 began adding support to NanoUI for combining multiple nano templates together when showing a UI.  Unfortunately the implementation resulted in unintended side effects ( VOREStation/VOREStation#3471 ).
Upon digging deeper I discovered that in fact nanoui **already** supported multiple templates!
Therefore I am reverting PR #5105 , going back to the already existing feature. 

To view the *actual* changes made by this PR, **view the diff of f508841fffe99320582f5976be8e39c066ebd68c**

## Example Implementation
- As a demonstration of the partials inclusion feature, the HTML for the atmospheric scan screen (which was nearly identical between PDAs and Communicators) was moved into a separate file and then included into both pda.tmpl and communicator.tmpl
- Also added a README.dm with the very basics of nanoui and template transclusion.

### DIFF
--- a/nano/README.md
+++ b/nano/README.md
@@ -1,0 +1,39 @@
+# NanoUI Templates
+
+NanoUI uses doT (https://olado.github.io/doT/index.html) as its templating engine.
+
+## Template Markup Tags
+
+Markup tags are used to add dynamic content to the template.
+TODO - This documentation is incomplete.
+
+### Print Tag
+- The print tag outputs variable as text to the UI.
+`{{:data.variable}}`
+
+### If Tag
+- The if tag displays content conditionally based on the provided expression being true.
+- When combined with the else tag the if tag can also show content if the provided expression is false.
+- The else tag can optionally have an expression provided (e.g. "`{{else expression2}}`"), giving it "elseif" functionality.
+
+`{{if expression}} <expression true content> {{/if}}`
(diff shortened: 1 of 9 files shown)


nano/README.md
```diff
@@ -0,0 +1,39 @@
...
+- When combined with the else tag the if tag can also show content if the provided expression is false.
+- The else tag can optionally have an expression provided (e.g. "`{{else expression2}}`"), giving it "elseif" functionality.
+
+`{{if expression}} <expression true content> {{/if}}`
+`{{if expression}} <expression true content> {{else}} <expression false content> {{/if}}`
+`{{if expression1}} <expression1 true content> {{else expression2}} <expression2 true content> {{/if}}`
+
+### For Tag
+- Loop through entries in an array (an array is a list with a numeric index (it does not use strings as keys).
+- Each time the `for` tag iterates though the array it sets a variable (default "value") to the data of the current entry (another variable, default "index", contains the index). An example of this is using the print tag to print the contents (e.g. `{{:value.key1}}` and `{{:value.key2}}`).
```
Atermonera: Maybe show the format to rebind the iterator name here too?

Leshana: Yeah it would be good to document more of this.   There are more features that havn't been documented, but at least there is some instructions instead of zero right?

PR #5224, opened 2018-04-28
Welding fuel has a sprite again

Atermonera:
The iconstate of fuel is "fuel", but the state name of the appropriate sprite was "vaguesplotch". iconstate hasn't been changed for 6 years according to blame, but blame doesn't work on binary files like .dmi's, so I can only easily tell that it was last touched feb 15.

This renames the appropriate state to "fuel", as there are no other instances in code (0 results for a github search of this repo for "vaguesplotch") where the "vaguesplotch" state name is expected.

Compiled and tested:
toast


Belsima:
I believe that would be #4827.

PR #5227, opened 2018-04-29
Updates changelog

Atermonera:
@Schnayy the format is
> changes:
> > [changes]

The `changes:` is required, that's how the parser knows it's about to read changes (This is part of why I proofread).


Schnayy:
Oh, oops! Thanks for catching that!

PR #5234, opened 2018-04-30
Fix signlang being totally broken

Arokha:
As reported in https://github.com/VOREStation/VOREStation/issues/3584, signlang is totally broken. The reason being is that it uses get_mobs_and_objs_in_view_fast but passes it a mob when it expects a turf.

Also makes two other tweaks: signlang only signs to people 'in' the thing you're 'in' if you're not on a turf (aka lockers, ~stomachs~), and signlang doesn't produce an extra 'X gestures' emote 30% of the time, given that it always produces either the message if they know signlang, or a 'gestures a long message' or whatever if they don't.

### DIFF
--- a/code/modules/mob/hear_say.dm
+++ b/code/modules/mob/hear_say.dm
@@ -302,7 +302,7 @@
 			else			adverb = " a very lengthy message"
 		message = "<B>[speaker]</B> [verb][adverb]."
 
-	src.show_message(message)
+	show_message(message, type = 1) // Type 1 is visual message
 
 /mob/proc/hear_sleep(var/message)
 	var/heard = ""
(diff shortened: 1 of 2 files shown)


Atermonera:
I mean, I would argue that if you're in a closed locker or a stomach, it's kinda hard for someone outside of that to actually see you signing, but gameplay>realism

Anewbe:
Seems fine to me. Worse comes to worst, easy to revert.

Arokha:
>I mean, I would argue that if you're in a closed locker or a stomach, it's kinda hard for someone outside of that to actually see you signing, but gameplay>realism

That's what this does, though! Maybe I wasn't clear. This makes it so if you're in a locker, people outside said locker can't see you signing. Previously, they could.

PR #5239, opened 2018-05-02
Corrects missing manifold and pipe in Distro loop

Lorwp:
From this:
image
To this:
image


Lorwp:
Actually since it's not actually my end might as well reopen this.

`travis ded pls restart`

PR #5241, opened 2018-05-02
Changes lizard name

Belsima:
Tokay geckos change color to match their surroundings, and thus would likely be far from green- Carolina anoles are small, more lizardy and match the sprite better, and their color is based on social stress rather than physical surroundings.

Also, they're actually the right shade of green.

Anolis cuvieri might make more sense but they're less common, will swap if asked.


elgeonmb:
make there a small chance of the name being H. unathi

Belsima:
i could do that but also strictly oppose weird tooltip names
also whats the H

elgeonmb:
Homo

PR #5247, opened 2018-05-03
Revert "Dufflebag of Holding"

elgeonmb:
Reverts PolarisSS13/Polaris#5207

Yeah not to be a pendant but a) the description is bad lorewise and b) the sprite is pretty explicitly a devsprite. If my resident spriter's computer will stop acting up I can see about improving the sprites and the desc but as it stands...


Neerti:
The PR to add it was up for more than a week (way too long imo) and the only complaint that was raised against it was its sprite, which it wasn't super good but it wasn't a huge issue and it can be fixed with a new sprite.
The desc can also get modified instead of a wholesale revert, and imo its better to make a PR to change the minor issues instead of reverting it entirely when _it just got merged_ and it isn't causing super horrible issues like crashing the server.

Belsima:
It was a functionally purely cosmetic addition, but the cosmetic choices didn't make sense
it was a copy-paste from the regular bag of holding which uses a different perspective and thus looks fucky on the duffelbag
it also doesn't change the duffelbag as a whole as all other 'x of holdings' do

I don't think it should have been merged because it adds nothing but a cosmetic option whilst the cosmetics were clearly not a finished product

Belsima:
the inventory one is the one i speak of, btw

PR #5248, opened 2018-05-03
Colorblindness

Arokha:
Adds it! In three traits. I assume there's nothing else I need to do to add it to your traits? If so, lemme know.

### DIFF
--- a/code/__defines/misc.dm
+++ b/code/__defines/misc.dm
@@ -238,7 +238,6 @@
 #define USE_FAIL_NOT_IN_USER 6
 #define USE_FAIL_IS_SILICON 7
 
-
 // This creates a consistant definition for creating global lists, automatically inserting objects into it when they are created, and removing them when deleted.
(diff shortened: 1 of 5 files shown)


Arokha:
Can you restart that travis build? It's whatever PoI has an active edge causing it.

Belsima:
Have you fixed Monochromacy?

Arokha:
Fixed in what way?

Belsima:
It just plain didn't work, IIRC, no? except on HUD items

Arokha:
All forms of colorblindness were broken on Vorestation after the planes update. I fixed that over there, and at the same time split this off so Polaris could use it here.

Arokha:
Thankye whoever reran travis.

Atermonera:
_Shakes boot angrily at travis_

KasparoVy:
How the heck did you calculate those matrices? :thinking: 
It took me forever to figure out Taj/Vulp colourblindness and I'm pretty sure I went about it HELLA WRONG

PR #5252, opened 2018-05-05
Fixes .44 magazine icons

SpadesNeil:
The icon was wrong.


Woodratt:
Appears to already be fixed?

SpadesNeil:
Did somebody else fix them? Because there's conflicts now.

PR #5253, opened 2018-05-06
Originalize window title

qwertyquerty:
Reasons:

* It looks better
* Many servers do it
* I want to add support for Polaris on [ss13rp](https://github.com/qwertyquerty/ss13rp)
* Why not?


Belsima:
Polaris Station 13 sounds super wrong and bad

Atermonera:
If I had to pick something, I'd personally prefer `SS13 Polaris`, but I think it does warrant further discussion, agreeing with belsima that `Polaris Station 13` doesn't sound right

qwertyquerty:
I'm fine with anything that has the word polaris in it.

PR #5255, opened 2018-05-08
Adds a client FPS setting

Arokha:
Title, really! But, it adds a client FPS setting in the global tab of character preferences, that allows you to adjust your client FPS. Takes immediate effect if adjusted while you're already in a mob. Re-applied each time you switch mobs for safety's sake.


Arokha:
@Atermonera Beep

PR #5256, opened 2018-05-08
Gives basic medical access to the Detective

Lorwp:
Due to the morgue now being blocked behind medical doors.
Either this or just revoking morgue access to resolve this oversight

### DIFF
--- a/code/game/jobs/job/security.dm
+++ b/code/game/jobs/job/security.dm
@@ -54,8 +54,8 @@
 	supervisors = "the head of security"
 	selection_color = "#601C1C"
 	idtype = /obj/item/weapon/card/id/security/detective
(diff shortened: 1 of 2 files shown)


Lorwp:
Guessing the build will have to be restarted on travis to resolve active edges ¯\_(ツ)_/¯

Atermonera:
_Applies boot to travis, medium style_

Anewbe:
I'm fairly certain that Detectives having Morgue access is a holdover from TG. I'd argue that if anything should be done here, it'd be the inverse of this PR, aka removing the access where it's superfluous. I don't know how that'll affect our downstreams, though, and as such it seems to me that the best option here is to continue to ignore the fact that they have said access.

PR #5257, opened 2018-05-08
Laser pointers

PrismaticGynoid:
Adds laser pointers, ported from tg/oldaurora (but heavily modified). They can be found in the loadout (for 2 points), as well as printed by R&D. By default, they come with the basic microlaser (rating 1), but higher-powered microlasers (also from R&D) can be used to upgrade it, making it more powerful (still only situationally useful, though).

In addition to its intended use of pointing at things, you can also use it to: (with laser ratings of [1/2/3])
- Shine it into someone's eyes for a [20/40/60]% chance of causing [momentary blindness/minor eye damage/a small chance to stun] to them, the effects of which can be counteracted with eye protection.
- Shine it at the sensors of silicon mobs for a [20/40/60]% chance to disable them for around 1 or [1/2/3] seconds.
- Shine it at a camera lens ([20/40/60]% chance of hitting) for a [33.3/50/100]% chance of blinding that camera for about [30/45/90] seconds. It does set off a camera alarm, though.
- Shine it at the ground within a tile of a cat to get them to pounce on it (or at least catch their attention).

Exact numbers and probabilities are subject to change if necessary. To make all these features work, I also had to fix 3 issues:
- Cat resting sprites only faced east. Now they're multidirectional.
- Cameras could only be EMPed once, and it wouldn't disable them. Now it does and it can happen more than once.
- Calling flash_eyes() on a mob wouldn't properly display an effect. Now it does (disabled for silicons, for some reason, but that can be changed too).


Belsima:
>Shine it at the sensors of silicon mobs for a [20/40/60]% chance to disable them for around 1 or [1/2/3] seconds.

I'm hoping this isn't ranged, or else that's a huge way to just shut down any traitor synth?

Belsima:
Im not certain a ranged instastun has much capability to be balanced. Generally, even up close ones are super hard to balance.

Also, whats the cooldown?

PrismaticGynoid:
I'll probably just remove the stun from being the main effect on silicons.
Cooldown is 1 second between uses, and using it as fast as possible will only give you around 9-10 uses before it runs out of energy and has to wait a while to recharge itself.

PrismaticGynoid:
Updated the numbers, stun on silicons is now made much, much less likely.

zekesulastin:
I'm personally a bit concerned about R&D being able to spam someone with a pointer and wreck their eyesight (not sure what other sources for R 2/3 diodes are around), since they need either a specialised chemical or surgery to fix.

I'll let someone else like Belsima worry about the stun balancing :p

PrismaticGynoid:
Fair, though anyone can achieve a similar effect on an unprotected person with a fork or screwdriver. It'll probably get you arrested by security/ talked to by staff/ beaten to death by an angry mob if you upgrade a laser pointer specifically to go around dealing small amounts of eye damage to people. I'm pretty sure high-powered micro lasers can only be found in telecomms machines and from R&D, and ultra-high-powered micro lasers only from R&D, and those use uranium. I could still remove/reduce the eye damage if you're concerned, though.

code/game/objects/items/devices/laserpointer.dm
```diff
@@ -146,9 +146,11 @@
 		var/mob/living/silicon/S = target
 		//20% chance to actually hit the sensors
```
Belsima: whats the current chance to actually fuck them

PrismaticGynoid: If you're talking about the [1/2/3] stun?
(prob(effectchance * diode.rating)) * (prob(3 * diode.rating))
[20/40/60]% chance of hitting the sensors, and a further [3/6/9]% chance of doing the stun. Combined, that means it's a [0.6/2.4/5.4]% each time you shine the laser pointer to actually do the stun. It's really not very likely.

Belsima: yeah that sounds solid to me thanks

code/game/objects/items/devices/laserpointer.dm
```diff
@@ -0,0 +1,218 @@
...
+
+/obj/item/device/laser_pointer/attackby(obj/item/W, mob/user)
+	if(istype(W, /obj/item/weapon/stock_parts/micro_laser))
+		if(!diode)
+			user.drop_item()
+			W.loc = src
+			diode = W
+			to_chat(user, "<span class='notice'>You install a [diode.name] in [src].</span>")
+		else
+			to_chat(user, "<span class='notice'>[src] already has a cell.</span>")
```
Atermonera: `Cell` may be misleading here, as every other instance I can think of uses it in reference to a _power_ cell, as opposed to a _laser_ cell here.

code/game/objects/items/devices/laserpointer.dm
```diff
@@ -0,0 +1,218 @@
...
+		var/obj/machinery/camera/C = target
+		if(prob(effectchance * diode.rating))
+			C.emp_act(4 - diode.rating)
+			outmsg = "<span class='notice'>You shine the [src] into the lens of [C].</span>"
+			add_attack_logs(user,C.name,"Tried disabling using [src]")
+		else
+			outmsg = "<span class='info'>You missed the lens of [C] with [src].</span>"
+
+	//cats!
+	for(var/mob/living/simple_animal/cat/C in view(1,targloc))
```
Atermonera: `viewers()` only returns a list of mobs, slightly more appropriate here

code/game/objects/items/devices/laserpointer.dm
```diff
@@ -0,0 +1,218 @@
...
+		..()
+	return
+
+/obj/item/device/laser_pointer/afterattack(var/atom/target, var/mob/living/user, flag, params)
+	if(flag)	//we're placing the object on a table or in backpack
+		return
+	laser_act(target, user)
+
+/obj/item/device/laser_pointer/proc/laser_act(var/atom/target, var/mob/living/user)
+	if(!(user in (viewers(world.view,target))))
```
Atermonera: Feels backwards to check if user is in the mobs that can see the target (List of mobs), but probably better than checking if the target is in the things that can be seen by the mob (List of atoms)

code/game/objects/items/devices/laserpointer.dm
```diff
@@ -0,0 +1,218 @@
...
+			add_attack_logs(user,C.name,"Tried disabling using [src]")
+		else
+			outmsg = "<span class='info'>You missed the lens of [C] with [src].</span>"
+
+	//cats!
+	for(var/mob/living/simple_animal/cat/C in view(1,targloc))
+		if(prob(50) && !(C.client))
+			C.visible_message("<span class='warning'>[C] pounces on the light!</span>", "<span class='warning'>You pounce on the light!</span>")
+			step_towards(C, targloc)
+			C.lay_down()
```
Atermonera: Does it need to be called twice?

code/game/objects/items/devices/laserpointer.dm
```diff
@@ -0,0 +1,218 @@
...
+				C.lay_down()
+		else
+			C.set_dir(get_dir(C,targloc))
+			C.visible_message("<span class='notice'>[C] watches the light.</span>", "<span class='notice'>Your attention is drawn to the mysterious glowing dot.</span>")
+
+
+	//laser pointer image
+	icon_state = "pointer_[pointer_icon_state]"
+	var/list/showto = list()
+	for(var/mob/M in range(world.view,targloc))
```
Atermonera: `range()` finds things through walls and other dense objects

code/game/objects/items/devices/laserpointer.dm
PrismaticGynoid: Didn't notice that, I'll change it.

code/game/objects/items/devices/laserpointer.dm
PrismaticGynoid: Change view to viewers, then?

code/game/objects/items/devices/laserpointer.dm
PrismaticGynoid: It's to make the cat "pounce" on the laser dot, but stand back up when the dot disappears. There's probably a cleaner way of doing it, but I'm not sure what.

code/game/objects/items/devices/laserpointer.dm
PrismaticGynoid: Is there a better way of doing this I could use?

Atermonera: `viewers()` would be more appropriate, since only the mobs that can see `src` can see `src` pointing a laser at someone

PR #5258, opened 2018-05-08
Fix weak mobs being able to smash anything they want.

Mechoid:
Weak mobs (Below 10 damage output) cannot smash destructible structures like Closets. Goddamnit Mice.


Atermonera:
`damage < 10` captures `!damage`, because 0<10

PR #5261, opened 2018-05-09
Ports 'fake' SSAO from /tg/ as a preference

Arokha:
Allows players to pick whether they want this visual feature or not. Basically adds drop-shadows to objects and stuff. It's off by default.

## Disabled (Default)
image

## Enabled
image


PrismaticGynoid:
Looks nice, but how does it affect performance? And, can it be disabled for objects like holopads which probably shouldn't have a shadow behind them?

Arokha:
Won't affect server performance at all, all this is rendered clientside.

For that, sure. Just set those to TURF_PLANE and ABOVE_TURF_LAYER and you should be good. I'm not ... suuuper sure I really want to 'accept a list of suggestions' because this PR will stay open forever though as people tack on one more thing at a time.

Maybe just compile a list amongst yourselves and tack those on?

PR #5269, opened 2018-05-16
Makes lighting a little fancier, reduces the amount of light sources with a power greater than 1

deathride58:
Title. The new values are based off of my own studies of lights in the real world. This should make lighting look a lot more realistic, instead of lighting being pure white against pure black in most areas.
This PR also reduces quite a few light power values. With robustlight2/goonlights/vglights, the light_power variable is essentially the alpha of a light source, so having it exceed 1 can result in some pretty binary light behavior. Reducing the light_power variable will result in the station being darker across the board, but will make most light sources behave in a slightly more realistic manner.
This also makes some minor tweaks to how flashlights handle their light values. This allows lights to actually shine a little brighter when their brightness is set to high, and makes flashlights glow darker when their brightness is set to low.

Before
image

After
image

### DIFF
--- a/code/__defines/lighting.dm
+++ b/code/__defines/lighting.dm
@@ -74,5 +74,12 @@
 #define LIGHT_COLOR_TUNGSTEN   "#FAE1AF" //Extremely diluted yellow, close to skin color (for some reason). rgb(250, 225, 175)
 #define LIGHT_COLOR_HALOGEN    "#F0FAFA" //Barely visible cyan-ish hue, as the doctor prescribed. rgb(240, 250, 250)
 
+//Lighting values used by the station lights
+#define LIGHT_COLOR_FLUORESCENT_TUBE "#E0EFFF"
+#define LIGHT_COLOR_FLUORESCENT_FLASHLIGHT "#CDDDFF"
+#define LIGHT_COLOR_INCANDESCENT_TUBE "#FFEEDD"
+#define LIGHT_COLOR_INCANDESCENT_BULB "#FFDDBB"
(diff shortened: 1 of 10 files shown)


Atermonera:
Would increasing the power of the lights lessen the darkening, or would a fix for that require mapping more lights?

Have you tested this with Diana, to ensure that it's not a significant nerf to them? They get nutrition from lights, and especially with flashlights being battery powered, they really can't reliably supply their own light source. Diona move slowly enough without hunger slowdown to hamper them even further when walking in main hallways.

Belsima:
I greatly prefer the system, but it looks dark for a clean NT station- mayhaps buff lights so that maint stays darker, though? I've noticed as of now its hard to find _any_ darkness, even in maint.

deathride58:
@Atermonera Yep yep, increasing the power variable of individual lights will brighten the station up, though the default brightness range of lights combined with the lacking light density seems to be the primary contributing factor to the overall darkness of the station with this PR. The most recent commit I've made to this PR should help make the average light level of the main station match how it currently is on live, as I've increased the default range of tube lights to accommodate for the station's light density.

Before the most recent commit:
image

After the most recent commit:
image

After a little bit of testing on my part, I've found that with the old brightness range values, dionae ended up suffering during casual strolling without a flashlight. The new brightness range values alleviate this, letting dionae function mostly the same as they have prior to this PR.

Arokha:
Any new thoughts on this?

Anewbe:
Unfortunately it'll probably take until this approaching Sunday to get an official ruling on this.

Anewbe:
Discussed and approved, sorry for the long wait.

PR #5273, opened 2018-05-17
Ports the Stasis Only Sleeper

Anewbe:
Ported from Virgo.
Actually mapping them in will come later.

### DIFF
--- a/code/game/machinery/Sleeper.dm
+++ b/code/game/machinery/Sleeper.dm
@@ -345,3 +345,9 @@
 			to_chat(user, "The subject has too many chemicals in their bloodstream.")
 	else
 		to_chat(user, "There's no suitable occupant in \the [src].")
+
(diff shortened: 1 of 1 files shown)


Arokha:
Those have special sprites they use, too. You mightcould grab them. They are from /tg/ so open source yadda yadda.

Anewbe:
Honestly not a fan of those sprites, don't fit in too well in my opinion.

Arokha:
Maybe? But to what end? There's no control panel to control injections/dialysis/etc.

Atermonera:
Unless someone builds a console next to it, in which case it will automatically link afaik

Atermonera:
It also doesn't override the `stasis_choices` list, so if you build a sleeper console next to it, it'll function exactly the same as a regular sleeper

PR #5274, opened 2018-05-18
Adjust Promethean limbs.

Mechoid:
One of many PRs I've made in response to a thread on the forums. (Because people think I don't read them or dismiss them?) Took a while to pinpoint the exact issues. This is likely not a final version as nothing is, but it's the numbers I've concluded are 'suitable for testing at this point'.
General:
- Limbs can spread their damage to parent and children limbs via the spread_dam var.

Promethean:
- Limbs have a higher max health of 20, rather than 5. They no longer only take 5 damage on their limbs, then stop taking damage. This Is Not A Buff.
***This is the likely source of 90% of complaints regarding Promethean passive regeneration that I have pinpointed, despite repetitive prodding of it. They never stored a large enough amount of damage on their limbs to make it slow.***
- Limbs can splatter any time they take 5 damage more of brute or burn than the maximum, in addition to normal limb severing.
- Limbs will spread their damage when reaching max damage, to their parent and child at reduced rates, without their sharp or edge bonuses. This damage overflow can splatter limbs as per normal damage.

### DIFF
--- a/code/modules/organs/organ_external.dm
+++ b/code/modules/organs/organ_external.dm
@@ -24,6 +24,7 @@
 	var/brute_dam = 0                  // Actual current brute damage.
 	var/burn_dam = 0                   // Actual current burn damage.
 	var/last_dam = -1                  // used in healing/processing calculations.
+	var/spread_dam = 0
 
 	// Appearance vars.
 	var/nonsolid                       // Snowflake warning, reee. Used for slime limbs.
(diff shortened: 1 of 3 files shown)


Mechoid:
I should note the Torso and Groin DO NOT spread their damage at this time, because they are Vital, and thus have no maximum damage limit despite the var being set.

Mechoid:
Reopening to kick Travis because Active Edges??? Again.

PR #5277, opened 2018-05-19
Seed storages are hackable.

Mechoid:
Seed storages in the garden and xenoflora are now hackable for More Goodies, with a few possible sets of concerning seeds to choose from. Due to this, and the possibility of someone just going through all of them in hopes each one was rolled, the electronics are 'secure', and thus randomized between them.

### DIFF
--- a/code/datums/wires/seedstorage.dm
+++ b/code/datums/wires/seedstorage.dm
@@ -1,0 +1,60 @@
+/datum/wires/seedstorage
+	holder_type = /obj/machinery/seed_storage
+	wire_count = 4
+	random = 1
+
+var/const/SEED_WIRE_SMART = 1
+var/const/SEED_WIRE_CONTRABAND = 2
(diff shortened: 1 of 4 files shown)


code/datums/wires/seedstorage.dm
```diff
@@ -0,0 +1,60 @@
+/datum/wires/seedstorage
+	holder_type = /obj/machinery/seed_storage
+	wire_count = 4
+	random = 1
+
+var/const/SEED_WIRE_SMART = 1
```
Atermonera: Do these need to be `var/const` as opposed to `#define`?
Defines use less memory (i.e., no memory as opposed to however much is required for a var) because the preprocessor parses them to their defined values at compile-time

And anyways global variables are generally bad.

Mechoid: I've never touched wires before this, so I am not entirely certain on this. Will check.

PR #5279, opened 2018-05-21
Laptops don't eat IDs

Atermonera:
Laptops will now relinquish their prey.
Pathing is now absolute, which involved changing the indentation of basically every single line in all of those files. Dualslot cardslots are subtypes again, because that's proper application of OOP principles.
Cleaned up grammar of various conditional statements, reordered a handful to be more concise.


Atermonera:
tested

PR #5285, opened 2018-05-22
Change accessories to use bitflags

Arokha:
Allows accessories to 'block' multiple slots. An example of why this is useful:

(Example given in current system)
A uniform contains "utility" as a valid accessory slot, and "utility" is also 'restricted' on the uniform (which means that only 1 can be installed at a time). Webbing and Holster cannot be used together, they are both slot "utility". Webbing and holster are both limited to 1 each per uniform.

I want to add an item (particularly a machette sheath) that can be used with webbing, but not with holsters. This is impossible given the above system. If I make the machette "utility", it cannot be used with webbing. If I make a new slot type "weapon" and put holster in it along with machette, then holsters can now be used with webbing, an unintended change.

So, changing them to bitfields allows them to essentially occupy more than one slot. Holsters now occupy 'utility' and 'weapon' which has the same effect as before. Basically this PR does nothing that anyone will notice, but makes it so you can make accessories exclude each other better.

### DIFF
--- a/code/__defines/items_clothing.dm
+++ b/code/__defines/items_clothing.dm
@@ -22,19 +22,23 @@
 #define SLOT_TIE        0x4000
 #define SLOT_HOLSTER	0x8000 //16th bit - higher than this will overflow
 
-#define ACCESSORY_SLOT_UTILITY  "Utility"
-#define ACCESSORY_SLOT_ARMBAND  "Armband"
-#define ACCESSORY_SLOT_RANK     "Rank"
-#define ACCESSORY_SLOT_DEPT		"Department"
-#define ACCESSORY_SLOT_DECOR    "Decor"
-#define ACCESSORY_SLOT_MEDAL    "Medal"
-#define ACCESSORY_SLOT_INSIGNIA "Insignia"
-#define ACCESSORY_SLOT_ARMOR_C  "Chest armor"
-#define ACCESSORY_SLOT_ARMOR_A  "Arm armor"
-#define ACCESSORY_SLOT_ARMOR_L  "Leg armor"
-#define ACCESSORY_SLOT_ARMOR_S  "Armor storage"
-#define ACCESSORY_SLOT_ARMOR_M  "Misc armor"
-#define ACCESSORY_SLOT_HELM_C	"Helmet cover"
+#define ACCESSORY_SLOT_UTILITY	0x1
+#define ACCESSORY_SLOT_WEAPON	0x2
+#define ACCESSORY_SLOT_ARMBAND  0x4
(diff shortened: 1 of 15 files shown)


code/modules/clothing/under/xenos/vox.dm
```diff
@@ -1,8 +1,8 @@
 /obj/item/clothing/under/vox
 	has_sensor = 0
 	species_restricted = list(SPECIES_VOX)
-	valid_accessory_slots = list("vox")
-	restricted_accessory_slots = list("vox")
+	valid_accessory_slots = "vox"
```
Anewbe: Will these even work now, since it's not a bitflag?

code/__defines/items_clothing.dm
```diff
@@ -22,19 +22,23 @@
...
-#define ACCESSORY_SLOT_HELM_C	"Helmet cover"
+#define ACCESSORY_SLOT_UTILITY	0x1
+#define ACCESSORY_SLOT_WEAPON	0x2
+#define ACCESSORY_SLOT_ARMBAND  0x4
+#define ACCESSORY_SLOT_DECOR    0x8
+#define ACCESSORY_SLOT_MEDAL    0x20
+#define ACCESSORY_SLOT_TIE		0x40
+#define ACCESSORY_SLOT_INSIGNIA 0x80
+#define ACCESSORY_SLOT_OVER		0x100
+//Should these really be 'accessory' accessories
```
Anewbe: I'd be open to other suggestions.

Arokha: It's fine for now imo, might as well not worry about it until we run out of flags 🤡

Anewbe: I always worry about that sort of thing, is the issue.

code/modules/clothing/under/xenos/vox.dm
Arokha: I was hoping you wouldn't notice thaaaaat. ;( You told me once Vox aren't really used.

I can use up a flag on them, but that onnneee item is the only thing that does that. Is it a big deal if other people can put on Vox webbing? ;v

Anewbe: At Some Point ™️ there'll be a handful of OP vox accessories to make up for their lack of backpacks, so it's kinda bad.

code/__defines/items_clothing.dm
Arokha: There's no other "good" way to do it, really. The other way is expensive.

Worst comes to worst, you override the 'can attach accessory' proc on armor and offload all the armor flags to some other 'armor slots' var.

PR #5286, opened 2018-05-23
Paranoid logging and byond account age

Arokha:
Adds the PARANOIA_LOGGING setting to config.txt (and PANIC_BUNKER which was missing for some reason) and updates example.

PARANOIA_LOGGING (also available as a verb for admins in Server) begins notifying admins of any 0day byond accounts, or users who have never joined before. For when you don't quite want to panic bunker.

Also makes this information available in player panel. Requires SQL connection.

image


Arokha:
Travis pls

PR #5287, opened 2018-05-23
Adds 'Smite' verb and pray button

Arokha:
For easier adminbuse in response to things. c:

* Break legs
* Lightning strike
* Combust
* Bluespace Arty

### DIFF
--- a/code/modules/admin/verbs/smite.dm
+++ b/code/modules/admin/verbs/smite.dm
@@ -1,0 +1,83 @@
+/client/proc/smite(var/mob/living/carbon/human/target in player_list)
+	set name = "Smite"
+	set desc = "Abuse a player with various 'special treatments' from a list."
+	set category = "Fun"
+	if(!check_rights(R_FUN))
+		return
+
+	if(!istype(target))
+		return
+
+	var/list/smite_types = list(SMITE_BREAKLEGS,SMITE_BLUESPACEARTILLERY,SMITE_SPONTANEOUSCOMBUSTION,SMITE_LIGHTNINGBOLT)
+
+	var/smite_choice = input("Select the type of SMITE for [target]","SMITE Type Choice") as null|anything in smite_types
+	if(!smite_choice)
+		return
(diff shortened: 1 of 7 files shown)


code/modules/admin/verbs/smite.dm
```diff
@@ -0,0 +1,83 @@
...
+		
+		if(SMITE_SPONTANEOUSCOMBUSTION)
+			target.adjust_fire_stacks(10)
+			target.IgniteMob()
+			target.visible_message("<span class='danger'>[target] bursts into flames!</span>")
+		
+		if(SMITE_LIGHTNINGBOLT)
+			var/turf/T = get_step(get_step(target, NORTH), NORTH)
+			T.Beam(target, icon_state="lightning[rand(1,12)]", time = 5)
+			target.adjustFireLoss(75)
```
Anewbe: Could this not be done with electrocute_act() ?

Arokha: It'd only shock hands, and be resisted by gloves.

Anewbe: electrocute_act can be passed a bodypart to target, can it not?

Arokha: I guess, but then what? I just pick() from the list of all their body parts and randomly only their right leg gets burned? :P I dunno, the overall seems like a nice smitey touch. Probably won't sever any limbs since all spread out.

Anewbe: Was thinking head or chest. Might not sever/ash a limb, but that's I think still 75 burn and 10 seconds of weaken, and burns that bad are fairly nasty.

Anewbe: Not really a big deal, but it feels more thematic, and if I really want to screw someone over for praying something silly, I have more dangerous methods.

Arokha: Did it!

PR #5289, opened 2018-05-24
Prometheans can now be revived as originally intended.

fludd12:
I remember reading a long while ago that they were supposed to be revivable by injecting their core with... *something*. This was obviously intended to happen at *some point*, given their brains were marked as open containers in potentially the worst way available, so I simply filled in the rest.

You can now revive them by injecting their core with 40u of phoron. Doing so makes them not have the reprocussions they normally get from cloning or sleeving, but they spawn moderately hungry, damaged, and limbless. They have exactly enough nutrition to try to regenerate once if they're quick.

Open to suggestions for improvement or alteration of values!

### DIFF
--- a/code/modules/organs/internal/brain.dm
+++ b/code/modules/organs/internal/brain.dm
@@ -137,9 +137,102 @@ GLOBAL_LIST_BOILERPLATE(all_brain_organs, /obj/item/organ/internal/brain)
 	icon_state = "green slime extract"
 	parent_organ = BP_TORSO
 	clone_source = TRUE
+	flags = OPENCONTAINER
 
 /obj/item/organ/internal/brain/slime/is_open_container()
 	return 1
+
+/obj/item/organ/internal/brain/slime/New()
+	..()
+	create_reagents(50)
+
+/obj/item/organ/internal/brain/slime/proc/reviveBody()
+	var/datum/dna2/record/R = new /datum/dna2/record()
+	R.dna = brainmob.dna
+	R.ckey = brainmob.ckey
+	R.id = copytext(md5(brainmob.real_name), 2, 6)
+	R.name = R.dna.real_name
+	R.types = DNA2_BUF_UI|DNA2_BUF_UE|DNA2_BUF_SE
+	R.languages = brainmob.languages
+	R.flavor = list()
+	for(var/datum/modifier/mod in brainmob.modifiers)
+		if(mod.flags & MODIFIER_GENETIC)
+			R.genetic_modifiers.Add(mod.type)
+
+	var/datum/mind/clonemind = brainmob.mind
+
+	if(!istype(clonemind, /datum/mind))	//not a mind
+		return 0
+	if(clonemind.current && clonemind.current.stat != DEAD)	//mind is associated with a non-dead body
+		return 0
+	if(clonemind.active)	//somebody is using that mind
+		if(ckey(clonemind.key) != R.ckey)
+			return 0
+	else
+		for(var/mob/observer/dead/G in player_list)
+			if(G.ckey == R.ckey)
+				if(G.can_reenter_corpse)
+					break
+				else
+					return 0
+
+	for(var/modifier_type in R.genetic_modifiers)	//Can't be revived. Probably won't happen...?
+		if(istype(modifier_type, /datum/modifier/no_clone))
+			return 0
+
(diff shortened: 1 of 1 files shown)


fludd12:
Frick, what went wrong

fludd12:
`>active edges failed unit test`
`>I didn't touch mapping`
Someone's due for a paddling.

Atermonera:
It's a known issue, we're kicking travis as necessary to make it pipe down until the issue is resolved.

fludd12:
Glad to know I didn't cause it!

code/modules/organs/internal/brain.dm
```diff
@@ -137,10 +137,103 @@ GLOBAL_LIST_BOILERPLATE(all_brain_organs, /obj/item/organ/internal/brain)
...
+
+	H.UpdateAppearance()
+	H.sync_organ_dna()
+	if(!R.dna.real_name)	//to prevent null names
+		R.dna.real_name = "promethean ([rand(0,999)])"
+	H.real_name = R.dna.real_name
+
+	H.nutrition = 260 //Enough to try to regenerate ONCE.
+	H.adjustBruteLoss(40)
+	H.adjustFireLoss(40)
```
Mechoid: Has this been tested and confirmed to not double to 80 damage? As most if not all modifications of burn or brute are modified by their burn or brute species mods.

Mechoid: I might suggest the nutrition be set to 300, as it's a single regen, and should give enough leeway if someone isn't paying attention as per below.

code/modules/organs/internal/brain.dm
```diff
@@ -137,10 +137,103 @@ GLOBAL_LIST_BOILERPLATE(all_brain_organs, /obj/item/organ/internal/brain)
...
+	return 1
+
+/datum/chemical_reaction/promethean_brain_revival
+	name = "Promethean Revival"
+	id = "prom_revival"
+	result = null
+	required_reagents = list("phoron" = 40)
+	result_amount = 1
+
+/datum/chemical_reaction/promethean_brain_revival/can_happen(var/datum/reagents/holder)
```
Mechoid: Assumedly this will take the ghost regardless of in-body (in-brain?) or not, and use the phoron similarly?

code/modules/organs/internal/brain.dm
fludd12: Since they're both doing the same amount of damage, and Prometheans have just as much resistance to brute as they do weakness to burn, the end result is them being put near-crit-but-not-quite regardless.

As for the leeway, they have something like 5-10 seconds to start regenerating, and I wanted regeneration *or* healing, but not both. If you think they should get 300 nutrition, though, I'll gladly make that change!

code/modules/organs/internal/brain.dm
fludd12: Correct - this is mostly the same code that cloning machines used. The chemical reaction calls the `reviveBody()` proc on the brain, which proceeds to try and shove the ghost back in the brain if it isn't there already. It'll fail if the ckey belongs to a different mind or if the mind is already in a living body, though, so no need to worry about body shenanigans.

Mechoid:
That open container thing is why Mechoid is not allowed to do important things with reagents.

fludd12:
*pat pats* Reagent system is hellish sometimes, don't worry.

code/modules/organs/internal/brain.dm
Atermonera: `40/2+40*2 = 100 != 40*2` as a note

fludd12: Oh.

Well, in testing, they were still at least conscious enough to use the regenerate ability, so :woman_shrugging:

Mechoid:
As a confirmation, you have indeed tested this?

fludd12:
It's currently live on a downstream server or two, and so far is working fine. The only note is that the reformation throws the same runtimes that cloning appears to do normally, but that seems to be a sort of higgs-bugson level issue with limb code.

Wickedtemp:
So, limbless, but not without heads?

fludd12:
Correct. I considered also removing head, but figured I'd give medbay a break - grinding up food and injecting it isn't what I'd consider fun.

Mechoid:
No other changes planned on this yeah?

fludd12:
Not unless you can think of anything!

PR #5295, opened 2018-05-24
Spacesuits are now in their own dmi file

Anewbe:
Should be working.
Basically, moved a bunch of spacesuits into another file, to stop suit.dmi and suits.dmi from eating things for a little while. There's probably a better way to do this, but it seems to work on my local.


Arokha:
I don't think there's a better way to do it tbh. Not that I'm aware of anyway.

PR #5302, opened 2018-05-26
Ports Baystation12 newest item pickup animations. Fancy!

drexample:
Demonstration video: https://streamable.com/z2evr

Original PR by thasc: https://github.com/Baystation12/Baystation12/pull/21092

The animation is a little faster than in video (very little, almost unnoticeably)

### DIFF
--- a/code/game/objects/effects/item_pickup_ghost.dm
+++ b/code/game/objects/effects/item_pickup_ghost.dm
@@ -1,0 +1,17 @@
+/obj/effect/temporary/item_pickup_ghost
+	anchored = 1
+	unacidable = 1
+	mouse_opacity = 0//just in case something dumb happens
+	var/lifetime = 0.2 SECONDS
+
+/obj/effect/temporary/item_pickup_ghost/proc/assumeform(var/obj/item/picked_up)
+	icon = picked_up.icon
+	icon_state = picked_up.icon_state
+	pixel_x = picked_up.pixel_x
+	pixel_y = picked_up.pixel_y
+	color = picked_up.color
+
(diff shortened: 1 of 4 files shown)


Arokha:
So fancy!

nerdass:
balance issue with traitor objectives. motion draws attention, this directly nerfs being able to sneakily pocket something off a table, because the object moving on the screen will make it a lot easier to notice.

Anewbe:
I don't see that as a serious concern. If you're pocketing something worth stealing and there are other people watching, there's pretty good chance you'll get called on it, anyway.

Atermonera:
It's also a very short animation. If there's only one thing on the table then it's considerably harder to pocket it, but if there's a ton of stuff in a pile, so long as the thing you're swiping isn't on top (Which can be done, move-to-top is a verb), there won't actually be much information about the object pocketed available to other people.
It's generally much more informative when you stow it in a bag and generate the "X stows Y in their bag"

drexample:
I can port Pickpocket's Gloves as syndicate items with such ability, to prevent such animation happening (also with bonuses they give), if you think that's required.

Arokha:
You could also do it more like attack animations, where it's not an object at all, just a flicked overlay, and you could use a different range to flick the overlay in, based on w_class, so that smaller items can only be seen to be picked up from closer, and larger from further (which is somewhat similar to how backpack 'X puts Y into Z' works)

drexample:
It uses already existing effect system and can be adapted to produce that effect. Is it really needed though? Would the change be really noticeable, considering item is put in hands immedeately.

Arokha:
I don't personally care, I think it's fine as-is. I was just mentioning that it's possible to do it with different ranges, using the existing attack animation system.

PR #5305, opened 2018-05-27
Prevent autosay() from creating/deleting AIs

### LINKED ISSUE #5271: AI creation spam due to announcements
Every time an announcement happens, a /mob/living/silicon/ai is spawned to say a line, then deleted. AIs all get hard del()'d because they are mobs, and it takes ~1 second to delete them, thus creating an annoying lagspike whenever it happens.

Not lag whenever there's an announcement.

#### What actually happened
Lag.

#### Steps to reproduce
Announce.

Arokha:
Now it gets one dedicated AI it can move around instead.

Fixes #5271

### DIFF
--- a/code/game/objects/items/devices/radio/radio.dm
+++ b/code/game/objects/items/devices/radio/radio.dm
@@ -247,14 +247,12 @@ var/global/list/default_medbay_channels = list(
 	if (!connection)
 		return
 
-	var/mob/living/silicon/ai/A = new /mob/living/silicon/ai(src, null, null, 1)
+	var/static/mob/living/silicon/ai/announcer/A = new /mob/living/silicon/ai/announcer(src, null, null, 1)
 	A.SetName(from)
 	Broadcast_Message(connection, A,
 						0, "*garbled automated announcement*", src,
 						message, from, "Automated Announcement", from, "synthesized voice",
 						4, 0, list(0), connection.frequency, "states")
-	qdel(A)
-	return
 
 // Interprets the message mode when talking into a radio, possibly returning a connection datum
 /obj/item/device/radio/proc/handle_message_mode(mob/living/M as mob, message, message_mode)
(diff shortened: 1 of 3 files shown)


code/game/objects/items/devices/radio/radio.dm
```diff
@@ -247,14 +247,12 @@ var/global/list/default_medbay_channels = list(
 	if (!connection)
 		return
 
-	var/mob/living/silicon/ai/A = new /mob/living/silicon/ai(src, null, null, 1)
+	var/static/mob/living/silicon/ai/announcer/A = new /mob/living/silicon/ai/announcer(src, null, null, 1)
```
Atermonera: And the `static` here means the Singleton AI will be found when the proc is called by other objects, or will `new` still make fresh copies?

Arokha: The way it appears to work (I was not aware of this before working on this issue, actually) is that declaring a static inside a proc makes it a sort of 'global static' that is only initialized once. I believe it makes it literally a global var, but somehow scoped to just this proc. This line is basically 'skipped' after it's initialized once (in truth, it's initialized when the game starts, this line 'isn't really here). It's a little weird how Byond handles this, but, it works and it's apparently intended to behave like I've described.

Atermonera: Ok. Byond reference docs don't describe static, or if they do it's somewhere hidden, but it's not surprising there's support for a Singleton-style object management

Arokha: Oh, sorry. `static` is an alias for `global`, maybe I misunderstood your question.
They are the same thing.

Anewbe:
As a question, this announcement AI doesn't show up on any sort of visible interface, a la Law Interface thing?

Arokha:
It should not. For example the silicon law editor admin verb does:
`var/mob/living/silicon/S = input("Select silicon.", "Rename Silicon.") as null|anything in silicon_mob_list`
And this AI removes itself from all those mob lists when it's created.

Anewbe:
Works for me.

PR #5310, opened 2018-05-28
Fixed arcades dispensing their own board

### LINKED ISSUE #5298: Arcade cannot be deconstructed after first win
Arcade removes its board from itself as a first prize. To start deconstruction of arcade it must have a board inside.

Arcade not to literally dispense its most important part. And to be deconstructable.

#### What actually happened
Won a board that should have stayed inside the machine, screwdriver does nothing anymore.

#### Steps to reproduce
- Win arcade once
- Try deconstructing it.

Heroman3003:
Fixes #5298

### DIFF
--- a/code/game/machinery/computer/arcade.dm
+++ b/code/game/machinery/computer/arcade.dm
@@ -45,7 +45,7 @@
 		qdel(src)
 
 /obj/machinery/computer/arcade/proc/prizevend()
-	if(!contents.len)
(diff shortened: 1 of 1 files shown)


Anewbe:
Did you test this? Because I feel like a better method would be to define a list of prizes, rather than picking randomly from contents-circuit.

Atermonera:
Especially considering we have random spawner objects that you could use to pick something for your

PR #5313, opened 2018-05-29
Move character join log statement

Arokha:
Was only logging latejoins due to where this line was placed. Moved it into job controller to catch any time anyone is assigned a job period.


Anewbe:
Travis issue isn't just active edges, can't easily copy the error over as I am on my phone.

PR #5317, opened 2018-05-30
Dead Ringer + PoI

drexample:
Adds:
Dead Ringer - A fancy watch, that can fake owner's death.
PoI for it

Death Prevention goes like this - whenever user is attacked, they turn invisible and a fake corpse is spawned. After a while, corpse disappears and invisibility ends. 

(also tweaks mob AI a little to include ignoring invisible mobs) 

default


Anewbe:
There are a number of concerns I personally have with this PR, and I want to bring this up to the rest of the Polaris Staff (which will take some time) before this gets properly reviewed.

My concerns are as follows:
- I don't want any more human simple mobs showing up in our PoIs, especially things that would otherwise be antag rounds. It raises some IC ethical issues, OOC sanity issues, and it also imbalances any PoI the mob is in, unless certain specific things are done.
- Both item ideas are cool, even though they're directly from TF2. Until someone properly reviews them (which, again, may take some time), I'm concerned about having them. What exactly would a nearby player experience if someone else uses either watch?
- If you can't think of another way to add these items, I still definitely like having them around, and they may be properly added at a later date, or presumably somewhere downstream.

drexample:
@Anewbe
Well, let’s try to find a compromise, then. 
In lore channel loremasters told me all things in PoIs are not canon and cultmobs are allowed so:
I can make cultsist non-human shades in cult attire, disappearing after being defeated. 
I also can replace cultists with their remains and just put more angry mobs around. 

As for items ideas - only Dead Ringer is from TF2, the stopwatch is just a thing made up by me. Dead Ringer has similar, but not completely same functions it has in TF2. Neither of watches’ descriptions point out their effects, so as far sanity of the is concerned, they’re extremely complex high-tech devices, that were supposed to be guarded. 

If you have any better ideas how to balance stuff out, I’d be glad to hear them.

Anewbe:
- Cultmob means Faithless, bats, and the blob things. Your shade solution is probably fine, though.
- My question was "If someone uses them near me, what happens to me, what do I see?"

drexample:
Dead Ringer (when activated) triggers on user taking any physical (burn or brute damage), turning them invisible and spawning a fake corpse. People near will just see them die (with deathgasp triggering) and all. Corpse disappears in a small cloud of harmless smoke shortly after. 

Stopwatch creates a timestop field that freezes all mobs, projectiles and thrown objects on short period of time in small area. They see a hologram of a bigger version of the watch, indicating affected area.

drexample:
active edgelords have arrived and screwed travis check

drexample:
@Anewbe Can I have "Do not merge" label removed, or are there more changes required?

Anewbe:
This was brought up at the June 10th Staff Meeting. There's a majority in favor of having the Dead Ringer added, but there are numerous concerns about the Stopwatch. If you can separate the two of them, the Dead Ringer will likely be added, the Stopwatch is going to require us going through the code with a fine-toothed comb, which takes time.

drexample:
Okay, sure. 
What should I do with the stopwatch? Just remove it or comment it out?

Anewbe:
If you could split the stopwatch code into a separate PR, that would be the best solution.

drexample:
Got it.

Anewbe:
Thank you.

drexample:
@Anewbe sorry for a long holdup, end of academic year is hell.

code/modules/artifice/deadringer.dm
```diff
@@ -0,0 +1,183 @@
...
+	var/fireloss_prev = 999999
+	var/mob/living/carbon/human/corpse = null
+	var/mob/living/carbon/human/watchowner = null
+
+
+/obj/item/weapon/deadringer/New()
+	..()
+	processing_objects |= src
+
+/obj/item/weapon/deadringer/Destroy() //just in case some smartass tries to stay invisible by destroying the watch
```
Anewbe: You need to remove the item from the processing_objects list somewhere in this proc.

code/modules/artifice/deadringer.dm
```diff
@@ -0,0 +1,183 @@
...
+	uncloak()
+	..()
+
+
+/obj/item/weapon/deadringer/dropped()
+	if(timer > 20)
```
Anewbe: Why is this check here? If you drop the source of your cloak, you should probably automatically uncloak.

code/modules/artifice/deadringer.dm
```diff
@@ -0,0 +1,183 @@
...
+			to_chat(H, "<font color='blue'>You press a small button on [src]'s side. It starts to hum quietly.</font>")
+			bruteloss_prev = H.getBruteLoss()
+			fireloss_prev = H.getFireLoss()
+			activated = 1
+			return
+		else
+			to_chat(H,"<font color='blue'>You press a small button on [src]'s side. It buzzes a little.</font>")
+			return
+	if(activated)
+		to_chat(H,"<font color='blue'>You press a small button on [src]'s side. It stopped humming.</font>")
```
Anewbe: `stops humming`

code/modules/artifice/deadringer.dm
drexample: It may be silly, but I use timer this way:
When it's above 20, it keeps its owner invisible, when it less or equal 20, it just goes into cooldown. Decided to conserve resources by using just one timer. If you feel there's a better solution to this, please tell.

code/modules/artifice/deadringer.dm
drexample: done

SpadesNeil:
neat

PR #5322, opened 2018-06-02
Move planets to SSplanets

Arokha:
Fuggit! Just make this into a subsystem. It probably works.

If it doesn't, what's the worst that will happen? ;v


Anewbe:
Notice any lag with this?

Arokha:
Nop, SSplanets since it's now an SMC subsystem shows the % tick used by the planets subsystem and it's generally averaging to like 1%.

Anewbe:
This is sort of on hold, need to have some background discussion about the possible IC side-effects, and how we'll adjust for them.

While this is happening, could you make a quick changelog that says something like "Weather processes faster now." so people get at least some warning about it?

Anewbe:
Staff seems to be generally in favor, but we'll need to change a few weather things before it gets implemented. If you care to help us out on that front, I can list them, otherwise it will take some small amount of time.

Anewbe:
Waiting on #5350, sorry for the delay.

Anewbe:
And I can't actually test my thing properly without this, so merging!

PR #5324, opened 2018-06-02
Deconstructing a Technomancer Core now does Bad Things

Anewbe:
Currently either makes a Supermatter explosion or a smallish Tesla ball, directly on the deconstructor after a five second delay. I couldn't think of anything else suitable to have happen, so until then it's a two item switch. Sorry 'bout that.

### DIFF
--- a/code/game/gamemodes/technomancer/core_obj.dm
+++ b/code/game/gamemodes/technomancer/core_obj.dm
@@ -134,6 +134,23 @@
 		wards_in_use -= ward
 		qdel(ward)
 
+// This makes a variety of bad things happen if someone decons the core.
+/obj/item/weapon/technomancer_core/decon_act(var/mob/M)
+	var/turf/T = get_turf(src)	// This should only be used for the coordinates, below. Everything else should use get_turf(src) to stop research borgs from fleeing their own deconstructors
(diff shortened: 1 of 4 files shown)


zekesulastin:
Can we like not wipe out that entire side of the station (including medical!) because someone used R&D?  What in the hell.

elgeonmb:
I mean if you permafuck an antag (potentially the _only_ antag) by putting their Antag Box in your Science Box and you don't like, at least take the measure of putting it in a Science Box that's not right next to medical that's kind of on you

zekesulastin:
I mean it's not like you can't just get rid of the tech levels or something to make it non deconstructable like half the other goodies in the game, no, you have to fuck everyone because "muh antags".

elgeonmb:
man why not make the engine just vaporize engineers as a penalty for not maintaining it correctly, why does it have to fuck cargo and anyone who enjoys having power too

drexample:
Uhh... You can deny the antag by just throwing the core into space. 
I think there needs to be a hard bomb-defusing level way to disable the core before deconstruction.

zekesulastin:
If you're seriously going to compare the entire engine setup and warning process to this, elgeon, then I really don't know what to say to you.

Anewbe:
If a better solution can be found, this will be discarded.

Anewbe:
Closing this, we'll be seeking a solution that prevents Core deconstruction, rather than being horribly destructive.

PR #5327, opened 2018-06-03
Changes "Setup Character" to "Character Setup"

Cameron653:
Since that's what is used everywhere else. **including the window that pops up when you click it.**


Rekeme:
Best change NA.

PR #5328, opened 2018-06-04
Communicator Glasses

MisterLayne:
The future is **now**.


Anewbe:
I'm hesitant to merge this, because it's contrary to our long-standing plans to make communicators entirely replace PDAs, including the slots they can occupy. Communicator watches can be easily made to fit into this plan, but an eyepiece is a bit much.

MisterLayne:
Is that a definite 'no', then?

Anewbe:
I'll let @Atermonera give you a final verdict on that, but from what I can tell it leans that way.

Atermonera:
How can you do messaging with commglasses? You can't say speaking unless you code it to act as such, like how radios do it.
How do you fit an ID card into it? (planned feature that's part of my ongoing project)
Where do cartridges attach to to not be big bulky awkward things? 
How do you use something like a medical scanner , which gives a readout, with a pair of glasses?
(Possibly more questions to come, I have to get back to work. Unless you can provide a satisfactory answer to all of them this will be denied)

MisterLayne:
Rest in peace, Google Glass.

PR #5335, opened 2018-06-07
Ports some Bay hair code

Anewbe:
- Ports the Very Short hair feature, which means hats and helmets with BLOCKHEADHAIR will allow certain hairstyles to hang around rather than overwriting them. Doesn't seem to have any notable effect on any of the few headpieces we have with said var, but it's there.
- Ports the Tie Hair verb. Certain hairstyles will allow you to switch between those hairstyles using a verb.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_powers.dm
+++ b/code/modules/mob/living/carbon/human/human_powers.dm
@@ -1,6 +1,38 @@
 // These should all be procs, you can add them to humans/subspecies by
 // species.dm's inherent_verbs ~ Z
 
+/mob/living/carbon/human/proc/tie_hair()
+	set name = "Tie Hair"
+	set desc = "Style your hair."
+	set category = "IC"
+
+	if(incapacitated())
+		to_chat(src, "<span class='warning'>You can't mess with your hair right now!</span>")
+		return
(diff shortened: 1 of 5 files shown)


code/modules/mob/living/carbon/human/human_powers.dm
```diff
@@ -1,6 +1,38 @@
...
+			to_chat(src, "<span class ='warning'>Your hair isn't long enough to tie.</span>")
+			return
+		else
+			var/list/datum/sprite_accessory/hair/valid_hairstyles = list()
+			for(var/hair_string in hair_styles_list)
+				var/list/datum/sprite_accessory/hair/test = hair_styles_list[hair_string]
+				if(test.flags & HAIR_TIEABLE)
+					valid_hairstyles.Add(hair_string)
+			selected_string = input("Select a new hairstyle", "Your hairstyle", hair_style) as null|anything in valid_hairstyles
+		if(incapacitated())
```
Atermonera: Byond is single-threaded, so unless something between the prevent `incapacitated()` check and here could cause `src` to become incapacitated (Which I don't see, if something like that's here), this second check is redundant

PR #5337, opened 2018-06-07
Folders props, and adds the laser prism, receptor, and sealed doors.

Mechoid:
- Adds the Laser Prism, an object meant for PoIs with laser puzzles. 
- Adds the Laser Receptor, basically a giant 'lock' for future laser-based puzzles. Requires an emitter blast, or a sufficiently Big laser of other types. I.E., a cannon.
- Adds the sealed door, basically an indestructible blast door for puzzle rooms.

Props are given a prop-er base object, and the common functionality moved there.


code/game/objects/structures/props/beam_prism.dm
```diff
@@ -0,0 +1,218 @@
...
+		if(new_bearing <= -1 || new_bearing > 360)
+			to_chat(user, "<span class='warning'>Rotating \the [src] [new_bearing] degrees would be a waste of time.</span>")
+			return
+	else
+		var/choice = input("What point do you want to set \the [src] to?", "[name]") as null|anything in compass_directions
+		new_bearing = round(compass_directions[choice])
+
+	var/rotate_degrees = new_bearing - degrees_from_north
+
+	if(new_bearing == 360) // Weird artifact.
```
Atermonera: In case you didn't test values over 360 and don't want it turning more than a full revolution in one action for whatever reason, `new_bearing %= 360` will force it to be on [0, 359]

code/game/objects/structures/props/beam_prism.dm
```diff
@@ -0,0 +1,218 @@
...
+	if(rotation_lock)
+		visible_message("<span class='notice'>\The [src] shudders.</span>")
+		playsound(src, 'sound/effects/clang.ogg', 50, 1)
+		return
+
+	visible_message("<span class='notice'>\The [src] rotates to a bearing of [new_bearing].</span>")
+
+	var/rotate_degrees = new_bearing - degrees_from_north
+
+	if(new_bearing == 360)
```
Atermonera: See above re: `%= 360`

code/game/objects/structures/props/beam_prism.dm
```diff
@@ -0,0 +1,218 @@
...
+	var/two_stage = 0
+	if(rotate_degrees == 180 || rotate_degrees == -180)
+		two_stage = 1
+		var/multiplier = pick(-1, 1)
+		rotate_degrees = multiplier * (rotate_degrees / 2)
+
+	playsound(src, interaction_sound, 50, 1)
+	if(two_stage)
+		animate(src, transform = turn(src.transform, rotate_degrees), time = 3)
+		sleep(3)
```
Atermonera: I think putting this inside a `spawn(3)` will have the same effect, and be desirable because spawn's nonblocking

code/game/objects/structures/props/beam_prism.dm
```diff
@@ -0,0 +1,218 @@
...
```
Atermonera: Above re: `spawn()`

code/game/objects/structures/props/beam_prism.dm
```diff
@@ -0,0 +1,218 @@
...
+
+		Proj.redirect(new_x, new_y, curloc, null)
+
+/obj/structure/prop/prism/incremental
+	free_rotate = 0
+	description_info = "This device is capable of redirecting any beam projectile, but only locks to specific positions in rotation."
+
+/obj/structure/prop/prism/incremental/externalcont
+	external_control_lock = 1
+	description_info = "This device is capable of redirecting any beam projectile, but only locks to specific positions in rotation when rotated by a control dial."
```
Atermonera: `but can only be rotated by a control dial to specific positions.`
Current text implies it can be rotated to any position if you don't use the dial

code/game/objects/structures/props/beam_prism.dm
```diff
@@ -0,0 +1,218 @@
...
+
+	var/free_rotate = 1
+	var/list/compass_directions = list()
+	for(var/obj/structure/prop/prism/P in my_turrets)
+		if(!P.free_rotate)
+			free_rotate = 0
+		else
+			free_rotate = 0
+			compass_directions = P.compass_directions
+			break
```
Atermonera: For each prism I control, if it can only rotate to cardinals, then I can only rotate to cardinals. 
If it can be rotated to non-cardinals, then I can only be rotated to cardinals, and the directions I can rotate to are the cardinals
(Such that cardinals includes composites, like NW)

That's the description of this for loop. I don't think it's what you're intending, but I could be wrong

code/game/objects/structures/props/puzzledoor.dm
```diff
@@ -0,0 +1,92 @@
+// An indestructible blast door that can only be opened once its puzzle requirements are completed.
+
+/obj/machinery/door/blast/puzzle
+	name = "puzzle door"
+	desc = "A large, virtually indestructible door that will not open until some form of requirement(s) is or are met."
```
Atermonera: `will not open unless certain requirements are met.`

Try reading what you have out loud. It's awkward

code/game/objects/structures/props/beam_prism.dm
Mechoid: I want 360 to be an applicable input, and reset to 0 as it is. 360 when input into the rotation method for some reason points it randomly one degree to the left or right, but 0 doesn't.

code/game/objects/structures/props/beam_prism.dm
Mechoid: This is the same method that crawling uses, which is what I based the entirety of the animation code on, but I can change it if necessary

PR #5350, opened 2018-06-12
Weather is shorter, hail is less deadly.

Anewbe:
Weather now has a pair of variables attached to it that tell it how long to run before it tries to make a change, rather than having a hardcoded timer on all weather. Weather timers have been dropped from 20-30 minutes down to 5-10. Variety is the spice of life, after all.

Hail has also been tweaked, in anticipation of #5322, which will make it tick far more often, turning a relatively minor annoyance into a deadly storm.
- Hail checks for new weather every 2-5 minutes, so people won't be trapped in shelter waiting for it to change.
- Hail's damage has been dropped from 5-10, down to 1-3. That _should_ give people time to get to cover, but if not, other systems can be implemented later.

### DIFF
--- a/code/modules/planet/sif.dm
+++ b/code/modules/planet/sif.dm
@@ -184,6 +184,7 @@ datum/weather/sif
 		)
 
 /datum/weather/sif/snow/process_effects()
+	..()
 	for(var/turf/simulated/floor/outdoors/snow/S in SSplanets.new_outdoor_turfs) //This didn't make any sense before SSplanets, either
 		if(S.z in holder.our_planet.expected_z_levels)
 			for(var/dir_checked in cardinal)
@@ -207,6 +208,7 @@ datum/weather/sif
 		)
 
 /datum/weather/sif/blizzard/process_effects()
+	..()
 	for(var/turf/simulated/floor/outdoors/snow/S in SSplanets.new_outdoor_turfs) //This didn't make any sense before SSplanets, either
 		if(S.z in holder.our_planet.expected_z_levels)
 			for(var/dir_checked in cardinal)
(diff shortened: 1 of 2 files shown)


code/modules/planet/weather.dm
```diff
@@ -25,7 +25,7 @@
 	if(current_weather)
 		old_light_modifier = current_weather.light_modifier // We store the old one, so we can determine if recalculating the sun is needed.
 	current_weather = allowed_weather_types[new_weather]
-	next_weather_shift = world.time + rand(20, 30) MINUTES
+	next_weather_shift = world.time + rand(current_weather.timer_low_bound, current_weather.timer_low_bound)
```
Atermonera: `current_weather.timer_low_bound, current_weather.timer_**high**_bound`

Anewbe: Dammit. Brainfritz again.

code/modules/planet/sif.dm
```diff
@@ -330,7 +332,7 @@ datum/weather/sif
 			if(amount_soaked >= 10)
 				continue // No need to apply damage.
 
-			H.apply_damage(rand(5, 10), BRUTE, target_zone, amount_blocked, amount_soaked, used_weapon = "hail")
+			H.apply_damage(rand(1, 5), BRUTE, target_zone, amount_blocked, amount_soaked, used_weapon = "hail")
 			to_chat(H, "<span class='warning'>The hail smacks into you!</span>")
```
Neerti: It might be worth putting an ``if(prop(some_number))`` statement here to reduce the chatlog getting filled with hailspam.

Anewbe: An excellent catch, but since rain and storms do the same thing, I'm setting up a timer instead.

PR #5351, opened 2018-06-13
Fixes "glass of cup of coffee"

MagmaRam:
also changes porcelain to ceramic because let's be real here


elgeonmb:
how dare you

MagmaRam:
I broke this way back when and now I'm fixing it

PR #5354, opened 2018-06-13
Fulton balloons

Anewbe:
This is a straight up port from Virgo. Should be good to merge, **but not be added anywhere**, due to concerns about aspects left over from their port from TG. I'm opening this to save me time in the near future, and to get the code reviewed to find any immediate bug-or-runtime related problems.

Adds the Fulton Recovery System, AKA Fulton Balloons. These come in a pack of three, and they can be attached to most things to have them sent almost immediately to a selected Beacon.


code/modules/mining/fulton.dm
```diff
@@ -0,0 +1,191 @@
...
+		to_chat(user, "<span class='notice'>You start attaching the pack to [A]...</span>")
+		if(do_after(user,50,target=A))
+			to_chat(user, "<span class='notice'>You attach the pack to [A] and activate it.</span>")
+			/* No components, sorry. No convienence for you!
+			if(loc == user && istype(user.back, /obj/item/storage/backpack))
+				var/obj/item/storage/backpack/B = user.back
+				B.SendSignal(COMSIG_TRY_STORAGE_INSERT, src, user, FALSE, FALSE)
+			*/
+			uses_left--
+			if(uses_left <= 0)
```
Atermonera: `if(--uses_left <= 0)`
Should the user just drop it when it runs out of charges, or should it delete itself?

Anewbe:
As mentioned in the original comment, this won't be added anywhere, because there's a lot of work that I need to do to it before it works the way it needs to. I opened this pull request to get the assets into the code, so I don't need to deal with it later.

PR #5355, opened 2018-06-14
Bioprinters and Cloners now use Biomass reagent, maps in a Bioprinter

Anewbe:
Adds a new reagent called `biomass`, which is made from protein, sugar, and phoron.
- The Cloner now uses liquid biomass for clones, rather than meat. The one on the map starts out full, with three bottles of biomass, but its capacity can be increased simply by replacing the beakers with larger ones. As always, Science is your friend if you want the really, really silly stuff.
- Bioprinters now use liquid biomass to print limbs. Like the Cloner, it starts with a full bottle of biomass, and its capacity can be increased by getting a bigger beaker.

Biomass, and by extension the Bioprinter, is deliberately far more costly in materials than the Prosthetics Fabricator, as part of the compromise that gets the Bioprinter  added to the map. The cloner is also more expensive now, but the presence of defibrillators means that cloning occurs far less often, and the noticeable change should be relatively minor.


code/game/machinery/cloning.dm
```diff
@@ -348,10 +346,66 @@
...
+					biomass_count += R.volume
+
+	return biomass_count
+
+// Removes [amount] biomass, spread across all containers. Doesn't have any check that you actually HAVE enough biomass, though.
+/obj/machinery/clonepod/proc/remove_biomass(var/amount = CLONE_BIOMASS)		//Just in case it doesn't get passed a new amount, assume one clone
+	var/to_remove = 0	// Tracks how much biomass has been found so far
+	if(LAZYLEN(containers))
+		for(var/obj/item/weapon/reagent_containers/glass/G in containers)
+			if(to_remove < amount)	//If we have what we need, we can stop. Checked every time we switch beakers
```
Mechoid: Does this not always return 1, as to_remove starts out as 0, less than the amount needed? If to_remove is how much has already been found, it would make sense if it checked it for being greater than what has been found.

Anewbe: Look at what the line below does. If the machine hasn't found enough biomass, it runs through the beakers and removes 60 total.
If 'to_remove >= amount', then we just return, because we've removed enough.

Mechoid: ..Yes I'm just an idiot and now question why I'm doing this tired.

PR #5366, opened 2018-06-19
Fix AI-drone runtimes. Get angry at commits.

Mechoid:
Fixes some runtimes with AI-drone languages.

I hate commits. _I just want to use the same branch for the same project is that too much to ask._


Anewbe:
You can force a new branch of the same name over the old branch.

PR #5369, opened 2018-06-21
Only slimes may squish.

MisterLayne:
Plain and simple. 
Also it's still keeping my old commits here even though it shouldn't be, send help.


Atermonera:
Ok if it's doing that then you need to reset your master branch to a commit before the clandestine commit, pull from upstream master to bring it back up to date, and then push to origin to update your fork.

To reset, you need to open the `Show Log` menu on tortoisegit, find a commit before the clandestine commit, right click it, and select "Reset 'master' to this commit..." while on your master branch

code/modules/mob/living/carbon/human/emote.dm
```diff
@@ -79,7 +79,7 @@
 
 		//Promethean-only emotes
 		if("squish")
-			if(!species.bump_flag == SLIME) //That should do, yaya.
+			if(!(species.bump_flag == SLIME)) //No. This should do it.
```
Zuhayr: `species.bump_flag != SLIME`

PR #5376, opened 2018-06-27
Changelings Now Succ Traits

### LINKED ISSUE #4850: Changeling Transform doesn't remove Large/Small Traits

When transforming, all transformations will be under the large/small trait you chose, dispelling the ability to blend in with them.

#### Additional info:

MisterLayne:
Changelog not included.
Fixes #4850 (Probably)


Neerti:
It would probably be better to copy someone's 'genetic' modifiers instead of their mind's traits, since traits are actually just things to tell the game to do something specific when that character is instanced at roundstart (which currently is just 'give me this modifier').
Using someone's traits means that their previous form's modifiers are not removed and it might be possible to be holding a combination of modifiers that don't really make sense (like high metabolism + low metabolism).
Modifiers have a 'genetic' flag that tells certain pieces of code to preserve it for later (e.g. cloning, MMI stuff). It would be better in my opinion for lings to use that as well.
An example on how to iterate over a mob's modifiers and get the genetic traits can be seen here. https://github.com/PolarisSS13/Polaris/blob/c72807f458d0897a69aae83db52ed224ff17a422/code/game/machinery/computer/cloning.dm#L342.
You'd also need to remove all genetic modifiers before transforming, but then if a ling absorbed someone small, then turns into them, they would become small. If they transform into their old form, they'd lose the small modifier and get to normal size.

PR #5384, opened 2018-07-01
Saber fix.

lbnesquik:
This add sprites along with changes to magazines.dm, automatic.dm, weapons.dm


Anewbe:
https://github.com/PolarisSS13/Polaris/blob/4a0acb27f178257ada1566c033a2e5b36692b858/code/modules/projectiles/guns/projectile/automatic.dm#L85

That's the Saber? It's working perfectly fine, I don't see why you say this is a fix, and it honestly doesn't add anything I find necessary.

Anewbe:
I stand corrected. This could have been avoided if you'd actually written out the specifics on what you're trying to change in your original post, as point 6 of the [Coding Rules](https://github.com/PolarisSS13/Polaris/wiki/Code-Submission-Rules) requires.

This still isn't going to be merged as-is. Instead of correcting a base class, you've decided to add a nearly identical subclass with very few changes, and that's unnecessary. If the issue is a missing magazine, correct that on the base class.
- Since you're adding a magazine, with the proper `load_method`, you don't need the `max_shells` var. The magazines themselves will have the maximum ammo allowed for the weapon.
- On top of that, you're replacing the base bullet type with an AP variant. That wasn't mentioned in the original post, and even if it was, this won't be merged until that's corrected.

lbnesquik:
I made a subclass for the saber as it was quite hard to spawn as is.
I don't personally expect the research SMG to be in  /obj/item/weapon/gun/projectile/automatic instead of something like  /obj/item/weapon/gun/projectile/automatic/research SMG.
Maybe /obj/item/weapon/gun/projectile/automatic could be given to a basic automatic firearm instead.

I will remove the max shells var in the update.

As for the bullet issues, apologies for not explaining this prior, it was to give the weapon some interesting property because to me, a simple 20 shots SMG doesn't have much justification for diamond and silver. It's not particularly innovative in any way.

SpadesNeil:
Trying to make `/obj/item/weapon/gun/projectile/automatic` fixed up to have magazines is a bad idea, and I'll tell you why.

Now **all automatic weapons** will inherit the traits of `/obj/item/weapon/gun/projectile/automatic` and that is going to break a lot of things irreparably and also prevent a lot of future flexibility with the code.

Trust me, I already tried this years ago, and ended up with the same solution of making a file path called `/obj/item/weapon/gun/projectile/automatic/saber` and I just changed the RnD SMG over to it. There are certain things in the code that are really not meant to be spawned and will invariably be broken. This is one of those things. It's a base class that shouldn't be appearing in the game.

Atermonera:
Point re: base class being just a template, full of markers to indicate when someone didn't properly override a particular variable, is very much valid and good practice, we already do that with a number of items, and honestly this is the sort of thing that sooner or later should be refactored to follow suit.
That said, if you're having trouble with inheritance then I'm gonna squint at you and not really believe you're overriding every variable correctly unless I get a solid example of code that's not working.

lbnesquik:
So in this case, keep the /saber but fix the issues with loading and maybe revert the ammo type back to standard?

Atermonera:
Yes, BUT also change variables on the base class to be more indicative that it's a template class, instead of a feature (More in line with what `/obj/item/weapon/gun/projectile` has for name, desc, etc), and comment what `multi_aim` and `burst_delay` do, as well as what the values in a `firemode` list represent

lbnesquik:
Right, i'll do that tomorrow

code/modules/projectiles/guns/projectile/automatic.dm
```diff
@@ -1,25 +1,41 @@
...
 	multi_aim = 1
 	burst_delay = 2
 
-//	one_handed_penalty = 15
-
 	firemodes = list(
 		list(mode_name="semiauto",       burst=1, fire_delay=0,    move_delay=null, burst_accuracy=null, dispersion=null),
 		list(mode_name="3-round bursts", burst=3, fire_delay=null, move_delay=4,    burst_accuracy=list(0,-15,-15), dispersion=list(0.0, 0.6, 1.0))
 //		list(mode_name="short bursts",   burst=5, fire_delay=null, move_delay=4,    burst_accuracy=list(0,-15,-15,-30,-30), dispersion=list(0.6, 1.0, 1.0, 1.0, 1.2)),
-		)
```
Atermonera: Missing indentation. May technically be valid, only passed travis because this is the last line of the scope

Atermonera:
Also, this is on your master branch. You should never code on your master branch, but rather on new branches off of your master branch with names at least vaguely related to the code you're working on that branch.
Reason for this is that, if you have commits to master in a PR that hasn't been merged yet, and you DO make a branch, your branch now also has those commits, and it's a much more complex set of git operations to fix this issue than it is to properly use branches in the first place.

code/modules/projectiles/guns/projectile/automatic.dm
```diff
@@ -1,25 +1,40 @@
-/obj/item/weapon/gun/projectile/automatic //Hopefully someone will find a way to make these fire in bursts or something. --Superxpdude //Except burstfire isn't fit for an rp server --Mark
+/obj/item/weapon/gun/projectile/automatic //This should never be spawned in, it is just here because of code necessities. 
+	name = "daka SMG"
+	desc = "A small SMG. You really shouldn't be able to get this gun. Uses 9mm rounds."
+	icon_state = "c05r"	//Used because it's not used anywhere else
+	load_method = SPEEDLOADER
+	ammo_type = /obj/item/ammo_casing/a9mm
+	projectile_type = /obj/item/projectile/bullet/pistol
+
+//Burst is the number of bullets fired; Fire delay is the time you have to wait to shoot the gun again, Move delay is the same but for moving after shooting. .
```
Anewbe: These comments are unnecessary, as the code performed here is done on the base gun, rather than on automatics.

code/modules/projectiles/guns/projectile/automatic.dm
```diff
@@ -1,25 +1,40 @@
...
+	icon_state = "saber"//Still ugly
 	w_class = ITEMSIZE_NORMAL
-	load_method = SPEEDLOADER //yup. until someone sprites a magazine for it.
+	load_method = MAGAZINE  //This should fix it
 	max_shells = 22
 	caliber = "9mm"
 	origin_tech = list(TECH_COMBAT = 4, TECH_MATERIAL = 2)
 	slot_flags = SLOT_BELT
-	ammo_type = /obj/item/ammo_casing/a9mm
+	magazine_type = /obj/item/ammo_magazine/m9mmR/saber/ap
```
Anewbe: No. Give it the base bullets.

code/modules/research/designs/weapons.dm
```diff
@@ -47,9 +47,10 @@
 
 /datum/design/item/weapon/smg
 	id = "smg"
+	desc = "An compact reliable SMG firing armor piercing ammo."
 	req_tech = list(TECH_COMBAT = 4, TECH_MATERIAL = 3)
-	materials = list(DEFAULT_WALL_MATERIAL = 8000, "silver" = 2000, "diamond" = 1000)
-	build_path = /obj/item/weapon/gun/projectile/automatic
+	materials = list(DEFAULT_WALL_MATERIAL = 8000, "silver" = 2500, "diamond" = 1000)
```
Anewbe: I'm not sure why you've changed the cost here.

lbnesquik:
Cost was changed because of ap bullets because otherwise it has absolutely nothing that is really research like

code/modules/projectiles/guns/projectile/automatic.dm
lbnesquik: Someone said I should put them in anyway

Anewbe:
Please change the base cost and the bullets back to what they were, or this PR will be closed in accordance with our [Coding Rules](https://github.com/PolarisSS13/Polaris/wiki/Code-Submission-Rules), which you've already been linked.

lbnesquik:
No need to get on a horse about it. I've been holding the changes so i could get some actual feedback on the purpose and relevance of the saber as a _research_ item. Not to give it a stealth buff.

PR #5390, opened 2018-07-05
Fixes some planets controller runtimes

Arokha:
Comes up if you use outdoor turfs in places other than planet z-levels


Arokha:
Whoops, I had fixed that, but forgot to commit it.

PR #5391, opened 2018-07-05
Tweak i_beam qdel logic

Arokha:
Seems sketchy!

Doesn't return a qdel() hint and apparently sometimes enters infinite loops, perhaps next is sometimes circular? This makes it slightly more robust, anyway.

### DIFF
--- a/code/modules/assembly/infrared.dm
+++ b/code/modules/assembly/infrared.dm
@@ -250,9 +250,8 @@
 	return
 
 /obj/effect/beam/i_beam/Destroy()
+	. = ..()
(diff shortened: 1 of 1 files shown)


gozulio:
This took 3 years to fix. Here's the commit where the proc was apparently coded.
https://github.com/PolarisSS13/Polaris/commit/e7d85f8edc7a38caccfd3e82a2fbba9c36595fa1

Also relevant issue down stream. https://github.com/Repede/YawnWiderStation/issues/175
Apparently someone did something with an IR laser, and the processor melted into hot slurry.

PR #5409, opened 2018-07-12
Adds roofing tiles

Atermonera:
They're used to place plating overhead without going up a floor, and to remove weather effects. Must be placed such that the new section of roofing/plating is over a wall or adjacent to an existing section, i.e., actually physically supported by something. Roofs still won't collapse, however, that sounds computationally expensive and out of scope.

Changing a turf with ChangeTurf() will correctly inherit the old outdoors variable, so that roofing tiles can check it when deciding whether to roof an area

Tested and working

Oh, and a bunch of `to_chat()`s while I was in there

### DIFF
--- a/code/game/objects/items/stacks/tiles/tile_types.dm
+++ b/code/game/objects/items/stacks/tiles/tile_types.dm
@@ -199,3 +199,9 @@
 	throw_range = 20
 	flags = 0
 	no_variants = FALSE
+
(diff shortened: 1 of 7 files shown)


Neerti:
You appear to be missing a ``'`` inside two spans, which Travis is sad about.
There's a fancy new ``span()`` function that adds most of the text for you that might be helpful.

PR #5415, opened 2018-07-16
Fixes Stairs in BYOND 512.1436

Leshana:
Port of VOREStation/VOREStation#4022

In version 512.1436 BYOND no longer calls Uncross reliably on stairs, which breaks them.
To fix this we switch to using CheckExit and Bumped.  This solution is probably the right way to do it anyway, so no harm done even if BYOND makes Uncross work again or that version isn't being used.

### DIFF
--- a/code/modules/multiz/structures.dm
+++ b/code/modules/multiz/structures.dm
@@ -141,20 +141,23 @@
 		if(!istype(above))
 			above.ChangeTurf(/turf/simulated/open)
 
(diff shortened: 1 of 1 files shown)


Serithi:
I WARNED YOU ABOUT STAIRS BRO

etc.

Anewbe:
Does this work on .1414, or will we have to update if this is merged?

PR #5416, opened 2018-07-16
Hopefully fixes seeing and hearing AOOC bug

Anewbe:


code/game/antagonist/antagonist_add.dm
```diff
@@ -36,7 +36,7 @@
 			and it otherwise has no bearing on your round.</span>"
 	player.current.verbs |= /mob/living/proc/write_ambition
 
-	if(can_use_aooc)
+	if(can_hear_aooc)
```
Neerti: Wouldn't ``can_speak_aooc`` be the better thing to check?

PR #5417, opened 2018-07-16
Gender name adjustments

kartagrafi:
for the sake of accuracy and making sense, this changes "gender identity" to "pronouns", and "biological gender" to "biological sex". could also just change it to "sex", but i feel like "biological sex" is clearer for the uninitiated.
i don't know shit so if there's anything broken in here, please lmk, though it's just names so there shouldn't be a problem


elgeonmb:
🏳️‍🌈

PR #5429, opened 2018-07-21
Adds Marking Order Changing at Character Creation

KasparoVy:
QOL for folks with numerous markings on their characters.

Testing/adding/manipulating markings becomes massively time consuming the more you have, especially if you're wanting to put it deep in the layer stack and are testing speculatively.
This introduces much convenience to that process: Now you can click a button and move a chosen marking up or down a layer at will, allowing you to simply create a marking as normal then shift it to the desired layer in the stack without having to re-do a ton of other markings just to get it where you think it might work best.

========================================

Clicking the back `<` button on a marking will trade its place with the _previous_ marking or _move it to the end after_ all other markings if it's the _first_ marking.

Clicking the forward `>` button on a marking will trade its place with the _next_ marking or _put it in first before_ all other markings if it's the _last_ marking.

I know the pictures contain non-Polaris markings but that's because the Virgo codebase served as my test-bed. Works the exact same here

**_What the buttons actually look like after being adjusted. Below simply demonstrates the functionality_**
virgo marking mover proper up and down arrows

_No buttons unless you've got more than one marking._
virgo marking mover one marking

_Moving markings_
marking mover 2

### DIFF
--- a/code/modules/client/preference_setup/general/03_body.dm
+++ b/code/modules/client/preference_setup/general/03_body.dm
@@ -303,7 +303,7 @@ var/global/list/valid_bloodtypes = list("A+", "A-", "B+", "B-", "AB+", "AB-", "O
 
 	. += "<br><a href='?src=\ref[src];marking_style=1'>Body Markings +</a><br>"
 	for(var/M in pref.body_markings)
-		. += "[M] <a href='?src=\ref[src];marking_up=[M]'><</a> <a href='?src=\ref[src];marking_down=[M]'>></a> <a href='?src=\ref[src];marking_remove=[M]'>-</a> <a href='?src=\ref[src];marking_color=[M]'>Color</a>"
+		. += "[M] [pref.body_markings.len > 1 ? "<a href='?src=\ref[src];marking_up=[M]'><</a> <a href='?src=\ref[src];marking_down=[M]'>></a> " : ""]<a href='?src=\ref[src];marking_remove=[M]'>-</a> <a href='?src=\ref[src];marking_color=[M]'>Color</a>"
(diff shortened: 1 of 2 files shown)


Atermonera:
Would it be possible to use `^/v` instead of `</>`, or some more symmetric pair of characters. The left and right indicators aren't explicit in function when the list is sorted vertically (i.e., you have to use them at least once to determine the actual direction the buttons move the element in).

Does it display if there's only one marking pattern? Minor QoL thing, mostly just curious, esp. because of the aforementioned lack of clarity on function. If you just add one marking and there's these buttons that don't visibly do anything, then you may conclude that they do not, in fact, do anything.

Oh, and we don't have that changelog functionality, you'll have to make an actual changelog .yml file in `/html/changelogs`

KasparoVy:
I tried `^/v` first but it looked even jankier because the carot `^` was notably smaller. I figured `</>` would make enough sense and since they had symmetry it'd be fair enough.

I'll try to whip up a preview image while I do the changelog.

KasparoVy:
We're restricted to ASCII characters only, right?

Comparison:
virgo marking mover leftright
virgo marking mover updown
virgo marking mover ud

Atermonera:
Ok, long shot but try inserting `&#708;` into the spot for the upwards-arrow character. NanoUI supports unicode, I've used it for the weather app for communicators, but I'm not sure a byond html window will. If it works, `&#709;` is the matching downwards arrow.

KasparoVy:
Excellent, here goes.

KasparoVy:
Excellent, here goes. I don't have my hopes up, though.

Edit: Aha! Worked like a charm. Thanks @Atermonera 
virgo marking mover proper up and down arrows

PR #5443, opened 2018-07-26
Two pAI fixes

PrismaticGynoid:
Fixes 6 pAI faces - PAIs are no longer coded to only recognize 9 emotions. This enables the use of the neutral, silly, nose, smirk, exclamation points, and question mark faces as a pAI. Before, those buttons did nothing.

Unfolded pAIs can be picked up - Because who doesn't want a robotic cat on their head? Adds in the missing code to allow pAIs to actually be picked up, and held sprites for the newer 4 chassis options.


Anewbe:
Tested?

PrismaticGynoid:
Yes, I wouldn't put up a PR without taking time to test it first.

PR #5448, opened 2018-07-27
Fixes #5447 (Deadspy submap not overwriting mapgen)

### LINKED ISSUE #5447: Deadspy PoI spawns in the rocks
The turfs in the PoI submap aren't actually set, so it doesn't override whatever the terrain generation makes

A cavern containing the contents of the PoI

#### What actually happened
This
`Request for Help: ____/(_____)(?)(PP)(VV)(SM)(JMP)(CA)(TAKE) there's a turret INSIDE of a rock that's shooting at me `

#### Steps to reproduce
Make a mountains submap that looks something like this
Spawn it in the mountains.
Observe the buried nature of various contents

#### Additional info:
Protip: When making a PoI, set the turf.

drexample:
Fixes #5447 
Adds walls and floor to the submap


Atermonera:
Suspect this isn't based on your master branch, or if it is, then you need to do a hard reset to upstream master to clear those old commits from the history.

PR #5452, opened 2018-07-28
Adds a hairstyle and alcohol tolerance trait

kartagrafi:
adds my oc's very specific hairstyle and adds the alcohol tolerance trait, where it can be easier or harder to get drunk depending on your tolerance
feedback/fixes appreciated!
also i can't be arsed to use goddamn COMMAND LINE to change commit names so would anyone else mind doing that for me thanks


code/modules/mob/_modifiers/modifiers.dm
```diff
@@ -45,6 +45,7 @@
 	var/icon_scale_percent				// Makes the holder's icon get scaled up or down.
 	var/attack_speed_percent			// Makes the holder's 'attack speed' (click delay) shorter or longer.
 	var/pain_immunity					// Makes the holder not care about pain while this is on. Only really useful to human mobs.
+	var/alcohol_effect				// Affects strength of alcohol.
```
PsiOmegaDelta: As it's currently used, because there is no check for 0, you probably want the default value to be 1.
Otherwise, having any modifier besides alcohol tolerance will make you immune.

kartagrafi: okay, considering i'm entirely new to coding and have no clue what i'm doing and did this with heavy help from neerti: setting the default would involve adding "= 1" next to that, yes?

PR #5459, opened 2018-07-31
Makes Storms Thunderous

Neerti:
The storm weather type will now periodically cause lightning strikes on a random outdoor turf.
This causes a big bolt to come down from the heavens and do various mean things to whatever is nearby. It produces a soft flash effect for all outdoor tiles, and a thunder sound is played to people on the planet.
Getting struck directly shouldn't kill you instantly (by itself) but you will want to go to medical.
Being sufficiently indoors will protect you from strikes, or having a grounding rod nearby.
Grounding rods and tesla coils attract lightning. They both redirect the strike to their tile, but do different things. The grounding rod stops any bad effects from happening, while the coils cause the bolt to bounce and get absorbed into their powernet, as if it was a tesla ball hitting them. The tesla coil has higher priority.
Simple animals that get struck get hurt much more, for that RNGesus saves feeling when it strikes the combat drone thats about to kill you (if that ever happens).
Admins have a fun button to do a cosmetic or real lightning strike, anywhere.
Adds ten thunder sounds, from various files on freesound.org with C0 (public domain) licenses.

Video: https://streamable.com/se0hk


MisterLayne:
What about the lightning being more likely to happen on turf with a tree, and then destroying the tree? Isn't that how lightning works-- it likes things that are closest to the sky, easiest path, something something I didn't pay attention in science class.

Atermonera:
It's specifically the path of least electrical resistance. If a tree happens to be more electrically conductive than air, then it's more likely to be struck.

I don't know how electrically conductive trees are, but since this is a spacefuture setting on an alien world we can fluff whether the local trees are or aren't conductive enough to be lightningrods in their own right

PR #5461, opened 2018-08-01
Refactors supply controller

Atermonera:
Supply consoles now run nanoUI
[Fully](https://puu.sh/B6vqz/193869bc42.mp4) [tested](https://puu.sh/B6vxG/eb1c52a99e.mp4)
Export reports:
example

I may have stayed up until 6 in the morning this morning getting the shuttle part to work. And then I went to bed because I realized it was past first light


Neerti:
I don't see anything wrong here.

PR #5471, opened 2018-08-04
Slightly cleans up hooded suit code

Anewbe:
- All of the hoods are now in a file specifically for hoods. The base hood item is defined, because I couldn't find it anywhere and they all do basically the same stuff.
- Hooded suits are now in a file for hooded suits. Their base code is no longer in `toggles.dm`.
- Hooded suits have had a var renamed and a var removed. `suittoggled` is now the more sensible `hood_up`, which should be using TRUE/FALSE now, and `hooded` has been removed because it's never actually checked anywhere.
- Nuked some old commented out code for pre-accessory sweatervests.


code/modules/clothing/suits/hooded.dm
```diff
@@ -0,0 +1,182 @@
...
+	MakeHood()
+	..()
+
+/obj/item/clothing/suit/storage/hooded/Destroy()
+	qdel(hood)
+	return ..()
+
+/obj/item/clothing/suit/storage/hooded/proc/MakeHood()
+	if(!hood)
+		var/obj/item/clothing/head/hood/winter/W = new hoodtype(src)
```
Atermonera: Not sure the `/winter/` is strictly necessary here. Not sure it'll do anything harmful either.

PR #5476, opened 2018-08-05
Ambience Refactor

Neerti:
Changes how ambience is assigned to areas in code. Instead of having to type a bunch of file paths, now you can just type something like ``ambience = AMBIENCE_HIGHSEC``, just like in the TG codebase.
Separates the ambience into a large number of themes.
Ports a lot of new ambience files from TG.
Removes some unused/defunct/horrible ambience that we don't need anymore.
Moves most ambience files into separate folders for easy of organization.
Applies new ambience files to various PoIs, and station areas such as the Armory.
Most importantly, no more god damn ear-bleeding high-pitched droning ambience.


Atermonera:
Changelog so people know that it's maybe not so earsplitting anymore

PR #5479, opened 2018-08-05
I don't know how we keep breaking falling

Anewbe:
- planetary_fall teleporters now try to find their target planet the first time they're used.

This isn't at all the prettiest code, but I don't have any idea how it worked before it broke, or even when it broke.


Nalarac:
Obviously the game wishes to be spess and one can not fall in spess 👍

PR #5481, opened 2018-08-06
Adds new job/department specific teshari clothing to the loadout

kcin2001:
since you guys seemed to like my last teshari clothing pr so much ill just drop it here first instead, images of the new clothes are attatched 
deptjacket
deptcloak
iaa
iaaa


code/modules/clothing/under/xenos/seromi.dm
```diff
@@ -34,134 +34,252 @@
 	icon_state = "seromi_rainbow"
 
 /obj/item/clothing/under/seromi/undercoat
+	name = "you shouldnt be seeing this"
 	desc =  "A Teshari traditional garb, with a modern twist! Made of micro and nanofibres to make it light and billowy, perfect for going fast and stylishly!"
```
Atermonera: Typically for prototypes that shouldn't be available in game, the name is set to something generic, and the description reflects the fact that it shouldn't be available. So we don't get an ahelp about something along the lines of
> This is X
> They are wearing a you shouldnt be seeing this

or
> What in the hells is a 'you shouldnt be seeing this'?

Atermonera:
Sorry for delay, have been travelling.
I feel I should note that the addition of uniforms for headstaff roles does not indicate the sudden allowance of teshari to play headstaff roles beyond what is currently allowed. They are being added for completeness.
If at some point teshari are permitted to play more headstaff roles, you can expect an official announcement from their lore maintainer.

code/modules/clothing/under/xenos/seromi.dm
kcin2001: ok i can change that, i mostly changed that for myself when i noticed an odd "under" option in the selection and i recompiled to see that it was the base item

kcin2001:
i mainly made it for a server downstream but i thought id save you guys the trouble of grabbing it and then us needing to remove the downstream specific copy

Atermonera:
Yeah, I figured that was the case, and do appreciate the effort on your part. Just wanted to make it very clear to our players that keep up with the github.

PR #5508, opened 2018-08-19
Fixes the bug(s) that allow meat bodies to have metal brains

Anewbe:
This is in fact a lore-breaking bug, forgot to get around to patching it until recently.

If there is _great demand_ for some sort of brain-implant thing, for character reasons, that can be added separately from this.


SpadesNeil:
https://www.youtube.com/watch?v=mh5NSmVs_uk

PR #5511, opened 2018-08-20
Ports character physical descriptors from Bay

Anewbe:
- Adds a height and weight selector to character setup. This is a relative value based on species, rather than any sort of actual measurement system, which means your options are roughly "Taller or Shorter than Average", and "Heavier or Thinner than Average".
- Said selectors show a comparative message to people who examine you, so if someone shorter than you is examining you, it tells them that you're tall. This is modified by species, so, for now, Unathi are taller and heavier than a human with the same setting.
- **Disables the Size traits.** This is intended as a replacement for said traits. The size-trait system isn't tied to any sort of metric, so someone at 6' with the Larger trait looks taller than someone at 6'2" with the Large trait, which is confusing. Additionally, the scaling works poorly on human mobs, for whatever reason, so those aforementioned traited people look like a full-body sensor blur.


Neerti:
I'm very much not in favor of this if its removing the resizing ability.

Woodratt:
Probably needs the staff input tag

Atermonera:
Looks like it just disables those traits, the underlying system to scale sprites is untouched

Anewbe:
Correct, my wording was inaccurate. Editing.

Belsima:
Character physical descriptors? Like... flavor text?

Atermonera:
It will, but it will also immediately provide more useful information than is provided in flavor text. Flavor text never gives values relative to you, the examiner, but rather in an absolute metric or relative to averages (About which you have to be mindful of your own character's values in order to actually compare).
This system will automatically tell you, "X is taller and fatter than you", in a much cleaner and more uniform manner than the icon-scaling traits (Which imo were ugly, but to each their own).

Belsima:
Ah, fair enough. I liked the scaling traits as a way to give information based on a glance, though would have preferred if it was just 'large' and 'small' given the common discrepancy pointed out in the OP.

Naidh:
This seems like it would work better if it was used to augment the size traits much rather than remove the ability to use them which would be as easy as suggesting what heights should take the trait, using average heights of males or females as a baseline. The icon heights are a little strange, but they do manage to help do the same job by serving as immediate visual cues, but that's just my two cents.

elgeonmb:
size traits ugly as heck

this good decision

PR #5515, opened 2018-08-21
Fixes a UI bug with emptying currently-open bags into smartfridges...hopefully.

Shophaune:
All seems to work on my end, but then I'm no expert - would love to know if there was a better way to fix this. Bug was that if you empty a plant bag with plants in it into a smartfridge whilst looking in the bag (i.e. you can see the plants) the icons of the plants remain frozen above your hands even when the bag is closed.


Mechoid:
Only other way would be force-closing the bag when you go to stock, this is as good a fix as that though.

PR #5516, opened 2018-08-22
Adds a new subset of grenades that shoot projectiles

Anewbe:
Pathed as `grenade/shooter` because we've got ten billion things with projectile in their path.

Basically, does hacky, terrible code to steal how frag grenades work, and replaces horrible pellets of death with any sort of projectile, including different horrible pellets of death. Yes, I did in fact try to make the fragmentate proc and assorted pellet projectiles work more generally, but that was a lot of math and I'm not confident I could have gotten it right.

Laser grenade description is actually what I want it to be (because why not?), until some madperson decides to make these obtainable.


code/game/objects/items/weapons/grenades/projectile.dm
```diff
@@ -0,0 +1,75 @@
...
+		return
+
+	src.launch_many_projectiles(O, spread_range, projectile_types)
+
+	qdel(src)
+
+
+
+/obj/item/weapon/grenade/shooter/rubber
+	name = "rubber pellet grenade"
```
Belsima: Just as a note, I usually hear these referred to as 'sting grenades', though that might be based on vernacular and locality.

PR #5522, opened 2018-08-26
Implements commcards (Communicator cartridges)

Atermonera:
STAFF INPUT: Do I implement the MULEbot control? They're rarely used, and on the occasions that they are, they frequently break.

Everything is tested, but the actual testing is over a rather long period of time and on multiple computers, so I can't easily locate and post links to it. It has all been documented on discord as I've worked on this, so people should generally be aware of this project.

Commcard features generally mirror PDA functions, but are more efficient and improved where feasible. 

I haven't done the antag uplink, or the machinery side of this project for message spoofing, etc, in part because I'm less certain on the extent of that, and because this is still a pretty sizeable chunk in itself.
TODO:
- [x] Make it comply with GLOB


Atermonera:
Well, rebase appears to have worked

PR #5523, opened 2018-08-27
Still more simple_mob conversions

Anewbe:


Atermonera:
vault4.dm, one instance each of `/mob/living/simple_animal/hostile/alien/drone` and `/mob/living/simple_animal/hostile/alien`

PR #5526, opened 2018-08-28
Catching is no longer guaranteed, accuracy code is more general

Anewbe:
- Catch chance now starts at 90%. This is decreased by the same ailments that affect gun and melee accuracy, ie Eye conditions.
- Sharp objects are harder to catch.
- Accuracy code from Eye conditions (blind, blurry, confused) is now generalized, rather than being copied in four places.
- You can no longer catch things using hands that do not exist.


mistyLuminescence:
make catching guaranteed for beach balls and basketballs tbh

PR #5527, opened 2018-08-28
Adds a Neural implant for future implementation.

Mechoid:
Neural implant to fit the 'Brain Implant' role.

- Changes the icon of the affected brain to an assisted version.
- Chance to damage the brain when EMPed.
- Failure of the implant, or removal will result in severe damage and side-effects.
- Robotic brains (Posi, Drone) will only get warnings about worse damage from the implant, aside from standard damage signs.

Promethean brains have been updated to fit the icon scheme.


Nalarac:
Is this implant removed using the hemostat extracting surgery? If so, it could be an issue if someone who has this also has, for example, shrapnel in their skull.

Mechoid:
..I think that may be more an issue with 'There's no way to determine if this is shrapnel or an implant before removing it', because this then happens with chem / uplink / tracking / death implants as well.

code/game/objects/items/weapons/implants/neuralbasic.dm
```diff
@@ -0,0 +1,106 @@
...
+	desc = "A small metal casing with numerous wires stemming off of it."
+	var/obj/item/organ/internal/brain/my_brain = null
+	var/target_state = null
+	var/robotic_brain = FALSE
+
+/obj/item/weapon/implant/neural/post_implant(var/mob/source)
+	if(ishuman(source))
+		var/mob/living/carbon/human/H = source
+		if(H.species.has_organ["brain"])
+			var/obj/item/organ/internal/brain/possible_brain = H.internal_organs_by_name["brain"]
```
Anewbe: O_BRAIN

mistyLuminescence:
@Mechoid maybe give the surgeon a list of embedded items, and allow them to choose which ones to remove? so if you're operating on someone and you don't want to remove the implant, you just.. don't (or if you're an antag surgeon and you're into that kinda thing, you do)

PR #5541, opened 2018-08-31
Submaps can be rotated

Atermonera:
Default direction is SOUTH, they can be rotated to any cardinal direction.
Submaps loaded at map generation will be randomly rotated, for more variety. Experience PoIs in orientations you've never seen before*!

* Unless you're a maniac that actually rotates the view.

Tested. Took a while to figure out how information went through the map loader but I got there eventually. And unlike the stuff posted to the server and discord, the maps don't overlap!

### DIFF
--- a/code/modules/maps/tg/reader.dm
+++ b/code/modules/maps/tg/reader.dm
@@ -39,6 +39,7 @@ var/global/use_preloader = FALSE
 	//How I wish for RAII
 	if(!measureOnly)
 		Master.StartLoadingMap()
+		world << "/dmm_suite/load_map()\t\tPlacing at ([x_offset], [y_offset], [z_offset])"
 	space_key = null
 	#ifdef TESTING
 	turfsSkipped = 0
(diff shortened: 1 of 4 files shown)


code/modules/maps/tg/reader.dm
```diff
@@ -132,27 +136,83 @@ var/global/use_preloader = FALSE
...
+
+					if(xcrd >= 1)
+						var/model_key = copytext(line, tpos, tpos + key_len)
+						line_keys[++line_keys.len] = model_key
+						#ifdef TESTING
+						else
+							++turfsSkipped
+						#endif
+						CHECK_TICK
+					maxx = max(maxx, ++xcrd)
```
Arokha: This line makes all loaded maps think they are 1 wider than they actually are and causes loading new z-levels from full width/height dmms to fail, because it believes they are 1 wider than maxx but they are not.

It should be xcrd++ not ++xcrd

PR #5546, opened 2018-09-02
Adds new set of cyborg sprites

Cerebulon:
For medical, science, default, security and combat modules.
Since we may or may not be losing one sprite option. May need to wait for that to be resolved and do a dmi merge.

Sprites:
https://i.gyazo.com/5d7da775141dcb0f03feac667d5f2bf9.png
(Combat module eyes not showing since it was spawned weirdly, but would be the same orange as security's eyes)

Mostly based on an obscure robot from Timesplitters:
https://vignette.wikia.nocookie.net/timesplitters/images/3/35/Inseticksd12.gif/revision/latest?cb=20110509223311


Atermonera:
Style reminds me of Risk of Rain

PR #5548, opened 2018-09-05
Ports /vg/ instrument frame work, adds client based sound pref toggle.

Poojawa:
Also shifts sound files which is 98% of the bulk here.

🆑 Poojawa
add: Ported /vg/ instrument code, improved the UI of instruments.
add: Added a client side pref that mutes instruments being played for you, which was a frequent request downstream.
/🆑 

This also allows the porting of tg or vg instruments, should the want be there. All that would be required is to add the object and the sounds into `sounds/instrument/[instrument]` and it should work right on the get-go.

<3


code/game/objects/items/devices/violin.dm
```diff
@@ -1,392 +1,41 @@
 //copy pasta of the space piano, don't hurt me -Pete
-
-/obj/item/device/violin
-	name = "space violin"
-	desc = "A wooden musical instrument with four strings and a bow. \"The devil went down to space, he was looking for an assistant to grief.\""
+/obj/item/device/instrument
```
Poojawa: I hestiated to just rename this to instrument.dm, but should any hand-held instruments be added, it may be prudent to rename it then.

Poojawa:
All of the travis failures aren't actually my fault, near as I can tell. it compiles cleanly and produces no runtimes in game.

mafana:
Is there be a mechanic for... muting them on my end? I find people are obnoxious with instruments in SS13 a lot.

mafana:
My last comment was made despite the answer very clearly being in the changelog, somehow I missed it despite it being half the text. Disregard.

Poojawa:
https://github.com/tgstation/tgstation-github-automation-tools just adapt as needed <3

but uh, I guess I'll do a changelog file.

Anewbe:
Do you happen to know what license this is under?

Anewbe:
I cannot for the life of me find where VG's license information is.

Poojawa:
https://github.com/vgstation-coders/vgstation13/blob/Bleeding-Edge/LICENSE-GPL3.txt

Poojawa:
also just in case you needed tg's as well, since it was more or less copied from tg who ported it from vg. 

https://github.com/tgstation/tgstation/blob/master/LICENSE

Atermonera:
/vg/ is under GPL, /tg/ is under AGPL unless the code was committed before Dec 31, 2014, 16:38 PST, in which case it is also GPL.

Anewbe:
Seems fine, merging.

PR #5549, opened 2018-09-05
Yet Another Circuit Update

Neerti:
Ports new assemblies from TG. Most of these are cosmetic.
Wall-mounted assemblies can be attached to walls by clicking on the wall with it. Wrenching will unattach from the wall.
Adds circuit clothing. You can now wear an assembly in most slots.
Adds empty circuit clothing to the loadout, since they look fancy. The chestpiece is not available in the loadout.
Ports ability to color assemblies with a coloring tool from TG.
TG lacked an option for hot pink. I have fixed that here.
Ports some of TG's circuit controller things, mainly the printer improvements.
Ports color input pad from TG.
Reorganizes sprite file locations for circuits.
Adds power draw/surplus information to assembly UI.
Ports rgb2hex and hsv2hex from TG.
Ports debug version of circuit printer, with all upgrades applied and infinite material.
Fixes bug that broke default pin values from being set on initialization.
Adds holographic projector circuit, which can copy any movable atom it can see, and make a holographic copy of it.
Adds powernet interface circuit, allows the assembly to draw or give power to the powernet, if anchored over a knotted wire.


code/modules/integrated_electronics/passive/power.dm
```diff
@@ -156,11 +156,88 @@
...
+	if(assembly && assembly.anchored && assembly.battery)
+		var/should_act = get_pin_data(IC_INPUT, 1) // Even if this is false, we still need to update the output pins with powernet information.
+		var/drawing = get_pin_data(IC_INPUT, 2)
+
+		if(should_act) // We're gonna give or take from the net.
+			if(drawing)
+				var/to_transfer = min(throughput, assembly.battery.amount_missing() / CELLRATE) // So we don't need to draw 10kW if the cell needs much less.
+				world << "Transferring [to_transfer]."
+				var/amount = IO.draw_power(to_transfer)
+				world << "Giving [amount] to cell."
```
Atermonera: Debug code?

PR #5553, opened 2018-09-09
Vehicle Expansion 1: Zoom Harder

Mechoid:
FRESH FROM VOXKEA, THE BRAND NEW NOT-STOLEN HOVERBIKE! AND ATV!

https://puu.sh/Br4b5/be163c257e.gif

Refactor cargo trains, they're just normal trains now. 

Adds the Quad, a type of 'train', and its trailer.
- Launches people who get hit by or run over by it, sending them flying to an adjacent tile with a very short stun and low damage.
- Slower in snow, dreadful in water, okay
Re-enables the Spacebike, and fixes it. Can now actually go over normal turf, just at a brisk running speed.
- Actually uses a cell now.
- Faster over water.

Both of these are orderable from cargo for rather high point costs, in addition to being buildable in Robotics from the mecha fabricator.

..You can, in fact, hold them in your hands, but they have slowdown 10, so I don't know why you would.


Atermonera:
Inb4 bodybuilder teshari deadlifting ATVs

lolman360:
When are we giving Security segways so they can be the true space mall cops we all know they are?

PR #5582, opened 2018-09-19
Boot knives fit in boots

Anewbe:


Nalarac:
Seems like the quarantine shuttle PoI has a _/obj/item/weapon/material/knife/boot_ which you changed to be a child of tacknife and that's what is causing the error.

Anewbe:
I know, I just didn't want to fix it at the time, will probably be done within a day.

PR #5575, opened 2018-09-18
Rig and Spacesuit additions

Woodratt:
- Added the 'military' Rigsuit from Bay.
- Added 'pmc' rigsuits
- Added exploration and pilot voidsuits along with alternate sprites (alternate sprites done by Naidh)
- Addition of suit cyclers for exploration and pilot voidsuits

None of these suits are currently accessible in game outside of admin bus. Currently only the pilot voidsuits have other species sprites. Exploration suits are missing sprites for Teshari (Naidh made some for their alternates but I have to add them) and the rig suits are human only. This is to be fixed in the near future, just wanted to get the make workload actually in the game first.

PR #5583, opened 2018-09-20
Adds tails to Unathi rigsuits (again)

Cerebulon:
Fixes #5551 
(Was unintentionally reverted by #5575 )


Woodratt:
Thought I had incorporated that.

Woops...

PR #5589, opened 2018-09-22
Adds the Biopsy Scanner to the Surgery Kit

Wickedtemp:
Since it's actually used in a surgical operation, it should proooobably be in the kit. Iunno if the kit is really used here aside from antag shuttles, but they might as well have it.


Wickedtemp:
aaaaaAAAAAA**AAAAAAA**

Will poke tomorrow.

Mechoid:
..I knew I was forgetting something.

PR #5590, opened 2018-09-22
Make the Statue 'better'.

Mechoid:
The statue is no longer entirely immortal, and actually has a piece of nice loot for dying, that isn't just its normal marble.

- Health dropped from 50,000 to 3,000 because GOD LORD MAN, it took like 15-20 minutes with an emitter pummeling it to get it to die.
- Non-rubber bullets should be able to damage the statue, as should higher power lasers.
- Drills, plasma cutters, and high-brute melee weapons can now damage the statue.
- Summoned player-statues are now assigned their creator, so they can move in range of them.
- The statue dies properly.

### DIFF
--- a/code/modules/mob/living/simple_animal/aliens/statue.dm
+++ b/code/modules/mob/living/simple_animal/aliens/statue.dm
@@ -22,8 +22,8 @@
 	environment_smash = 2
 	can_be_antagged = 1
 	speed = -1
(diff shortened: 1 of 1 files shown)


Anewbe:
So here's a weird idea for this abomination of code: Let's skip the "Nothing below 20 damage can hurt it" and just give it armor and soak?

PR #5602, opened 2018-09-24
Creates the is_knife proc

Anewbe:
Okay, time for more bloat.
- Objects now have an `is_knife()` proc, much like the `is_[tool]()` procs added semi-recently.
- Hatchets are no longer machetes, nor knives.
- The Unathi Dueling Knife is no longer a hatchet, nor a machete, nor a knife, nor even a material weapon. I was going to make them wooden knifes, but then they did some weird crap with breaking, which may or may not be bugged. I can swap them back to actual knives if need be.
- The umbrella is no longer a melee weapon.
- Updates the pathing on PoIs, the Southern Cross, and the Northern Star.


code/game/objects/items/weapons/umbrella.dm
```diff
@@ -0,0 +1,36 @@
...
+	icon_state = "umbrella_closed"
+	addblends = "umbrella_closed_a"
+	flags = CONDUCT
+	slot_flags = SLOT_BELT
+	force = 5
+	throwforce = 5
+	w_class = ITEMSIZE_NORMAL
+	var/open = FALSE
+
+/obj/item/weapon/melee/umbrella/New()
```
Atermonera: Missed a `/melee`

Anewbe: Righto.

Anewbe:
Marking this DNM until the map tweaks are done.

PR #5613, opened 2018-09-28
Custom RIG Framework

Mechoid:
I'm starting this off by saying good lord rigs are weird. Okay that's all.
- Adds a Rigsuit customization kit, which uses the same parameters as the suit kit which it is based on. This kit does NOT mask the type of RIG it is, by leaving the components with their original name + 'customized'.
- Adds four custom component DMI files because rigs are annoying and use the same base state name for every piece by core design, and this isn't meant to be hyper intrusive.

https://github.com/PolarisSS13/Custom-Items/pull/119 should be merged after this.


Anewbe:
Have you tested that this works properly? I'm not seeing where things get pointed to the right files.

code/game/objects/items/paintkit.dm
```diff
@@ -149,6 +149,52 @@
 		return
 	return ..()
 
+/obj/item/device/kit/suit/rig
+	name = "rig modification kit"
+	desc = "A kit for modifying a rigsuit."
+	uses = 1
+
+/obj/item/device/kit/suit/rig/customize(var/obj/item/I, var/mob/user)
```
Mechoid: Pointers set here in customize. Only the RIG uses the kit's icon, because they're horrifying and use all the same icon_state. Changing this would mean touching RIGs way more than I'd like, because it's already a streamlined for loop to assign everything the icons/states in multiple places and on creation.

Mechoid:
Tested as much as I physically can with my local stuff.

PR #5624, opened 2018-10-04
Moved all cooking related files into code/module/food

lbnesquik:
Makes for easier searching and everything.


Anewbe:
I would much rather that the file `Chemistry-Reagents-Food_Drinks.dm` stays with the other chemistry code, because the rest of the food code is just a delivery system for said reagents.

lbnesquik:
I dunno, it feels more organised this way given you rarely are looking for beverages at the same time you are coding for medicines. And at the inverse is true. If you insist i'll move it.

Anewbe:
Please do so.

PR #5626, opened 2018-10-05
Move almost everything food related into the kitchen module.

lbnesquik:
Not moving reagents and tools.


lbnesquik:
Oh uh, those two things keep popping up in all of my commits, I usually recall not to push or commit them

lbnesquik:
Resolved @Anewbe

PR #5635, opened 2018-10-05
Fix yes emote not working correctly.

lbnesquik:
It's a lazy fix but it works.


Anewbe:
So if I'm trying to do the `yes` emote, do I need to do `*yes` or `*yess`?

PR #5636, opened 2018-10-05
[WIP] Ports /tg/station pixel projectiles, processing subsystems, timer subsystems, and some misc stuff to make it all work

kevinz000:
What it says on the tin.
For those of you unfamilier: Pixel projectiles on /tg/ gives projectiles true angular motion, and while you may say that's already a thing with the current system it doesn't take into account where you clicked on a tile, and only takes into account which tile you clicked so I'm porting the tg one. Also, the new projectiles system is processed on a projectile subsystem, and is 100% accurate and more realistic! Yeah.
Anyways does this place have an auto changelogger or do I have to manually put changelog files? :/


Anewbe:
We don't have an automatic changelog thing, sadly.

kevinz000:
ouch

kevinz000:
relies on stuff from 5677

Atermonera:
> code/modules/projectiles/projectile/magnetic.dm:130:error: kill_count: undefined var
> code/modules/projectiles/projectile/hook.dm:12:error: step_delay: undefined var
> code/modules/projectiles/projectile/hook.dm:31:error: kill_count: undefined var
> code/modules/projectiles/projectile/hook.dm:29:error: launch: undefined proc

kevinz000:
Needs to be reverified after merge.

kevinz000:
fixed fixed bugtested

maps/example/example-2.dmm
```diff
@@ -1,69 +1,1324 @@
...
-"B" = (/obj/structure/cable/yellow{d1 = 4; d2 = 8; icon_state = "4-8"},/turf/simulated/floor/tiled/dark,/area/aisat_interior)
-"C" = (/obj/machinery/power/apc{dir = 4; name = "east bump"; pixel_x = 24},/obj/structure/cable/yellow{d2 = 8; icon_state = "0-8"},/obj/structure/cable/yellow{d1 = 2; d2 = 8; icon_state = "2-8"},/obj/machinery/light{icon_state = "tube1"; dir = 4},/turf/simulated/floor/tiled/dark,/area/aisat_interior)
-"D" = (/obj/structure/cable/yellow{d1 = 1; d2 = 2; icon_state = "1-2"},/obj/effect/floor_decal/corner/lime/diagonal,/turf/simulated/floor/tiled/dark,/area/aisat_interior)
-"E" = (/turf/simulated/floor/tiled/red,/area/aisat_interior)
-"F" = (/obj/effect/floor_decal/sign/two,/turf/simulated/floor/tiled/dark,/area/aisat_interior)
-"G" = (/obj/machinery/light,/obj/effect/floor_decal/corner/lime/diagonal,/turf/simulated/floor/tiled/dark,/area/aisat_interior)
-"H" = (/obj/structure/lattice,/obj/structure/cable/yellow{d1 = 32; icon_state = "32-1"},/turf/simulated/open,/area/aisat_interior)
+//MAP CONVERTED BY dmm2tgm.py THIS HEADER COMMENT PREVENTS RECONVERSION, DO NOT REMOVE
```
Neerti: This might cause problems if we don't run the tgm stuff currently.

code/game/atoms_movable.dm
```diff
@@ -45,51 +47,234 @@
...
+
+	for(var/i in loc)
+		if(i == src) // Multi tile objects
+			continue
+		var/atom/movable/thing = i
+		thing.Crossed(src)
+//
+////////////////////////////////////////
+
+/atom/movable/Move(atom/newloc, direct)
```
Neerti: Is this supposed to be overridden twice?

Neerti:
Looks neat, did some local testing and some bugs came up. Gonna post them here from discord so you don't need to scroll up a lot if you forget some of them.

[1:11 PM] Neerti: @A Runescriber Testing the PR locally, it looks like the projectile is not being turned towards where the projectile is going(edited)
[1:11 PM] Neerti: so far the ion bolt and phase rifle have that issue
[1:12 PM] Neerti: if I shoot to the left the projectile points down as it flies left of me
[1:14 PM] Neerti: Okay so shooting to the right also has the same issue, projectile points south
[1:14 PM] Neerti: shooting south of me makes it point north
[1:14 PM] Neerti: shooting north works fine
[1:18 PM] Neerti: Arcing projectiles don't appear to call on_impact() when they hit the ground.
[1:20 PM] Neerti: Projectiles arcing in the air also are turned incorrectly but that might be the first bug as well
[1:21 PM] Neerti: firing an arcing projectile with a gun also causes a trying to divide null error because target_distance is not set, presumably because its using old_style_target()

Otherwise this seems mostly fine.

PR #5637, opened 2018-10-05
Kates syringe fix

lbnesquik:
This fix syringes being capable of slicing apples (and probably some other errors no one reported). This doesn't affect syringes gun because they use darts for some reason 
~~despite having the same sprite as ones who use standard syringes and no cue in the description~~
https://github.com/VOREStation/VOREStation/pull/4304

### DIFF
--- a/code/modules/projectiles/guns/launcher/syringe_gun.dm
+++ b/code/modules/projectiles/guns/launcher/syringe_gun.dm
@@ -64,7 +64,7 @@
 
 /obj/item/weapon/gun/launcher/syringe
 	name = "syringe gun"
-	desc = "A spring loaded rifle designed to fit syringes, designed to incapacitate unruly patients from a distance."
+	desc = "A spring loaded rifle designed to fit syringes, designed to incapacitate unruly patients from a distance. Does not fit standard medical grade syringes."
 	icon_state = "syringegun"
 	item_state = "syringegun"
 	w_class = ITEMSIZE_NORMAL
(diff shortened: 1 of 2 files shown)


Mewchild:
Syringes use darts, loaded into cartridges. Make sure you're using the correct syringe gun.

lbnesquik:
So it turns out that syringes TURN sharp whenever they are put into a dart case. So there is no issue. Thanks, bug checking coder from the past!

Mewchild:
With your change makin syringes no longer sharp, check and see if they still cause bleeding/cuts on harm intent, as they should.

lbnesquik:
They probably don't.

lbnesquik:
I could probably just hack it by adding sharp = 1 at the start of the harm process and then sharp =initial(sharp)

code/modules/projectiles/guns/launcher/syringe_gun.dm
```diff
@@ -129,12 +129,14 @@
 		C.loc = src
 		darts += C //add to the end
 		user.visible_message("[user] inserts \a [C] into [src].", "<span class='notice'>You insert \a [C] into [src].</span>")
-	else
+	else if(istype(A, /obj/item/weapon/reagent_containers/syringe))
+		var/obj/item/weapon/reagent_containers/syringe/B = A
+		user << "<span class='warning'>[B] is too long and fragile to work with [src]! You'll need special darts for this device!</span>"
```
Anewbe: `to_chat(user, ` is the preferred way to do this.

lbnesquik:
`else if(W.sharp)` is all that is checked before slicing sliceable.

Neerti:
Could change it to check for both sharp and edge.

PR #5640, opened 2018-10-07
Fixes laptop blocking grown adults.

lbnesquik:
https://github.com/VOREStation/VOREStation/pull/4319


Atermonera:
It would also be preferable imo to have them use their own circuit board, whereby one could _actually_ de-/reconstruct them

PR #5648, opened 2018-10-09
Prevents people from HREF exploiting around the R&D console lock.

king5327:
Fixes: A dirt-easy HREF exploit that lets people who think like me ignore RD console security.
Fixes: << becomes to_chat

### DIFF
--- a/code/modules/research/rdconsole.dm
+++ b/code/modules/research/rdconsole.dm
@@ -155,7 +155,7 @@ won't update every console in existence) but it's more of a hassle to do. Also, 
 	add_fingerprint(usr)
 
 	usr.set_machine(src)
-	if((href_list["lock"] || screen < 1 || (screen == 1.6 && href_list["menu"] != "1.0")) && !allowed(usr)) //Stops people from HREF exploiting out of the lock screen, but allow it if they have the access.
+	if((screen < 1 || (screen == 1.6 && href_list["menu"] != "1.0")) && (!allowed(usr) && !emagged)) //Stops people from HREF exploiting out of the lock screen, but allow it if they have the access.
 		usr << "Unauthorized Access"
 		return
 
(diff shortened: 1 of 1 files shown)


code/modules/research/rdconsole.dm
```diff
@@ -155,6 +155,10 @@ won't update every console in existence) but it's more of a hassle to do. Also,
 	add_fingerprint(usr)
 
 	usr.set_machine(src)
+	if((screen < 1 || (screen == 1.6 && href_list["menu"] != "1.0")) && (!allowed(usr) && !emagged)) //Stops people from HREF exploiting out of the lock screen, but allow it if they have the access.
+		usr << "Unauthorized Access"
```
Atermonera: `to_chat(usr, "<span class='warning'>Unauthorized Access</span>")`

king5327: I know. I chose to follow internal file consistency in this case. Fixing that would be a topic for another PR.

king5327:
Another option would be to change the access from access_research to access_tox so that only people with lab access can access it.

code/modules/research/rdconsole.dm
Atermonera: `<<` is technically non-compliant with our coding standards, see "Outputting Messages" section of our coding standards. It's not incredibly urgent that all instances be updated, but it is requested that any new instances use `to_chat()`, and that instances of `<<` be updated to `to_chat()` when related code is touched.

Atermonera: In case I wasn't clear, this is a formal request that you comply with our coding standards. I would sooner see usage of `<<` updated in the whole file in this PR if you're concerned about consistency of a single file (And I do as much in pretty much any project of mine), than the addition of another `<<` in the name of file consistency, when we're slowing working on updating all of '<<' to 'to_chat()' on an as-we-touch-each-file-tangentially basis, particularly when consistency within one file means very little against the broader inconsistency of the codebase at large.

PR #5650, opened 2018-10-09
Fix incorrect HUD/screen/hologram

lbnesquik:
Adds a missing pointer, giving proper screens to the machines.


lbnesquik:
https://github.com/VOREStation/VOREStation/pull/4312

PR #5654, opened 2018-10-12
Makes math helpers defines for performance

kevinz000:
What it says on the tin!


code/__defines/math.dm
```diff
@@ -1,18 +1,228 @@
...
+// Secant
+#define SEC(x) (1 / cos(x))
+
+// Cosecant
+#define CSC(x) (1 / sin(x))
+
+#define ATAN2(x, y) ( !(x) && !(y) ? 0 : (y) >= 0 ? arccos((x) / sqrt((x)*(x) + (y)*(y))) : -arccos((x) / sqrt((x)*(x) + (y)*(y))) )
+
+// Greatest Common Divisor - Euclid's algorithm
+/proc/GCD(a, b)
```
Atermonera: Why are you moving helper functions out of a file for helper functions and into a file for defines?

code/modules/mob/living/carbon/metroid/powers.dm
```diff
@@ -1,4 +1,4 @@
-/mob/living/carbon/slime/proc/Wrap(var/mob/living/M) // This is a proc for the clicks
+/mob/living/carbon/slime/proc/WRAP(var/mob/living/M) // This is a proc for the clicks
```
Atermonera: This isn't a define, so it shouldn't be all caps

kevinz000: Oops.

code/__defines/math.dm
kevinz000: I consolidated all math based stuff into a single math file, could put it back in a math.dm under helpers I guess?

code/__defines/math.dm
```diff
@@ -1,18 +1,228 @@
...
+// Cotangent
+#define COT(x) (1 / TAN(x))
+
+// Secant
+#define SEC(x) (1 / cos(x))
+
+// Cosecant
+#define CSC(x) (1 / sin(x))
+
+#define ATAN2(x, y) ( !(x) && !(y) ? 0 : (y) >= 0 ? arccos((x) / sqrt((x)*(x) + (y)*(y))) : -arccos((x) / sqrt((x)*(x) + (y)*(y))) )
```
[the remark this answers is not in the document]
kevinz000: This is the old Atan2 but as a define. 
/proc/Atan2(x, y)
	if(!x && !y) return 0
	var/a = arccos(x / sqrt(x*x + y*y))
	return y >= 0 ? a : -a
#define ATAN2(x, y) (
 !(x) && !(y) ? 0 : 
(y) >= 0 ? arccos((x) / sqrt((x)*(x) + (y)*(y))) :
 -arccos((x) / sqrt((x)*(x) + (y)*(y))) )

PR #5665, opened 2018-10-13
Port some noises from baystation.

lbnesquik:
- Plays keyboard noises for every computers
- Button noises for the chem master
- Mini framework for easier noises put in. Couldn't put in everything without a lot more work than i felt like doing this Saturday.


lbnesquik:
Am i supposed to fix the conflict myself or something? I never had this sort of issues before.

Atermonera:
Yes. The nature of conflicts means that git cannot reconcile them itself, so it's on whoever gets the conflict to fix them.
Also requisite question about sound sources/licenses, because that's not something we want to get in trouble for

Neerti:
I know TG has keyboard sounds too, Bay might've ported them.

PR #5677, opened 2018-10-15
[WIP]Replaces GoonPS scheduled_tasks and obj processing with SMC processing subsystems and timer

kevinz000:
Separated from the projectile PR due to this being a larger project than I thought.


Atermonera:
It's generally preferred that you poke a dev to reboot travis, rather than making an empty commit

PR #5680, opened 2018-10-15
Add the heavy Duty drill, a sidegrade to the current mining tools.

lbnesquik:
This PR adds a fuel powered handheld, animated, upgradable drill meant to be a decent side grade to current mining tools. 
Many buzzwords wow. Anyway, this works like the chainsaw if you ever got to use it. Need welding fuel to work but it can be easily be converted to use something less available. It should be perfectly functional as is but i would like to have the modular element well working before merging.
It is in it's own file because it's pretty long.

Stuff to ~~not~~ validate before merging:
- [ ] Handheld sprites. (Current ones are the chainsaw ones.)
- [x] Make it so you can fill the drill with beakers and such instead of having to click on fuel tanks.
- [ ] Get icons for modular parts.
- [x] Make it so modular parts can be removed and switched relatively easily.
- [x] Make them actually relevant in the drill process.
- [x] Make it possible for drill parts to be ordered or researched.

Edit.
Since it may not have been clear, the modulable part is meant for upgrades. For example you could get another drill chassis with less slowdown or more fuel. Or both. You can get better filters so your engines jams less. You can get better engines either for better fuel consumption or faster drilling as well. And you can do something useful by changing drill bits.


lbnesquik:
https://cdn.discordapp.com/attachments/170364397238222848/501506772247511041/2018-08-11_08-56-38.gif vrum vrum

lbnesquik:
Many colors

lbnesquik:
Big push

lbnesquik:
Cargo parts can now be ordered.

lbnesquik:
Well for now the only 3 real issues is, lack of parts sprites, lack of handheld sprites and the parts don't correctly update their stats.

PR #5683, opened 2018-10-16
Adds Mods/Dev perms to ticket permission checks

Neerti:
Currently, moderation staff can't answer tickets. That's pretty bad.
This changes it so that moderators, admins, and developers can answer tickets (Remember, an adminhelp isn't always about 'pls ban he', but might also be questions on how to do something, which a dev is generally an expert in how the game ~~doesn't~~ work).

Downstream servers might want to skip on this PR if their staff structure/permission setup is radically different from ours.

### DIFF
--- a/code/modules/admin/topic.dm
+++ b/code/modules/admin/topic.dm
@@ -11,7 +11,7 @@
 		return
 
 	if(href_list["ahelp"])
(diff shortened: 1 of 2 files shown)


Woodratt:
Kind of a priority to get this merged.

PR #5689, opened 2018-10-18
Remplace pills made by the condimaster with bouillon cubes.

lbnesquik:
Just a couple lines remplace so the condimaster can create pills which look like bouillon cubes.
The .DMI change is just me adding the monkey cube sprite with a different name. It was fair easier than to have the template look into another file for the icon.

example

example2


lbnesquik:
https://github.com/PolarisSS13/Polaris/pull/5689

Neerti:
You appeared to have linked to this exact PR (#5689), which I don't think you intended to do.

lbnesquik:
Not really, no!

code/modules/reagents/Chemistry-Machinery.dm
```diff
@@ -231,9 +231,14 @@
 			var/amount_per_pill = reagents.total_volume/count
 			if (amount_per_pill > 60) amount_per_pill = 60
 
-			var/name = sanitizeSafe(input(usr,"Name:","Name your pill!","[reagents.get_master_reagent_name()] ([amount_per_pill] units)") as null|text, MAX_NAME_LEN)
-			if(!name) //Blank name (sanitized to nothing, or left empty) or cancel
-				return
+			if(condi)//For the condimaster
+				var/name = sanitizeSafe(input(usr,"Name:","Name your cube!","[reagents.get_master_reagent_name()] ([amount_per_pill] units)") as null|text, MAX_NAME_LEN)
+				if(!name) //Repeated because i suck
```
Atermonera: You can pull both of these `if(!name)`s out of the if else block and put it after the else. The code needs to run regardless of the outcome of the if, so it shouldn't go in the if

lbnesquik: If there is no name, a return occurs and no pill is produced.

code/modules/reagents/Chemistry-Machinery.dm
```diff
@@ -231,10 +231,15 @@
 			var/amount_per_pill = reagents.total_volume/count
 			if (amount_per_pill > 60) amount_per_pill = 60
 
-			var/name = sanitizeSafe(input(usr,"Name:","Name your pill!","[reagents.get_master_reagent_name()] ([amount_per_pill] units)") as null|text, MAX_NAME_LEN)
+			if(condi)//For the condimaster
+				var/name = sanitizeSafe(input(usr,"Name:","Name your cube!","[reagents.get_master_reagent_name()] ([amount_per_pill] units)") as null|text, MAX_NAME_LEN)
```
Atermonera: This only exists within the scope of this if statement. The variable declaration needs to be moved outside of it

PR #5692, opened 2018-10-18
Correctly fix surgery caps selection this time.

lbnesquik:
Honestly this one is so much more clear than the other that we should always use it.


lbnesquik:
https://github.com/VOREStation/VOREStation/issues/4420

lbnesquik:
For reference, the code is the same as the ring selection one

PR #5698, opened 2018-10-22
Prometheans can process blood without exploding.

Mechoid:
Prometheans now process blood like a weak nutrient/nutriment, and if they ever get the 'non-blood blood' for their 'vessel', it will still provide some restoration a la normal nutrients.


Mechoid:
It takes 250 nutrition to regenerate limbs, blood gives 20 per unit, so not counting nutrition loss from casual hunger, 12.5 units. That's more than nutriment at 30 -> 8.4.

Wickedtemp:
Ohneat! I'm glad this is being added, then.

PR #5699, opened 2018-10-22
Kates fish port 2 electric boogaloo

lbnesquik:
Ports paradise fish with most previous change requests fullfilled.
I've left the special cat interaction in the code, commented out for reference purposes. 
Currently the one big issue remains is that our glass code is (very likely) not modular, making the aquariums pretty hard to obtain directly. But fear not, we can always simply put some with map tools, particularly the larger ones. I *may* try to generate recipes with glass alongside the unique window interface but otherwise, i am not willing to touch it.


lbnesquik:
https://github.com/PolarisSS13/Polaris/pull/5627

lbnesquik:
Changelog included.

lbnesquik:
1 stalled build. the rest are ok

lbnesquik:
Well would you look at that.

lbnesquik:
DO remember to check my comments in the previous PR before looking at this one.

Atermonera:
If a build has stalled, then you don't need to make a commit to fix it, devs will see that travis is confused and restart jobs as necessary if/when they see it

lbnesquik:
Got it.

PR #5702, opened 2018-10-24
Toilet BUFF

lbnesquik:
Make it possible to put reagents into a toilet cistern, allowing for tained swirlies. Wanna do acid swirlies?
You sure can! Also there is now a flushing noise. Alt click to flush.

### DIFF
--- a/code/game/objects/structures/watercloset.dm
+++ b/code/game/objects/structures/watercloset.dm
@@ -12,8 +12,12 @@
 	var/w_items = 0			//the combined w_class of all the items in the cistern
 	var/mob/living/swirlie = null	//the mob being given a swirlie
 
+	var/water_supply = 1
+
 /obj/structure/toilet/New()
 	open = round(rand(0, 1))
+	create_reagents(300)
+	reagents.add_reagent("water", 100)
 	update_icon()
 
 /obj/structure/toilet/attack_hand(mob/living/user as mob)
@@ -47,11 +51,28 @@
 	if(I.is_crowbar())
 		to_chat(user, "<span class='notice'>You start to [cistern ? "replace the lid on the cistern" : "lift the lid off the cistern"].</span>")
 		playsound(loc, 'sound/effects/stonedoor_openclose.ogg', 50, 1)
-		if(do_after(user, 30))
+		if(do_after(user, 15))
 			user.visible_message("<span class='notice'>[user] [cistern ? "replaces the lid on the cistern" : "lifts the lid off the cistern"]!</span>", "<span class='notice'>You [cistern ? "replace the lid on the cistern" : "lift the lid off the cistern"]!</span>", "You hear grinding porcelain.")
 			cistern = !cistern
 			update_icon()
 			return
+
+	if(I.is_wrench())
+		water_supply = !water_supply
+		to_chat(user, "You [water_supply ? "close" : "open"] \the [src] water supply. It won't refill after being flushed now.")
+		return //So you don't put the wrench insides
+
+	if(istype(I, /obj/item/weapon/reagent_containers))
+		if(cistern)
+			if(I.reagents.total_volume < 1)
+				user << "\The [I] is empty."
+				return
+			else
+				to_chat(user, "<span class='notice'>You empty 20 units out of \the [I.name] into \the [src].</span>")
+				I.reagents.trans_to_obj(src, 20)
+				return
+		else
+			user << "\The [src] cistern is closed."
 
 	if(istype(I, /obj/item/weapon/grab))
 		user.setClickCooldown(user.get_attack_speed(I))
@@ -71,6 +92,8 @@
 						user.visible_message("<span class='danger'>[user] gives [GM.name] a swirlie!</span>", "<span class='notice'>You give [GM.name] a swirlie!</span>", "You hear a toilet flushing.")
 						if(!GM.internal)
 							GM.adjustOxyLoss(5)
+						src.reagents.splash(GM, reagents.total_volume)
+						flush(user)
 					swirlie = null
 				else
 					user.visible_message("<span class='danger'>[user] slams [GM.name] into the [src]!</span>", "<span class='notice'>You slam [GM.name] into the [src]!</span>")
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/watercloset.dm
```diff
@@ -47,12 +51,29 @@
...
 
+	if(I.is_wrench())
+		water_supply = !water_supply
+		to_chat(user, "You [water_supply ? "close" : "open"] \the [src] water supply. It won't refill after being flushed now.")
+		return //So you don't put the wrench insides
+
+	if(istype(I, /obj/item/weapon/reagent_containers))
+		if(cistern)
+			if(I.reagents.total_volume < 1)
+				user << "\The [I] is empty."
```
Poojawa: to_chat(user, instead of <<

code/game/objects/structures/watercloset.dm
```diff
@@ -47,12 +51,29 @@
...
+				return
+			else
+				to_chat(user, "<span class='notice'>You empty 20 units out of \the [I.name] into \the [src].</span>")
+				I.reagents.trans_to_obj(src, 20)
+				return
+		else
+			user << "\The [src] cistern is closed."
```
Poojawa: to_chat(user, instead of <<

code/game/objects/structures/watercloset.dm
lbnesquik: what are you doing reviewing pr's 'round here, partner

code/game/objects/structures/watercloset.dm
```diff
@@ -47,12 +51,29 @@
...
-		if(do_after(user, 30))
+		if(do_after(user, 15))
 			user.visible_message("<span class='notice'>[user] [cistern ? "replaces the lid on the cistern" : "lifts the lid off the cistern"]!</span>", "<span class='notice'>You [cistern ? "replace the lid on the cistern" : "lift the lid off the cistern"]!</span>", "You hear grinding porcelain.")
 			cistern = !cistern
 			update_icon()
 			return
 
+	if(I.is_wrench())
+		water_supply = !water_supply
+		to_chat(user, "You [water_supply ? "close" : "open"] \the [src] water supply. It won't refill after being flushed now.")
```
Screemonster: "You open the water supply. It won't refill after being flushed now."

????

code/game/objects/structures/watercloset.dm
```diff
@@ -47,12 +51,29 @@
...
+		return //So you don't put the wrench insides
+
+	if(istype(I, /obj/item/weapon/reagent_containers))
+		if(cistern)
+			if(I.reagents.total_volume < 1)
+				to_chat(user, "\The [I] is empty.")
+				return
+			else
+				to_chat(user, "<span class='notice'>You empty 20 units out of \the [I.name] into \the [src].</span>")
```
Screemonster: Why not use the transfer volume from whatever the container is?

Arokha: Which is the extremely long `var/amount_per_transfer_from_this`

SpadesNeil:
I sit here and think, "This is the kind of nonsense Vorestation comes up with."

Screemonster:
to be fair I've thought giving someone a polytrinic swirly on /tg/ would be a brutal way for a traitor to kill their target. Hell, I'd probably trade my telecrystals for an unanchored toilet if I rolled traitor just for the spectacle of dragging a toilet around the halls and swirlying people to death in it.

lbnesquik:
@mistyLuminescence This doesn't interact at all. You would empty the beaker for sure.

code/game/objects/structures/watercloset.dm
lbnesquik: Electrical terms. Circuit being open = not complete = not working = not going to refill. Im going to change that.

code/game/objects/structures/watercloset.dm
lbnesquik: I am not sure how to do that.

Atermonera:
So it DOES interact by outright removing that function for reagent containers?

lbnesquik:
Er, *yes?* You can't hide beakers into toilets, or any reagent containers for that matter. I felt like adding another variable to control whether you pour the contents or put the beaker in would had been too complex and frankly, unneeded.

Screemonster:
If it's already possible to hide beakers then continue to make them hideable _only_ if the lid is on.

Seriously, you're putting an open container into a toilet cistern. Full of water. The contents of that container are not gonna stay contained.

lbnesquik:
But @Screemonster  im lazyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy. 
I doubt anyone ever hid any beaker of liquid anyways. How often do people get reagents that they want to hide in toilets? The only reason you do that is if you are expecting to be arrested. But, if there is insistent request, i'll make it so you can hide beakers as well.

code/game/objects/structures/watercloset.dm
```diff
@@ -47,12 +51,30 @@
...
-		if(do_after(user, 30))
+		if(do_after(user, 15))
 			user.visible_message("<span class='notice'>[user] [cistern ? "replaces the lid on the cistern" : "lifts the lid off the cistern"]!</span>", "<span class='notice'>You [cistern ? "replace the lid on the cistern" : "lift the lid off the cistern"]!</span>", "You hear grinding porcelain.")
 			cistern = !cistern
 			update_icon()
 			return
 
+	if(I.is_wrench())
+		water_supply = !water_supply
+		to_chat(user, "You [water_supply ? "open" : "close"] \the [src] water supply valve. It won't refill after being flushed now.")
```
Screemonster: Still says "it won't refill after being flushed now" regardless of the open/closed state.
"you open the valve, now it won't refill"
"you close the valve, now it won't refill"

lbnesquik: Oh i get the issue now.

lbnesquik:
Now it works btw.

code/game/objects/structures/watercloset.dm
```diff
@@ -47,12 +51,33 @@
...
+	if(I.is_wrench())
+		water_supply = !water_supply
+		if(water_supply)
+			to_chat(user, "You open \the water supply valve. It can flush again now.")
+		else
+			to_chat(user, "You close \the [src] water supply valve. It won't refill after being flushed now.")
+		return //So you don't put the wrench insides
+
+	if(istype(I, /obj/item/weapon/reagent_containers))
+		if(cistern)//If 1, it is open.
```
Screemonster: Wait, so you have to crowbar the cistern open to add reagents? Not just pour them into the bowl?

PR #5703, opened 2018-10-24
Semifinal AI

Neerti:
Almost done with the rewrite.
All it needs after this is another unclusterfucking with master, server tests, bug fixes from the tests, then probably another unclusterfucking, and then finally a merge into master.


Neerti:
I hit the wrong branch :sad:

PR #5704, opened 2018-10-24
Semifinal AI : Electric Boogaloo

Neerti:
Almost done with the rewrite.
All it needs after this is another unclusterfucking with master, server tests, bug fixes from the tests, then probably another unclusterfucking, and then finally a merge into master.
Now targeting the correct branch.


code/modules/mob/living/simple_mob/subtypes/animal/pets/cat.dm
```diff
@@ -0,0 +1,136 @@
...
+					visible += O
+			if(visible.len)
+				var/atom/A = pick(visible)
+				visible_emote("suddenly stops and stares at something unseen[istype(A) ? " near [A]":""].")
+
+// Instakills mice.
+/mob/living/simple_mob/animal/passive/cat/apply_melee_effects(var/atom/A)
+	if(ismouse(A))
+		var/mob/living/simple_mob/animal/passive/mouse/mouse = A
+		if(mouse.getMaxHealth() > 20) // In case a badmin makes giant mice or something.
```
Anewbe: Is this supposed to be more than 20, or less than?

Neerti: Should be correct now, goal was so mice wouldn't be instakilled if they had more than 20 max health.

PR #5684, opened 2018-10-16
[WIP]Ports (almost) all global variables to managed globals

kevinz000:
More refactors for the refactor god.
This is going to take a while.

PR #5720, opened 2018-10-29
Add round end stats about the station activity.

lbnesquik:
See titles. Examples include cans opened, number of items vended by machines or limbs lopped off.
Feel free to suggest more.
stuff


code/roundstats.dm
```diff
@@ -0,0 +1,46 @@
...
+	var/lost_limbs_shift = lost_limbs_shift_temp
+	//var/mouse_spawned_shift = mouse_spawned_shift_temp //Virgo
+	var/seed_planted_shift = seed_planted_shift_temp
+	var/step_taken_shift = step_taken_shift_temp
+	//var/number_people_walked_over = number_people_walked_over_temp //Also virgo
+	var/destroyed_research_items = destroyed_research_items_temp
+	var/items_sold_shift = items_sold_shift_temp
+	var/disposals_flush_shift = disposals_flush_shift_temp
+
+	world << "<B>Shift facts!</B>"
```
Atermonera: `to_chat()` everywhere in here. It's currently just a macro for `<<` so it should work fine. If it's expanded later, then whoever does that needs to be sure that they have a case for output to large numbers of mobs.

code/global.dm
```diff
@@ -185,3 +185,17 @@ var/static/icon/ingame_hud_med = icon('icons/mob/hud_med.dmi')
 //Keyed list for caching icons so you don't need to make them for records, IDs, etc all separately.
 //Could be useful for AI impersonation or something at some point?
 var/static/list/cached_character_icons = list()
+
+var/cans_opened_temp = 0
```
Atermonera: I feel like these are the sort of thing that's supposed to live on GLOB

code/roundstats.dm
lbnesquik: Are you sure it would be visible to everyone?

code/global.dm
lbnesquik: I have no idea what i am doing, really.

lbnesquik:
@Atermonera Only the ones like steps taken or elevator usage should grow to high numbers. Then it just depend on what is the biggest number a variable can hold.

code/roundstats.dm
Arokha: There may be a `to_world` macro already. /tg/ has one anyway.

Arokha:
Incrementing the variable is probably pretty quick. I'm more concerned with the silly name instead of something like Paradise's 'score_stuffharvested' etc which makes it more obvious.

Atermonera:
See #5684 for how to put stuff on GLOB. It's a controller for global variables, something something performance, I don't really know why I think bay made it but apparently it's something we want to steal.

Oh, and globals represent an imperfect knowledge of what you're doing, because there's usually a better way to do it, that might be a part of it.

Neerti:
I am a bit concerned that this could end up flooding the chat at the end of the round with information that might not really be worth noting, compared to more impressive things that might deserve a mention like 'mining got 600 points worth of ore this shift' or something like that.

Bromuzl:
Most games with server performance as foremost priority seem to use a set of 3-6 "if/else/or" functions to draw the most interesting stats.

Rare or lucky events such as " Steve activated a memetic happiness artifact, a 0.28% chance find! Praise Steve °-° " taking priority over weighing via equation if "Urist mined 17 diamond turfs this shift!" is more impressive than "Urkle took damage from x unique sources for 830 combined damage, and a low of -134.53 health without dying this round!" for example.

Also worth considering repurposing data we **already** collect or at least retooling them to minimize performance/clutter.
- **Performance Award** Function finds and announces lowest minutes spent idle by department/antag role.
- **Adbuse Driver** __honk honk!__ Most admin logs generated, Gamemaster rank exempt to exclude event managers and their help.
- **..but they still can't hear you scream** Highest count of "follow" and "jump-to" verbs executed on given ckey's mobs by observers. Count excludes aghost, winner excludes those banned/kicked. Exclusive to Cult/Cata modes, and rounds with 3+ deaths occurring to PC mobs on space tiles

lbnesquik:
Given I didn't want people to stumble on the variable++ and have a hard time identifying them, I gave them an unique suffix. Anyone searching this in a proper editor would figure it out quickly

lbnesquik:
@neerti It is likely possible to hard check for the bigger things like outstanding amounts of material or money made and then have a function pick a few random trivia lines to output. It would keep things a bit more novel and a lot less spammy.

lbnesquik:
So a couple days ago, i tried to move stuff to glob and it just failed horribly. I get new errors that i don't understand. So it would be cool to stick to the tried and tested var/stuff instead.

lbnesquik:
I have them changed. I just can't push my code as it's all buggy GLOB. stuff now.

code/_global_vars/roundstats.dm
```diff
@@ -0,0 +1,78 @@
...
+//Virgo specific
+//GLOBAL_VAR_INIT(mouse_spawned_shift_roundstat, 0)
+//GLOBAL_VAR_INIT(number_people_walked_over_roundstat, 0)
+//GLOBAL_VAR_INIT(trash_piles_searched_roundstat, 0)
+//GLOBAL_VAR_INIT(rare_trash_found_roundstat, 0)
+
+
+/hook/roundend/proc/RoundEnd()
+
+	var/lights_switched_on = GLOB.lights_switched_on_roundstat
```
Cerebulon: Why are you saying "This variable = That variable", why not just output GLOB.lights_switched_on_roundstat?

code/_global_vars/roundstats.dm
```diff
@@ -0,0 +1,78 @@
...
+	if(destroyed_research_items > 0)
+		to_world("[destroyed_research_items] objects were destroyed in the name of Science! Keep it up!")
+	if(items_sold_shift > 0)
+		to_world("The vending machines sold [items_sold_shift] items today.")
+	if(disposals_flush_shift > 0)
+		to_world("The disposal system flushed a whole [disposals_flush_shift] times for this shift. We should really invest in waste treatement.")
+	if(rocks_drilled > 0)
+		to_world("Our strong miners pulverized a whole [rocks_drilled] piles of pathetic rubble.")
+
+	//virgo specific aswell
```
Cerebulon: We don't need code for Virgo in here.
You or they are going to have to make a separate PR to uncomment these anyway, so just take them out and add them in a PR on Virgo otherwise we have to take it out which is... Weird.

code/_global_vars/roundstats.dm
lbnesquik: Last time i tried, i had issues where the variables wouldn't actually update. But it may not be necessary.

lbnesquik:
Removed the virgo stuff and the duplicates.

lbnesquik:
@Anewbe Beep

Atermonera:
This was discussed at the staff meeting on 8 Dec 2018.
The consensus is that we aren't particularly interested in the tracked stats or the code with which they're tracked.

PR #5723, opened 2018-10-30
Promethean Regen tweaks. (The one with the devil's deal.)

Mechoid:
Promethean passive regeneration and universal active regeneration now cause pain. Active regeneration is, percentage wise, less painful than passive regeneration at 50% pain conversion to passive 66%. This makes both regeneration forms something better for use outside of combat or high-risk situations, as it will eventually cause incapacitation if left unchecked, on top of normal injury pain.

Promethean brute and burn regeneration will not function in dangerously low pressure, however toxin and oxy (the latter not currently used) will continue as normal. Voidsuits and other sealed suits keep the regeneration active.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_powers.dm
+++ b/code/modules/mob/living/carbon/human/human_powers.dm
@@ -307,7 +307,7 @@
 				organ_data["descriptor"] = O.name
 				to_chat(src, "<span class='notice'>You feel a slithering sensation as your [O.name] reform.</span>")
 
-				var/agony_to_apply = round(0.66 * O.max_damage) // 60% of the limb's health is converted into pain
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/human/human_powers.dm
```diff
@@ -305,6 +306,10 @@
 				var/obj/item/organ/O = new limb_path(src)
 				organ_data["descriptor"] = O.name
 				to_chat(src, "<span class='notice'>You feel a slithering sensation as your [O.name] reform.</span>")
+
+				var/agony_to_apply = round(0.66 * O.max_damage) // 60% of the limb's health is converted into pain.
```
Atermonera: `66%`, I think you mean

PR #5736, opened 2018-11-10
Fixes missing hollow point ammo for 9mm

SpadesNeil:
Alternate title:
## Ace annoys Anewbe with another gun PR C:

- 9mm didn't have HP ammo for some reason, which is dumb, considering HP ammo makes 9mm not useless while not being overpowered as hell.
- Fixed inconsistent naming with hollow point ammo in the code, being labeled /hollow instead of just /hp like how armor piercing is /ap and not /armor, because that would be stupid.
- Also made some comments for stuff I have to come back and do in a future PR when the feature freeze is over.

All I did was add in the missing projectiles and relevant casings. How they get implimented if at all is up to another PR.

### DIFF
--- a/code/modules/projectiles/ammunition/rounds.dm
+++ b/code/modules/projectiles/ammunition/rounds.dm
@@ -149,7 +149,11 @@
 
 /obj/item/ammo_casing/a45/hp
 	desc = "A .45 hollow-point bullet casing."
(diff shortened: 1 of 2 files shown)


SpadesNeil:
Oh, and remind me to throw in an admin abuse .50AE Desert Eagle for shits and giggles at a later date.

PR #5738, opened 2018-11-12
Add scheduled callbacks

Arokha:
And include a callback for setting vars.

You'd use it like `schedule_callback_in(5 SECONDS, VARSET_CALLBACK(src, myvarname, FALSE))` if you wanted to set `myvarname` to `FALSE` in 5 seconds.

This was originally going to be a bugfix to you but I guess I totally forgot that I didn't actually send this to you originally.


Neerti:
Will this conflict with @kevinz000 's eventual goal to have the TG timer system?

kevinz000:
please just use the tg timers instead

Issue #5708, opened 2018-10-25
PoI rotation doesn't respect powernet special dir variables


Atermonera:
#### Brief description of the issue
PoI rotation code, specifically found here: https://github.com/PolarisSS13/Polaris/blob/master/code/modules/maps/tg/reader.dm#L311
Doesn't have a special case for wires and other powernet objects. When a PoI is rotated, the sprites of the various wires are updated, but the `d1` and `d2` vars are not, so the powernet cannot construct itself.

#### What you expected to happen
PoI rotation to not break PoI powernets

#### Steps to reproduce
I'm specifically asking for a special case to be added to the above-linked code that checks for any special directional variables, such as `d1` and `d2`, on powernet objects, that will rotate those in the same manner as the `dir` variable is rotated.

#### Additional info:
 - I should fix this, but due to school, I don't have time.
 - **Server Revision**: Found using the "Show Server Revision" verb under the OOC tab.
 - **Anything else you may wish to add** (Location if it's a mapping issue, etc)

Issue #5707, opened 2018-10-25
Fulltile windows are of the same type as directional windows


Atermonera:
#### Brief description of the issue
Any operation that sets the dir of a full tile window object, such as PoI rotation, forces them to use directional-window properties

#### What you expected to happen
Fulltile windows to remain as such when dir is set

#### What actually happened
PoI rotation breaks fulltile windows

#### Steps to reproduce
I'm specifically asking that fulltile windows be made a different type than directional windows. Inheritance makes it logical to make one the child of the other. Window construction would need to be updated, and any mapped objects of the changed type would need to be remapped, including in all PoIs

#### Additional info:
 - I should fix this, but due to school, do not have time.

Verkister:
This also affects roundstart engine rotation where applicable and sounds more like a startup initialization issue rather than mapping issue.
https://i.gyazo.com/d8b8f0046ee7f7e4f8091033fab49511.png

PR #5741, opened 2018-11-20
Fixes unnecessary submap item rotation on load

Verkister:
The comment said "Rotate dir _if_ orientation isn't south (default)". The loading was forcibly rotating everything that either had no relevance of direction, or had been intentionally set to not face a _sideways_ direction (full windows etc), to south during load regardless of conditions.
While the issue priority here was marked low as a PoI exclusive thing, this thing also affects roundstart engine rotation _(and thus half of the engineering)_ where enabled among other things.

### DIFF
--- a/code/modules/maps/tg/reader.dm
+++ b/code/modules/maps/tg/reader.dm
@@ -311,8 +311,6 @@ var/global/use_preloader = FALSE
 			// Rotate dir if orientation isn't south (default)
 			if(fields["dir"])
 				fields["dir"] = turn(fields["dir"], dir2angle(orientation) + 180)
-			else
-				fields["dir"] = turn(SOUTH, dir2angle(orientation) + 180)
 
 			//then fill the members_attributes list with the corresponding variables
 			members_attributes.len++


Atermonera:
If dir is south, then +180 will result in a 0 degree rotation. If dir isn't specified, then it defaults to south. Things that don't care about dir shouldn't care if they get rotated, but things that DO care about dir but weren't mapped with dir specifically set won't get properly rotated without the else statement.

I insist that the full-tile windows are an issue more with full tile windows specifically having no dir and directional windows being the same object, just with a _non-0_ dir, rather than me forcing them to have a rotation, because again, something that is _necessarily directionless_ shouldn't care if it gets rotated. That it does seems like an oversight to me. This is why I wrote #5707 

Furthermore, any object, like wires, which has additional dir-variables _beyond_ `dir` still won't be rotated properly because either those need to be refactored or special cases need to be added to the rotation code, as specified in #5708 

So, with all that in mind, is this tested?

Verkister:
Yeah, tested for roundstart submaps at least. Not sure if I can reliably test the rotation thing before the other issue of the code throwing the submap vertical dimensions out of window is fixed.
Other thing is, _why should the necessarily directionless things get rotated in the first place?_

Atermonera:
They don't need to rotated, but not everything that's mapped without a direction is directionless (since south is the default). They should be rotated because when rotating a submap, it makes far more sense to rotate everything uniformly than to have a huge and ever-changing list of things that are directionless specifically so that I can check that the thing I want to rotate actually cares if it gets rotated. It only makes the algorithm slightly more complicated if you use a list like that, but that list has to be defined somewhere and kept up to date with every new thing that's directionless.

That fulltile windows DO care that dir is 0 means that their implementation _isn't_ directionless, even though it should be.

If the default direction _wasnt_ a valid assignment of one or more directionless botflags (i.e., 0), then every mapped object that cares about direction would need it manually defined, and this would be a non-issue and the else statement you're removing here wouldn't have been needed in the first place. Sadly, I don't know if that's strictly possible in byond, or if it's feasible to implement such a standard.

Atermonera:
As an example of the issue I believe this change would introduce, can you prove with screenshots that the highlighted chairs in the image below are properly rotated with your change? The map file for the Drugden wilderness poi and text representation thereof are provided, showing that they are not mapped in with**out** a direction, and therefore would be rotated only by the else clause you're removing here.
Drugden.dmm

Verkister:
I'll check it out when I can. Not entirely sure if I got this part _"showing that they are not mapped in with**out** a direction, and therefore would be rotated only by the else clause you're removing here."_ right, but the else part would **only** rotate things that actually **are** mapped in with**out** a direction, and that in the screenshot there is an existing dir of 2 showing.

Neerti:
#5748 might be a viable alternative for downstreams to utilize until the rotation issues can be fixed.

Atermonera:
The screenshot shows dir=2, but the 2 isn't bolded, indicating that it wasn't mapped in that way, that's just the default value. If you look at the text representation, the line with a blue dot specifically, you can see that the definition of those chairs does not include a {dir=2}.
That {dir=2} is what the if() clause catches, where the item has an explicitly mapped direction, and that needs to be rotated. The else clause catches the chair case, where dir is not explicitly mapped in, and rather implicitly inherited from the object's code definition.

PR #5747, opened 2018-12-01
Makes fulltile windows seperate from dir

### LINKED ISSUE #5707: Fulltile windows are of the same type as directional windows

Neerti:
Also cleans up some unused code for building windows, and makes it so there aren't two separate procs for determining if a window is a fulltile one or not.
It also should fix window weirdness in PoI rotation.

Fixes #5707

### DIFF
--- a/code/game/atoms_movable.dm
+++ b/code/game/atoms_movable.dm
@@ -119,7 +119,7 @@
 				// Special handling of windows, which are dense but block only from some directions
 				if(istype(A, /obj/structure/window))
 					var/obj/structure/window/W = A
-					if (!W.is_full_window() && !(turn(src.last_move, 180) & A.dir))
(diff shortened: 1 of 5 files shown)


Neerti:
Also a note: Window health might need to be looked at since the PR also fixes what I assume is an oversight where full tile windows both had their health define doubled, as well as their health doubled again when instantiated.

PR #5748, opened 2018-12-01
Adds ability to selectively disable submap rotation.

Neerti:
Submap rotation can be disabled globally with a config change now.
They can also be disabled on a per-map basis, in case there is a submap that really should not be rotated.
Note that admins loading in submaps can still rotate the submap if they choose to do so regardless of what the config or the submap datum says.

The beacon Point of Interest now has a fixed orientation, as changing the orientation caused the perspective to change, and as a result caused the meaning of the map to be loss.


elgeonmb:
*screaming*

PR #5757, opened 2018-12-04
Adds several new map/event oriented objects

Neerti:
This PR adds several things to make maps and events better.
- Invisible objects which do a specific thing, including;
  - Permanent lighting assignable from the map editor independent of any physical object visible in the world.
  - Radiation emitters, that turn an area into Fallout [whatever game you liked the best].
  - Effect emitters, that emit things like smoke or sparks over time.
    - Adds several smoke types and emitters that do bad things like lighting you on fire when touched. Wish we had better sprites for clouds on fire or freezing mist but oh well.
  - Sound emitters, that make one or more sounds every so often.
  - Screen shakers, which do what the name says, every so often.
  - Beam points, which make a visual beam between two or more points. They can be permanent or turn on or off on a timer. The beams can even react to being touched.
    - Includes a lightning beam that shocks people who touch it, as a demo for how to make non-mob based hazards in PoIs. Video example here: https://streamable.com/5k4g8
- Adds ability to have lights automatically flicker when someone is somewhat nearby for maximum spooky. They will stop doing so if nobody is around to see it to not make the lighting controller cry.

Other things:
- Fixes bug/oversight causing ``inflict_shock_damage(amount)`` to only hurt the hands due to how ``electrocute_act()`` handles not providing a targeted bodypart on human mobs.
- Adds helper proc to test if a mob can breathe. For now everything except FBPs and lings with their self-breathing ability active will pass it.
- Only things that breathe will cough in smoke, so FBPs will stop electronically coughing for no reason.
- Adds helper proc to test for nearby player presence. Can exclude ghosts and AFK people if desired with proper arguments.


Anewbe:
> Adds helper proc to test if a mob can breathe. For now everything except FBPs and lings with their self-breathing ability active will pass it.
_Can_ breathe, or _must_ breathe?

PR #5762, opened 2018-12-07
Capitalizes initialize() to Initialize() because Destroy() is capitalized and this is pretty nerve-grinding

kevinz000:
'least I'm honest..


Atermonera:
As I noted on discord, instead of making instances where initialize() is called on an atom, I would make a proc initialize() on atom that just calls Initialize(), in keeping with the semantics of New() and Destroy(), where the proc names are capitalized, but calling them requires not capitalizing:
>/atom/proc/initialize()
> > Initialize()

PR #5764, opened 2018-12-07
Event / Future PoI / Other weaponry & tools.

Mechoid:
Tweaks & Fixes
- Energy weapons can be set to allow using different types of power cells. The cell size influences speed.
- Energy weapons' icons update when their cells are returned, even if they don't have a charge meter

Additions
- Phoron Bore- A heavy-duty mining tool. Breaks rock and deals heavy damage to walls, fueled by phoron and powercells. Hopefully going to be eventually available through the mining point vendor.
- Portable MHD Howitzer- A massive two-handed weapon used to punch through armor and walls. Uses powercells, and takes 10,000 units of power to fire. Charges shots, and laser-designates its target turf for three seconds before firing.
- Normal and Combat 'monorifles'- Weapons utilizing similar capacitors and focus lenses as LWAP rifles, however limited to one or two shots, and minimal-zoom sights. Incredibly inefficient power-wise, though the single shot variant isn't terribly inaccurate when hip-fired to compensate.
- Gravity Whip- A multi-purpose device used for moving people and objects about. A proof of concept for the idea, and has the ability to be powerful in the right hands.
- Ominous Ring- A probably-not-too-vague reference to something about lanterns and rainbows, but also an example of combining the hook projectiles with others in different fire-modes.

[Apparently I typed faster than I expected. This covers it.]


code/modules/projectiles/projectile/hook.dm
```diff
@@ -0,0 +1,192 @@
...
+				if(!target_mob)
+					return
+
+				if(Bump(target_mob, forced=1))	//If we hit a turf, try to force an interaction with a mob on the turf.
+					done_mob_unique = TRUE
+				success = TRUE
+			else if(firer)
+				var/obj/T
+
+				if(original in target.contents && istype(original, /obj))
```
Atermonera: Original doesn't appear to be defined anywhere, so I'm not sure how this compiled.

code/modules/projectiles/projectile/special.dm
```diff
@@ -257,3 +257,84 @@
...
+		if(ishuman(L))
+			var/mob/living/carbon/human/H = L
+
+			var/target_armor = H.getarmor(def_zone, check_armour)
+			var/obj/item/organ/external/target_limb = H.get_organ(def_zone)
+
+			var/armor_special = 0
+
+			if(target_armor >= 60)
+				var/list/throw_dirs = list(1, 2, 4, 8, 5, 6, 9, 10)
```
Atermonera: `var/turf/T = get_step(H, pick(alldirs - src.dir))`
`alldirs` is a global list containing the human-readable dirs:
`var/list/alldirs = list(NORTH, SOUTH, EAST, WEST, NORTHEAST, NORTHWEST, SOUTHEAST, SOUTHWEST)`

code/modules/projectiles/projectile/special.dm
```diff
@@ -257,3 +257,84 @@
...
+	if(istype(A,/turf/simulated/shuttle/wall) || istype(A,/turf/simulated/wall) || (istype(A,/turf/simulated/mineral) && A.density) || istype(A,/obj/mecha) || istype(A,/obj/machinery/door))
+		var/blast_dir = src.dir
+		A.visible_message("<span class='danger'>\The [A] begins to glow!</span>")
+		spawn(2 SECONDS)
+			var/blastloc = get_step(A, blast_dir)
+			if(blastloc)
+				explosion(blastloc, -1, -1, 2, 3)
+	..()
+
+/obj/item/projectile/beam/tungsten/Bump(atom/A as mob|obj|turf|area, forced=0)
```
Atermonera: mob, obj, turf, and area are all subtypes of atom, so you don't need to cast it as any of them.

code/modules/projectiles/guns/magnetic/bore.dm
```diff
@@ -0,0 +1,142 @@
+/obj/item/weapon/gun/magnetic/matfed
+	name = "portable phoron bore"
+	desc = "A large man-portable tunnel bore, using phorogenic plasma blasts. Point away from user."
+	description_fluff = "An aging Grayson Manufactories mining tool used for rapidly digging through rock. Mass production was discontinued when multiple of the devices were stolen and used to break into a high security facility by Boiling Point drones."
```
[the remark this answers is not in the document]
Mechoid: I can english I swear.

code/modules/projectiles/projectile/hook.dm
Mechoid: Original is the projectile's 'original' target, the thing the player clicked on to fire the gun.

PR #5776, opened 2018-12-12
Adds pressure protection bounds to clothing.

### LINKED ISSUE #5739: Firefighter-Gear Spaceproof
Wearing the Figherfighter Suit with gasmask, internals and Helmet makes you spaceproof.
Death by decompression after fatal Toxins Screwup.
#### What actually happened
I am watching the stars, and live.
#### Steps to reproduce
Wear the full firefighter-suit (Helmet maybe optional?) And spacewalk. Remember to have air.
#### Additional info:

Atermonera:
Protection falls off when exceeding the pressure limits from 100% protection at the boundary to 0% at 10% in excess (above the max or below the min) boundary. Currently, only hat and suit slots are checked for this protection (No change from prior behaviour). In order for the protection to work, both the max and min pressure variables must be non-null.
Anomaly suits (The orange ones) now offer limited pressure protection in case anomalies start making or draining air. Firesuits are no longer spaceproof, but still offer protection against the high pressures of fire.
Tested.
Fixes #5739

### DIFF
--- a/code/__defines/items_clothing.dm
+++ b/code/__defines/items_clothing.dm
@@ -53,11 +53,10 @@
 
 //Flags for items (equipment) - Used in /obj/item/var/item_flags
 #define THICKMATERIAL          0x1  // Prevents syringes, parapens and hyposprays if equipped to slot_suit or slot_head.
-#define STOPPRESSUREDAMAGE     0x2  // Counts towards pressure protection. Note that like temperature protection, body_parts_covered is considered here as well.
-#define AIRTIGHT               0x4  // Functions with internals.
-#define NOSLIP                 0x8  // Prevents from slipping on wet floors, in space, etc.
-#define BLOCK_GAS_SMOKE_EFFECT 0x10 // Blocks the effect that chemical clouds would have on a mob -- glasses, mask and helmets ONLY! (NOTE: flag shared with ONESIZEFITSALL)
-#define FLEXIBLEMATERIAL       0x20 // At the moment, masks with this flag will not prevent eating even if they are covering your face.
+#define AIRTIGHT               0x2  // Functions with internals.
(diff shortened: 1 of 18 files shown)


zekesulastin:
I'm not quite certain what the point of firesuits are anymore, considering engineering has access to the milieu of engineering voidsuits and that they will no longer protect you from a toxins breach/venting an area to stop a large fire, but whatever; they had a great run!

Also, while I see that you attempted to differentiate the CE rigsuit by giving it 100x atmos protection compared to every other suit's 20x protection to make it "comparable to the atmos suit", you forgot to actually buff the atmos suit as far as I can tell.  I would kind of figure that all the engi stuff in general might be a bit more resistant than a plain old softsuit since any engineer would likely be going into a fire instead of running away, but eh.

Atermonera:
Forgetting to increase the resistance of the atmos suit was an oversight, and the uniform 20 atmosphere value is a testing value while other people with more to say about balance actually discuss what values each suit should get. I don't know what pressure a toxins fire can reach, so I don't actually know what the engi/atmos suits should be expected to protect against.
That said, I acknowledge that the upper bounds are quite overkill, and generally, the values could probably stand to fall by up 90%. I.e., softsuits can only protect against 1-2 atmospheres. Voidsuits should be noticeably higher, perhaps up to 10-15, so on and so forth, but these values do need discussion.

Atermonera:
https://forum.ss13polaris.com/viewtopic.php?f=14&t=687

Atermonera:
Box move ate part of the forum discussion, but I'm comfortable with these values. I know in the last batch of testing that a firesuit with internals in hard vacuum caused bruteloss-induced unconsciousness (as opposed to oxyloss) after 2m45s. I think without internals it should work out to roughly a minute.

PR #5779, opened 2018-12-13
Emergency tanks start full

Novacat:
It always bugged me how emergency tanks that you are kind of supposed to keep on you start off mostly empty, and you have to remember to fill it every round.

This PR makes emergency tanks start off full.

### DIFF
--- a/code/game/objects/items/weapons/tanks/tank_types.dm
+++ b/code/game/objects/items/weapons/tanks/tank_types.dm
@@ -139,7 +139,7 @@
 
 /obj/item/weapon/tank/emergency/oxygen/New()
 		..()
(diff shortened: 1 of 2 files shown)


Woodratt:
How long will a emergency tank last now, and with it full, how long does it last?

Novacat:
This does not change the maximum volume of the tanks. If you refilled your emergency tank at an oxygen canister, there will be no change. The main benefit is that if, say, there is an actual emergency and you need to grab internals out of an internals locker or your emergency internals box, the oxygen tanks there will start off with a little over three times the amount of air that they usually have, as emergency oxygen tanks currently start with 30% of their maximum oxygen capacity.

I do not have exact numbers on how long the tanks last. I will run some tests.

PR #5781, opened 2018-12-13
Apparently fixes the PoI squishing for good.

Verkister:
Just another functionally vital little snippet that was removed but never replaced.

### DIFF
--- a/code/modules/maps/tg/reader.dm
+++ b/code/modules/maps/tg/reader.dm
@@ -99,6 +99,11 @@ var/global/use_preloader = FALSE
 			var/ycrd = text2num(dmmRegex.group[4]) + y_offset - 1
 			var/zcrd = text2num(dmmRegex.group[5]) + z_offset - 1
 
+			if(orientation & (EAST | WEST))
+				xcrd = ycrd // temp variable
+				ycrd = xcrdStart
+				xcrdStart = xcrd
+
 			var/zexpansion = zcrd > world.maxz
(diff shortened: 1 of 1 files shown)


Verkister:
Yeah I'm gonna let ya do your thing there, but feel free to consider whether this bit can be of relevant use after the refactor.

Atermonera:
I'm... pretty sure this code just mirrors the start position about the line y=x. I can say this, because I'm pretty sure I had similar code when I was first developing the rotation system, and I removed it because it caused POIs to overlap during map generation.
It might've been placed somewhere else in the code, but I'm rather certain that I flipped the x and y coordinates if it was rotated 90 degrees, and that caused overlapping POIs.

Verkister:
Possibly. Seemed like this was the only thing in there that was in some way preserving the y-dimensions in the cases where a sideways rotation would occur since the y-part of the thing was being set here before the code would even begin checking for rotation and recalculating the x-part. There was also another little bug that pretty much "replaced" the squishing while I was testing this out, that I feel like probably was also the cause behind the overlapping. In these cases, the startup log for the submap placement reported inverted x/y coordinates for the PoIs that ended up sideways https://i.gyazo.com/e65629a3db29bb5943130129ed9fc1fb.png

Atermonera:
The y-dimension gets preserved when it generates the new key-list, by swapping the dimensions of the array and populating it from [j][i], where the original key-list is iterated across [i][j]. Or at least, that's what it should be doing.

PR #5784, opened 2018-12-16
Renames "phase pistol" cabinets to "explorer weapon cabinet"

Lorwp:
because there's multiple phase weapons now ¯\_(ツ)_/¯


Naidh:
Would probably be better if it was just renamed to "explorer gun cabinet".

PR #5786, opened 2018-12-17
Fixes a bug with holographic projectors.

avabee:
Would not properly null out the variable, meaning it just waits to be garbage collected. Bad. Very bad.

The other change is because the Github web editor wanted to convert it to UTF-8. As you see, there is no visible change, though I may need to fix that if it causes display issues in-game. (I believe it doesn't work anyway.)

### DIFF
--- a/code/modules/integrated_electronics/subtypes/output.dm
+++ b/code/modules/integrated_electronics/subtypes/output.dm
@@ -300,7 +300,7 @@
 		text_output += "\an [name]"
 	else
 		text_output += "\an ["\improper[initial_name]"] labeled '[name]'"
-	text_output += " which is currently [get_pin_data(IC_INPUT, 1) ? "lit <font color=[led_color]>¤</font>" : "unlit."]"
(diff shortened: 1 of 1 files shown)


avabee:
Fixes issue #5788.

code/modules/integrated_electronics/subtypes/output.dm
```diff
@@ -462,7 +462,7 @@
 
 
 /obj/item/integrated_circuit/output/holographic_projector/proc/destroy_hologram()
-	hologram.forceMove(src)
+	hologram = null
```
Mechoid: Will this not cause the qdel on the next line to simply be attempting to delete a null object?

Neerti:
It might be better to remove the hologram using the ``QDEL_NULL()`` macro.

code/modules/integrated_electronics/subtypes/output.dm
avabee: Ah, you're right. It should be in the opposite order. I suppose I could've just used `qdel_null()`.

PR #5790, opened 2018-12-21
Drop the number of shots of normal and sec-xeno tasers back down to 5.

Mechoid:
On the tin.
Tweaks:
- Taser shot count dropped to 5 from 10.
- Security-Issued xeno taser shot count dropped to 5 from 10.

This means that targets have a higher likelihood of escape should the officer(s) pursuing them miss, as they must either change weapons or reload their tasers.

### DIFF
--- a/code/modules/projectiles/guns/energy/stun.dm
+++ b/code/modules/projectiles/guns/energy/stun.dm
@@ -4,6 +4,7 @@
 	icon_state = "taser"
 	item_state = null	//so the human update icon uses the icon_state instead.
 	projectile_type = /obj/item/projectile/beam/stun
+	charge_cost = 480
 
 /obj/item/weapon/gun/energy/taser/mounted
 	name = "mounted taser gun"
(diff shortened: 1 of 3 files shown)


elgeonmb:
good now remove weapon cells

Anewbe:
Did we discuss this anywhere? Can't remember, feel like that's something people have words on.

Mechoid:
I've spoken about it in a meeting prior though it was ages ago, and I've spoken about it on the Teamspeak and Discord. I've not received any negative response save for concerned neutrality due to the larger issue of "Security & Station v Antag".

CameronWoof:
Considering the fairly massive RNG miss chance on weapons, this is going to be _horrifyingly_ unfun if you get even a little bit of bad luck.

Mechoid:
The intention is to make missing actually matter again, as 10 shots is _Incredibly_ lenient to the point of foolproof, especially with taser reloadability.

Anewbe:
Discussed in the Staff Meeting; if further tweaks are needed here or elsewhere, we can do that.

PR #5793, opened 2018-12-24
Upgrades SDQL2 and refactors it to a datum

kevinz000:
Documentation included!
Removes old SDQL too.
Adds a SDQL2 admin panel for when there's queries alive.


kevinz000:
fixing in a bit

PR #5796, opened 2018-12-29
Adds New PoI Features + Bonus PoI

Neerti:
This PR adds various things to spice up Points of Interest.
* Cliffs, which give a visual illusion of depth. Cliffs can be fallen off of if not careful near the edge while on top. Flying mobs can pass through cliffs without risk of injury. Projectiles trying to go 'uphill' may hit the cliff instead, making it harder to fight things above you. Sprites sourced from Baystation.
* Climbing boots, which let you climb cliffs as if they were tables. Not currently obtainable. Sprite is a mix of a magboot sprite from TG and winter boots here.
* Lava, which burns you severely when touched, and even a full firesuit will not completely protect you (although it will help a lot). Flying mobs are safe. Sprites and most of the code ported from TG.
* Fences, which let projectiles pass through, and can be cut open with wirecutter-like tools after a period of time. Also includes gates and a variant that is locked. Keys and lockpicking sold separately. Ported from TG, with a few tweaks and additions.
* Adds system to mark and unmark a turf tile as 'potentially dangerous' due to specific objects it may have in its contents. This is done to make the AI not walk off of cliffs (unless they can fly). Also made it work for landmines. (Later I might add the ability to bury landmines so the AI cannot avoid them).
* Refactors crystal generation to be cleaner and adds lava crystals as decor.
* Adds a brand new Point of Interest which shows off all of the new features as well as some map_effects. This PoI is somewhat hazardous, due to the environment instead of being full of monsters, which is really rare and I hope to make these kinds of PoIs more common later.

Image Preview


Screemonster:
For our purposes, it may be worth putting "flying" into the checks to allow flying people to pass the turfs, since this checks "hovering" and flight for players isn't a thing upstream. Is there a difference?

Anewbe:
@Screemonster Hovering basically means you don't touch the ground, and can ignore gravity while moving in the z-axis.

PR #5798, opened 2018-12-30
Adds Alien Reality pods, aka AR pods, for use in events or PoIs.

Mechoid:
**Tweaks:**
Organs can now add or remove verbs.
Organs now can have special effects on processing.
Xenomorph plasma vessels now regenerate phoron passively, at an incredibly slow rate. Can be accelerated by consuming liquid phoron.
Virtual Reality pods cleaned up.

** Additions:**
Xenomorph organs now give their respective abilities when implanted.
Alien Realiy pod added for PoIs or admins to use. Functions similarly to a VR pod, except it will produce a living body on the pod's turf.
Three replicant species added for event and PoI use. They are shape-shifters each with different abilities due to their organs.
Replicant organs added, versions of many organs that do not reject in their host. Two currently exist that give special passives.


code/game/machinery/virtual_reality/ar_console.dm
```diff
@@ -0,0 +1,135 @@
...
+
+/obj/machinery/vr_sleeper/alien
+	name = "strange pod"
+	desc = "A strange machine with what appears to be a comfortable, if quite vertical, bed. Numerous mechanical cylinders dot the ceiling, their purpose uncertain."
+	icon = 'icons/obj/Cryogenic2.dmi'
+	icon_state = "alienpod_0"
+	base_state = "alienpod_"
+
+	eject_dead = FALSE
+	var/mirror_first_occupant = TRUE	// Do we force the newly produced body to look like the occupant?
```
Anewbe: Could these vars not go on the base VR sleeper? Might be helpful/reduce copypasta.

code/game/machinery/virtual_reality/vr_console.dm
```diff
@@ -1,227 +1,260 @@
...
+
+/obj/machinery/vr_sleeper/emp_act(var/severity)
+	if(stat & (BROKEN|NOPOWER))
+		..(severity)
+		return
+
+	if(occupant)
+		// This will eject the user from VR
+		// ### Fry the brain? Yes. Maybe.
+		if(prob(severity * 15) && occupant.species.has_organ[O_BRAIN] && occupant.internal_organs_by_name[O_BRAIN])
```
Anewbe: Looks like this severity thing might be backwards? High Severity means a less powerful EMP pulse. Severity = 1 is the strongest it can be.
As written this makes minor EMPs 4 times as likely and 4 times as damaging as a direct hit would be.

code/game/machinery/virtual_reality/vr_console.dm
```diff
@@ -1,227 +1,260 @@
...
+		M.loc = src
+		update_use_power(2)
+		occupant = M
+
+		update_icon()
+
+		enter_vr()
+	return
+
+/obj/machinery/vr_sleeper/proc/go_out(var/forced = 1)
```
Anewbe: Why does forced default to 1? In combination with the above point on the forced var above, this seems to me like it'll always pass 1/TRUE into this argument, which means the check on 187 will never occur?

code/game/machinery/virtual_reality/vr_console.dm
```diff
@@ -1,227 +1,260 @@
...
+
+/obj/machinery/vr_sleeper/verb/eject()
+	set src in view(1)
+	set category = "Object"
+	set name = "Eject VR Capsule"
+
+	if(usr.incapacitated())
+		return
+
+	var/forced
```
Anewbe: Is this a YES/NO var? If so, further references should be TRUE/FALSE, and this might be better if it's set to FALSE by default.

code/game/machinery/virtual_reality/vr_console.dm
```diff
@@ -1,227 +1,260 @@
...
+	// No mob to transfer a mind from
+	if(!occupant)
+		return
+
+	// No mind to transfer
+	if(!occupant.mind)
+		return
+
+	// Mob doesn't have an active consciousness to send/receive from
+	if(occupant.stat == DEAD)
```
Anewbe: Why was this changed?

code/modules/mob/living/carbon/human/species/outsider/replicant.dm
```diff
@@ -0,0 +1,123 @@
...
+	inherent_verbs = list(
+		/mob/living/carbon/human/proc/shapeshifter_select_shape,
+		/mob/living/carbon/human/proc/shapeshifter_select_hair,
+		/mob/living/carbon/human/proc/shapeshifter_select_gender,
+		/mob/living/carbon/human/proc/shapeshifter_select_colour,
+		/mob/living/carbon/human/proc/shapeshifter_select_hair_colors,
+		/mob/living/carbon/human/proc/exit_vr,
+		/mob/living/carbon/human/proc/corrosive_acid,
+		/mob/living/carbon/human/proc/neurotoxin,
+		/mob/living/carbon/human/proc/acidspit
```
Anewbe: Are these verbs not supposed to be added by the organs they've got?

code/modules/organs/organ.dm
```diff
@@ -399,4 +414,19 @@ var/list/organ_cache = list()
 		return 0
 	if(robotic && robotic < ORGAN_LIFELIKE)	//Super fancy humanlike robotics probably have sensors, or something?
 		return 0
-	return 1
\ No newline at end of file
+	return 1
+
+/obj/item/organ/proc/handle_organ_mod_special(var/removed = 0)	// Called when created, transplanted, and removed.
```
Anewbe: Can removed not be a TRUE/FALSE?

code/modules/organs/organ.dm
```diff
@@ -399,4 +414,19 @@ var/list/organ_cache = list()
...
+	if(!istype(owner))
+		return
+
+	if(!removed && organ_verbs)
+		for(var/verb_path in organ_verbs)
+			owner.verbs |= verb_path
+	else if(organ_verbs)
+		for(var/verb_path in organ_verbs)
+			owner.verbs -= verb_path
```
Anewbe: I may be wrong, but I'm pretty sure this code means that, in the off chance you have two organs with the same verb(s), removing one will totally clear the verb(s) from your mob. Is this intended?

code/modules/organs/subtypes/replicant.dm
```diff
@@ -0,0 +1,171 @@
...
+		I.rejecting = max(0, rejecting - modifier)
+
+/obj/item/organ/internal/metamorphgland
+	name = "morphoplastic node"
+	desc = "A strange clump of meat that doesn't quite stay in place."
+	organ_tag = O_VENTC
+	parent_organ = BP_GROIN
+	icon_state = "innards"
+	organ_verbs = list(
+		/mob/living/proc/ventcrawl
```
Anewbe: Did you test this? If so, what were the results? I'm immensely curious.

code/modules/organs/subtypes/xenos.dm
```diff
@@ -29,6 +29,27 @@
...
+	if(!istype(owner))
+		return
+
+	var/modifier = 1 - 0.5 * is_bruised()
+
+	if(owner.bloodstr.has_reagent("phoron"))
+		adjust_plasma(round(2 * modifier))
+
+	if(owner.ingested.has_reagent("phoron"))
+		adjust_plasma(round(4 * modifier))
```
Anewbe: Why is this higher than the bloodstream one? Since this isn't part of phoron ticking, or seemingly related to human reagent containers beyond looking for phoron, this should probably be half of what the bloodstr offers, because it'll be hanging around longer.

code/modules/organs/subtypes/xenos.dm
```diff
@@ -61,6 +82,12 @@
 	icon_state = "acidgland"
 	organ_tag = O_ACID
 
+	organ_verbs = list(
```
Anewbe: You sure these all belong on the same organ? Last I checked the different xenos aren't all supposed to have all three of those powers.

code/modules/organs/subtypes/replicant.dm
Mechoid: The person can ventcrawl with the organ, though it doesn't allow them to crawl with things like clothing or other items still.

code/modules/organs/organ.dm
Mechoid: Not intended, I'm just an idiot and didn't consider it.

code/game/machinery/virtual_reality/ar_console.dm
Mechoid: Mirror, probably, but the species is explicitly only ever Virtual Reality Avatar for the VR pods, the others just exist to get icons from IIRC.

code/game/machinery/virtual_reality/vr_console.dm
Mechoid: The only time it isn't forced is when the eject verb is used, so defaulting to TRUE means it's less things to type, and I like most people want to type as little as possible.

code/modules/organs/subtypes/xenos.dm
Mechoid: This only gives the Drone the two spits, which considering how easily it dies, and the fact the AI drones spit, isn't the biggest of concerns in my opinion. If it needs to be changed later, I can make a new drone organ.

code/modules/organs/organ.dm
Mechoid: Should note the organs now remove their verbs if there's no other organs that add the verbs.

code/__defines/mobs.dm
```diff
@@ -209,7 +209,10 @@
 #define O_ACID     "acid gland"
 #define O_EGG      "egg sac"
 #define O_RESIN    "resin spinner"
-#define O_ALL list(O_STANDARD, O_MOUTH, O_CELL, O_PLASMA, O_HIVE, O_NUTRIENT, O_STRATA, O_RESPONSE, O_GBLADDER, O_POLYP, O_ANCHOR, O_ACID, O_EGG, O_RESIN)
+#define O_AREJECT  "immune hub"
+#define O_VENTC    "morphoplastic node"
+#define O_VRLINK   "virtual node"
```
Anewbe: Oh, right, should this also be in the list below?

Anewbe:
Bump?

PR #5802, opened 2019-01-01
Dragging mobs now causes slowdown

Neerti:
Dragging mobs behind you now applies a variable amount of slowdown. This is done to stop people from dragging big and heavy people to speed them up, particularly as a tiny weak bird.

Specifically, when calculating your own mob's 'item slowdown', if you are dragging someone else, your item slowdown will be set to their item slowdown, if it is higher than yours (e.g. if you have 0.5 item slowdown, and they have 3, if you drag them then you get 3).

If they have zero item slowdown, you will receive 1 slowdown as a floor (this is required for the next step due to multiplication).

Afterwards, the item slowdown is modified depending on your species. Species like Unathi get their slowdown from items reduced significantly, where as species like Teshari get it increased a lot. This allows big stronk things to drag people easily, and for tiny birds to struggle doing so, in a mechanical way.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_movement.dm
+++ b/code/modules/mob/living/carbon/human/human_movement.dm
@@ -1,8 +1,8 @@
+#define HUMAN_LOWEST_SLOWDOWN -3
+
 /mob/living/carbon/human/movement_delay()
 
 	var/tally = 0
-
-	var/item_tally = 0
 
 	if(species.slowdown)
 		tally = species.slowdown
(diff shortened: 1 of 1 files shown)


SpadesNeil:
While I know this is being done for a good reason, as it has basically been an exploit to drag instead of grab, I am still upset to find that zippy bois can no longer give me a speed boost by pulling me along.

SpadesNeil:
Come to think of it, anything with wheels on it should not cause a slowdown.

This would give a new use to the neglected trash crate.

Neerti:
This does not change dragging objects like roller-beds.

PR #5803, opened 2019-01-01
Guns sounds refactor and changes

SpadesNeil:
Give me a minute to fill in the description here with all the relevant changes.


SpadesNeil:
Of course, it would help if I actually changed the file names like I thought I did last night.

I fell asleep at my keyboard, shut up. I finished this when I was dreaming. (Give me a minute. :U)

SpadesNeil:
Compiles on my end because my .rsc file is derp

code/modules/projectiles/gun.dm
```diff
@@ -636,11 +633,18 @@
 	return launched
 
 /obj/item/weapon/gun/proc/play_fire_sound(var/mob/user, var/obj/item/projectile/P)
-	var/shot_sound = (istype(P) && P.fire_sound)? P.fire_sound : fire_sound
```
Atermonera: `var/shot_sound = fire_sound ? fire_sound : ((istype(P) && P.fire_sound)? P.fire_sound : [generic gunshot sound])`
Copypasta bad

SpadesNeil: Welp, it's not copypasta anymore.

SpadesNeil: Actually, on second thought, give me a minute. This line just pisses me off the more I look at it.

code/modules/projectiles/guns/projectile/pistol.dm
```diff
@@ -132,6 +132,7 @@
 	item_state = "deagle"
 	force = 14.0
 	caliber = ".44"
+	fire_sound = 'sound/weapons/Gunshot_deagle.ogg'
```
SpadesNeil: w h o o p s

almost forgot that

code/modules/projectiles/gun.dm
Atermonera: Use the provided line in place of the highlighted line, and remove the `if(fire_sound) else` block, you just need the one piece of
> if(silenced)
> > [quiet sound]
> else
> > [regular sound]

SpadesNeil: Wait, what? _Why_ would I do that?

SpadesNeil:
```if(fire_sound) // If the gun has its own fire-sound, use that.
		if(silenced)
			playsound(user, fire_sound, 10, 1)
		else
			playsound(user, fire_sound, 50, 1)
	else // If it has no special sound, default to the sound that the projectile uses.
		if(P.fire_sound) // Catches possible runtime where fire_sound is null for some reason.
			if(silenced)
				playsound(user, P.fire_sound, 10, 1)
			else
				playsound(user, P.fire_sound, 50, 1)```

The if/else block is the reason this PR exists. If the **gun** has a fire sound, **use that sound**. If that var is null, it will go to the else part, and use the projectile sound.

Comment (SpadesNeil) [reply code/modules/projectiles/gun.dm]: Okay now I get what you're saying after a Discord convo. I got totally lost.

**Comment by SpadesNeil** [comment]
My choice not to add a default sound is deliberate. I considered at least adding a debug log but decided not to.

One, somebody might want to make a soundless gun on purpose.
Two, a gun not having any sound is going to be pretty fucking obvious and quickly patched if it's not intentional.

File: code/modules/projectiles/gun.dm [hunk 3]
```diff
@@ -636,7 +633,13 @@
 	return launched
 
 /obj/item/weapon/gun/proc/play_fire_sound(var/mob/user, var/obj/item/projectile/P)
-	var/shot_sound = (istype(P) && P.fire_sound)? P.fire_sound : fire_sound
+	var/shot_sound = fire_sound
+
+	if(!shot_sound && istype(P) && P.fire_sound) // If the gun didn't have a fire_sound, but the projectile does...
+		shot_sound = P.fire_sound
+	else
```
Atermonera: else catches:
> if the gun has a fire sound
> if the projectile doesn't exist
> if the projectile doesn't have a fire sound

You probably want an
> if(!shot_sound)
> > return

SpadesNeil: Yeah that's probably a bit better.

PR #5804, opened 2019-01-02
IC News update for 1st Jan

Cerebulon:
Just adds the new article to the codex on behalf of elgeon who can't branches.


elgeonmb:
i can branches my brain is just rotten right now

PR #5822, opened 2019-01-10
This literally can not work without runtime maploading. Yay.

kevinz000:
Just close this if y'all don't use TGM, was sure you guys did but I might be wrong so


Anewbe:
Leave PR names as they are when you close them, please.

PR #5823, opened 2019-01-10
Egun stun beams are weaker

Anewbe:
In an effort to make eguns not just Taser+:tm:, decreases the stun done by an egun beam to 30, which means that it should now take 4 egun shots to down someone, compared to the taser's 3.


elgeonmb:
good now remove weapon cells

Anewbe:
No.

SpadesNeil:
This makes sense. :+1:

PR #5825, opened 2019-01-10
Adds ability to put butts in boxes

Neerti:
Specifically cigarette butts in cigarette boxes, and cigar butts in cigar boxes.


Atermonera:
But can you photocopy them

PR #5844, opened 2019-01-15
Adds new cargo crates(Correctly now!)

krumpmeboyz:
Adds three new crates: A cat crate, a mini crate, and a plushie crate. **No longer adds the hydroponics and misc crate directories to the polaris main folder**, as it did originally because I'm incompetent.

Yay!


krumpmeboyz:
God dammit.

PR #5849, opened 2019-01-15
Fixes a few mob AI bugs

### LINKED ISSUE #5834: AI mercs are too forgiving
When a mob enters a merc's sight, the merc is supposed to yell at them to go away, and then fight them if they don't go away or come back too soon. The second part does not appear to function correctly, so they yell at the mob every time, even if they were gone for just a second. This makes them super-easy to cheese.

For the merc mobs to open fire immediately if the other mob came back too soon.

#### What actually happened
Get out of here ~~stalker~~.

#### Steps to reproduce
Spawn BSD PoI.
Run up to merc.
Hear them yell.
Run out of sight.
Run into sight again very fast.
Hear them yell.

#### Additional info:
Server revision: master - 2019-01-07
81c3036140d2f565d57f039fe32dc71fcc8a6a26 

Anewbe:
Should fix #5834 
Should fix retaliate not causing mobs to retarget until their first target is dead
Should also stop all mobs from attacking SSD players, and most mobs from attacking unconscious players, meaning you can once again play dead.

### DIFF
--- a/code/modules/ai/ai_holder_targeting.dm
+++ b/code/modules/ai/ai_holder_targeting.dm
@@ -238,7 +238,7 @@
 /datum/ai_holder/proc/on_attacked(atom/movable/AM)
 	if(isliving(AM))
 		var/mob/living/L = AM
-		if(!(L in attackers))
+		if(!(L.name in attackers))
 			attackers |= L.name
 			last_conflict_time = world.time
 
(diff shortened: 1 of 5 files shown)


code/modules/ai/ai_holder_targeting.dm
```diff
@@ -220,10 +236,11 @@
 
 // Sets a few vars so mobs that threaten will react faster to an attacker or someone who attacked them before.
 /datum/ai_holder/proc/on_attacked(atom/movable/AM)
-	last_conflict_time = world.time
 	if(isliving(AM))
 		var/mob/living/L = AM
-		attackers |= L.name
+		if(!(L in attackers))
```
Neerti: Not sure this is the correct check since ``attackers`` contains mob names and not mob instances.

PR #5851, opened 2019-01-16
Fix Technomancer Refunding

Mechoid:
Tin. Gives the catalog a var that controls which Z-number it can refund on.


Neerti:
I think it'd be better to use the ``using_map.admin_zlevels`` or whatever that var is called since it avoids hardcoding in the z-level in case it changes again (which is what originally caused the bug in the first place)

PR #5837, opened 2019-01-13
Kraut Space Magic™ caseless gun, and other gun changes

SpadesNeil:
This is probably going to need a changelog but I forget how to do that. Somebody guide me through that later.

# So first, the stuff that requires explaination

- 5.45 and 7.62 are now .223 and .308 respectively.

Why am I using caliber as opposed to millimeter? To avoid confusion. What if we _want_ to have 5.45mm in the code one day? What about the SVD that uses 7.62x54R? How confusing would it be to, at a glance, simultaneously have a gun chambered in 5.56? And furthermore, we **already** have a problem where every rifle that has 7.62mm bullets is assumed to be the same caliber, _even though there's at least 3 different kinds of 7.62mm, **none** of which are compatible!_ When I say .308, future devs and players know **exactly** what kind of bullet this gun is supposed to take. Not AK ammunition, not Mosin ammunition, but the commonplace .308 cartridge.

- SVD now uses "high velocity" 7.62x54mmR ammunition.

"High velocity" is what I am calling sniper ammo, and I am calling it that because "sniper ammo" is hitscan. This makes the SVD no longer absolutely useless as a scoped gun. This feature, which isn't new, but has so far been unique to the PTR, means that we have a chance to test it in practice and see how it works. Maybe the DMR should use hitscan ammo in the future.

- Added some comments under bullets.dm to give devs an idea of what bullet they should use for their new cartridges.

It was kind of stupid that we had 7.62 and 5.45 projectiles, when players never ever actually see the name of those projectiles anywhere at any time. I altered this so that instead a dev can use a small set of only three or four projectiles for a _wide_ variety of ammo cartridges, but they consistently know what to use for whichever gun they are making. Why would I add a new projectile for 7.62x54mmR when the one for .308 does exactly the same damage? I also added some small caliber stuff that could be used later.

- Added the "prototype caseless rifle"

Like the advanced SMG, research can build them, but they don't _spawn_ with ammo. You can however produce both stun rounds or live bullets for it in 30 round magazines. This gun is semi-automatic, and has about the power of a medium level pistol. **Note I still need to decide how the gun is balanced, so please lend some opinion on that or I am going to wing it.**

## And now for everything else.

- Add: "ratshot" projectile for future use. Basically harmless to people, but kills pesky mice from across the bar.
- Add: New Semi-automatic rifle, which is a semi-auto only version of the STS-35 automatic rifle (formerly assault rifle). Might make this civilian version illegally modifiable in future.
[...]

PR #5873, opened 2019-01-20
5.45 is now .223

SpadesNeil:


SpadesNeil:
This doesn't break anything. I tested it thoroughly.

PR #5874, opened 2019-01-20
7.62 is now .308

SpadesNeil:


SpadesNeil:
This doesn't break anything. I tested it thoroughly.

It's also not unnecessary. Developers keep thinking that 7.62mm is the same in every gun they think uses 7.62, from AK variants to sniper rifles. This ends that confusion and had positive feedback in the original PR.

Atermonera:
Or we can separate the different 7.62 calibers into spacefuture calibers, because 500 years of science, engineering, and culture makes the persistence of modern calibers kinda questionable.

SpadesNeil:
I would be okay with that. I just don't want one caliber referencing _three rifle calibers and one pistol caliber_.

PR #5876, opened 2019-01-20
10mm is now 10mm Auto

SpadesNeil:
https://github.com/PolarisSS13/Polaris/pull/5837


SpadesNeil:
This one is probably unnecessary but it will **not** break things.

PR #5877, opened 2019-01-20
Unfucks Vintage Rifle

SpadesNeil:
Also removes redundant code, and this gun can no longer speedload. I'm going to slap whoever copypasted this blindly.

Also this really shouldn't be 7.62/.308 but that's pretty firmly out of scope ¯|-(ツ)-|¯

### DIFF
--- a/code/modules/projectiles/guns/projectile/boltaction.dm
+++ b/code/modules/projectiles/guns/projectile/boltaction.dm
@@ -55,14 +55,11 @@
 	icon_state = "leveraction"
 	max_shells = 5
 	caliber = "7.62mm"
-	origin_tech = list(TECH_COMBAT = 1)// Old as shit rifle doesn't have very good tech.
-	ammo_type = /obj/item/ammo_casing/a762
-	load_method = SINGLE_CASING|SPEEDLOADER
-	action_sound = 'sound/weapons/riflebolt.ogg'
+	load_method = SINGLE_CASING
 
 /obj/item/weapon/gun/projectile/shotgun/pump/rifle/lever/vintage
 	name = "vintage repeater"
 	desc = "An iconic manually operated lever action rifle, offering adequate stopping power due to it's still powerful cartridge while at the same time having a rather respectable firing rate due to it's mechanism. It is very probable this is a replica instead of a museum piece, but rifles of this pattern still see usage as colonist guns in some far off regions. Uses 7,62mm ammo."
-	item_state = "levercarabine"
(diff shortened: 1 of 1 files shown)


code/modules/projectiles/guns/projectile/boltaction.dm
```diff
@@ -55,14 +55,11 @@
...
-	load_method = SINGLE_CASING|SPEEDLOADER
-	action_sound = 'sound/weapons/riflebolt.ogg'
+	load_method = SINGLE_CASING
 
 /obj/item/weapon/gun/projectile/shotgun/pump/rifle/lever/vintage
 	name = "vintage repeater"
 	desc = "An iconic manually operated lever action rifle, offering adequate stopping power due to it's still powerful cartridge while at the same time having a rather respectable firing rate due to it's mechanism. It is very probable this is a replica instead of a museum piece, but rifles of this pattern still see usage as colonist guns in some far off regions. Uses 7,62mm ammo."
-	item_state = "levercarabine"
-	icon_state = "levercarabine"
+	item_state = "levercarbine"
```
Anewbe: This is going to break the sprite.

code/modules/projectiles/guns/projectile/boltaction.dm
```diff
@@ -55,14 +55,11 @@
...
+	icon_state = "levercarbine"
```
Anewbe: As is this.

elgeonmb: i think this is the sprite being un-broken

Anewbe: I went into the live files and checked, it's specifically spelled as `levercarabine`. If this is changed, and that file is not, it will break the sprite.

elgeonmb: god bless ss13, never change

SpadesNeil: The icon is _already_ broken, though. This fixed it when I tested it. Hang on. What the hell is going on here?

code/modules/projectiles/guns/projectile/boltaction.dm
SpadesNeil: The icon is already broken, though. This fixed it when I tested it. Hang on. What the hell is going on here?

SpadesNeil: Apparently I'm going crazy.

PR #5893, opened 2019-01-22
Additional Hair

Woodratt:
- Adds two new hairstyles New You and Parted Alt
- Hair is now in a semi-alphabetical order (some hairstyles were renamed). Facial hair was not touched and is still out of alphabetical order.

Tested, appears to work proper like.


code/modules/mob/new_player/sprite_accessories.dm
```diff
@@ -49,174 +49,41 @@
 	var/icon_add = 'icons/mob/human_face.dmi'
 	var/flags
 
+	eighties
```
Atermonera: Does this change the default hair option from bald?

Woodratt: Honestly, no idea. Didn't test for that. Since I assumed that was handled differently.

PR #5901, opened 2019-01-24
Unclogs powersink drains

Atermonera:
Powersinks drain first, damnit!

They were broken by whoever added the Object controller, because adding an item to that ticks the DF_ISPROCESSING flag on datum_flags. When you try to add an item to PROCESSING_POWER_ITEMS, it also ticks that flag. In either case, if the flag is already ticked, it fails to add the item to the appropriate list. Because powersinks are added to Obj first, power_items fails and the powersink drains no power. That said, if they were flipped, it'd still break, because then it wouldn't unset drained_this_tick and thereby only drain once.

Also, tested.


Atermonera:
*angry amend-last-commit button noises*

polaris.dme
```diff
@@ -2614,7 +2614,7 @@
 #include "code\ZAS\Zone.dm"
 #include "interface\interface.dm"
 #include "interface\skin.dmf"
-#include "maps\southern_cross\southern_cross.dm"
+#include "maps\example\example.dm"
```
Anewbe: Wrong map.

code/controllers/subsystems/machines.dm
```diff
@@ -155,6 +155,8 @@ SUBSYSTEM_DEF(machines)
 		src.current_run = global.processing_power_items.Copy()
 
 	var/list/current_run = src.current_run
+	world << "Processing power objects: [english_list(current_run)]"
```
Anewbe: Debug code

code/controllers/subsystems/machines.dm
```diff
@@ -155,6 +155,8 @@ SUBSYSTEM_DEF(machines)
...
+	world << "Actual power objects: [english_list(global.processing_power_items)]"
```
Anewbe: Debug code

code/game/objects/items/devices/powersink.dm
```diff
@@ -84,6 +85,7 @@
 			STOP_PROCESSING_POWER_OBJECT(src)
 
 /obj/item/device/powersink/pwr_drain()
+	world << "pwr_drain called"
```
Anewbe: Debug code

code/game/objects/items/devices/powersink.dm
```diff
@@ -95,11 +97,13 @@
 
 	if(!PN)
 		return 1
+	world << "draining"
```
Anewbe: Debug code

code/game/objects/items/devices/powersink.dm
```diff
@@ -95,11 +97,13 @@
 
 	if(!PN)
 		return 1
+	world << "draining"
 
 	set_light(12)
 	PN.trigger_warning()
 	// found a powernet, so drain up to max power from it
 	drained = PN.draw_power(drain_rate)
+	world << "drained [drained]"
```
Anewbe: Debug code

PR #5907, opened 2019-01-27
Flash rebalance.

Mechoid:
Tweaks:
- Flashes again run a non-negligible risk of burning out through consecutive use.
- Burned out flashes with repairs enabled can have repairs attempted [30% success] by using a screwdriver on them over the course of approximately fourty to sixty seconds. This is not on default flashes due to the sheer number available to Security at round-start. [22 not including on-spawn equipment, 7 in mounts or deployables.]
- Flashes now run on a non-removable internal cell in addition to their capacitor cooldown. 12 charges.
- Cyborg flashes do not burn out or limit use, but instead consume larger and larger amounts of power with each use over the limit [Around 12 flashes began eating percentages of a standard cell.]


Anewbe:
Why both the bulb burnout **and** the battery thing? Seems a bit much.

zekesulastin:
... I am somewhat surprised flashes are still being used enough to need to be nerfed **more**, but then I usually avoid redshirts.

FenodyreeAv:
Why is this necessary.

Mechoid:
Currently flashes have, at max, a 5% chance to break on their 10th use, the maximum amount of uses they can have before requiring a cooldown. I have never seen this happen since it was changed, when flashes would previously burn out after 3 or 4 usually.

At time of writing, this allows for 16 flashes - 4 people downed, per cell - if you don't over-use it and burn it out, requiring repair. Burn-out in testing took usually between six to eight flashes consecutively.

zekesulastin:
Oh hey, I remembered why I never noticed flashes before - the last few times people have tried to use them on me I've been flash immune thanks to job gear anyways and the redshirt didn't think to pull out their baton before I got out of range.  I wonder who got dunked.

Mechoid:
..Wait what is that

FenodyreeAv:
It's entirely possible i'm reading this wrong.
With every use of the flash there's an additional 3% chance it will break.
So, first use 3%, second 6%, fourth use 12%.

So, lets say I'm stunning two people, (or more commonly, the same person a second time)
That's 6 uses, with a total chance of 54.9% of burning out.
The main purpose of the flash is dealing with unruly prisoners, I'm going to break my flash multiple times every single round with these numbers.

Unless, after every arrest I flash myself until it breaks, then fix it. Which is dumb.

SpadesNeil:
I think this is a terrible idea that was only made because somebody (OP) got roflstomped.

It's hard enough to use flashes _optimally_. This nerfs them to the point of uselessness compared to other tools in Security...

Having a battery makes sense, but the rest of this is absurd.

elgeonmb:
"someone (OP) got roflstomped"

he was specifically asked by staff to make this pr but nice try

Anewbe:
~~Then he shoulda marked it Staff Request, dangit.~~

SpadesNeil:
>he was specifically asked by staff to make this pr but nice try

_oh_

I redact the statement.

Wickedtemp:
If it was asked by staff, then we now know the true reason.

Somebody in _staff_ got rofltstomped.

Anewbe:
See, that was my assumption too, but i asked basically everyone and they all responded negative. ~~Maybe sometimes staff finds the need to adjust things that they've been paying attention to for a long time, who knows.~~ Anyway, this PR is closed, we should all stop commenting on it.

PR #5910, opened 2019-01-28
Pilot consoles require pilot access

Anewbe:
The pilot ships being faster when driven was supposed to be a reason to play Pilot, not a reason to break into the cockpit.

Staff Request:   Mangled/(Mangled)(JMP): legit say I said "I cant be bothered yelling at people breaking into the cockpit anymore, ID lock that shit but make it emaggable" Screenshot this, right now and post it in the git thingie.


MisterLayne:
Is it emaggable?

Anewbe:
It is emaggable, just double checked.

PR #5913, opened 2019-01-29
Readds access restrictions to supply computer

Atermonera:
Because apparently we're not allowed to have nice things

### DIFF
--- a/code/game/machinery/computer/supply.dm
+++ b/code/game/machinery/computer/supply.dm
@@ -31,6 +31,8 @@
 /obj/machinery/computer/supplycomp/attack_hand(var/mob/user as mob)
 	if(..())
 		return
+	if(!allowed(user))
+		return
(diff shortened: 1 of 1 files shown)


mistyLuminescence:
Isn't that what the access-locked door to the cargo office is for?

Atermonera:
Since I removed the access lock, I've heard repeated complaints about people breaking in to use it, making cargo irrelevant.

FenodyreeAv:
*Making use of cargo during lowpop rounds when there is no cargo, HoP, or Captain.

Naidh:
Short of Engineering doing their thing when there isn't actually Cargo or someone capable of fulfilling that post present (e.g. Site Manager, Head of Personnel, anyone with native Cargo access, synthetics), I haven't really seen any break-ins necessitating this from my side of things.

PR #5919, opened 2019-01-31
Implements is_cable_coil() in cable coils.

RobotAlice:
One of the holdouts from the conversion of turning tool status into procs. This being missing breaks some downstream stuff, and breaks part of constructable_frame.dm too.

### DIFF
--- a/code/modules/power/cable.dm
+++ b/code/modules/power/cable.dm
@@ -154,6 +154,10 @@ var/list/possible_cable_coil_colours = list(
 // General procedures
 ///////////////////////////////////
 
+//Finally implements is_cable_coil():
+/obj/item/stack/cable_coil/is_cable_coil()
+	return TRUE
+
 //If underfloor, hide the cable
 /obj/structure/cable/hide(var/i)
 	if(istype(loc, /turf))


Anewbe:
The reason this wasn't implemented here is because in a lot of places you absolutely need the thing returning true to actually be cable, or the code inside the if check breaks, so generally it's better to use an istype.

Atermonera:
Furthermore, the intent of the is_X() procs are that they can be called on any object, and if you're just defining it on the thing that's supposed to return true, that's missing the point of the refactor

RobotAlice:
[I'm not touching the previous declaration common to all /obj/items/ to begin with, so I don't know where you're coming from.](https://github.com/PolarisSS13/Polaris/blob/d704e874c7503016512f310982121457c3078dee/code/game/objects/items.dm#L846-L847) All I'm doing is overriding the proc for cable coils: note in my code, I don't need to declare the proc via explicitly adding /proc/.

Neerti:
@Atermonera Is this good to go then?

Anewbe:
I'm not sure how or why anyone downstream would be using this? It was built as vestigial/TODO code, because, as far as I could figure out at the time, you'd still need a typecheck in every place this would be reasonably used.
Can you provide an example or a link to where it's being used?

Also, where specifically does it break constructable_frame.dm?

Anewbe:
@RobotAlice Any comments?

Anewbe:
This hasn't been responded to in almost two weeks. Closing it.

PR #5921, opened 2019-02-01
Fixes compile error.

Neerti:
I put the CLAMP()s to it.

Still confused why Travis didn't catch it in the reagent PR.


Neerti:
Gonna self merge because ~~I'm a tyrant~~ master not compiling is A Bad Thing :tm:

Anewbe:
I was just about to ask. :(

PR #5935, opened 2019-02-06
Nerfs Sonic Speed Wheelchair

TheFurryFeline:
Porting a few things from CHOMPstation, essentially to make it so that you're not going at Mach 3.

### DIFF
--- a/code/game/objects/structures/stool_bed_chair_nest/wheelchair.dm
+++ b/code/game/objects/structures/stool_bed_chair_nest/wheelchair.dm
@@ -5,6 +5,7 @@
 	anchored = 0
 	buckle_movable = 1
 
+	var/move_delay = null
 	var/driving = 0
 	var/mob/living/pulling = null
 	var/bloodiness
@@ -29,6 +30,13 @@
 
 /obj/structure/bed/chair/wheelchair/relaymove(mob/user, direction)
 	// Redundant check?
+ 	
+	var/calculated_move_delay
+	calculated_move_delay += 2 //TheFurryFeline: nerfs speed so you don't go like Sonic. >W>
+
+ 	if(world.time < move_delay)
+		return
+
 	if(user.stat || user.stunned || user.weakened || user.paralysis || user.lying || user.restrained())
 		if(user==pulling)
 			pulling = null
@@ -56,6 +64,11 @@
 	if(pulling && has_buckled_mobs() && (user in buckled_mobs))
 		user << "<span class='warning'>You cannot drive while being pushed.</span>"
 		return
+
+
+ 	move_delay = world.time
+	move_delay += calculated_move_delay
+
 
 	// Let's roll
 	driving = 1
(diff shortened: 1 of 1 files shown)


MisterLayne:
This also stops you from sumo-slamming the person against doors and windows.

Anewbe:
I don't see any code that'd do that, @MisterLayne ?

Also, out of curiosity, @TheFurryFeline , why specifically are you doing this as implemented?

TheFurryFeline:
Ack, sorry for being silent up to now, @Anewbe. Basically, it's pretty much what's on the tin, makes it so that you're not going like a.... Well, maniac on wheels. Ideally,  giving a few thoughts to it now, this would prevent characters from outrunning virtually anyone that are running.

TheFurryFeline:
Additionally, I'm wanting this to also be implemented on VOREstation but with the filename required, I'm over up here so it'll come out with a mirror if/when this is merged.

Anewbe:
Sorry for the vagueness on my part; my more specific question was why in that specific part of the code, rather than something like movement?

TheFurryFeline:
Ah, if I placed it in an odd area, then sorry. I pretty much just copy-pasted from another codebase into this, keeping the lines in their same places. I'm still rather new to coding.

Anewbe:
No worries.
Do you know how to test your code, at least?

TheFurryFeline:
I've already seen this code in action on CHOMPstation, actually, and with it, it has a noticeable difference in movement speed for wheelchairs and on VOREstation, it's still supersonic speed without this being part of the file.

Anewbe:
I don't disbelieve you, but I'd rather you actually test this code, as it applies specifically to Polaris.
I've got some concerns about the pushing wheelchairs as this is implemented, and I'd much rather hear from you if that does something weird, than have to find out for myself in a few months when I've forgotten about this PR, if that makes sense?

TheFurryFeline:
Fair enough, I'll clone the codebase and then open up a test client, try out the chair. Then change the file and compare the speed rate.

Anewbe:
Thank you

TheFurryFeline:
Huh..... Took a while for me to get around to it but without this in the code, it seems that wheelchairs aren't super fast here. Unless the tickrate's been changed here, as well?

Anewbe:
Tick rate is variable, ours hasn't been touched in a while.

TheFurryFeline:
Alright, got the .dme open, putting the lines in to test to see if there's any issues.

TheFurryFeline:
And again sorry for the delays, but it seems to work nicely. No issues that I noticed. :> As for pushing the wheelchairs, seems to be fine as well. Spawned human in, totally didn't bump 'em into walls and windows, and speed was normal as well as I could tell.

code/game/objects/structures/stool_bed_chair_nest/wheelchair.dm
```diff
@@ -29,6 +30,13 @@
 
 /obj/structure/bed/chair/wheelchair/relaymove(mob/user, direction)
 	// Redundant check?
+ 	
+	var/calculated_move_delay
+	calculated_move_delay += 2 //TheFurryFeline: nerfs speed so you don't go like Sonic. >W>
```
SpadesNeil: **>W>**

SpadesNeil:
Seriously though this is a good tweak.

PR #5940, opened 2019-02-08
Fix a div/0.

Mechoid:
Someone's an idiot, and it's probably me.


Atermonera:
This hasn't actually fixed the divide by 0, which occurs when the denominator, the thing after `/`, is 0. You're enforcing the numerator to be non-0, when it should be `if(charge_cost)`

Anewbe:
I'm bad at math, leave me alone.

PR #5942, opened 2019-02-08
Headgibbing is a config thing

Anewbe:
As discussed in or around the Staff Meeting

### DIFF
--- a/code/modules/organs/subtypes/standard.dm
+++ b/code/modules/organs/subtypes/standard.dm
@@ -258,7 +258,7 @@
 	joint = "jaw"
 	amputation_point = "neck"
 	gendered_icon = 1
-	cannot_gib = 1
+	cannot_gib = TRUE
 	encased = "skull"
 	base_miss_chance = 40
 	var/can_intake_reagents = 1
@@ -267,6 +267,11 @@
 	throwforce = 7
 
 	var/eye_icon_location = 'icons/mob/human_face.dmi'
+
+/obj/item/organ/external/head/Initialize()
+	if(config.allow_headgibs)
+		cannot_gib = FALSE
+	return ..()
 
 /obj/item/organ/external/head/robotize(var/company, var/skip_prosthetics, var/keep_organs)
 	return ..(company, skip_prosthetics, 1)
--- a/code/controllers/configuration.dm
+++ b/code/controllers/configuration.dm
@@ -130,6 +130,7 @@ var/list/gamemode_cache = list()
 	var/organ_regeneration_multiplier = 1
 	var/organs_decay
 	var/default_brain_health = 400
+	var/allow_headgibs = FALSE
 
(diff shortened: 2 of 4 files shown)


code/modules/organs/subtypes/standard.dm
```diff
@@ -268,6 +268,11 @@
 
 	var/eye_icon_location = 'icons/mob/human_face.dmi'
 
+/obj/item/organ/external/head/Initialize()
+	..()
```
Neerti: Generally ``Initialize()`` needs to return a thing and so usually at the bottom is ``return ..()``, similar to ``Destroy()``.

Anewbe: Righto.

Wickedtemp:
Neato, what's it do?

Anewbe:
If the config is turned on, damage that would give a normal limb can, when applied to the head, give the head, as it was in days long past.

Wickedtemp:
Whats the damage threshold for that? 60?

Anewbe:
🤷‍♂️

zekesulastin:
Oh cool, this crap is back.

mistyLuminescence:
but
why

Anewbe:
Done in an effort to stop people from treating death, and especially armed antags, as a trivial slap on the wrist.
Reworked as a config to allow it to be easily toggled off and on where and when appropriate.

zekesulastin:
Here we go, then.  In b4 some antag gets rekt by an exosuit or stacking burn damage and the one doing the wrecking gets banned for their trouble.

Woodratt:
Far as I'm aware it'll just be a reset to how things basically were before we removed head gibbing. 

As for complaints, if you have 'em, point them at Rando.

zekesulastin:
Yeah, and I was there when we removed it too.  I'm somewhat concerned with all the other stuff that hasn't been posted yet since the notes have no details.

Probably a good thing I missed the meeting where this was all discussed because it probably would not have been polite.

mistyLuminescence:
Isn't this why decapitation exists, though, and the super difficult surgery to fix it? This reads like a bad D&D DM trying to get people to die forever instead of, like, actually playing the round with the character they want to play. 

"You find the patient and they are dead! Very dead!"
"Oof. Can I try and bring them back? I've got a defib and expertise in Medicine."
"No, uh, their head... fell off. It's across the room."
"Alright, let's reattach it and get to work!"
"Oh, uh, their head fell off *real bad*. You'll have to do some really tricky surgery to save them now."
"Let's try the surgery... Nat 20! How'd I do?"
"Actually, I changed my mind, their head just explodes. There's nothing to reattach."

The answer to "antag violence is ineffective" is... which? 
A) "encourage antags to actually confirm their kills instead of shooting someone into crit and running off"
or B) "have the victims' heads comically explode into a shower of gibs and grey matter and encourage them to go watch Netflix for an hour"
A... or B?

PR #5949, opened 2019-02-11
Tweaks Daxons

Mechoid:
After receiving feedback regarding the new chems I added previously, I've tweaked their effects / values to be more worth the cost.

Tweaks
Respiro- Cordra- Gastiro- and Hepanephro- Daxons increased in process rate from 0.2 to 0.3 per cycle.
Repair amount changed from 2 per unit to 4.

If not conflicting with other chemicals:
Respirodaxon lowers losebreath
Cordradaxon lowers oxygen damage, dealing with bloodloss
Gastirodaxon lowers toxin at 10 per unit, 2 more than carthatoline
Hepanephrodaxon lowers toxin at 12 per unit, 4 more than carthatoline

Skrell Anti-Rejection processes at 0.15 rather than 0.2.

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine.dm
@@ -566,6 +566,8 @@
 				H.vomit(1)
 			else if(H.nutrition > 30)
 				H.nutrition = max(0, H.nutrition - round(30 * removed))
+		else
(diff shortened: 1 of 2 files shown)


Wickedtemp:
The reagents they conflict with are other daxons, yeah? Peri, and the new ones.

Mechoid:
Correct, though not universally. Hepanephro can be used with respiro, and gastro can be used with cordra if I'm remembering properly.

PR #5962, opened 2019-02-15
Prevents people from joining with the same character

GeneriedJenelle:
Just a check on late spawn to prevent people from rejoining constantly as the same char.

image

### DIFF
--- a/code/modules/mob/new_player/new_player.dm
+++ b/code/modules/mob/new_player/new_player.dm
@@ -163,6 +163,13 @@
 		ViewManifest()
 
 	if(href_list["SelectedJob"])
+	
+		//Prevents people rejoining as same character.
+		for (var/mob/living/carbon/human/C in mob_list)
+			var/char_name = client.prefs.real_name
+			if(char_name == C.real_name)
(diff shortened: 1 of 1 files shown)


mistyLuminescence:
Sometimes people join as the wrong job, then cryo/elevator/gateway etc and rejoin as the correct job. Would this affect that, or is it specifically only for people who respawn a bunch for free stuff?

GeneriedJenelle:
> Sometimes people join as the wrong job, then cryo/elevator/gateway etc and rejoin as the correct job. Would this affect that, or is it specifically only for people who respawn a bunch for free stuff?

When someone cryos, they turn into a ghost. I do believe the ghost deletes itself when the person respawns, Either way, this code checks for human mobs so it shouldn't affect observers. I'll test this after work tonight.

`		for (var/mob/living/carbon/human/C in mob_list)`

GeneriedJenelle:
> Sometimes people join as the wrong job, then cryo/elevator/gateway etc and rejoin as the correct job. Would this affect that, or is it specifically only for people who respawn a bunch for free stuff?

Right. Tested this out. 

Works as intended, will not affect people who have cryo'ed, or get in the way if there's a ghost with the same name. All is good.

PR #5967, opened 2019-02-17
Fixes submap footprint miscalculations

Verkister:
Submap sanity checks will finally get properly oriented bounds to work with.


code/modules/maps/tg/map_template.dm
```diff
@@ -27,8 +27,12 @@
 /datum/map_template/proc/preload_size(path, orientation = 0)
 	var/bounds = SSmapping.maploader.load_map(file(path), 1, 1, 1, cropMap=FALSE, measureOnly=TRUE, orientation=orientation)
 	if(bounds)
-		width = bounds[MAP_MAXX] // Assumes all templates are rectangular, have a single Z level, and begin at 1,1,1
-		height = bounds[MAP_MAXY]
+		if(orientation & (EAST | WEST))
```
Atermonera: `if(orientation%180)`
Submaps were recently changed to use degrees rotated from north, instead of forcing the direction to a preset and then trying to figure out how much that actually rotated

PR #5970, opened 2019-02-18
Moves crashed medical shuttle PoI to deep caves

Neerti:
Been meaning to do this for awhile. Now the 'easy' side will be less dangerous as intended.


mistyLuminescence:
:+1: but why not just drop the amount of spiders in it? The OG PoI had a ton of spiders to balance out the good medical loot, and now it doesn't have good medical loot, so it doesn't really need a ton of spiders either.

PR #5975, opened 2019-02-20
Promethean tuning

Anewbe:
- Promethean regeneration stops working if the promethean is too hot, too cold, or too hungry (<50 nutrition). Being wet still stops regeneration and hurts them, as before.
- Promethean body temperature now tries to stabilize at "Room Temperature". In addition, their cold_level_2 threshold has been bumped up to 240, from 220. Hopefully this means that cold (especially from cryotoxin) will be a bit of a threat, rather than the non-issue it is now.
- Prometheans are no longer shock resistant.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/prometheans.dm
+++ b/code/modules/mob/living/carbon/human/species/station/prometheans.dm
@@ -58,17 +58,16 @@ var/datum/species/shapeshifter/promethean/prometheans
 	cloning_modifier = /datum/modifier/cloning_sickness/promethean
 
 	cold_level_1 = 280 //Default 260 - Lower is better
-	cold_level_2 = 220 //Default 200
+	cold_level_2 = 240 //Default 200
 	cold_level_3 = 130 //Default 120
 
 	heat_level_1 = 320 //Default 360
 	heat_level_2 = 370 //Default 400
 	heat_level_3 = 600 //Default 1000
 
-	body_temperature =      310.15
+	body_temperature = T20C	// Room temperature
 
-	siemens_coefficient =   0.4
-	rarity_value =          5
+	rarity_value = 5
 
 	genders = list(MALE, FEMALE, NEUTER, PLURAL)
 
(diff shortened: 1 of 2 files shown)


Dunmeritude:
I really feel like the shock resistant bit should be nixed. There's a reason security carries xeno tasers. They're going to be made useless if you can drop a Promethean just as easily with your perun. Two goddamn shots from the X-Taser is /instant crit/.

...Furthermore cryotoxin already /is/ a serious danger for Prometheans. I've seen it kill a good number of prom explorers. Prometheans already have heavily reduced healing if they're hungry at /all/, and can't regenerate. This feels like a lot of aggravating nerfs to aspects of Promethean play that are already treacherous as it is-- Not 'tweaking' or '~rebalancing~'.

Schnayy:
I took a cryotox spider to the face twice in one sitting as a Promethean and got right back up so I think that's not working as intended.

Anewbe:
In regards to the xeno tasers, I've head enough complaints about `Security Inventory Tetris` to last me several hundred lifetimes. They already have to carry around a bunch of other gear, having to carry a second gun specifically to deal with Prometheans is a pain.

As for cryotoxin, it's not about what _might_ be a threat to them, it's about what they very handily survive nearly every time. I've seen plenty of prometheans get cryotoxin injected into them and walk it off, as is their usual MO. I've also seen a Promethean get injected with about 50 units of the stuff, and have to literally be told by staff to care about them being chilly, because they heal faster than it damages them. That's a problem for something that's supposedly weak to cold.

elgeonmb:
i continue to be strongly of the belief that standard kit (bullets and lasers) should be decently effective on everyone, so, the taser vuln is good. everything else is basically implementing behavior i assumed was already a Thing so, like, you know. Constant regen is a _hell_ of a thing.

Dunmeritude:
Tasers already /are/ decently effective. This is just making the xeno tasers useless for anyone but Xenobio, and pointless to be stocked in security to begin with. They're there for a reason. People should use them instead of whining about "BUT MUH VALIDS" because they can't two-shot a slimebag or use their .45 to kneecap the damn thing.

PR #5978, opened 2019-02-23
Unpicks some mining spaghetti

Screemonster:
Something of an attempt to disentangle the spaghetticode surrounding mineral turfs.

Moved a bunch of the junk from the horrible nested if statement in attackby() into separate procs. With any luck these can be unpicked further.

Replaced the snowflaked emitter-beam code with a generic projectile-mining effect that uses the same "mining depth" as pickaxes - allows for the addition of generic ranged mining tools if someone wants to do that. Emitters are given a mining depth of 70 so it'll take 3 shots to mine a turf as before.

Gives some mining tools (namely, advanced and diamond drills and the sonic jackhammer) a "destroy artefacts" property. Guaranteed destruction of any artefact finds, but at least you don't have to deal with rocky debris or getting assblasted with rads completely at random provided you geared up with an appropriate tool. (same as a ripley drill basically, it just drills and doesn't care what it's drilling)

### DIFF
--- a/code/modules/mining/mine_items.dm
+++ b/code/modules/mining/mine_items.dm
@@ -56,7 +56,6 @@
 	origin_tech = list(TECH_MATERIAL = 3, TECH_POWER = 2, TECH_ENGINEERING = 2)
 	desc = "Cracks rocks with sonic blasts, perfect for killing cave lizards."
 	drill_verb = "hammering"
-	destroy_artefacts = TRUE
 
 /obj/item/weapon/pickaxe/gold
 	name = "golden pickaxe"
--- a/code/modules/mining/mine_items.dm
+++ b/code/modules/mining/mine_items.dm
@@ -47,7 +47,6 @@
 	origin_tech = list(TECH_MATERIAL = 2, TECH_POWER = 3, TECH_ENGINEERING = 2)
 	desc = "Yours is the drill that will pierce through the rock walls."
 	drill_verb = "drilling"
-	destroy_artefacts = TRUE
 
 /obj/item/weapon/pickaxe/jackhammer
 	name = "sonic jackhammer"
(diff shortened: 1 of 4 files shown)


code/modules/mining/mine_items.dm
```diff
@@ -29,6 +29,7 @@
 	sharp = 1
 
 	var/excavation_amount = 200
+	var/destroy_artefacts = FALSE // some mining tools will just trash artefacts completely. At least you won't get assblasted by rads.
```
Anewbe: Keep comments serious and clean, please.

code/modules/mining/mine_items.dm
```diff
@@ -29,6 +29,7 @@
 	sharp = 1
 
 	var/excavation_amount = 200
+	var/destroy_artefacts = FALSE // some mining tools will destroy artefacts completely while avoiding side-effects.
```
Anewbe: As per the March 3rd Staff Meeting, we like everything in this PR other than this particular feature. If you'd rather leave it in, at least disable it on the assorted drills.

Screemonster: It's disabled by default on the standard drill (not sure at what point regular pickaxes became drills) - which things, if anything, would you prefer it be enabled on?

Anewbe: Didn't mean to resolve that. We'd rather it not be active on anything, for the time being.

Screemonster: Alrighty. Probably gonna enable it on some things downstream because it was a major headache for expeditions given that we have a lot of underground exploration maps.

Anewbe: Makes perfectly good sense, in your situation, which is why it makes sense to me to leave the code proper in.

PR #5981, opened 2019-02-27
Handtele no longer relies on teleporter being mapped in in very specific way

Heroman3003:
Seriously, that logic for finding the computer is dumb.


mistyLuminescence:
While we're at it, can we have the hand teleporter display the teleporter destination? "8029 (Inactive)" means... very little.

Mechoid:
8029 in that case is actually the Teleporter's ID. You can change it with a multitool IIRC.

PR #5773, opened 2018-12-10
Adds Identification System, New Autoinjectors

Neerti:
In order to continue trying to ~force this game to be a poorly made roguelike~ make exploration more interesting while also making the rest of the station be more relevant for explorers, this PR adds two main features, an identification system, and a bunch of new autoinjectors with random spawners that start unidentified.

In the Future :tm:, Points of Interest are going to receive various pieces of loot that start 'unidentified', meaning that players will not know what it actually does.

Unidentified objects can be identified in a few ways, the easiest being with a specific machine that will be placed in various departments. For now, as only hyposprays can start unidentified, there is only a new chemistry machine that will identify any unidentified reagent-based object.

Later on in the Far Future :tm: :tm:, when more objects start unidentified, other ways of discovering the functionality of an unidentified object will be possible, such as using something blindly that isn't a once-off consumable, or using a rare device that identifies other devices (and hopefully isn't lying to you).

Other things this PR does;
- Maps in a new machine in chemistry. A table was removed in order to make room, the things that were on the table (a box of syringes and a screwdriver) have been moved to the locker in the same room. Chemistry now looks like this: https://i.imgur.com/WpDP5tV.png
- Makes buttons layer over windows so they won't be hidden underneath a window.
- Reagent scanners cannot scan things that have its reagents sealed, like closed beakers or the new big autoinjectors, in order to not trivialize Chemistry's new role.
- Refactors hypospray injection logic to make it cleaner and to make it easier to override to do special things.
- Moves the basic biginjectors out of technomancer files and into the hypospray one.
- Ports the two helper .bat files in mapmerger folder to .sh files so I can use them on my preferred OS.

PR #5983, opened 2019-03-01
Ultra-Advanced Reagent Scanners

mistyLuminescence:
So, two types of autoinjectors- the small ones, which usually contain inaprovaline, and the big ones, which contain other stuff (e.g. clotting agent). Prior to #5773, they worked exactly the same, except that the big ones were bigger- 5773 removed the ability to scan and syringe from the big ones to accommodate unidentified autoinjectors.

This PR reworks un/identified autoinjectors.

* Identified biginjectors are back to being scannable/syringeable containers again!
* Unidentified ones remain unchanged. You still can't cheat these with a PDA reagent scanner.
* When identified using the machine in Chemistry, an injector is unsealed and can be analyzed/syringed.
* Adds a tier 3 reagent scanner to R&D which can analyze sealed containers, but requires silver and uranium to produce.

So there's now two ways to ID unidentified autoinjectors- the chem machine, and the t3 ultra-advanced reagent scanner. When you ID them using the chem machine, they become unsealed, and you can do what you want with the reagents inside. The UARS doesn't unseal them, though, just tells you what's in them.

And here's the sprite: https://i.imgur.com/n2P635V.png


code/modules/reagents/Chemistry-Machinery.dm
```diff
@@ -602,6 +602,12 @@
 					continue
 				to_chat(user, span("notice", "Contains [R.volume]u of <b>[R.name]</b>.<br>[R.description]<br>"))
 
+
+		// Last, unseal it if it's an autoinjector.
+		if(istype(I,/obj/item/weapon/reagent_containers/hypospray/autoinjector/biginjector) && I.flags == 0)
```
Atermonera: `!(I.flags & OPENCONTAINER)`
Enforce that it does not have the targeted flag, instead of no flags at all.

code/modules/reagents/Chemistry-Machinery.dm
```diff
@@ -602,6 +602,12 @@
 					continue
 				to_chat(user, span("notice", "Contains [R.volume]u of <b>[R.name]</b>.<br>[R.description]<br>"))
 
+
+		// Last, unseal it if it's an autoinjector.
+		if(istype(I,/obj/item/weapon/reagent_containers/hypospray/autoinjector/biginjector) && I.flags == 0)
+			I.flags ^= OPENCONTAINER
```
Atermonera: `I.flags |= OPENCONTAINER`
Only open it. If this code were to run a second time the container would be resealed.

Anewbe:
The handheld scanner isn't wanted, and was specifically shot down during the March 3rd Staff Meeting.

The rest of this fundamentally touches a system in a way that I'm not sure works the way you think it does, so I would like you to open a new PR _specifically_ changing whatever else in this PR that you want looked over.

You have until the beginning of this Sunday (March 10th, at 00:00 EST) to do so.

mistyLuminescence:
Remade as #5990 by @Anewbe 's request.

PR #5985, opened 2019-03-02
Subspace Jammer now blocks signalers, suit sensors, lasts longer

Anewbe:
Buffs the Subspace Jammer traitor item, so hopefully it'll be used occasionally.
- Signalers can't send or receive while in range.
- Suit Sensors don't show up on the console if they're in jamming range.
- Battery now lasts 10 minutes.


mistyLuminescence:
Does it block communicators and PDAs too? If not, it totally should.

zekesulastin:
Wouldn't ten minutes make it really powerful in conjunction with replacing the cell in it, especially since you're removing suit sensors from it?

PR #5988, opened 2019-03-04
Copy character slots

PrismaticGynoid:
Adds a button to the setup menu that can be used to copy a character over another slot, allowing for easier duplicating, rearranging, and wiping of character slots. After you copy a character to another slot, you must then save to overwrite what's already on that slot - or don't, if you've changed your mind.

Essentially what it does behind the scenes is load a character slot without loading the preferences for that character slot, instead leaving the preferences of the previous slot open for you to overwrite the loaded slot with. Tested a bunch and it doesn't seem to break anything/ mess up the save files any.

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -217,7 +217,8 @@ datum/preferences
 		dat += "<a href='?src=\ref[src];load=1'>Load slot</a> - "
 		dat += "<a href='?src=\ref[src];save=1'>Save slot</a> - "
 		dat += "<a href='?src=\ref[src];reload=1'>Reload slot</a> - "
-		dat += "<a href='?src=\ref[src];resetslot=1'>Reset slot</a>"
+		dat += "<a href='?src=\ref[src];resetslot=1'>Reset slot</a> - "
+		dat += "<a href='?src=\ref[src];copy=1'>Copy slot</a>"
 
 	else
 		dat += "Please create an account to save your preferences."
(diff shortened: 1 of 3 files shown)


Anewbe:
I'm curious as to what specifically the goal of this is?

Neerti:
I imagine it'd be useful for those 'in the future (more so than usual)' rounds.

PrismaticGynoid:
> easier duplicating, rearranging, and wiping of character slots

is the goal of this.

Anewbe:
Well yes, but for what reason do you find this necessary?

PrismaticGynoid:
Because I and several other people have expressed desires for an easier way to move characters around or copy them for future round usage than having to transfer over each detail in the character setup manually, which becomes very tedious. Especially when dealing with characters with fully-filled-out records/flavor text/loadouts/etc.

zekesulastin:
I would find the feature useful as it would make it easier to make separate character slots for, say, Farala the engineer, Farala the moonlighting bartender, and Farala the visitor each with difference appearance and loadout items; and to expedite this process with newer characters.

Anewbe:
Fair enough on all counts.

Atermonera:
It also allows simpler reordering of character slots, since the copying is handled automatically

Mechoid:
It also makes it easier to delete slots, if you keep one default on all-random!

Novacat:
Downstream approves of this PR.

PR #5990, opened 2019-03-05
Unidentified Auto Tweak 2: Electric Boogaloo

mistyLuminescence:
Remakes #5983 by @Anewbe 's request. It's basically the same PR except this time without the R&D scanner thing.

In very simple words delivered in the form of space-rap and interpretive dance, here's how it works:
*Listen up, explorers, to the Research Director,*  
*But remember, you're the one with an autoinjector!*  
*What's that, you didn't get it identified?*  
*How are you gonna know what chems are inside?!*  
*That pen is nasty, ain't safe to use!*  
*You can't even scan it! That gives me the blues!*  
*Now, Chem has a machine that can fix your probs,*  
*But only Chem can do it, so respect those jobs!*  
*Before the rework, you could syringe all day,*  
*Replace it with chloral to make your enemies pay!*  
*But now you can't do that in case it's P-O-I,*  
*Doesn't that just make you... wanna cry?*  
*So here's the deal, the biznit, the rap-*  
*We're tweaking a bit, and this PR ain't crap!*  
*Now it's like before, with syringes ahoy,*  
*And scanners too, for girl, NB or boy!*  
*You can do what you want, to an autoinjector,*  
*But only ID'd ones, says the Research Director.*  
*If it's from Sif, some neat loot from the plains,*  
*It's still unID'd so the old system remains!*  
*Does that make sense? It's the same as before!*  
*Except if you want to scan an auto, just count to four!*  
*One! Is it identified?*  
*Yes? You can scan it and syringe it alright!*  
*Two! Take it to Chemistry!*  
*They'll scan it and prod it and put it in their machine!*  
*Three! Get your autoinjector back!*  
*Or maybe Chem keeps it, that'd be quite the hack!*  
*And FOUR! Your injector is safe to use!*  
*And you can syringe it! Scan it! Even fill it with booze!*  
*Did you get all that? The autoinjector rap?*  
*Now take it from me, this PR ain't crap.*  

Or, in slightly longer but less fun words:
**How it used to work**: any autoinjector was scannable and syringeable.
**What the rework was supposed to do**: make unidentified autoinjectors unscannable/unsyringeable.
**What the rework actually did**: make *any* biginjector unscannable/unsyringeable.
**What this PR does**: makes unidentified autoinjectors unscannable/unsyringeable until identified.

Yes, I've tested it, and it works perfectly.

### DIFF
--- a/code/modules/reagents/Chemistry-Machinery.dm
+++ b/code/modules/reagents/Chemistry-Machinery.dm
@@ -602,6 +602,11 @@
 					continue
 				to_chat(user, span("notice", "Contains [R.volume]u of <b>[R.name]</b>.<br>[R.description]<br>"))
 
+		// Last, unseal it if it's an autoinjector.
+		if(istype(I,/obj/item/weapon/reagent_containers/hypospray/autoinjector/biginjector) && !(I.flags & OPENCONTAINER))
+			I.flags |= OPENCONTAINER
+			to_chat(user, span("notice", "Sample container unsealed.<br>"))
+
 		to_chat(user, span("notice", "Scanning of \the [I] complete."))
 		analyzing = FALSE
 		update_icon()
--- a/code/modules/reagents/reagent_containers/hypospray.dm
+++ b/code/modules/reagents/reagent_containers/hypospray.dm
@@ -184,7 +184,6 @@
 	volume = 15
 	origin_tech = list(TECH_BIO = 4)
 	filled_reagents = list("inaprovaline" = 15)
-	flags = 0 // Removed OPENCONTAINER so you can't extract things to cheese the identification system in unidentified versions.
 
 /obj/item/weapon/reagent_containers/hypospray/autoinjector/biginjector/brute
 	name = "trauma hypo"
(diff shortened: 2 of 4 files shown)


Atermonera:
A bunch of people (notably no less than half the dev team) were unable to attend the staff meeting (self included) so I don't actually know if it got discussed or what any discussion resulted in.
Will unfortunately likely require discussion at the next staff meeting, sorry for the delay

Wickedtemp:
... Just to clarify, did that change result in every single injector becoming "locked", unable to be syringe'd and filled with something else, until you threw it into chemistry's scanner?

Atermonera:
Only big ones, i.e., hypos

mistyLuminescence:
Almost. The change resulted in every single biginjector (e.g. clotting injectors, PoI injectors) being "locked" (no syringing or scanning with reagent scanners), period. Even when scanned in Chemistry's scanner.

So if you found, say, an unidentified autoinjector and found out it had osteodaxon in it, you wouldn't be able to syringe out the osteodaxon and scan it into an Odysseus.

Prior to this, biginjectors worked just like normal injectors. Hyposprays haven't been touched at all, either by this or by a previous change.

Wickedtemp:
So if for example I ordered a clotting kit but decided I'd rather have a general stabilizer mix in the 15u injector, that just isn't possible because I can't use syringes on it anymore?

If that's the case... this sounds absurdly stupid and I'm at a loss as to why this actually requires staff discussion.

mistyLuminescence:
Yeah, pretty much. With this PR, only unidentified autos are 'closed' (no scan/syringe), and only until they're identified- so it's exactly how it used to be originally, except with a minor tweak to allow for unID'd autos to be sneaky still.

In hindsight, making a bug report for this might've got it merged three weeks ago.

PR #5999, opened 2019-03-10
Winter coat suit storage expansion

mistyLuminescence:
Departmental winter coats can now hold the same items as their regular variants- labcoats for medsci, hazard vests for engi, armored vests for security and CD, aprons for hydro, and mining suits for mining. Oh, and they can all hold suit cooling units too, for those of you who play robutts.


Dunmeritude:
don't think cooling units are necessary given you shouldn't overheat entirely in them anyways? The rest of this is so good though and will solve literally all my fucking woes

Mechoid:
While this does now have a mechanical influence at round-start, their Original Versions are also available to anyone in load-out.

PR #6002, opened 2019-03-16
Belt-Worn RIGs

mistyLuminescence:
Lays the foundation for RIGs to be worn on the belt slot instead of the back slot. Also, lets certain items (none yet, but the functionality now exists) be worn on the belt slot without a jumpsuit.

* Modules are fully functional.
* Deployment works as intended.
* Can be surgically removed as usual.
* Two RIGs cannot be worn at once- it just won't fit.
* MMIs, AIs and pAIs can be installed as usual, and operate as usual.

When making a RIG available from the belt, just add SLOT_BELT as usual. When making an item fit in the belt slot without a jumpsuit, set 'suitlink' to 0 (default = 1). Highly recommended for belt-RIGs to avoid bugs.

Again, this doesn't actually change anything for players, so you're not going to notice the difference unless someone VV's a RIG (which is encouraged for testing purposes!). I've tested it as much as I can on my end, but it could use a once-over.

Possible PRs if this gets merged:
* Add this functionality to the offstation (vox, merc, ninja, ERT) rigs?
* Allow actual belts (e.g. toolbelt, sec-belt) to be worn without jumpsuits?

### DIFF
--- a/code/modules/clothing/spacesuits/rig/rig_verbs.dm
+++ b/code/modules/clothing/spacesuits/rig/rig_verbs.dm
@@ -6,7 +6,7 @@
 	set category = "Hardsuit"
 	set src = usr.contents
 
-	if(wearer)
+	if(wearer && (wearer.back == src || wearer.belt == src))
 		ui_interact(usr)
 
 /obj/item/weapon/rig/verb/toggle_vision()
--- a/code/modules/clothing/spacesuits/rig/rig_verbs.dm
+++ b/code/modules/clothing/spacesuits/rig/rig_verbs.dm
@@ -16,7 +16,7 @@
 	set category = "Hardsuit"
 	set src = usr.contents
 
-	if(!istype(wearer))
+	if(!istype(wearer) || (!wearer.back == src && !wearer.belt == src))
 		usr << "<span class='warning'>The hardsuit is not being worn.</span>"
 		return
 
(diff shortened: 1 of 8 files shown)


Wickedtemp:
So is this making RIGs belt-only, or is it just *adding* belt slot in addition to back?

mistyLuminescence:
Neither- it changes absolutely nothing for players. RIGs were hardcoded to _only_ work from the back slot, even if they'd been modified to be wearable on belts (to stop people using them while holding them in-hands). This changes that so wearing them on belts works too... if, somehow, you have a way to actually get it on your belt.

So until a RIG with SLOT_BELT gets added, nothing changes- this just gets the framework in place.

code/modules/clothing/spacesuits/rig/rig_verbs.dm
```diff
@@ -6,7 +6,7 @@
 	set category = "Hardsuit"
 	set src = usr.contents
 
-	if(wearer && wearer.back == src)
+	if(wearer)
```
Mechoid: Is there any particular reason these are losing the second check, instead of getting one for the belt slot in addition?

mistyLuminescence: Not really tbh, fixed

PR #6010, opened 2019-03-20
Removes spammy keyboard noise from arcade machines

Heroman3003:
Those get pretty spammy and the noise really doesn't fit the machine its attached to anyway at all.


MisterLayne:
I vote on using a different sound instead. I personally enjoy the clicky noises- or perhaps changing the sound _and_ adding a client preference for whether or not your want to hear click sounds.

Neerti:
TG has a bunch of sounds for the arcade machines, however they're based on events that occur inside the arcade machine's game instead of just clicking on the machine itself, I think.

Porting that is probably out of scope for this, however.

PR #6012, opened 2019-03-21
[DNM, DOESN'T FEEL SO GOOD] A Series of Unfortunate Techlevels.

Mechoid:
Everything in this is subject to change now, and in the case it gets merged somehow, till forever, as it contains some Interesting and !!FUN!! things that may not fit well into the current balance with respect to time, player, and material investment to achieve.

Tweaks:
- Marble now spawns as a poor ore, and in underground veins. Relatively rare.
- Lead has been added as a smeltable material, found in mining. Relatively rare.
- Drill upgrades actually matter for the purposes of finding Marble, Lead, and Verdantium [More on that later]
- R&D machines can now accept a wider range of more materials. None of these new materials were added as requirements for previously existing designs. Some are hidden until placed into the machine.
- TECH_ARCANE is now repurposed as 'anomaly' research, only acquired from PoIs and Xenoarch finds.
- Crafting steps for the Phazon. Still inaccessible.
- Marble is less-murdery, as it is now accessible.
- More xenoarch sprites. Some xenoarch finds have TECH_ARCANE 1, or in the case of unique items, TECH_ARCANE 2, providing an 'Anomalous Breakthrough'. Currently, un-used.
- Sleepers gain new chemicals based on their tier level, starting at tier 2. Alien-Tech components will cause possibly Concerning results.
- Random E-swords can now have white blades.

Additions
- Verdantium ore, a rare crystalline dust that stores energy. Dangerous if it enters your eyes, **incredibly rare**.
- 'Anomalous Matter' can be detected via the handheld ore scanners. Currently only Verdantium exists in this type.
- Particle Smasher, a modified singularity generator that needs to be energized by two P.A.s, or one hacked P.A. Gives off rads and light when energized, requires certain energy levels and possibly reagents or items to convert materials from one type to another. A high-energy beam is required to initiate the reaction. _**Requires direct interaction from an operator in order to empty or fill,**_ meaning chambers cannot be permanently inhospitable or inaccessible, and easy to make livable. **This is a pain.**
- Morphium, produced in the Particle Smasher. Eerily motile metallic crystal mesh derived from Verdantium. Capable of absorbing high levels of energy in short bursts.
- Valhollide, a pre-supermatter material used in the production of supermatter. Requries a bluespace beaker full of phoron, and a sheet of phoron.
- Crusted supermatter chunks, a highly dangerous, relatively stable form of Supermatter, due to its dense 'crust' formed from waste matter. Requires **intense** heat and energy levels in the Smasher in order to craft.
- Prototype Phazon Janus Class, or 'Janus' for short. A highly advanced, and costly, exosuit requiring the production of crusted supermatter, and the acquisition of Ancient Alien Technology by explorers and archaeologists,
- TECH_PRECURSOR added, common 'precursor' or ancient items should have at max 1, rarer items from PoIs may have more.
- Tier 4 and 5 components, Anomalous and Precursotech respectively. Precursotech components can give either random boons or maluses in machines they are used in, due to the power they provide.
- R&D can now build a charge-based energy sword, which requires cell replacement via screwdriver.
- Three more nanite types, used in the tier 5 sleeper upgrade, all using the same display name, with mostly-dangerous effects. All are unscannable.


MisterLayne:
> Marble is less-murdery, as it is now accessible.
Why was marble murdery before? I'm interested.

> Verdantium ore, a rare crystalline dust that stores energy. Dangerous if it enters your eyes, incredibly 
> rare.
How does one get Verdantium in their eyes- would it be dangerous enough to be used by antags? Permanent and un-repairable eye damage? 

> Valhollide, a pre-supermatter material used in the production of supermatter.
Does this mean a supermatter crystal is now craftable, using the particle smasher or otherwise? 

> Prototype Phazon Janus Class, or 'Janus' for short.
I love this name and I will create portal sprites if I can do that.

zekesulastin:
It's pocket sand: https://github.com/PolarisSS13/Polaris/pull/6012/files#diff-9f17a6655b99dc5d7d374bde261e9008R86 temporary blindness and blurriness.

My question with that is: it looks like goggles/voidsuits/etc. do not protect against it - is that intentional?

Mechoid:
It's a 1:1 copy of Sand's throwing, which also doesn't check goggles/helmets for some reason. I'll go stare at it. 

You can craft a 'Supermatter Shard' which is a slightly smaller supermatter IIRC, though it requires more than one chunk, and thus more than one use of the smasher in Hell-temperatures.

DasIrrlicht:
> Sleepers gain new chemicals based on their tier level, starting at tier 2

That looks interesting and probably will R&D get out more to replace parts, or supply Assistants/Engies/Anyone willing to do the job/ with the parts to buff up the sleepers. Especially in the shuttle and outpost.

Atermonera:
Per 31 March staff meeting, to be broken into multiple pr's

PR #6016, opened 2019-03-23
Explorer Update - Cataloguers

Neerti:
# Overview
This PR aims to give the explorer job a purpose in being, instead of just existing as a thing to get loot to get more loot. To do this, explorers will be given a goal to fulfill, which is to use a new device called a Cataloguer (sprite by Mechoid) to scan things that look interesting. Upon doing so, they will get a paragraph or two of lore about the thing they just scanned, which they can review at any time. They will also get 'exploration points', which currently do nothing except perhaps for measuring epeen. In the future I intend to add the ability to buy explorer things with the points.

A majority of mobs explorers will encounter can also be scanned. This is considerably trickier if the mob is hostile, since you will need to ensure your survival while you scan the monster. Fortunately, you are able to move while scanning. The catch is that you must remain within a certain range in order to scan something. A helpful box is shown around the mob you are scanning to help you with this, which only you can see. If you step out of the box, the scan will stop. Fortunately, if you scan the same mob again, it will resume from where it left off, so feel free to step out if you need to dodge a nasty attack that's coming for you.

The Cataloguer can also highlight valid objects to scan in sight, useful to avoid fruitlessly clicking on everything to find the thing that actually has lore written for it.

Explorers near each other will automatically share discoveries, and as a result get the full amount of points for each person, meaning you will not miss out on points if you're with friends. Conversely, you will collectively have more points than if you go alone.

Video (Item sprite was changed, and the box turns green as the scan gets closer to finishing now): https://streamable.com/xrzqf

## Dev notes
Adding a catalogue entry is fairly easy, and just requires writing a new datum. Generally there's three vars to worry about, and occasionally a fourth one. Note that there are several category subtypes. Your new entry needs to fit in one of them, or you will need to make a new category datum for the ``/datum/category_item/catalogue`` to go into.

- name - What the entry will be named as.
- desc - The actual writing for the entry. Supports HTML formatting.
- value - How many points to award for unlocking the entry. Suggested to use the CATALOGUER_REWARD_* defines for easy tweaking.
- unlocked_by_any - A list of types that, if they are discovered, will also discover this entry.
- unlocked_by_all - Similar to above, but all types must have been discovered for this entry to be discovered.

After defining your datum in the code, you will need to attach it to an object, so that it can be scanned. The easiest way to do this is to find the object you want to be able to be scanned, and override the ``catalogue_data`` variable with a list containing the catalogue ``/datum/category_item/catalogue`` subtype you made. e.g. ``catalogue_data = list(/datum/category_item/catalogue/technology/positronics)``.

If you need conditional catalogue_data retrieval, you should override ``get_catalogue_data()`` to do this. Humans and borgs override it to get the correct entry based on what kind of human or borg they are, as an example.

## Other things this PR does
- Adds argument to ``do_after()`` and ``do_mob()`` to allow for movement to not interrupt.
- Small borg braintype cleanup with defines.
- Adds sequencer looping sounds, which repeat sounds in a defined pattern.
  - Adds Morse code sequencer datum from above.
    - Adds transmitter prop which emits Morse code sounds according to a set message, for use with comm PoIs.
- Cleans up turret icon code to allow for more than one sprite.
  - Alien UFO turrets have a new sprite, by Subber. 
- Adds blackbox props, for use in crashed shuttle PoIs.
  - Adds and maps in blackboxes for the quarantined shuttle and crashed med shuttle.
- Adds fake AI props so that I can finally spawn one easily without using the exploding decoy AI.
  - Replaces fake AI core inside crashed med shuttle with a "real" fake AI object.
- Adv. Dark Gygax primary attack now homes in on intended target, however the projectile is now much slower, enabling a sort of bullet-hell method of dodging the attack.
- Shantaks now have 5 melee armor soak, to match their lore.


mistyLuminescence:
**THERE IS A SANCTUARY TARGET HERE, HUNTER. I SENSE IT. DO NOT LET IT ESCAPE!**

Mechoid:
I've read over this multiple times, and if I missed something I'm going to shriek.

PR #6017, opened 2019-03-24
Ports Stasis Cages from Baystation

Novacat:
The idea behind this is that these can be used by explorers to 'trap' mobs that have been netted, and brought back for Xenobiology to have fun with.

Edit: Stasis cages now mostly finished and operational!

### DIFF
--- a/code/game/objects/structures/stasis_cage.dm
+++ b/code/game/objects/structures/stasis_cage.dm
@@ -2,7 +2,7 @@
 	name = "stasis cage"
 	desc = "A high-tech animal cage, designed to keep contained fauna docile and safe."
 	icon = 'icons/obj/storage.dmi'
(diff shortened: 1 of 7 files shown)


mistyLuminescence:
[insert Simaris quote here]

Mechoid:
I assume this has been tested properly?

Novacat:
So far, I just tested basic functionality. I stuffed a mob in, took a mob out. The only thing I have not tested was to see if the mob took atmos damage while inside the cage.

Mechoid:
Please do so, if you would.

Novacat:
Tested. Went out into vacuum with a cat in a stasis cage. After spending 10 minutes outside, cycled back in. Cat was completely unharmed.

PR #6019, opened 2019-03-26
Adds New Alert Levels

Novacat:
Orange/Violet were ported from Baystation, Yellow is a new creation..

Code Yellow: Allows Security to openly display weapons, but not search persons.
Code Yellow is primarily for fauna-based emergencies, spiders, carp, that kind of thing. Basically there is no reason to rescind privacy laws and allow random searches.

Code Orange: Engineering Emergency.
Code Orange is for emergencies involving a lot of structural damage to the station. Meteors, massive depressurization of multiple decks.

Code Violet: Medical Emergency.
Code Violet is for medical emergencies. Usually just viruses or other contagions.

Code Blue: Security Emergency.
Code Blue is for security emergencies not involving fauna, such as murder.

Code Red is kind of fuzzy, but, basically, multiple threats, or a large enough threat, would warrant going from one of the above codes to Code Red. Ideally, Code Red is not a security-specific code, but rather, is an 'all hands on deck' multi-department code. 

Adds new status display icons for all alert levels, except yellow (because I cannot sprite). Changing alert level now automatically switches the status display to the alert level.


mistyLuminescence:
👍 but only if we rename them to have funky names.

**Code Chartreuse, I repeat, Code Chartreuse!**

Wickedtemp:
When would we use Violet? If a large number of people are wounded, there's probably a general emergency, or at least one involving Security or Engineering as well. 

Iunno, it just seems like we're getting too specific. When you have a bad guy who shot some people and blew something up, you have three or more departments all responding to different threats. Would that warrant a Code Blue Orange Violet, or would it go straight to Red? If mob swarms are attacking, would that be Yellow, Blue, or Red? Who decides if a thing is an emergency? Does the CD need to call a Code Violet or does the CMO do it from their office?

Novacat:
Code Violet would be primarily used for situations involving contagions. On Baystation it is one of the more rarer alert types called, but it still gets called every so often.

When you have a bad guy who shot some people and blew something up, you have three or more departments all responding to different threats.:
Code Blue. It is primarily a security threat.

If mob swarms are attacking, would that be Yellow, Blue, or Red?:
Yellow. Unless things go really bad, in which case it would probably go straight to Red.

Who decides if a thing is an emergency? Does the CD need to call a Code Violet or does the CMO do it from their office?:
The way codes change has not changed. Any code except for red and delta can be called from the console.

Wickedtemp:
Suggestions:
- Allow change of department-specific alerts from the Offices of the relevant Heads. If there's a virus for example and I'm one of the only Medical crew, I'd probably just make a CMO Announcement and work on antibodies instead of running to the bridge to change the alert level and then run back to Medbay. The bridge is, honestly, rarely if ever actually manned by more than the CD, HoP and a maybe a Secretary. All other Heads tend to stay in the departments.

- Suit sensors should probably be mentioned regardless of whether or not they'd be *required*. Currently, Blue requires Vitals and Red requires Tracking. I would think that Violet/Red requires tracking, Blue would require vitals, Yellow/Orange could reasonably require vitals but I can see an argument for it not to, in which case, "privacy rights are still enforced, but we suggest switching Sensors to Vitals or higher." should be added, because it's...still a hazard that would warrant a reminder.

Novacat:
To fix the first suggestion, I could map in command consoles to every head of staff's office.

PR #6024, opened 2019-04-01
Precurso / Anomalotech Sleepers

Mechoid:
Sleeper component of #6012 
Reliant on #6023, DNM until post.

* Upgrading the manipulator of a sleeper will allow it to inject more chemicals.
- Tier 2, Leporazine, the temperature stabilizing chemical.
- Tier 3, Spaceacillin, the antibiotic.
- Tier 4 (Anomalotech), Immunosuprizine, the base anti-rejection drug.
- Tier 5 (Precursotech), A random nanite type, Brute/Burn causing, Radiation inducing, Brain-eating, or universally-beneficial Nanites.


Wickedtemp:
Ethylredoxrazine isn't really anti-alcohol in the way it's intended. Separate issue, but when it was 'fixed' last time, the reduced effectiveness on alcohol actually in the stomach - you know, the only reasonable place alcohol actually goes - is negligible. Instead, it focuses on cleansing alcohol from the bloodstream, which doesn't exist, unless you got antagged and jabbed with syringes of alcohol. But, for people who drank a keg? Nothing to do but OD them on ethyl and shove carbon in their face and babysit them until it slowly...slowly works.

Wickedtemp:
Leporazine is next to useless. I'd personally recommend:
- Sticking with Ethyl and fixing it so it actually removes alcohol from the stomach

- Imidazoline and/or Alkysine.

- Hyronalin

Not quite basic, not quite advanced, actually useful and used.

Mechoid:
The idea for Leporazine is that it'd actually serve a purpose for the shuttle / groundside sleepers to be upgraded with it, in the event someone is Literally Frozen, or actively processing frost-venom. I tried to avoid anything touching organs repair wise, as that would make the sleeper basically an autodoc for the organ in question. I considered hyronalin or arithrazine, but those are both a bit more general.

Wickedtemp:
That actually does make sense - I was thinking more of first-aid sleepers, triage station sleepers, instead of shuttle sleepers. It'd be good to have in that case.

code/game/machinery/Sleeper.dm
```diff
@@ -351,3 +394,7 @@
 	desc = "A limited functionality sleeper, all it can do is put patients into stasis. It lacks the medication and configuration of the larger units."
 	icon_state = "sleeper"
 	stasis_level = 100 //Just one setting
+
+/obj/machinery/sleeper/survival_pod/New()
```
Neerti: I think this should use ``Initialize()`` instead.

PR #6028, opened 2019-04-01
Precursotech Tools

Mechoid:
Tools component of #6012 
Reliant on #6023, DNM until post.
Adds the Precursotech Tools
* Tools have a reach of 2 tiles. Chance to fail on use and irradiate their tile.
* Welding tool will burn through 2 layers of eye protection, meaning a welding mask acts as though it doesn't exist. Welding goggles + mask, or Goggles + voidsuit helmet are required.


Wickedtemp:
It sounds like the protection and risk outweighs the reward. Is it just a tool with increased range or does it do anything else?

Mechoid:
A welding tool that has infinite fuel, a quarter the time required to do work, has a reach of two tiles, and can be used as a weapon, is pretty good.

PR #6030, opened 2019-04-02
Reverts chemistry and nanoui back to processes due to slowness

Heroman3003:
The conversion to subsystems was with a good heart, but the extreme slowdown in workings of reactions and nanoui updates makes it feel plain bad. In addition to that nanoui not initializing if you are around too early is also extreme annoyance. In general, this conversion just made both systems ultra-slow. Revert keeps other subsystem conversions and any refactors/code cleanups done along the way, only reversing it for nanoui and chemistry subsystems.


Wickedtemp:
Chemistry is a job that you can spend an hour on even with zero lag. 

If it were any longer, even *I'd* have a hard time having fun with it.

Neerti:
@kevinz000 says they're looking into the issues from those things being made into substations, and I'd rather that they/someone else fix the subsystems instead of going back to processes, since it gets in the way of the eventual future goal of removing processes in favor of pure subsystems.

So hopefully kevins delivers.

Heroman3003:
I agree with that, but until that fix is up and tested, I am just leaving this PR to hang, just in case.

Wickedtemp:
Is there any issue with merging this as a temporary solution until kevinz is done with the actual fix?

Wickedtemp:
Update. I actually played a bit. Holy *fuck* it's bad. This needs to get fixed ASAP.

SpadesNeil:
https://www.youtube.com/watch?v=1Isjgc0oX0s

Atermonera:
#6151 is merged

PR #6037, opened 2019-04-05
Fixes export reports not mentioning crates' cost

Heroman3003:
Now the total cost tally up will include cost of crates in which stuff was sent.

### DIFF
--- a/code/controllers/Processes/supply.dm
+++ b/code/controllers/Processes/supply.dm
@@ -100,6 +100,7 @@ var/datum/controller/supply/supply_controller = new()
 		EC.name = "\proper[MA.name]"
 		EC.value = 0
 		EC.contents = list()
+		var/base_value = 0
 
 		// Must be in a crate!
 		if(istype(MA,/obj/structure/closet/crate))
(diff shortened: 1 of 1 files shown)


code/controllers/Processes/supply.dm
```diff
@@ -100,13 +100,16 @@ var/datum/controller/supply/supply_controller = new()
...
+		var/base_value = 0
 
 		// Must be in a crate!
 		if(istype(MA,/obj/structure/closet/crate))
 			var/obj/structure/closet/crate/CR = MA
 			callHook("sell_crate", list(CR, area_shuttle))
 
 			points += CR.points_per_crate
+			if(CR.points_per_crate)
+				base_value = CR.points_per_crate
```
Atermonera: Editing EC.value directly with `EC.value += CR.points_per_crate` makes the `base_value` var unnecessary

Heroman3003: CR is only defined in that IF statement and if the MA is a crate in the first place, and EC.value has to only be edited after the whole ifelse statement is done and points are assigned.

Atermonera: I see. Nevermind, then.

PR #6042, opened 2019-04-06
Research QoL changes

LiquidFirefly:
Moved to Polaris as per instructions in https://github.com/VOREStation/VOREStation/pull/4902

- Lowers stock part print times to 1/3rd of what they were.

- Adds an Advanced Part Exchange device at materials: 5, engineering: 5, which can hold 200 parts.

This thread is mostly to spark discussion about the topic. Most of what this does is reduces the time required to finish RnD by a few minutes and greatly increases QoL as it relates to materials lying around the lab, RPED routes, and most importantly printing parts for RPED routes.  The ARPED in it's entirety was a change to the initial formula that I was asked to make.  I'd rather simply upgrade the capacity of the RPED itself.


TheDavestDave:
having to wait longer than it takes me to do R and D to print parts of is why I hate upgradeing things. I'm happy to do work faster and go do my own things

Mechoid:
As someone who is actively touching Research, I'm mostly questioning lowering times universally, as part replacement makes things go pretty fast anyway, and increasing the material capacity, as upgrades also do this.
Lowering stock-part build time is not quite an issue to myself as they are pretty universally small and are 'simple' parts of machines.

LiquidFirefly:
Part replacement essentially takes you from a full minute for the part replacer and nano manips you need to actually an almost bearable speed.  It's annoying to be limited by the game instead of my own memorization.  The material capacity is far more of a 'hey, why not.'  There's no reason that someone should actively be forced to manage it, and it just removes a number to memorize.  I should also note that part replacement is just as powerful as it was, fractionally, and you still get a rather meaningful net positive in your RnD time by incorporating it.   Again, all that these changes amount to for your actual RnD run is about 2-5 minutes of difference.  The part where it starts slicing off time in chunks is waiting for all of your stock parts to print out.

Mechoid:
Material capacity and universal time decrease are the two points of my contention. Stock-part specific time decrease isn't particularly a concern to me as it stands.

> The material capacity is far more of a 'hey, why not.' There's no reason that someone should actively be forced to manage it, and it just removes a number to memorize.

Then there is no point to having a material limit, or having matter bins doing anything at all to the machines, by that logic. It is a constraint on what R&D can actively have in the machine at different tiers of upgrade, meaning they may in fact have to dump some material or swap in different sheets to build different things early on. That is a buff, not a QoL.

It should be pointed out that any change that makes R&D faster is also a buff, as it's already completed by people who memorize or follow guides within a few minutes, meaning the only balancing point is the time it takes for them to get stuff out of the machines, and get materials.

LiquidFirefly:
See, the thing is by the time you're actually getting materials back from mining, you already have supermatter bins and can fit in twenty sheets of everything sans Steel and glass, which you slam in as much as possible.  Twenty is kind of an awkward number, but it's enough that you barely run out anyway- and in cases of materials like osmium, you only really slam in five to make room for shit that you actually use.  There's never a case where you're actively swapping materials, because if there were, you set up the machine wrong.   You can call it a buff all you want, but in reality removing the middleman on the sum total of two seconds it takes to pop out 30/50 sheets of the gold someone slammed in there like an idiot so you can actually put in diamond isn't really a buff?

And again, I know exactly how long it takes to complete RnD.  With optimizations it can take about 10-12 minutes until levels are set with these changes, and about 14-17 minutes without.  However, the feeling of setting levels is less like a chore and more like you're actually doing the job afterward.  It feels so much nicer to play with just a little bit of grease added to the gears.  Most of that time removed is spent just kind of staring at the lab and waiting for the machine to stop being slow long enough so that you can print the next manipulator after having dodged the protolathe for as long as humanly possible, either using the autolathe or the circuit imprinter.   That's still the best strat for research.  If a few minutes before levels are set is the difference between an antag winning around and not, you have bigger balancing issues than research. 

As you mentioned, the limiting point for research was never how long it takes to finish RnD.  It's always been how quickly they're able to procure materials from mining.  In all likelihood, I can only think of a few niche situations where the two seconds it takes to print some required item is going to make a massive difference.  It just makes the researcher not roll their eyes and tap their feet as they wait five seconds per item for it all to print.

Mechoid:
If you've already got the super matter bins by the time mining shows up, then hooray, you can store a ton of materials. Otherwise, removing that limitation also removes an opportunity for IC interactions from mechanical factors. I.E., someone shoving a ton of one material in and not caring just to print one thing, and then getting told off when someone else tries putting another material in and can't.

Scientists _should_ have to care about the machine's limitations, and take it into account for what they're doing, even if those limitations are more lax by the time mining actually shows up.  If you have materials lying around the lab, excellent, someone can come and pinch them for robotics or a golden throne if they so want, and have that thread of interactions go as it would.

I do not believe lowering stock part times specifically is an issue as it stands, as I said before. For the number you need to make to do anything useful, it takes ages. Things like weapons or advanced tools should stay their long-duration sans upgrades however as they are more likely to have more of an impact / complexity.

I remain opposed to doubling the material storage across the board, as having the limitation as it is now provides, in my eyes, a wider array of possibility for IC interaction and meaningful operation of the machines than not having it provides.

LiquidFirefly:
I mean, stock parts are the bulk of this.  I'm willing to tone down or remove the other aspects, but I've already made my point about why I think that they're for the better.  All in all, stock parts and making the use of stock parts not mind mindbogglingly horrible to deal with was the point.  I'd like to see opinions or a statement from a collaborator or maintainer, but that's likely something for a later date.

Mechoid:
I am in fact a Polaris Dev.

LiquidFirefly:
Oh, nah then.  I'll just slap all of that stuff out

LiquidFirefly:
Why is github screaming about the protolathe file, I swear

LiquidFirefly:
Wait, literally nothing changed.   Github hates me, 'kay

PR #6043, opened 2019-04-06
Fixes way elevator handles deleting z-shadows

Heroman3003:
Should stop the bug where going from first to third floor on elevator would leave a HUD icon behind in place.

Problem was caused by the fact that shadow 'stack' was not deleted properly.

### DIFF
--- a/code/modules/turbolift/turbolift.dm
+++ b/code/modules/turbolift/turbolift.dm
@@ -97,6 +97,8 @@
 			if(istype(AM, /mob/living))
 				var/mob/living/M = AM
 				M.gib()
+			else if(istype(AM, /mob/zshadow))
+				AM.Destroy()		//prevent deleting shadow without deleting shadow's shadows
 			else if(AM.simulated && !(istype(AM, /mob/observer)))
 				qdel(AM)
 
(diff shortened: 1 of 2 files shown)


code/modules/turbolift/turbolift.dm
```diff
@@ -97,6 +97,8 @@
 			if(istype(AM, /mob/living))
 				var/mob/living/M = AM
 				M.gib()
+			else if(istype(AM, /mob/zshadow))
+				AM.Destroy()		//prevent deleting shadow without deleting shadow's shadows
```
Neerti: Should this not use qdel() instead?

Heroman3003: The problem with it using qdel() was that it deleted the object, but if there were shadows of that shadow, they would not get deleted. For example, person is on level 1, moving up. Initially they have a shadow on level 2, and that shadow has a shadow on level three. The elevator moves up, and qdel's the shadow on level 2. The shadow on level 3 is NOT immedeately deleted in process and only after a long delay during automatic cleanup. Calling Destroy instead properly handles entire 'stack' of shadows and deleting shadows of shadows along the way.

Heroman3003: That is also why this exception was even added. The bug was caused by the fact that shadows were handled like any non /mob/living and just qdel'd

PR #6047, opened 2019-04-07
Fixes mob swapping bugs

Heroman3003:
Allows to swap with mobs on help if you are both on help. Does not allow to swap with mob on the table while you are on the ground anymore. Generally should help the swapping system be consistant.


Neerti:
Woodrat will be happy this finally got fixed.

PR #6057, opened 2019-04-08
Fixes borgs not being unlockable with PDA holding ID

Heroman3003:
Could only use ID outside PDA. Fixes that.


Atermonera:
Can you use a wallet containing an id?

Neerti:
I think there's supposed to be a generic proc to get the ID but I can't recall what its called.

Heroman3003:
Put in the proc. Wallets and any potential future id-holding item like wallet or PDa will work now.

PR #6058, opened 2019-04-08
Removes umbrella permastuns. Adds wind mechanics.

Neerti:
When outdoors, and in certain types of weather, the wind can make things easier or harder for you. If there is any wind, moving in the same general direction as it will give you a speed boost. Conversely, if moving against the wind, it will slow you down. The wind strength and direction changes each time the weather changes, and a message is shown to anyone outside when wind appears, and what direction its blowing in.

This was done to replace the 'Umbrellas in storms make you unable to move' thing, which was problematic due to the stun being very long and easily resulting in permastuns until the storm stopped or you threw the umbrella way and had to sit still for a long time before being able to move indoors, which is a huge issue if you're a promethean or getting attacked.

I feel this is a more interesting and fair downside to being in hostile weather, as instead of just making you stunned or unable to use an umbrella, it now means running in a certain direction can get harder, but also running in the opposite direction can get much easier, which can be helpful when running from big angry saviks.

Another this this PR does is fix a strange, and apparently very old issue where if an FBP received a stun[1], it was set to zero on the next Life() tick. That made me think this umbrella issue was already fixed since my character would never get stunned by it, but I've recently been told otherwise. Now FBPs should respect stun sources again (which are rather rare nowadays, but its still a thing that should work).

[1] In the game code, there are three types of stuns, all of them named badly. There is ``weaken``, which is what most people think of when thinking of stuns, since its the one that makes you fall over and unable to act. Additionally, there is ``paralysis``, which works like weaken, but makes your screen turn mostly black, essentially knocking out your character. Finally, there is ``stunned``, which merely makes your character unable to move or act, but NOT fall over. FBPs were effectively immune to the last type, but this PR fixes that.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_movement.dm
+++ b/code/modules/mob/living/carbon/human/human_movement.dm
@@ -140,16 +140,17 @@
 			. += turf_move_cost
 
 	// Wind makes it easier or harder to move, depending on if you're with or against the wind.
-	var/datum/planet/P = SSplanets.z_to_planet[z]
-	if(P)
-		var/datum/weather_holder/WH = P.weather_holder
-		if(WH && WH.wind_speed) // Is there any wind?
-			// With the wind.
-			if(direct & WH.wind_dir)
-				. = max(. - WH.wind_speed, -1) // Wind speedup is capped at roughly taj speed.
-			// Against it.
-			else if(direct & reverse_dir[WH.wind_dir])
-				. += WH.wind_speed
+	if(T.outdoors)
+		var/datum/planet/P = SSplanets.z_to_planet[z]
+		if(P)
+			var/datum/weather_holder/WH = P.weather_holder
+			if(WH && WH.wind_speed) // Is there any wind?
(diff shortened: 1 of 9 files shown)


code/modules/mob/living/carbon/human/human_movement.dm
```diff
@@ -135,6 +139,19 @@
...
 
+	// Wind makes it easier or harder to move, depending on if you're with or against the wind.
+	if(T.outdoors)
+		var/datum/planet/P = SSplanets.z_to_planet[z]
+		if(P)
+			var/datum/weather_holder/WH = P.weather_holder
+			if(WH && WH.wind_speed) // Is there any wind?
+				// With the wind.
+				if(direct & WH.wind_dir)
+					. = max(. - WH.wind_speed, -1) // Wind speedup is capped at roughly taj speed.
```
Anewbe: Taj speed is -0.5., Teshari speed is -1.

PR #6068, opened 2019-04-11
Re-adds the Space Worm.

Mechoid:
On the tin.

- The worm now must use a timed toggle for it to be able to devour as it did in Ye Olden Days.
- Mobs are no longer deleted, instead taking BIOACID damage.
- Doors and Shields have a longer delay for devouring, both making messages that there is something dangerous occuring.
- Worms in devour mode move slower than normal, meaning it is possible to outrun them if not severely hindered.
- Worms can safely move from Z to Z without bisecting.
- Worms move freely in space. The head is 'hovering' and unanchored, but the rest of the body is not. Will hit landmines.
- Segments and the head have overall higher health, and do not bisect if one dies. Instead it cuts the ability for the creature to digest things by dead segments immediately dumping their stomach contents to the tile they are on.
- Killing a segment kills all previous segments.
- The worm cannot move directly backwards, similar to a cargo tug, but it can cross itself. It cannot devour its own segments.
- Will not be bisected by having an airlock or other object cross its body. It will force itself through the airlock or object to continue on to prevent further Universe Implosions.
- Still makes Phoron, because the Phoron must flow.


Novacat:
Since catalogers is now a thing now, this should have catalog entries, no?

Mechoid:
Currently not giving the worm a log entry until I make a proper AI for it, as it won't show up outside of Admin-bus until then.

code/modules/mob/living/simple_mob/subtypes/animal/space/worm.dm
```diff
@@ -0,0 +1,428 @@
...
+						break
+
+					if(do_after(src, 5))
+						D.visible_message("<span class='danger'>Something crashes against \the [D]!</span>")
+						D.take_damage(2 * melee_damage_upper)
+					else
+						objectOrMob = null
+						break
+
+					if(D && (D.stat &= BROKEN|NOPOWER))
```
Atermonera: `(D.stat & (BROKEN | NOPOWER))`
Using &= will turn off all other flags

MisterLayne:
Can have picture of worm? Am curious.

PR #6072, opened 2019-04-12
[RNG is the new RNG] Converts a bunch of events to the Game Master datum system.

Mechoid:
1-1 converts the remaining events to the Game Master system, to the best of my ability. Some may be tweaked.

Some disabled events are _re-enabled_ by this conversion. Specifically
- Virus outbreaks and normal infections.
- Wormholes
- Window Break
- Airlock electrification


code/modules/gamemaster/actions/atmos_leak.dm
```diff
@@ -0,0 +1,76 @@
...
+// Decide which area will be targeted!
+/datum/gm_action/atmos_leak/set_up()
+	severity = pickweight(EVENT_LEVEL_MUNDANE = 8,
+	EVENT_LEVEL_MODERATE = 5,
+	EVENT_LEVEL_MAJOR = 3
+	)
+
+	var/gas_choices = list("carbon_dioxide", "sleeping_agent") // Annoying
+	if(severity >= EVENT_LEVEL_MODERATE)
+		gas_choices += "phoron" // Dangerous
```
zekesulastin: Last time I checked, phoron was removed from the default air alarm filtering for `reasons`, which puts it at the same level as volatile_fuel per the comments.

code/modules/gamemaster/actions/spontaneous_appendicitis.dm
```diff
@@ -0,0 +1,13 @@
+/datum/gm_action/spontaneous_appendicitis
+	name = "grid check"
```
zekesulastin: Appendicitis isn't a grid check ...

code/modules/gamemaster/actions/atmos_leak.dm
Mechoid: Volatile Fuel requires panic syphon, IIRC.

code/modules/gamemaster/actions/electrified_door.dm
```diff
@@ -0,0 +1,70 @@
...
+
+/datum/gm_action/electrified_door/start()
+	..()
+	if(!chosen_door)
+		return
+	if(prob(33))
+		chosen_door.visible_message("<span class='danger'>\The [chosen_door]'s panel sparks!</span>")
+	chosen_door.set_safeties(0)
+	if(severity >= EVENT_LEVEL_MODERATE)
+		chosen_door.electrify(-1)
```
zekesulastin: Yeah, I somehow read over this without actually writing something down although I vaguely remember talking about it on TS.  In the interest of stuff not getting lost in Discord, can we please have some sort of permanent SOMETHING visibly wrong with the door so that Joe Blow joining after the radio announcement (... is there even a radio announcement?) and not seeing the sparks doesn't get randomly screwed because we decided to run meteor shields?  Thanks.

((Actually, that's a good question - how do you know where to start looking if you don't see the sparks?  Not entirely sure how notifications work for these))

code/modules/gamemaster/actions/electrified_door.dm
```diff
@@ -0,0 +1,70 @@
...
+			if(!chosen_door || !chosen_door.arePowerSystemsOn())
+				return
+			chosen_door.visible_message("<span class='warning'>\The [chosen_door]'s hydraulics creak.</span>")
+			playsound(get_turf(chosen_door), 'sound/machines/airlock_creaking.ogg', 50, 1)
+			sleep(rand(30 SECONDS, 10 MINUTES))
+			if(!chosen_door || !chosen_door.arePowerSystemsOn())
+				return
+			chosen_door.visible_message("<span class='danger'>\The [chosen_door] emits a hydraulic shriek!</span>")
+			playsound(get_turf(chosen_door), 'sound/machines/airlock.ogg', 80, 1)
+			spawn(rand(5 SECONDS, 30 SECONDS))
```
zekesulastin: Going with the above, the door bolt light should blink or there should be a custom announcement or something if actually stopping the explosion is at all desired so that you can find the door even if you miss the 3 random events.  (also, since I cannot fully parse DM, does cutting the power stop the event altogether or do you have to keep it unpowered for 15 minutes to be safe?)

PR #6078, opened 2019-04-13
Fixes resisting out of vehicles causing problems

Heroman3003:
Yeah, without properly removing mobs, it got really weird if somebody tried hitting it later for example. Fixes that.


Neerti:
Vehicle code is a horrid mess and I'm surprised you found a solution.

Heroman3003:
Tbh, solution is far from the best, as it just makes resisting be snowflake in regards to specifically vehicles. But its practical and simple one. Although after taking a look at this one, I dread looking into whole 'vehicles float over open space thing' later on.

PR #6081, opened 2019-04-14
Starts work on Organs affecting Reagent Processing

Mechoid:
Tweaks:
- Hearts can have a variable pulse level, utilizing the standard PULSE_NONE to PULSE_THREADY system.
- Reagent processing speed in the bloodstream is affected by the percentage of your pulse compared to your heart's normal pulse, if your species has a heart. If your heart has no 'pulse' normally, then reagents will process at 80% speed compared to normal.
- Reagents can have 'filtration organs' assigned, which slow the absorption of chemicals into the body through the bloodstream.


Wickedtemp:
Is overall great, making organs more relevant in general is something our medical code kinda lacks. They don't do a whole lot. I do think either having some type of easy-to-use device available in medical or a built-in trait of the OR table which causes the pulse to stabilize would be neato, especially for heart-transplant things.

PR #6092, opened 2019-04-16
Fixes fireaxe cabinets

Heroman3003:
Oh god this is horrible I hate it I hate it so much.

Anyway, should fix some bugs related to it.

No longer a closet either.

EDIT: To be specific. Fixes bugs where it would display wrong state of contents as icon and bug where it would spawn without an axe inside.


Novacat:
One of the pillars of SS13 programmer rage has fallen. The end is neigh.

PR #6100, opened 2019-04-17
Fixes silicate making fulltile windows invisible

Heroman3003:
Tin says it all


Heroman3003:
Tbh, it'd be much better to just have silicate overlay be its own icon, like additional shine or something, with how fulltile windows generate their icon.

PR #6102, opened 2019-04-17
Fixes taperolls not being usable on tiles with directional windows

Heroman3003:
To clarify, if window is directly blocking path of tape, it will still be unusuable. But if you can freely go in direction parallel to tape on same tile, you can now put tape onto it with no problem.

### DIFF
--- a/code/game/objects/items/weapons/policetape.dm
+++ b/code/game/objects/items/weapons/policetape.dm
@@ -140,14 +140,14 @@ var/list/tape_roll_applications = list()
 /obj/item/taperoll/attack_self(mob/user as mob)
 	if(!start)
 		start = get_turf(src)
-		usr << "<span class='notice'>You place the first end of \the [src].</span>"
+		to_chat(user, "<span class='notice'>You place the first end of \the [src].</span>")
 		update_icon()
 	else
 		end = get_turf(src)
 		if(start.y != end.y && start.x != end.x || start.z != end.z)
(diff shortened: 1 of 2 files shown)


code/game/objects/items/weapons/policetape.dm
```diff
@@ -162,10 +162,13 @@ var/list/tape_roll_applications = list()
 					for(var/obj/structure/window/W in T)
 						if(W.is_fulltile() || W.dir == reverse_dir[dir])
 							possible_dirs += dir
+			for(var/obj/structure/window/window in start)
+				if(istype(window) && !window.is_fulltile())
+					possible_dirs += window.dir
```
Atermonera: Possible_dirs is using dirflsgs, so shouldn't it use || instead of += to prevent double-counting a dir?

Atermonera: And shouldn't it also specifically exclude the dir of the window on this turf? Like, if there's a window facing west, you cant take tape in that direction

Heroman3003: This area of code specifically is responsible for placing tape on a single tile, so starting and ending without moving. As a result it cant really be blocked by that window, and code is mostly there to calculate way it should appear.

As for being binary, some testing showed you are right. Was mostly going off the way it was done in nearby part of the file. Fixed that both in new and old.

PR #6106, opened 2019-04-17
ore bags will now automatically pick up ore and dump them into a box if one is being pulled.

chaoko99:
https://streamable.com/7acm4


chaoko99:
Has changelog.

PR #6107, opened 2019-04-17
Adds Config-Controlled Automated Solar Setup

### LINKED ISSUE #6075: Solar Controller Autostart doesn't Autostart
https://github.com/PolarisSS13/Polaris/blob/61df5b880bb7f9047a32f66517b962ac43dcc350/code/modules/power/solar.dm#L535-L541

Contents of this New() Proc should be in a Initialize() Proc so that the autostart tracker actually autostarts.
As of now they only autostart if spawned using the Spawn verb.

Neerti:
Adds functionality to solar computers, to set themselves up on roundstart, based on a new variable. The new var can be set to never start itself, to always start itself, or to check the config.
Tweaks the three finished solar arrays on the Southern Cross map to start themselves based off of the config setting. The config defaults to off.
SMES units in the solar control rooms have also been modified to start active.
Fixes #6075


code/controllers/configuration.dm
```diff
@@ -237,6 +237,7 @@ var/list/gamemode_cache = list()
 	var/radiation_lower_limit = 0.35 //If the radiation level for a turf would be below this, ignore it.
 
 	var/random_submap_orientation = FALSE // If true, submaps loaded automatically can be rotated.
+	var/autostart_solars = FALSE // If true, the
```
Atermonera: Incomplete comment?

Screemonster:
... Could this be set to a situational check like the low-population extended access config option? Would be neat to have solars automatically set up on lowpop shifts.

Neerti:
I don't believe we have automatic extended access coded here right now, that might be a downstream feature.
I know other servers like TG have it, as well.

PR #6110, opened 2019-04-17
Extended now requires one person.

Neerti:
Extended will now require at least two people to ready up, for it to start.


chaoko99:
Will this cause issues with testing, or is this a config only thing?

Neerti:
I'm thinking changing the code to have the minimum amount of required players to be controlled by the config, since this solution seems less than ideal to myself.

PR #6147, opened 2019-05-02
Fixes emergency pods not working

Heroman3003:
Lists are cursed. CURSED I SAY.


Atermonera:
The list is populated with `emergency_shuttle.escape_pods[name] = src`
Which makes it an associative list, for which you have to iterate over the indices.

Heroman3003:
Yeah. Thats what the fix does. Originally it just iterated over entries... and first thing it pulled were strings. Not the references.

Atermonera:
Yeah I don't know how it was supposed to work in days of yore. Two years ago, Neerti ported a refactor from bay, but before that it looks like they were still stored in `escape_pods` as an associative list, so they must've been broken for even longer before that.

PR #6148, opened 2019-05-02
Removes melee damage from snacks.

Verkister:
Not gonna ask why it was there but someone nearly lit the place on fire after jokingly mashing their sandwich at a borg downstream lmao.

### DIFF
--- a/code/modules/food/food/snacks.dm
+++ b/code/modules/food/food/snacks.dm
@@ -16,7 +16,7 @@
 	var/list/nutriment_desc = list("food" = 1)
 	center_of_mass = list("x"=16, "y"=16)
 	w_class = ITEMSIZE_SMALL
-	force = 1
+	force = 0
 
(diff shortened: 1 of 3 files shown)


PrismaticGynoid:
It's there to allow stinging plants, such as nettles, to be used as a weapon. You couldn't use them as such otherwise.

Verkister:
Did anyone ever consider that maybe they could define that var for those specific subtypes rather than _everyting_?

Atermonera:
They can't, because hydroponics procedurally generates the fruit

Verkister:
I think my fix here would make the part that assigns the stinginess to assign the force var as well while at it or something like that (gonna make it so if it doesn't already). Anyway they could have just as well assigned the force to the grown produce subtype in order to get it even a little bit closer to the stinging gimmick rather than just slap it to the _root of all snacks_.

Atermonera:
The real solution was always to rewrite hydroponics, but this'll do

PR #6153, opened 2019-05-05
Restricts camera access on MC

Heroman3003:
Heads and security get full access to networks.

Engineering gets access to Engine and alarm networks.

Rest is restricted.


Neerti:
That feel when you thought MC meant the Master Controller, IE the ticker, and not Modular Computers.

PR #6154, opened 2019-05-05
Sif Trees and Icon Scaling

Mechoid:
Icon scaling is now separated into X and Y directions.

Sif's trees will now scale randomly in the X and Y direction.
Sif trees now have more sprites to pick from, which will also show how much light they have dependant on how many bulbs they have.


Neerti:
Could you show us a picture of what the trees look like now?

Mechoid:
image

code/game/atoms_movable.dm
```diff
@@ -469,13 +470,16 @@
...
 	var/matrix/M = matrix()
-	M.Scale(icon_scale)
+	M.Scale(icon_scale_x, icon_scale_y)
 	M.Turn(icon_rotation)
 	src.transform = M
 
 // Use this to set the object's scale.
-/atom/movable/proc/adjust_scale(new_scale)
-	icon_scale = new_scale
+/atom/movable/proc/adjust_scale(new_scale_x, new_scale_y)
```
Anewbe: Is it possible to set one or both of inputs to 0? If so, might want a check to prevent that.

PR #6155, opened 2019-05-05
Restricts modular laptops to be openable only on tables

Heroman3003:
Not benches or racks either. Similar to how they worked before port.


chaoko99:
I can hardly see why a bench would be restricted. It is, indeed, a flat surface.

Light-Weave:
You should be able to open them on the ground- after all, they are laptops, not desktops. Any flat surface works, and floors are (usually...) flat.

chaoko99:
I'd argue that spacemen must not have elbows, because you can open a laptop in your hands IRL, but that's just me being a bit too much tbh. The intent of the loc check is to make it so you can't open it without setting it down.

Anewbe:
Marking as Staff Input so I remember to bring it up on the 26th, sorry for the delay.

Atermonera:
Per the 26 May staff meeting, we are not interested in this, as you can feasibly open a laptop on any flat surface.

PR #6158, opened 2019-05-07
Fixes bug with custom laptops/tablets

Novacat:
Fixes a problem with custom laptops and tablets not spawning with proper verbs.

Also raises cost of the elite model to 7 loadout points, due to concerns raised.


Atermonera:
The verbs aren't being updated by... tweak_item()?

PR #6159, opened 2019-05-08
Allows objects to be made uncatchable

Heroman3003:
Additional var and check in catching code.

Also fixes missiles fired being catchable what the hell.

### DIFF
--- a/code/game/mecha/equipment/weapons/weapons.dm
+++ b/code/game/mecha/equipment/weapons/weapons.dm
@@ -411,6 +411,7 @@
 	icon_state = "missile"
 	var/primed = null
 	throwforce = 15
+	catchable = 0
 	var/devastation = 0
 	var/heavy_blast = 1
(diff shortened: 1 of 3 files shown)


Neerti:
Not sure if this is a bit redundant with ``catch_chance = 0`` being possible, but this might be better at future-proofing, with regards to mobs being able to influence ``catch_chance`` in the hypothetical future.

chaoko99:
Perhaps it would've been smarter to make the rocket launcher into a projectile weapon though.

Heroman3003:
@Neerti catch_chance is not an item var, but car defined in this proc. The only thing on item that affects catching is 'sharp' var that reduces the catch chance by 50, but also has a lot of its own effects. I would have set catch_chance of item to 0, but items plain dont use that.

Neerti:
@Heroman3003 I'm a dummy then.

PR #6162, opened 2019-05-10
Fixes smasher recipies being microwave ones

### LINKED ISSUE #6141: Bad recipes produce iron ingots
When the microwave receives an incorrect recipe and is supposed to produce a burned mess, instead, an iron ingot comes out.
A burned mess
#### What actually happened
A single iron ingot
#### Steps to reproduce
Stick some random ingredients that don't form a recipe into the microwave. Turn it on and wait. Out comes a delicious warm ingot!
#### Additional info:
 - **Server Revision**: pr-6031 - 2019-04-28
a41e6343b60a7dfe31886bd6644a7ff81efc4b25 
 - **Anything else you may wish to add**: Only affects the microwave. Can still produce burned mess from the grill and oven.

Heroman3003:
Yep. That was a thing.

Fixes #6141

### DIFF
--- a/code/modules/power/singularity/particle_accelerator/particle_smasher.dm
+++ b/code/modules/power/singularity/particle_accelerator/particle_smasher.dm
@@ -31,7 +31,7 @@
 	prepare_recipes()
 
 /obj/machinery/particle_smasher/Destroy()
-	for(var/datum/recipe/particle_smasher/D in recipes)
+	for(var/datum/particle_smasher_recipe/D in recipes)
 		qdel(D)
 	recipes.Cut()
 	..()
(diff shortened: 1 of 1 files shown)


Atermonera:
You could make the microwave recipes a subtype of `/datum/recipe` and add strict subtype enforcement to the `/recipe`? Would be more work, but also more in line with OOP

Heroman3003:
@Atermonera main reason current option was taken is more because every other different machine recipe in the game follows its own type... And I do hold belief that with just how MUCH different smasher recipies are from microwave by nature, its worthwhile to separate them.

Atermonera:
*shrug* I'd consider it preferable to repath all recipes to be subtypes of /recipe purely for class organization, but that would probably end up a much larger refactor, and this at least works as a bugfix

PR #6171, opened 2019-05-22
Phase Weaponry - Energy-->Beam

mistyLuminescence:
Previously: phase weaponry fired energy projectiles, similar to that of the stun revolver, which would impact against windows.
Now: phase weaponry fires beam projectiles, similar to that of the taser, which passes through windows and leaves a visible trail.

You wanna use windows tactically to Not Die on Sif? Go for it.
You wanna not have to break into the Armory every time there's space carp and no Warden/HoS? Sure.
You wanna know where your explorer buddy is shooting so you can join in? Absolutely.
You wanna get your ass handed to you by the Sif wildlife because it won't save you in the slightest? Yep.

The biggest implication of this change is that, like tasers, phasers can now be fired through windows. This means more tactical and engaging explorer gameplay (hey, you can now use windows as cover!), the ability to take out space carp when there's nobody with Armory access, and probably Ian being shot a little more by antags. Sorry, Ian, it's for the greater good.

That's really all there is to it! Minor change, but a welcome one, I think. (You'll still get shredded by spiders, though.)

### DIFF
--- a/code/game/mecha/equipment/weapons/weapons.dm
+++ b/code/game/mecha/equipment/weapons/weapons.dm
@@ -161,7 +161,7 @@
 	description_fluff = "Essentially an Orion mounted inside a modified Gaia case."
 	icon_state = "mecha_phase"
 	energy_drain = 25
-	projectile = /obj/item/projectile/energy/phase/heavy
+	projectile = /obj/item/projectile/beam/phase/heavy
 	fire_sound = 'sound/weapons/Taser.ogg'
 
 	equip_type = EQUIP_UTILITY
--- a/code/modules/projectiles/guns/energy/phase.dm
+++ b/code/modules/projectiles/guns/energy/phase.dm
@@ -8,7 +8,7 @@
 	wielded_item_state = "phasecarbine-wielded"
 	slot_flags = SLOT_BACK|SLOT_BELT
 	charge_cost = 240
-	projectile_type = /obj/item/projectile/energy/phase
+	projectile_type = /obj/item/projectile/beam/phase
 	one_handed_penalty = 15
 
 /obj/item/weapon/gun/energy/phasegun/pistol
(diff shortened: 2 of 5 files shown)


Nalarac:
I just want to point out that you never had to break into the armory to deal with space carp. Carp can be killed with tasers (giant spiders as well), so you really only need the taser that every security officer gets in their locker. 

I believe the math is that tasers do half their stun damage as real damage to these simple mobs (so 20), which would mean 2 shots to kill a carp (25 hp).

Mechoid:
Should note that phase-weapons using beams would mean PoIs with windows might need to be tweaked to prevent easy "standard" clearing, since firing through a window-grille combo is, for one or two animals, functionally a no-risk victory if you have a spare battery.

Woodratt:
Going to request this gets a staff input tag, since this means a rather large balance concern when it comes to explorers and POIs

Anewbe:
Phase weapons were very deliberately _not_ made to pierce glass for balance reasons, namely those Mechoid and Nalarac brought up.

mistyLuminescence:
Forum thread up [here](https://forum.ss13polaris.com/viewtopic.php?f=14&t=791). One thing to note is that laser carbines, available from Cargo, Security or certain PoIs, can fire through windows and have 40 shots at 15 each on suppressive (for a total of 600 damage) versus the 10x50 (500) damage of a phase carbine or the 10x60 (600) of a phase rifle, plus being able to deal with humanoid or robotic threats. Still a flat upgrade, and not too uncommon, even if phasers could pierce windows.

mistyLuminescence:
From the town hall:
- Invalidates the niche the phase weapons fill, negating the actual need to swap for 'better' gear. (Hitscan, windowpierce, total damage output)
- Phase-weapons are currently superior to the example weapon, the carbine, in pure damage per-shot output.

Ideas are ongoing as to *even cooler things which may or may not involve shooting spiders through windows*.

PR #6172, opened 2019-05-22
Makeover Kit

mistyLuminescence:
Adds the makeover kit to the traitor uplink (5 TC) and infiltrator null crate bundle. It's a tiny box with a mirror and some contact lenses inside, for styling your hair and changing your eye color. Pretty good for disguises too, if you can duck into a side room and give yourself a magenta floor-length braid and bright red eyes before Security catches you.

Does *not* change fur or scales color for nonhuman races- it's just the hair and eyes.

### DIFF
--- a/code/datums/supplypacks/contraband.dm
+++ b/code/datums/supplypacks/contraband.dm
@@ -88,7 +88,8 @@
 					/obj/item/weapon/storage/box/syndie_kit/chameleon,
 					/obj/item/device/encryptionkey/syndicate,
 					/obj/item/weapon/card/id/syndicate,
-					/obj/item/clothing/mask/gas/voice
+					/obj/item/clothing/mask/gas/voice,
+					/obj/item/weapon/makeover
 					),
 			list( //the professional,
 					/obj/item/weapon/gun/projectile/silenced,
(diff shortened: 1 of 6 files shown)


Wickedtemp:
Considering you can basically already do this in any mirror, go for it

Atermonera:
Per the May 26th staff meeting, we are not interested in this object as something generally available to crew, especially for only 1 loadout point. We would accept it as an antag-only item for 5 TCs.
Reasons include, but may not be limited to:

- 1 point is distinctly low, for something leaning toward antagonists with sufficient mechanics to replace a common in-game structure.
- Allows someone to change their distinct outward appearance (minus skin-color) at any location for free.
- Mirrors' uses have, generally, no urgency. Radiation, Clone-hair, Choosing the wrong hair at loadout, changing hair for an event, and Traitor gimmicks were brought up.
- Would again, be an -amazing- cheap antag tool.
- (Organic) eye color is likely a security restriction.

mistyLuminescence:
:+1: PR updated, now available for 5 TC (and in the infiltrator bundle in null crates- the one with the chameleon kit and agent ID).

DasIrrlicht:
By the way, can one add it next to the holokit in the illegal ressearch printable by R&D?

Atermonera:
Would require further discussion.

PR #6175, opened 2019-05-25
Fixed mining drill brace deconstruction issue

Heroman3003:
Now will no longer runtime and create frames.


TheFurryFeline:
Deconstructing works, however, I don't think the reconstruction process of the brace itself should require a capacitor, powercell, etc..... That's what's happening when I tested this here.

Heroman3003:
Wrong board was being used. Fixed.

PR #6185, opened 2019-05-30
lowers the minimum treatment bound of medibots.

chaoko99:
Allows medibots to heal damages that are lower than 5 at minimum. Should prevent people wasting doctor time with boo boos.


polaris.dme
```diff
@@ -2642,7 +2642,7 @@
 #include "code\ZAS\Zone.dm"
 #include "interface\interface.dm"
 #include "interface\skin.dmf"
-#include "maps\southern_cross\southern_cross.dm"
+#include "maps\example\example.dm"
```
Novacat: This should probably be reverted.

chaoko99: fix

PR #6194, opened 2019-06-03
Adds two new Cargo crates!

mistyLuminescence:
Crate 1, for 100 points: xenoarcheology tech! An excavation drill, a xenoarch multitool, and an anomaly suit. If you can find a xenoarch shield gen, this crate contains everything you'll need to search for artifacts buried deep in the rock. Requires research access.

Crate 2, for 75 points: prototype 'flexitac' Tactical Light armor! Like regular tactical armor, except warmer, lighter, and a little less protective (40/40/60). It's a flat upgrade from standard-issue armor (40/30/30), but won't save you as much as the 'heavy' tactical armor (60/60/60) also available from Cargo. Instead of having less slowdown (because helmets don't have slowdown anyway), the helmet has a built-in flashlight too! It's as warm as a hoodie, which will save you against light and (most) moderate snow, but a blizzard will still freeze you solid- and don't even think about going into space. The slowdown is 0.2 for the legs and 0.3 for the body- a total of 0.5 (half that of regular tactical armor).

[Flexitac screenshots here](https://imgur.com/a/947zctK), also ft. tactical belt and jumpsuit.

Also, you can now store boot knives in leg guards, and xenoarch pickaxes no longer have disproportionately large force (from 15, down to 1-10). You can also splint the head, chest and groin now, so if there's no surgeon you're not completely screwed.


Woodratt:
Probably needs a staff review tag.

Wickedtemp:
Looks neato

mistyLuminescence:
Updated based on feedback from 09/06 staff meeting- flexitac now has cold protection equal to hoodies and bomber jackets (-20 C, so enough to protect against most moderate snow, but not blizzards) and a combined slowdown of 0.5, and instead of a bone repair injector crate, you can now splint the head, torso and groin as a stopgap measure until Real Surgeons arrive.

mistyLuminescence:
fuck wrong button

Wickedtemp:
"Until surgeons arrive" - as if we have a way to ask for them. I wish we could, it'd be neat

mistyLuminescence:
A 'crew request terminal' outside the HoP line could actually be really cool, but I doubt it'd make people play Surgeon- it's a very rare role outside of a couple of dedicated players. At least a broken head won't be an instant round-ender now.

Anewbe:
This needs to be split into multiple PRs, as per the Coding Rules.

code/modules/clothing/shoes/leg_guards.dm
```diff
@@ -6,6 +6,7 @@
 	species_restricted = null	//Unathi and Taj can wear leg armor now
 	w_class = ITEMSIZE_NORMAL
 	step_volume_mod = 1.3
+	can_hold_knife = 1
```
Anewbe: TRUE/FALSE

mistyLuminescence:
We can remove the DNM now, by the way.

PR #6200, opened 2019-06-06
Fixes some xenoarch bugs

Heroman3003:
Fixes the finds not breaking apart when they should if they are not of xenoarch_find type.

Fixes skulls and bones not spawning when found.

### DIFF
--- a/code/modules/xenoarcheaology/finds/find_spawning.dm
+++ b/code/modules/xenoarcheaology/finds/find_spawning.dm
@@ -605,9 +605,9 @@
 			new_item.origin_tech[TECH_ARCANE] = 1
 			new_item.origin_tech[TECH_PRECURSOR] = 1
 
-		var/turf/T = get_turf(new_item)
-		if(istype(T, /turf/simulated/mineral))
-			T:last_find = new_item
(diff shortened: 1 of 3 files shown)


code/modules/xenoarcheaology/finds/find_spawning.dm
```diff
@@ -605,6 +605,10 @@
 			new_item.origin_tech[TECH_ARCANE] = 1
 			new_item.origin_tech[TECH_PRECURSOR] = 1
 
+		var/turf/T = get_turf(new_item)
+		if(istype(T, /turf/simulated/mineral))
+			T:last_find = new_item
```
Neerti: I know this colon's right after a typecheck and so it probably won't be an issue, but I'd still prefer casting T into a mineral turf, since colon usage is discouraged by pretty much everyone iirc, and it'd only add one more line, but make the code cleaner and less likely for the colon to spread by copypasta.

Heroman3003: The main reason it was kept that way is because this code was copy and pasted from different section of xenoarch to be more broad, but sure.

PR #6202, opened 2019-06-06
Lawyers Are Back!

TheFurryFeline:
Adds a new job slot following a discussion downstream about whether lawyers are different from IAA. Also mapped in spawnpoints and placeholder stamps. May get more sprites for PDA, headset, ID, etc. along the line. Also will need additional teshari outfits for this, but for now I've put in additional access to the IAA ones as placeholders.


Heroman3003:
Ehh... I think return of the lawyers is a bad idea. IAA serves every positive aspect of the job while the negative aspects are being much less in focus.

Also, lawyer has no business ever entering IA offices OR having access to command communication channel. You've just copy-and-pasted IAA there.

I dont think we need "IAA, but instead of mediating, they take one single side and do their darndest to make other side's life harder".

TheFurryFeline:
As stated in the OP, I have several placeholders for a lot of this pull that can get changed as needed.

Haery70:
Popping in from the downstream that OP mentioned, after giving it over 24 hours of thought I feel like that, first of all: It's unnecessary to be making a pr about a downstream-related issue that wouldn't be an issue towards the upstream, regardless of what the staff downstream suggests since the issue isn't also an upstream one, as far as I know.

Second of all: This is literally taking the IAA role and renaming it, why does a Lawyer need the same privileges as the IAA? At the most, they're a third-party entity if they're not corporate-owned, which the names of the role and alt-title suggests. So, they wouldn't be granted the same privileges as the IAAs and would have the same privileges as an Assistant. At the very least, as one of the downstream staff suggested despite being against such an idea, it would be a lot easier to make them an alt-title for IAA. If you plan on altering it further during its review, then it would probably be best to grant them less privileges than the role mentioned to avoid making them just an IAA with a fancy name, despite both positions being similar in a few ways when you take mechanics out of the equation.

While I do like the idea of the role being a part of the jobs description, this pr has a lot of problems and it seems that there're more dislikes to it than likes if I'm judging by the emotes on the OP. For now, I'm not giving out my support for this one.

Ketrai:
This seems entirely unneeded. Just roleplay within a roleplay if you want proper court cases. Instead of risking to draw out security issues over the entire freaking shift.

Atermonera:
Staff meeting on 9-June-19 is unanimously "We object to this." Any further discussion on lawyers is best left to the downstreams that actually discussed it in the first place.

TheFurryFeline:
Alright, thank you for your consideration with this addition.

On Jun 9, 2019 17:37, "Atermonera" <notifications@github.com> wrote:

Staff meeting on 9-June-19 is unanimously "We object to this." Any further
discussion on lawyers is best left to the downstreams that actually
discussed it in the first place.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/PolarisSS13/Polaris/pull/6202?email_source=notifications&email_token=AJGMTYZ5WUUI2KOS56W4FRDPZVZZNA5CNFSM4HVKJ4P2YY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGODXISYNA#issuecomment-500247604>,
or mute the thread

PR #6206, opened 2019-06-08
tinytweaks

mistyLuminescence:
- tweak: "When the supermatter explodes, it now leaves behind a shattered plinth that continues to emit radiation. You should probably get rid of it."
  - bugfix: "If you destroy the supermatter early (e.g. through admin deletion shenanigans), it no longer irradiates everyone forever."
  - bugfix: "Welding lockers now actually updates their sprite properly."
  - tweak: "Every floor tile now has a minor (2%) chance to spawn with some dirt on it. The Janitor now has a job again."
  - tweak: "Similarly, every Sif grass tile now has a minor (5%) chance to spawn with some fancy eyebulb grass on it. It can be removed by clicking on it with harm intent."
  - rscadd: "Adds cats-in-boxes, which can be activated (once) to spawn cats. There's an orange tabby (Cargo cats) and a tuxedo (Runtime) version."
 - bugfix: "Fixes a material duplication exploit." 

it's 10:34 AM i'm sleepy so just have the changelog copy-pasted
cats-in-boxes can, in theory, be used to create custom item cats, if you're fine with your cat being either tabby or tux (note: only tux sprites exist for wearing cats so. go with tux)

Edited with requested changes and exploit fix from #6212 .


Woodratt:
> tweak: "When the supermatter explodes, it now leaves behind a shattered plinth that continues to emit radiation. You should probably get rid of it."

Not exactly what I asked or what the bug was. The bug was attempting to destroy the SM early would result in radiation being around forever. Not sure how this is a fix for it, but eh.

mistyLuminescence:
Whoops, misunderstood. Fix for the early-destruction bug is included too.

code/game/objects/structures/flora.dm
```diff
@@ -341,3 +341,11 @@
 /obj/structure/flora/sif/eyes/Initialize()
 	icon_state = "[initial(icon_state)][rand(1,3)]"
 	. = ..()
+
+/obj/structure/flora/sif/attack_hand(mob/user)
```
Anewbe: Is this meant to be specific to the eye plant? If so, you used the incorrect path. If not, please move the proc upwards into the 301-ish line, to keep it directly after the `structure/flora/sif` definition.

Additionally, the `visible_message` proc has inputs for what the user sees and what everyone else sees, those really ought to be used.

code/modules/mob/living/simple_mob/subtypes/animal/pets/cat.dm
```diff
@@ -121,16 +126,44 @@
 	gender = MALE
 	icon_state = "cat3"
 	item_state = "cat3"
-	icon_living = "cat3"
-	icon_dead = "cat3_dead"
-	icon_rest = "cat3_rest"
+	named = 1
```
Anewbe: Use `TRUE` or `FALSE` for things that only ever have two states.

code/modules/mob/living/simple_mob/subtypes/animal/pets/cat.dm
```diff
@@ -121,16 +126,44 @@
...
+		if(named)
+			to_chat(user, "<span class='notice'>\the [name] already has a name!</span>")
+		else
+			var/tmp_name = sanitizeSafe(input(user, "Give \the [name] a name", "Name"), MAX_NAME_LEN)
+			if(length(tmp_name) > 50)
+				to_chat(user, "<span class='notice'>The name can be at most 50 characters long.</span>")
+			else
+				to_chat(user, "<span class='notice'>You name \the [name]. Meow!</span>")
+				name = tmp_name
+				named = 1
```
Anewbe: Use `TRUE` or `FALSE` for things that only ever have two states.

code/modules/mob/living/simple_mob/subtypes/animal/pets/cat.dm
```diff
@@ -121,16 +126,44 @@
...
+	else
+		..()
+
+/obj/item/weapon/catinabox
```
Anewbe: Prefer the path to be `cat_box` for readability.

code/modules/power/supermatter/supermatter.dm
```diff
@@ -150,6 +151,8 @@
 				H.hallucination += max(50, min(300, DETONATION_HALLUCINATION * sqrt(1 / (get_dist(mob, src) + 1)) ) )
 	spawn(pull_time)
 		explosion(get_turf(src), explosion_power, explosion_power * 2, explosion_power * 3, explosion_power * 4, 1)
+		sleep(5) //to allow the explosion to finish
```
Anewbe: I think this should probably be a `spawn` proc.

code/modules/mob/living/simple_mob/subtypes/animal/pets/cat.dm
```diff
@@ -90,24 +95,24 @@
 	gender = FEMALE
 	icon_state = "cat"
 	item_state = "cat"
-	icon_living = "cat"
-	icon_dead = "cat_dead"
-	icon_rest = "cat_rest"
+	named = 1
```
Anewbe: Use TRUE or FALSE for things that only ever have two states.

code/game/machinery/autolathe.dm
```diff
@@ -19,6 +19,7 @@
 	var/busy = 0
 
 	var/mat_efficiency = 1
+	var/coeff = 1
```
Anewbe: Has this entirely replaced `mat_efficiency`, or what purpose do the two now serve? In-code comments are very helpful for this sort of thing.

Additionally, if this is a 1-or-0 thing, it should be TRUE/FALSE.

code/datums/autolathe/autolathe.dm
```diff
@@ -66,6 +67,7 @@ var/datum/category_collection/autolathe/autolathe_recipes
 	var/power_use = 0
 	var/is_stack
 	var/max_stack
+	var/no_scale
```
Anewbe: If this is a 1-or-0 thing it should use TRUE/FALSE, rather than numbers. I'm not seeing anywhere that this specifically _isn't_ set to 1, but I'll mark them just in case.

code/datums/autolathe/engineering.dm
```diff
@@ -93,6 +93,7 @@
 /datum/category_item/autolathe/engineering/rcd_ammo
 	name = "matter cartridge"
 	path =/obj/item/weapon/rcd_ammo
+	no_scale = 1 //prevents material duplication exploits
```
Anewbe: TRUE/FALSE?

code/datums/autolathe/general.dm
```diff
@@ -74,21 +74,25 @@
 	name = "steel sheets"
 	path =/obj/item/stack/material/steel
 	is_stack = 1
+	no_scale = 1 //prevents material duplication exploits
```
Anewbe: TRUE/FALSE?

code/datums/autolathe/general.dm
```diff
@@ -74,21 +74,25 @@
...
 
 /datum/category_item/autolathe/general/glass
 	name = "glass sheets"
 	path =/obj/item/stack/material/glass
 	is_stack = 1
+	no_scale = 1 //prevents material duplication exploits
```
Anewbe: TRUE/FALSE?

code/datums/autolathe/general.dm
```diff
@@ -74,21 +74,25 @@
...
 
 /datum/category_item/autolathe/general/rglass
 	name = "reinforced glass sheets"
 	path =/obj/item/stack/material/glass/reinforced
 	is_stack = 1
+	no_scale = 1 //prevents material duplication exploits
```
Anewbe: TRUE/FALSE?

code/datums/autolathe/general.dm
```diff
@@ -74,21 +74,25 @@
...
 
 /datum/category_item/autolathe/general/rods
 	name = "metal rods"
 	path =/obj/item/stack/rods
 	is_stack = 1
+	no_scale = 1 //prevents material duplication exploits
```
Anewbe: TRUE/FALSE?

code/game/machinery/autolathe.dm
mistyLuminescence: Numeric value, but I just realised we don't actually need it defined in this line anyway.

Anewbe: That doesn't explain the purpose of the variable, which was the more important part of my last comment.

mistyLuminescence: The comments in autolathe.dm (line 9) and the other autolathe.dm (line 86 and 255) do.

PR #6208, opened 2019-06-10
Serenity

mistyLuminescence:
What do you get when you mix a Gygax and an Odysseus?

If you answered "a lawsuit from VeyMed", you'd be right! But not any more- for they've finally released their designs for the *Serenity*, a modification on the classic Gygax security mech. Lose all that bulky armor plating and swap most of the weapons for an onboard medical analysis suite, and the Serenity is born.

True to its namesake, it's not made for fighting- so despite being based on a Gygax (and therefore able to equip one weapon), it's still almost as paper-thin as the Odysseus, making that beloved leg actuator feature a little more dangerous to use. When you've only got 150 health, every little sliver counts- but in the medical field, time is vital, and those few seconds might make a difference.

Just as the Durand is the natural evolution of 'a Ripley with a gun duct-taped to it', so too is the Serenity the natural evolution of the Odysseus. Is it fragile? Sure. Will it run out of power fast? Totally. Will Security try and arrest you because "hurr no mechs allowed"? Probably. But is it undeniably sexy as hell?

[You bet it is](https://cdn.discordapp.com/attachments/176163422751293441/587554323936837645/unknown.png).
 
Honestly, if you can afford it, you're probably better off with a Gygax for anything other than purely medical rescue missions, but that's the niche the Serenity fills- reliable, effective, and able to leap tall buildings without getting a wedgie. Get yours today for only three ninety-nine ninety-nine ninety-nine, all rights reserved.


Wickedtemp:
Related to mechs, but having a somewhat ranged healing option for mechs would be neat. 

We do have a syringe gun, and it doesn't seem to do damage or have an embed chance when you shoot at someone so all they'd get is whatever reagent its loaded with, but from an IC standpoint... you're still shooting your patient with syringes. 

And its clunky as hell because you have to analyze reagents beforehand, wait for the mech to create the reagents, etc and in a situation where everything around you is going on faster than your mech is capable of handling, its honestly pointless. Syringe gun can't even use harmful reagents for antags so there's no real point in it

Anewbe:
@Wickedtemp that's a suggestion for another time, place, or thread. In theory I like it. DM me the ideas you might have, and I'll see what can be done about it.

As for this PR itself, the Mech Rework was only approved after a long and very specific discussion about preventing exactly this sort of thing from being possible. That said, I do kind of want to see this thing in a _controlled setting_, so if you remove it from what the players can access, it may be reconsidered.

mistyLuminescence:
What *is* "this sort of thing", why is preventing it required, and what would make the Serenity viable for adding as an actual feature rather than a never-seen adminspawn mech?

It's explicitly not a combat exosuit- with its low health, minimal armor and single weapon slot, it'll prove ineffective in a proper fight (though better than nothing). It's more suited towards SAR operations and as a shiny upgrade to the Odysseus once Robotics gets materials. The main selling point is probably that shitcurity won't murder you for having it.

Wickedtemp:
...if it's not going to be accessible by players, then what's the point. I'm not sure what issues a "slightly better Ody" would bring to the table, especially with how little use mechs provide from a medical perspective.

Anewbe:
"This sort of thing" is a mech that does both combat and medical things, by design. Frankly nothing you can do to this PR will make this mech not become an issue.

You have a long history of drastically understating the capabilities and intricacies of the code you try to get in here, and I don't feel the need to expend further energy discussing it with either of you.

Either make it Aspawn only so we can try before we buy, or this will be close. You have 24 hours.

Wickedtemp:
...You can already 'try before you buy'... it's open-source, you can look at the code, compare it to existing mechs, and you can just... merge this into a private server and take it for a test drive to your hearts content.

Anewbe:
> I don't feel the need to expend further energy discussing it with either of you.

One final point before I lock this conversation and await a code change, then. There's a massive difference between seeing the code on GitHub and seeing what people do with the code once it's in game, and it's far more frustrating to attempt to correct things after the fact.

code/modules/research/designs/circuits.dm
```diff
@@ -468,6 +468,13 @@ CIRCUITS BELOW
 	build_path = /obj/item/weapon/circuitboard/mecha/gygax/targeting
 	sort_string = "NAACC"
 
+/datum/design/circuit/mecha/gygax_medical
```
Anewbe: This needs to be commented out.

PR #6209, opened 2019-06-11
June Map Adjustments

Woodratt:
- Turrets and respective controls added to each teleporter room
- Air tank on shuttles switched out with an air canister
- Fix for Dock 2 Transfer Shuttle airlock buttons not working
- Intercoms added to arrivals shuttle
- Fix Bridge Secretary Quarters tint
- Roundstart Tcomms SMES starts fully charged.
- Landmarks for Bluespacerift added
- Large Air tanks on station have the same amount in them as air canisters
- Expanded atmospheric tanks to be 2 by 3 instead of 2 by 2
- Wilderness Shelter Area added
- Wilderness Shuttle landing point shelter updated
- Wilderness Shuttle landing sides should no longer shiskabob shuttles
-Teleporter room hand tele can now spawn in one of 4 places
- Halved the warm up time when the autopilot on the shuttles start at round start from ten minutes to five and nine respectively. Should mean the first shuttle will be taking from the outpost when the second shuttle leaves the station
- Amount of phoron in the outposts reduced
- Current transit time 75 seconds with pilot, 150 seconds without. Reduced to 60 and 120 respectively.
- Reduction of the amount of steel and glass in engineering by half
- Reduction of the amount of steel and glass in EVA by half
- Ground Xenobio and Xenoflora torn down partially
- Gateway can no longer be used to exit a round (Can still be used to enter a round)
- Elevator can be accessed at centcomm


mistyLuminescence:
Copying feedback from discord- instead of making the teleporters inaccessible through turrets, make the shuttles more appealing. The appeal of the teleporter is its speed- so what can we give shuttles to make them equally viable as transit options? Reducing the shuttle delay is a start, but won't make them outclass teleporters.

Perhaps giving shuttles the ability to travel to the wilderness without a pilot, and when a pilot is present, allow for pinpoint shuttle placement and custom landing zones? Teleporter for speed, shuttle for precision.

"The teleporters shoot you to death if you try and use them" won't really make shuttles any more viable, after all, but "you can land directly next to BSD and use the shuttle as cover from the mercs" will.

zekesulastin:
Unless the gateway turns back on when the departure shuttle is active, might just be better off unmapping it altogether if nerfing it is because of "realism".  Doesn't really make any realistic sense to arrive there from the planet then take the shuttle (or lolcryo) off.

mistyLuminescence:
Slightly better idea regarding the gateway- have it require being on a wire dot (similar to a PACMAN) in order to function, allowing antags to cut the wire and easily sabotage the gateway if they really want to prevent someone from escaping.

The gateway has a ton of legitimate uses, and if an antag wants to do something with 'nobody can escape, mwuhahaha' they just have to sabotage the gateway. And if they want people to have the opportunity of escaping, then they can.

Mechoid:
> Gateway can no longer be used to exit a round (Can still be used to enter a round)

This should really be its own PR for discussion, as it is highly contentious.

Woodratt:
Its in its own PR now.

PR #6211, opened 2019-06-11
Nerfs NTNet Relay power usage

Novacat:
...It probably should not be drawing as much power as the rest of telecommunications combined. 1600 puts it as equivalent of the Telecommunications hub, which is, IIRC, the most power-intensive machinery in telecomms.

### DIFF
--- a/code/modules/modular_computers/NTNet/NTNet_relay.dm
+++ b/code/modules/modular_computers/NTNet/NTNet_relay.dm
@@ -3,7 +3,7 @@
 	name = "NTNet Quantum Relay"
 	desc = "A very complex router and transmitter capable of connecting electronic devices together. Looks fragile."
 	use_power = 2
-	active_power_usage = 1600
+	active_power_usage = 2000 //20kW, apropriate for machine that keeps massive cross-Zlevel wireless network operational.
 	idle_power_usage = 100
 	icon_state = "bus"
 	anchored = 1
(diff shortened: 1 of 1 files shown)


Anewbe:
The comment on the old version leads me to believe that the large power drain was very much intentional. I'm curious as to what exactly lead to this PR's creation, and question its necessity.

Novacat:
The main issue is that the NTNet Relay uses almost as much power as the rest of telecoms combined. The original power usage (and the comment) were imports from Bay, which, I am guessing Bay might have adjusted the power consumption of their telecoms machines.

Anewbe:
To reiterate, I don't see the large power drain as something that particularly needs to be altered. If you have a reasoning beyond "It seems odd" it may change my mind, but as things stand I don't think this needs to be done.

Novacat:
The main concern from this is that on the Tether map, the passive total power consumption across the whole map is very close to consuming more power than the primary power + solar power generation combined. Not sure if that issue applies to Southern Cross. Figured that, since the NTNet relay is consuming an abnormally large amount of power compared to other telecoms devices, that it should be tweaked upstream.

Anewbe:
It's really not an issue for us, should be adjusted on your end.

mistyLuminescence:
We do have a similar issue in that telecomms is guaranteed to fail on solar power if left unattended for long enough- increasing the SMES just prolongs its lifespan, and doesn't guarantee it. This also means it's possible to join a new round where telecomms (and most of deck one) is depowered. Given that our equivalent of a glorified wifi modem uses almost half the total power draw from telecomms, it might be worth fixing the inconsistency for the sake of new rounds alone.

Anewbe:
I'll keep an eye out for that, and if it's deemed necessary, changes will be made. Until then, this should be applied to the codebase where it's directly relevant, but thank you for bringing it to our attention.

PR #6213, opened 2019-06-12
Prescription Pills

mistyLuminescence:
Adds a bottle of prescription medicine to the loadout for 1 point. It does absolutely nothing, but if you want to pop some pills for RP value, you can. If you want *actual medicine* that has mechanical effects, go see the chemist- these just taste faintly of banana.

Will conflict with #6199 - if that gets merged first, I'll update this to match.

### DIFF
--- a/code/game/objects/items/weapons/storage/firstaid.dm
+++ b/code/game/objects/items/weapons/storage/firstaid.dm
@@ -271,3 +271,8 @@
 	name = "bottle of Iron pills"
 	desc = "Contains pills used to aid in blood regeneration."
 	starts_with = list(/obj/item/weapon/reagent_containers/pill/iron = 7)
+
+/obj/item/weapon/storage/pill_bottle/rx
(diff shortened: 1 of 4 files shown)


Cerebulon:
Leaving it this totally generic seems totally pointless and lazy rather than working on a proper new mini-feature.
I've been meaning to do (Soon, once I'm not doing exams) a proper meds-in-loadout/more flavour stuff for chemistry to Dispense thing for a while which would include existing psych drugs, plus a couple more, plus various common prescription drugs (Some with effects equivalent to the psych drugs where they just give a message, some with very minor effects) like mild painkillers, anti/coagulants, insulin, various hormone supplements, immunosuppresants etc.

mistyLuminescence:
Yeah, I'm not 100% sold on generic medicine either, but giving people actual meds seemed like a balance concern (starting with free tramadol, anyone?). If you've got Cunning Plans™, this one's all yours.

PR #6215, opened 2019-06-14
THIS PR IS 7+ DAYS OLD, SEE #6203 . Energy-Swords and Changeling Melee Weapons Can Deflect Projectiles

Mechoid:
Energy-Swords and Changeling Melee Weapons Can Deflect Projectiles, see: #6203

> On the tin.

Energy swords, changeling claws / blades, and ninja swords can now deflect projectiles (in some cases, again).

Recursively-Enhanced changeling melee weapons have greater blocking capability, both for melee and projectiles. Claws are, generally, inferior to arm-blades due to size.

Adds an over-rideable proc for parrying.

This time, not master branch.


code/game/gamemodes/changeling/powers/armblade.dm
```diff
@@ -62,6 +62,9 @@
 	var/weapType = "weapon"
 	var/weapLocation = "arm"
 
+	var/defend_chance = 30	// The base chance for the weapon to parry.
```
Anewbe: If you're going to use this var on multiple items, you may want to put it on the parent class, which in this case would be at least `obj/item/weapon/melee`

Also the `projectile_parry_chance` on everything that has it seems to be a tad high, would probably be better to start it low and increase it if the need presents itself.

PR #6217, opened 2019-06-14
POI Names

mistyLuminescence:
Adds the "POI - " prefix to the POI location names to make it easier for ghosts to teleport to them- e.g. "POI - Rocky2" rather than just "Rocky2". Very simple PR, and it should make things a little easier for anyone who wants to see a huge list of POIs.

Before:
Abandoned Shuttle
Armory
Dig Site
Engineering
Medbay
PascalB

After:
Armory
Engineering
Medbay
POI - Abandoned Shuttle
POI - Dig Site
POI - PascalB

(etc)

### DIFF
--- a/html/changelogs/mistyLuminescence - POInames.yml
+++ b/html/changelogs/mistyLuminescence - POInames.yml
@@ -1,0 +1,36 @@
+################################
+# Example Changelog File
+#
+# Note: This file, and files beginning with ".", and files that don't end in ".yml" will not be read. If you change this file, you will look really dumb.
+#
+# Your changelog will be merged with a master changelog. (New stuff added only, and only on the date entry for the day it was merged.)
+# When it is, any changes listed below will disappear.
+#
+# Valid Prefixes: 
+#   bugfix
+#   wip (For works in progress)
+#   tweak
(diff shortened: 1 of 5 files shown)


Woodratt:
Slightly concerned this might make it easier for someone to simply ghost, search through the PoIs for where the good loot is at, then join the round and beeline for those PoIs. That has been an issue in the past, and I’m not entirely certain that it is a good idea to make that easier for folk. As it stands now you kind of need to already know the name in order to search for a certain PoI.

There’s that, and it is a tad immersion breaking for anyone with a health scanner, but that’s a bit more minor.

mistyLuminescence:
Someone who wants to do that could already do it, because they'd likely know the names of the PoIs they're after anyway (e.g. Crashed Alien Vessel). Sure, it takes twenty seconds less now, but it's not like they could respawn immediately anyway, and if they're abusing metaknowledge they get a big fat slap. I'll concede it might be a little jarring to see "POI - [name]", but no more than seeing "Rocky2" or "Snow5" or "DoomP". Maybe the sensor console just prefixes it for the convenience of whoever's operating it, I guess.

PR #6219, opened 2019-06-17
Tweaks the NewTM event system further

Mechoid:
Tweaks the NewTM event system further, adds a weather shift event for when people are groundside and outdoors.

Appendicitis is now unable to occur without any, and incredibly unlikely to occur with one, MD.

Airlock event will now make an alert equivalent to the light-bursting event. Moderate or higher instances will shortly send an announcement regarding the localized area.

2 metric counting procs have been added, for counting the number of active players outdoors on the planet, or in space.

New event added utilizing the outdoor counting metric, a sudden weather shift. Incredibly unlikely to occur with only one person outdoors, becomes more likely the larger the group.


zekesulastin:
I'm still concerned about the door shocker being a blind up-to-120-burn-to-one-organ-surpriiiiise to players compared to most moderate events, but the random afterwarning helps a bit, I guess...?  Interestingly, lacking engineering staff to fix the problem will generally mean you are less screwed >.>

I wonder if there is a way to trigger a grid check to clear the electrification without having to go around and smack possibly every door on the station with insulated gloves; having to check for safeties is also not great but those hurt a lot less.

Mechoid:
The event does damage the door, making its panel spark.

zekesulastin:
AS IT TURNS OUT I CANNOT READ CODE WHATSOEVER AND THOUGHT THAT THE SPARKS MENTIONED WERE A ONE TIME OBJECT SPARKS NOT BREAKING THE CONTROL PANEL SPARKS ON THE ICON DISREGARD THINGS

PR #6221, opened 2019-06-19
Head/chest/groin splints

mistyLuminescence:
As of when this PR was made, it was already 16 days old - see #6194.

Splints now work on all areas of the body. You're no longer forced to cryo and respawn if you break a rib and there's no surgeon.


mistyLuminescence:
i can't count

PR #6225, opened 2019-06-21
Teleporter Turret Adjustments

mistyLuminescence:
So you know the PR that added turrets to the teleporters? Yeah, now you can't actually use the teleporters at all if you're unlucky. The turrets are usually non-dense, but if they're active or have been active recently, they're dense- which means they block the teleporter station, preventing the teleporter from being activated or test-fired. The turret controls use AI upload access rather than teleporter access, meaning that if someone gives you teleporter access you'll still be mercilessly gunned down by turrets. There's no way of telling if the turrets are active before entering the room, so you might be wasting your time - and pain. There's also no way of disabling the turrets from outside the room if they've been set to laser anyone who goes nearby (unlike telecomms and the AI upload, which both have exterior turret controls). Finally, if by some miracle you do get the teleporter working and you use it to travel to another teleporter, you'll immediately be shot by that teleporter's turret- and if you don't have access, you won't be able to hack out either, because you're too busy being shot by turrets.

This PR fixes all that by implementing the following changes:
- The turret control panel is now accessible from both the inside and the outside of the teleporter rooms (main tele, tcomms, outpost).
- The turrets have been moved to the corners of the room, giving them a good firing angle while not blocking access to the teleporter itself.
- The turret control panel now uses teleporter access rather than AI upload access.
- The turrets start deactivated, and require manual activation via synthetic unit or authorised personnel. This is perhaps the most important change here, as it means a player must actively decide to enable the turrets. This should only be done in situations that warrant it (e.g. an armed murderer is evading capture and may escape through the teleporter)- turrets are not a toy.

Note that emagging the turret will still prevent it from being controlled using the panel and will set it to light up anyone who enters.


Woodratt:
The turrets are set up to prevent unauthorized access to the teleporter room. Of these changes the only changes I'm inclined to agree with is

> The turret control panel is now accessible from both the inside and the outside of the teleporter rooms (main tele, tcomms, outpost).

Yet also against since I have a feeling this'll result in them being emagged more often then not since it'll be a target of opportunity.

> The turret control panel now uses teleporter access rather than AI upload access.

Which seems fine.

The other bits are non-issues since the turrets will only be active if you have the authority to be in there or antag fun stuff is occurring, which seems fine with me. Since the whole purpose of the bloody turrets are to prevent folks that don't have the authority to be in there from using the things. Starting deactivated is entirely against the reason they were added in the first place.

mistyLuminescence:
Regardless of if giving the teleporters their own turrets is consistent or sensible, having them be active at roundstart is a contrivance made solely to fuck people over. If you take a teleporter *into* a teleporter room, you won't be able to escape if the turrets are targeting you. And if there's nobody with access to the teleporter, you're boned.

At least now if the turrets target you, you know it's because some antag, comdom, or malicious AI/borg decided to turn them on. And if someone turned them on, that means there's someone out there with the access to turn them off and save you from your permastun predicament- in contrast to the current 'better ghost and respawn while you can'. If there's nobody with access to enable the turrets? Then the turrets *should not be enabled*.

Wickedtemp:
Adding turrets never made sense to start with anyway.

Anewbe:
I marked this PR as Requires Staff Input specifically to get in contact with Woodrat on the topic. Having done so, I find this PR unnecessary and unwanted. If Woodrat cares to direct you to make changes to what's included here, he's welcome to do so, otherwise I will be closing this in the next day or so.

Wickedtemp:
I mean you say unwanted and unneeded but I kinda want it, and wouldn't want teleporters to be completely disabled every round.

mistyLuminescence:
If the criteria for a PR being merged is "wanted and needed" (which would make sense), I distinctly remember objections and concerns being raised to the "let's give teleporters their very own turrets set to permastun anyone without access" idea when it was first suggested. Like the change to the gateway that was proposed, this is exactly the kind of controversial issue that warrants a forum thread *before* being implemented. We seem to have been moving on from the "my way or the highway" trend recently, and adjusting the most egregious parts of the change (see: this PR) while the community can provide input would be the next step in continuing this improvement.

mistyLuminescence:
Better PR incoming, brace for impact.

PR #6229, opened 2019-06-23
Teleporter 2: Electric Boogaloo

mistyLuminescence:
Here we go! Told y'all there'd be something good coming.

First up- you can now construct and deconstruct teleporters. You'll need certain subspace parts from telecomms storage or R&D, as well as high bluespace levels for the circuits. The hub uses a treatment disk, and the station uses an ansible and filter. 

Next, teleporter stations can now be locked and unlocked with a teleporter-access ID. A locked teleporter cannot be changed from its current state (if it's disabled, it stays disabled; if it's enabled, it stays enabled). Teleporters start locked- so you can only use them if you have the right access on your ID. Stationbound units (AIs/borgs) are able to bypass this restriction and toggle a teleporter on or off regardless of its lock state (unless the wires have been tampered with).
 
Next, teleporter stations now have six wires- one dummy wire, one electricity wire, one that controls the lock, one that enables and disables it, one that test-fires it, and one that can be pulsed to randomise the chances of someone being spaced. (Don't pulse that one.)

The dummy wire does nothing.
The electricity wire creates a cloud of electric smoke when pulsed, and electrifies the station when cut (like airlocks).
The lock wire will toggle the lock on/off and randomise the space chance (like the randomise wire) when pulsed, or fix the lock in that position when cut- so if you lock it and cut the lock wire, it cannot be unlocked, and vice versa.
The enable/disable wire will turn the teleporter on and off when pulsed, or completely disable the teleporter when cut.
The testfire wire will testfire the teleporter when pulsed, or disable testfiring when cut- you can still testfire it, you just won't gain any benefit from it. Deadly when combined with...
... the randomise wire, which randomises the chance of someone being spaced when pulsed. (Cutting it and mending it restores it to the base 5%.) Teleporters won't space you if they're testfired, though- unless, of course, you also cut the testfire wire!

Sound complicated? Good. You shouldn't be tinkering with teleporter wires unless you're willing to get spaced. If someone's tampered with it in advance, you're probably going to want an engineer or scientist to take a look at it and fix it... or, y'know, just take it apart and put it back together again. That works too.

Finally, because the teleporters are now secure on their own, NanoTrasen has removed the turrets from the teleporter rooms. We don't need them any more, since the teleporters start out ID-locked, and it's a lot less immersion-breaking to have a simple ID lock than it is to shoot people who don't have authorisation.


Anewbe:
I'm preemptively locking the conversation on this PR before it becomes contentious or heated. It will be discussed at the next Staff Meeting.

code/game/machinery/teleporter.dm
```diff
@@ -221,8 +228,9 @@
 		var/datum/effect/effect/system/spark_spread/s = new /datum/effect/effect/system/spark_spread
 		s.set_up(5, 1, src)
 		s.start()
-		accurate = 1
-		spawn(3000)	accurate = 0 //Accurate teleporting for 5 minutes
+		if(testfire_broken == FALSE)
+			accurate = 1
+			spawn(3000)	accurate = 0 //Accurate teleporting for 5 minutes... if the testfire system works
```
Mechoid: Considering tickrate is configurable, I'd use `5 MINUTES` instead.

code/game/machinery/teleporter.dm
```diff
@@ -209,7 +216,7 @@
 			O.show_message("<span class='warning'>Failure: Cannot authenticate locked on coordinates. Please reinstate coordinate matrix.</span>")
 		return
 	if(istype(M, /atom/movable))
-		if(prob(5) && !accurate) //oh dear a problem, put em in deep space
+		if(prob(failprob) && !accurate) //oh dear a problem, put em in deep space- and if someone's tampered with the teleporter, even testfires won't help
```
Mechoid: This should probably have a check for if `testfire_broken` is true or false.
I.E. `(prob(failprob) && (!accurate || testfire_broken))`

Anewbe:
Closed as per discussion in the July 7th Staff Meeting, we may rebuild parts of this later.

PR #6242, opened 2019-06-28
Adds Arcane Tech to Cult Items

Nalarac:
Cult blade gets combat 1 and arcane 1
All cult armor gets material 3 and arcane 1.
Soul stone received arcane 1


mistyLuminescence:
Think we could add soulstone shards to this? I know they exist as a random xenoarch drop.

Anewbe:
Marked as Staff Input, I'd like to at least bring it up for brief discussion at the next Staff Meeting.

PR #6244, opened 2019-06-28
Fixes glucose crate spawning

Heroman3003:
And brings some consistency in defining crate type for supplypacks along the way.


Anewbe:
Has this been tested?

Heroman3003:
Yes. Spawns crate without runtimes. Problem was missing / in front of path. As for the "" around paths, those were entirely optional as about half packs had them and another half didn't, and this is simply bringing it all up to uniform standard.

PR #6252, opened 2019-07-01
Fixes ghost riders

Heroman3003:
--...and other riders. Fixes the bike verbs to allow human-only interaction.--

Fixes to allow only living non-mice to interact, rather than human-only. Still fixes ghost riders.

### DIFF
--- a/code/modules/vehicles/bike.dm
+++ b/code/modules/vehicles/bike.dm
@@ -59,7 +59,7 @@
 	set category = "Vehicle"
 	set src in view(0)
 
-	if(!istype(usr, /mob/living/carbon/human))
+	if(!isliving(usr) || ismouse(usr))
 		return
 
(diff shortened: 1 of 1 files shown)


Anewbe:
Code's fine; Mechoid may have further thoughts on this, so I'm going to let him give input before I give any further thoughts.

Mechoid:
I'm mostly concerned about removing more things that borgs can use, since the spacebike is the only vehicle currently operable by them. (Which is something I would like to expand upon in the future)

Heroman3003:
Well, I mostly added the check matching the other vehicles moreso than just blocking off. That lack of check is reason ghosts could do stuff. I will just limit it to living/not-mouse instead if thats desired.

Mechoid:
You are my hero

PR #6257, opened 2019-07-01
Removes extraneous verbs from blob types that can't use them

spacemen-on-a-space-station:
In porting blob to urist I noticed that blobs that couldn't spawn factories still got 2 verbs for spawning factories, same for nodes, same for resources, and it resulted in a very cluttered verb tab. 

Not sure how often you guys even use player blob but this PR removes verbs from blob types that can't use them.


Mechoid:
I'm mostly concerned about losing those verbs in blobs due to the fact busing them would become more difficult. Busing referring to editing them in-round.

Mechoid:
Closing until further notice, as no response has been received for over a month.

PR #6262, opened 2019-07-03
Civilian Expansion 1: The Plantening

Mechoid:
Stage 1 of a Mechoid ™️ project.

Adds a large number of plants, food, and drinks requested here: https://forum.ss13polaris.com/viewtopic.php?f=14&t=804, though not all of them are 100% matching with the suggestions.

Fixes flower crowns, and makes them slightly more 'modular' at least to the point of not being hair-pulling. They use the flowers now instead of ONLY EVER THE BASE SEED AAAA

Examples: Sunflower seeds, roasted sunflower seeds. Kudzu Donburi. PB&J.
Plantxamples: Vanilla. Wabback. Durian, with a brand new reagent effect, 'paste', that leaves effects on the ground that smell horrific.


PrismaticGynoid:
Why would the peanut butter need milk to make? It doesn't actually contain any butter.

code/modules/hydroponics/seed_datums.dm
```diff
@@ -602,12 +648,103 @@
 	..()
 	set_trait(TRAIT_PRODUCT_COLOUR,"42ed2f")
 
+// Lettuce/varieties.
+/datum/seed/lettuce
+	name = "lettuce"
+	seed_name = "lettuce"
+	display_name = "lettuce"
+	chems = list("nutriment" = list(1,15))
+	kitchen_tag = "cabbage"
```
Atermonera: *squint*

Mechoid: This is to save the trouble of making 2 recipes for cabbage / lettuce salads.

PR #6170, opened 2019-05-21
Big Bugfix Bonanza

mistyLuminescence:
Been a while since I've been here! Have some bugfixes (and bug closes!).

Fixes #380 - prone people can no longer interact with an empty hand.
Fixes #658 - now checks for two-handed items on limb loss.
Fixes #1048 - mecha syringe guns now respect pierceproof clothing.
Fixes #3318 - securitrons will only react to being attacked when active.
Fixes #3386 - rechargers now benefit from upgraded capacitors.
Fixes #5434 - adds a subtype of medical records (and detective TV) computers.
Fixes #5948 - shrouds now retain their base name when damaged.

Closes #896 - unable to reproduce.
Closes #1065 - unable to reproduce.
Closes #1370 - unable to reproduce.
Closes #1754 - unable to reproduce.
Closes #1829 - fixed a while back, but never closed. Whoops.
Closes #2027 - unable to reproduce.
Closes #6015 - unable to reproduce.
Closes #6062 - intentional (see Cerebulon's comment below).

PR #6271, opened 2019-07-05
Un-Nerfs Odysseus Tools

Nalarac:
This requires staff input

Reverts part of #6170 that makes the mech syringe gun unable to pierce clothing. My opinion is that industrial syringe cannon can pierce clothing.
Partially reverts #2196 that makes it so the sleeper can load someone in 3 seconds instead of 5 (original change was 2 seconds to 5)


Wickedtemp:
First one is pointless. Syringe guns aren't effective tools for administering medicinal reagents, because from an IC standpoint, you really, really shouldn't be turning your patients into pincushions. We have functioning alternatives for mech ranged-healing already made, with a non-mech, kinda modular variant in the works, but those might not get added here. It's really only good for antags.

Second one is a bit nicer and is a good small improvement.

PR #6274, opened 2019-07-07
Returns logs from plants

Heroman3003:
Regular wooden logs have been unobtainable for a while due to the plantcode. Now that they can produce items, its only fair that logs come back to being available as base material.

### DIFF
--- a/code/modules/hydroponics/grown.dm
+++ b/code/modules/hydroponics/grown.dm
@@ -200,7 +200,24 @@
 				qdel(src)
 				return
 			else if(seed.chems)
-				if(!isnull(seed.chems["potato"]))
+				if(W.sharp && W.edge && !isnull(seed.chems["woodpulp"]))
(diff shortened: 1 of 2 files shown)


code/modules/hydroponics/grown.dm
```diff
@@ -200,24 +200,7 @@
 				qdel(src)
 				return
 			else if(seed.chems)
-				if(W.sharp && W.edge && !isnull(seed.chems["woodpulp"]))
-					user.show_message("<span class='notice'>You make planks out of \the [src]!</span>", 1)
```
Mechoid: I would leave this block in, as woodpulp can exist in plants that don't specifically have the item-product for logs. It makes sense for Towercaps to go straight to logs, but some other plants might need processing.

Heroman3003: Alright. Was just thinking it'd be mostly redundant, but will fix.

PR #6279, opened 2019-07-08
Pipe Clamps, Shutoff Valves, and Leaking Tubes, oh my.

Mechoid:
Ports Pipe clamps, shutoff valves, and pipe leakage from Baystation.

Pipe clamps are found in atmos techs' lockers, or 2 in the CE's locker.

Automatic shutoff valves can be acquired from the pipe dispenser, and will automatically close themself if they detect a leak on the pipe network they are attached to.

Pipes without endcaps will slowly leak their contents into the area around them, unless they are stasis clamped.


chaoko99:
Okay, so I'm going to say this as a head in from baystation: This is a terrible idea. shutoff valves have a nice little habit of irrepairably breaking. Stasis clamps literally do not work on 90% of pipe types, and just about everyone on bay hates this feature.  It soudns neat on paper, but in reality it just makes atmos hell, and even less fun than it already is.

chaoko99:
Would it be possible to link this to a config option?

code/game/machinery/atmoalter/clamp.dm
```diff
@@ -0,0 +1,150 @@
...
+	var/obj/machinery/atmospherics/pipe/simple/target = null
+	var/open = 1
+
+	var/datum/pipe_network/network_node1
+	var/datum/pipe_network/network_node2
+
+/obj/machinery/clamp/New(loc, var/obj/machinery/atmospherics/pipe/simple/to_attach = null)
+	..()
+	if(istype(to_attach))
+		target = to_attach
```
Atermonera: Multiple clamps attached to the same target will interfere with each other and potentially act unpredictably

code/game/machinery/atmoalter/clamp.dm
```diff
@@ -0,0 +1,150 @@
...
+		update_networks()
+		dir = target.dir
+	return 1
+
+/obj/machinery/clamp/proc/update_networks()
+	if(!target)
+		return
+	else
+		var/obj/machinery/atmospherics/pipe/node1 = target.node1
+		var/obj/machinery/atmospherics/pipe/node2 = target.node2
```
Atermonera: What if a clamp is attached to a manifold, 3- or 4-way?

code/game/machinery/atmoalter/clamp.dm
```diff
@@ -0,0 +1,150 @@
...
+		return
+
+	if (istype(A, /obj/machinery/atmospherics/pipe/simple))
+		to_chat(user, "<span class='notice'>You begin to attach \the [src] to \the [A]...</span>")
+		if (do_after(user, 30, src))
+			if(!user.unEquip(src))
+				return
+			to_chat(user, "<span class='notice'>You have attached \the [src] to \the [A].</span>")
+			new/obj/machinery/clamp(A.loc, A)
+			qdel(src)
```
Atermonera: Missing `..()`?

code/game/machinery/atmoalter/clamp.dm
```diff
@@ -0,0 +1,150 @@
+//Good luck. --BlueNexus
+
+//Static version of the clamp
+/obj/machinery/clamp
+	name = "stasis clamp"
+	desc = "A magnetic clamp which can halt the flow of gas in a pipe, via a localised stasis field."
```
Atermonera: Extended desc should include how to remove it, if not by wrench as most other atmos-related things use

code/game/machinery/atmoalter/clamp.dm
Mechoid: Not possible, as they're not 'pipe/simple'

chaoko99:
This still does not account for the requirement for a total remap to function sanely.

Atermonera:
Our current map already has manual valves separating each department. If we have to fiddle with it further, we are capable of doing so.

AaronSherwood117:
please do not

Andromeda-K22:
This is a bad idea, you should feel bad, please god no, this feature is terrible.

Natje0:
Please don't do this.

code/game/machinery/atmoalter/clamp.dm
```diff
@@ -142,9 +142,12 @@
...
+		var/C = locate(/obj/machinery/clamp) in get_turf(A)
+		if (do_after(user, 30, src) && !C)
 			if(!user.unEquip(src))
 				return
 			to_chat(user, "<span class='notice'>You have attached \the [src] to \the [A].</span>")
 			new/obj/machinery/clamp(A.loc, A)
-			qdel(src)
\ No newline at end of file
+			qdel(src)
+		if(C)
```
Atermonera: this should go above the do_after() and have a return. At present it finds a clamp, lets you attach another, then tells you there was already one there, without removing the new one

Mechoid: 146 has a check to ensure C does not exist, this `if(C)` is after to ensure the do_after has a chance to cycle before saying it fails. As in, the person tries to attach it, and then fails.

Atermonera: oh, I'm just blind

chaoko99:
You're making a mistake.

Atermonera:
All of the complaints posited are based upon a system that they acknowledge is broken. While Mechoid has admitted that this system is ported from that system, I am confident enough in his ability to code to trust that he has resolved those problems to his satisfaction. Some of the complainants play on servers that are unrelated to our code. In their cases, I am unsure what their cause for complaint is. Others of the complainants play on downstreams, and while their concern is valid, downstreams may choose which commits to keep and which to throw out. I have even gone so far as to squash this PR into a single commit, that you might find it easier to revert the changes, if you're so vehemently opposed to this.

If it happens that this PR retains the problems of Bay's implementation, despite mechoid's best efforts to test it, there's this nifty thing called `revert`. Yes, it can be said that while we throw this around every so often when something controversial comes up, only to ignore the actual results of the PR and keep the broken system. But that doesn't mean we will always do so.

There were concerns raised over our map requiring extensive remapping to accommodate this system, when that is not actually the case. Furthermore, if the valves aren't actually mapped in, the only part of this system affecting the server is the leaky pipes, and they are only a problem in cases of damage to atmospherics (Which carries much larger implications for the course of the round), damage to toxins (which honestly is more likely the fault of the scientist and this is a side-effect), or damage to pipes distributed about the station, which, at least given our default setup where distro is put under no greater pressure than 1 atmosphere, is _solely_ caused by a player _intentionally_ changing that. Which, as an atmos player, is frankly unnecessary.

chaoko99:
>Furthermore, if the valves aren't actually mapped in
So it's 2am so I'm just going to address this and get to it later: You'd be amazed by how little gas is actually in atmospherics. leaking pipes have this neat little thing where they interact with vacuum and quickly empty atmos of all airmix by spitting it all into space. This is not desirable in any situation, and I am sure you could understand why this is a huge problem. I should've clarified upon this, it's been a thing fixed with extensive (See: Halfassed) remapping of the Torch's pipenet, to the point that I actually nearly forgot about it, that was years ago.

zekesulastin:
Will any valves be mapped in before the next server update?  Since you didn't mention it when you dismissed Chako et al's concerns, I remember specifically discussing the engine room cooling loops which would require moving the TEGs to install in-round like we regularly do with lead these days.

chaoko99:
Did you know that Thermal electric generators have no check for pressure, and can be unsecured during use, allowing you for an instant one click super matter delamination

zekesulastin:
> 
> 
> Did you know that Thermal electric generators have no check for pressure, and can be unsecured during use, allowing you for an instant one click super matter delamination?
> 
> I know this because it's happened to me

I tested exactly that just now and the leak is slow enough that it took a few minutes running the automatic nitrogen setup to actually get to that point - seems like it would be quicker to break the reactor window somehow and leave the emitter on.  The pressure of the whole engine room is increasing at about 1 kPa/s.  Can't look any farther because I have to get to work.

(Also, and this is the "oh god Zeke is complaining about something" snark slipping in so y'all staff may want to read past it, the PR got tagged with bikeshedding and Atermonera just told all of you that Bay's experience either does not matter or does not match what our experience will be.  You basically got a very polite version of Mangled calling me a whiny reactionary baby when I complained about the headshot PR.  Save your breath; it'll be better for your blood pressure and it looks like some stuff was actually considered.)

chaoko99:
It's a bit more critical on a plasma set up personally, because it'll set the engine room on fire rather quickly, leading to a bit of a death spiral, you may or may not be aware that the super matters reaction to oxygen is significantly reduced on a nitrogen setup due to its mechanics. Making little seriously larger Hazard on other gases

chaoko99:
For the record, I believe bikeshed is rather... Less than apt here. This is something that needs CAREFUL thought due to the years of pain it can provide, and has provided for others. Sooner or later someone's going to make a mistake that screws everyone.

image

Heroman3003:
Well, so far this PR is showing to be a burning pile of manure from both gameplay standpoint and code quality standpoint.

PR #6295, opened 2019-07-11
Adds an advanced TTS circuit, and sign-camera.

Mechoid:
More people are doing robotics and using Sign heavily. So you know what that means. It's time for a _small_ expansion in the interim between them.

Adds the Advanced TTS, which utilizes cutting edge TECHNOLOGY to produce a voice capable of being heard by intercoms. For the more devious, or apt, it can also replicate the voice of any scanned mob, roughly. So people who have their larynx ripped out can talk with their own voice once more. (Eventually)

Adds the Sign-Language Translator camera, which can translate any sign language it sees into understandable text. It can translate some vocal speech through facial recognition tech, but due to its specialization, it's not perfect, and is more likely to spit out gibberish in these cases.

### DIFF
--- a/code/game/objects/objs.dm
+++ b/code/game/objects/objs.dm
@@ -173,6 +173,9 @@
 		*/
 	return
 
+/obj/proc/hear_signlang(mob/M as mob, text, verb, datum/language/speaking) // Saycode gets worse every day.
+	return FALSE
+
 /obj/proc/see_emote(mob/M as mob, text, var/emote_type)
 	return
 
--- a/code/modules/integrated_electronics/subtypes/input.dm
+++ b/code/modules/integrated_electronics/subtypes/input.dm
@@ -540,7 +540,57 @@
 	if(translated)
 		activate_pin(2)
 
+/obj/item/integrated_circuit/input/microphone/sign
+	name = "sign-language translator"
+	desc = "Useful for spying on people or for sign activated machines."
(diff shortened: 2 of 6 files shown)


DasIrrlicht:
[produce a voice capable of being heard by intercoms] This will be fun for antags.
[it can also replicate the voice of any scanned mob, roughly] This will be fun for Braun.
[It can translate some vocal speech through facial recognition tech] well, and that is at least nice for a bit of fluff.

Since you are poking at it, can I make a its-y bits-y request? Can you add a circut that recieves radio simelar how the microphone recives, maybe using a number-string for the channel?
If it being able to hear in in any channel with just the frequency is to strong (Given, I like the idea of being able to make it basically hijack security-conservations [I mean, if you do that and they notice it because 'you know things', well, I assume we can add 'hearing in in unauthorised channels' as 'espionage' for the Corp-Reg/Laws]) one maybe limits it to Common, Entertainment and the AI-Channel.

Atermonera:
That's actually a pretty big ask, because it'd require hooking a circuit into t-comms, which is notably one of the Great Artifacts with which mortal man best not trifle. Up to mechoid on whether they actually dare to do as much. As for channel limitations, you're extraordinarily unlikely to get no channel restrictions whatsoever, and far more likely to get the same channels as intercomms.

Mechoid:
A radio-receiver is planned for a later Expansion into the world of Integrated Circuits, unfortunately right now I'm busy with Skill stuff

PR #6303, opened 2019-07-15
Combat Mechs Can Punch More Things

Nalarac:
Combat mechs can punch pretty much everything now (some things don't take damage though), and changes the mech punch sound to the same as the juggernaut (punch4.ogg to heavysmash.ogg)

Makes take_damage a proc on /atom and changes all other uses of the proc to reflect that. Adds take_damage to things that would get damaged by attack_generic.

Cult pylons can now be destroyed with bullets

**Edit for More take_damage Stuff:**

Adds click delay on attacking barriers and simple doors.
Adds attack_generic to some of the things I added take_damage to (so simple mobs can now attack them)
Proper noises when attacking material doors and barricades. (wood vs resin vs metal)

### DIFF
--- a/code/game/objects/structures/simple_doors.dm
+++ b/code/game/objects/structures/simple_doors.dm
@@ -124,16 +124,16 @@
 		icon_state = material.door_icon_base
 
 /obj/structure/simple_door/attackby(obj/item/weapon/W as obj, mob/user as mob)
-	usr.setClickCooldown(DEFAULT_ATTACK_COOLDOWN)
+	user.setClickCooldown(DEFAULT_ATTACK_COOLDOWN)
 	if(istype(W,/obj/item/weapon/pickaxe))
 		var/obj/item/weapon/pickaxe/digTool = W
-		visible_message("<span class='danger'>[usr] starts digging [src]!</span>")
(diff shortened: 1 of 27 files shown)


Wickedtemp:
Tbh every mech should be able to punch stuff with varying degrees of damage. If you get hit by an arm of a Ripley, it doesn't matter if it "isn't made for combat", it's still a big metal arm slamming into you, it's going to do damage. Kitchen knives and frying pans aren't made to fight people with, but they do an excellent job of cutting and crushing.

code/game/objects/structures/simple_doors.dm
```diff
@@ -124,15 +124,22 @@
 		icon_state = material.door_icon_base
 
 /obj/structure/simple_door/attackby(obj/item/weapon/W as obj, mob/user as mob)
+	usr.setClickCooldown(DEFAULT_ATTACK_COOLDOWN)
 	if(istype(W,/obj/item/weapon/pickaxe))
 		var/obj/item/weapon/pickaxe/digTool = W
-		user << "You start digging the [name]."
+		visible_message("<span class='danger'>[usr] starts digging [src]!</span>")
```
Atermonera: These should continue to use `user` instead of `usr`

PR #6322, opened 2019-07-21
Allows synthetic names to start with a number

Heroman3003:
Because, really, why not? Special symbols (like +, -, etc) are all still forbidden at start of the names.


Atermonera:
Staff say yay

PR #6324, opened 2019-07-22
Adds Laser Tag Turrets Plus Changes

### LINKED ISSUE #6317: Turrets build with practice laser carbines can be lethal
Turrets sem to be able to shot lethal no matter what their gun is.
It being less lethal when only having no-lethal weapons.
#### What actually happened
Impressive firepower.
#### Steps to reproduce
Build a turret using a non lethal gun (like a practice laser carbine) and let it shot at people, turned to lethal.
#### Additional info:
 - **Server Revision**: master - 2019-07-15

### LINKED ISSUE #6318: Lethal Turrets not commited to their job.
Lethal turrets do not finish off unconscious players.
It not going to try an bleed me out, but keeping firing till dead.
#### What actually happened
Only died once I came back to senses and getting shot then.
#### Steps to reproduce
Turn a turret to lethal. Get shot till knocked out. (Maybe Sleep-Verb or chemical-induced sleep, too?) See how the turret retracts, having lost interest in you.
#### Additional info:
 - **Server Revision**: master - 2019-07-15

### LINKED ISSUE #6327: turret settings confusing/unrelated to actual targeting
Build a turret and set it to shot 'misc Lifeforms'.
It going after a mouse, for example.
#### What actually happened
Prometheans are, ID in hand or not, a valid target.
#### Steps to reproduce
Build the usual setup. Test the settings with different targets, ideally a mix of what should be valid and what shouldn't be.
#### Additional info:
 - **Server Revision**: master - 2019-07-21
 - **Anything else you may wish to add** I think everything of the turrets is broken by this point...

Nalarac:
Fixes #6317 Non-lethal weapons will no longer be able to create lethal turrets.
<strike> Turrets will now shoot you dead if they're set to lethal </strike>  **Removed**
Fixes #6327 Misc Lifeforms no longer includes small mobs arbitrarily. Prommies and Tesh are now safe.

Fixes laser tag projectiles from being "lastertag" to "lasertag"
Fixes laser tag beams causing eye blur
Places the alien pistol's fire noise on the projectile instead of on the gun. This means the turret will have the proper noise.
Changes how the icon_state works for turrets, allowing them to be more colors of death and fun!
E.g. lasers (generally) make turrets red, x-ray green, phase orange, and default is the standard blue

Adds a laser tag turret that can be ordered from cargo for 40 points. It only targets people wearing red or blue laser tag vests (depending what you set them to).

Changes the turrets.dmi to fit the way the icon_state is set.


Nalarac:
Before:
image

After:
image

code/game/machinery/portable_turret.dm
```diff
@@ -677,10 +729,10 @@
 	if(faction && L.faction == faction)
 		return TURRET_NOT_TARGET
 
-	if(!emagged && issilicon(L) && check_all == 0)	// Don't target silica, unless told to neutralize everything.
+	if(!emagged && issilicon(L) && check_all == FALSE)	// Don't target silica, unless told to neutralize everything.
 		return TURRET_NOT_TARGET
 
-	if(L.stat && !emagged)		//if the perp is dead/dying, no need to bother really
+	if(L.stat == DEAD && !emagged)		//if the perp is dead no need to bother really
```
Mechoid: IIRC this specifically checks general stat, so non-emagged turrets that are not on lethal don't stunlock someone.

Nalarac: This line of code checks if you're lying down and if the turret is set to lethal. If it's not on lethal, you're not a target if you're laying down.
```
768	if(L.lying)		//if the perp is lying down, it's still a target but a less-important target
769		return lethal ? TURRET_SECONDARY_TARGET : TURRET_NOT_TARGET
```

zekesulastin:
I kind of feel like the "fix" for 6318 should be its own separate PR since the comment indicates the previous behaviour was intentional, but I could very easily have missed any other discussion on it.

This would really impact a few PoIs if nothing else.

Atermonera:
I agree with zekes' request to split that off, pending staff discussion whether we want that to continue being intended behaviour (as I was under the impression that it was) or if it's now desirable for turrets to make rescue worthless

Nalarac:
Removed.

Nalarac:
Okay, fixed merge conflicts. I'll just let github say I deleted and re-pasted the laser.dm file

PR #6328, opened 2019-07-23
Makes Heavy-Duty Cell Chargers Constructable

Nalarac:
Fixes robots being unable to upgrade regular (and wall) rechargers.

The heavy-duty cell recharger is constructed the same way as the regular recharger (with a unique frame).

The heavy-duty recharger can now be upgraded like the standard rechargers.

**EDIT WITH "Fixes Robots putting things in rechargers"**

- Rechargers and cell chargers show what's inside and charge / maxcharge when examined.
- Changes messages in cell_charger to use to_chat
- Fixes cyborgs being able to put their goodies inside rechargers
- Cyborgs can remove items from rechargers now


code/game/machinery/recharger.dm
```diff
@@ -25,9 +25,6 @@ obj/machinery/recharger
 	return
 
 /obj/machinery/recharger/attackby(obj/item/weapon/G as obj, mob/user as mob)
-	if(istype(user,/mob/living/silicon))
```
Anewbe: I think this might screw with things like the secborg's baton, did/could you check that?

Nalarac: You are correct. I'll have to fix that.

Nalarac: Fixed with latest pull

PR #6331, opened 2019-07-23
Makes Husking Require More Burn

Nalarac:
Makes you need 250 burn instead of 200 to husk a (human) body. Should make killing people with lasers/burns slightly less likely to result in a husk.


Wickedtemp:
Was actually planning on doing this exact thing. Neat.

Nalarac:
I have no idea what the TODO is referring to

PR #6341, opened 2019-07-24
Adds automatic shutoff valves to distro

Atermonera:
Scrubbers loop was untouched, because space wasn't always available and it hasn't ever entered a state in which this PR would actually be problematic as far as I can remember in my entire time playing ss13.
Maps 3 stasis clamps into the engine monitoring room.
A pressure regulator is used before the shutoff valve to actually separate the networks, so that the valves don't all fire for a single leak. Yes, this means that distro takes longer to pressurize. Atmospherics has more tools than just distro to repressurize chambers. Deck 3 has just one valve for the entire floor, because there isn't space enough to break it into more than one network (And yes, the piping is very messy, but expanding that room had additional considerations). Decks 1 and 2 are in a cyclic loop running clockwise around the middle, such that if any one breaks, gas from the others can help refill it once the leaks are repaired.
This has been tested, to ensure that all regulators and valves are actually open at roundstart.
This has been tested with a breach, and the contents of the broken segment did not drain nearly as fast as advertised. Yes, in the time it takes someone to locate the break, it will probably empty, but starting at 100kpa, the drop was a handful of kpa every few seconds. The valve at the start of the broken segment did correctly close, and no others were affected, and the valve correctly reopened when the break was mended.
This has been tested with multiZ, where breaks in the segment on Deck 1 that contains the upwards connection to deck 2, which certainly did exist while I was rooting those, out did correctly cause the valve controlling the deck 2 segment containing the downwards connection, to close.

Pipe Changes (Just new, old replaces the pressure reg/valve with just a digital valve) ((Enough context has been provided that verifying the locations of each of these should be relatively straightforwards, I hope))
Deck 3:
Deck 3
Scrubber pipe was moved to run under the wall. No, that is not ideal, but with the space available that's what I had to do.

Deck 2:
Engineering Subloop breakoff
Engineering

Deck 2 Fore segment breakoff
Deck 2 Fore
Controls Fore central hall, Janitor, IA, tool storage

Security Subloop breakoff
Security

Research Subloop breakoff
Research
This also contains the deck 1 research areas, there wasn't space enough to easily break those into their own loop

Deck 2 Starboard segment breakoff
Deck 2 Starboard
Controls Starboard central hall.

Medical Subloop breakoff
Medical

Deck 2 Aft segment breakoff
Deck 2 Aft
Controls HoP office, teleporter, deck 1/2-central connections, and the docks. Yes, this is unfortunately the biggest loop, that's sorta just how the cookie crumbled. The central space doesn't have enough space without running pipes under walls, which I did try to avoid except where necessary, and breaking that off would still have a fairly small effect upon the size of the loop.

Civilian Subloop breakoff
Civ
Another particularly big loop, because Civilian is an entire spoke, where others are only half a spoke.
Controls Bar, Kitchen, Garden, Chapel, Holodeck, Library, Lounge, Security checkpoint, tool storage, locker room

Deck 2 Port segment breakoff
Deck 2 Port
Controls port central hall, a first aid station, emergency EVA, and contains the main connection to Deck 1

Cargo Subloop breakoff
Cargo
Also controls the Deck 1 cargo area, due to space considerations

Deck 1:
Deck 1 Port segment breakoff
Deck 1 Port
Controls port hallway, hangar 1, contains connection up to deck 2

Deck 1 Fore segment breakoff
Deck 1 Fore
Controls fore hallway, gym, bathroom, vault

Deck 1 Starboard segment breakoff
Deck 1 Starboard
Controls starboard hallway, hangars 2 and 3

Deck 1 Aft segment breakoff
Deck 1 Aft
Controls deck 1 docks, deck 1 security checkpoint, aft hallway, tech storage

Deck 1 Central subloop breakoff
Deck 1 Central
Controls pilot lockers, t-comms, t-comms teleporter

The valves are all named, I have plans to make monitoring them significantly easier in the next few days


Atermonera:
I should note that it IS entirely possible for some more leaks to have escaped my notice, as thorough as I tried to be.

The pressure regulator comes before the valve, so that it immediately creates a new network, then the valve can stop that regulator from dumping gas into it if there's a break. If there's a break in the previous segment, then the regulator can only take gas when the pressure is lower, and will not impose increased power requirements like a pump would.

Atermonera:
Recommit to remove step_y because ater doesn't map very often and thus was unaware "nudge [dir]" is verboten

chaoko99:
Honestly instead of this TERRIFYING pipe spaghetti, why not just make them layered, like other pipes are?

PR #6344, opened 2019-07-25
Manifolds now obey the laws of pipes.

Mechoid:
Manifolds will now leak, because I am a sausage and didn't give them the actual leaking code.


Heroman3003:
Did you actually make it possible to apply stasis clamp to them? Because that was also a thing that wasn't possible. And if it still isn't, now they leak with _no way of stopping it_

PR #6351, opened 2019-07-27
Fixes 3 appearing to blind people when surgery table is used

Heroman3003:
Happens whenever someone is put onto surgery table (unless you're one doing it) within the 'sightline' when you are blind.


Anewbe:
IT HAS BEEN LOCATED

Heroman3003:
Had same roots as '1' from Modular computers. Just an outdated visible_message proc whenever weird number just randomly appear for any reason is usually the cause.

PR #6353, opened 2019-07-27
Adds shutoff valve monitor

Atermonera:
Mapped into atmos monitoring and deck 1 engineering space. Didn't have room to put it into the engineering monitoring area in the lobby, but if engineers need to fix pipes they're already hacking at least one door to get the pipe dispenser in the first place.
Unfortunately my familiarity with nanoui isn't shared with bay computers, but if anyone else wants to make this a program for that by all means, go for it.

The monitor only gives the location and status of the valves at present, and cannot be interacted with.
tested


Novacat:
Bay computers use mostly the same nanoui stuff as a lot of other things. The only thing that will probably throw you off at first is that it is all located in a different directory.

code\modules\modular_computers\file_system\programs

PR #6358, opened 2019-07-30
Exploration Expansion 1: Or, How I Learned To Love The Tree

Mechoid:
Bugfixes:
 - "Exosuits are now targetted by turrets again."

Additions:
  - "Searing damage type added. Energy axe now uses searing damage."
  - "Survey points added. Vendor added for explorer use. Costs are universally 'lower' compared to the mining vendor due to the rarity of points."
  - "Swiping an ID card on a cataloguer will transfer the points to the card."
  - "Energy Axe prototype added to R&D."
  - "Anomalous 'hunting trap' added to R&D. Useful for an initial stun on mobs for explorer scanning."
  - "Emergency welding pack added. Incredibly slow, but requires no protection."
  - "Random turf modifier added for mapping use. Random humanoid remains spawner added for mapping use."
  - "Plants under obj/structure/flora now can be made to be harvestable / forage-able."
  - "Sif trees now occasionally have fruit containing useful microbes. Microbes slow blood flow resulting in slower chemical absorption and blood loss."
  - "Moss tendrils added to Sif, making eye-plants more rare."
  - "Multiple PoIs added. 'Blue shuttle down' variant, 'UFO Frigate', multiple geysers, cliffs, and two variants of SFR which take drastically different approaches to the SFR standard. A vault was added, too."
  - "Generic AI-controlled humanoids added. Subtype made for use on Sif, clad in weak armor and temperature resistance, used as 'mindless clones'."
  - "More artifice added."

Tweaks:
  - "Mining charge strength lowered, price increased. Can be upgraded through R&D laser components."
  - "Any melee energy weapon can now be made to use a cell and charge."
  - "Beartraps now stun for a quarter of a second, so they are actually useful again, due to resistability."
  - "Flora by default are now under mobs, on the same layer as turf decals. Flora can now also randomize their size, previously only codified in Sivian trees."
  - "Vault PoIs are now in a template group."
  - "Fultons made to be [somewhat] sane. Now only useable planetside."
  - "Graviton Visor moved to R&D from the mining equipment vendor. Silver and Diamond pickaxes removed."


Atermonera:
`The entry /datum/design/item/weapon/eaxe has invalid material type MAT_OSMIUM`
Suspect missing define

code/modules/artifice/cursedform.dm
```diff
@@ -0,0 +1,41 @@
...
+
+	if(P.lit && !user.restrained())
+		if(istype(P, /obj/item/weapon/flame/lighter/zippo))
+			class = "rose"
+
+		user.visible_message("<span class='[class]'>[user] holds \the [P] up to \the [src], it looks like [TU.hes] trying to burn it!</span>", \
+		"<span class='[class]'>You hold \the [P] up to \the [src], burning it slowly.</span>")
+
+		spawn(20)
+			if(get_dist(src, user) < 2 && user.get_active_hand() == P && P.lit)
```
Atermonera: can you not use a do_after() here?

PR #6361, opened 2019-07-31
Changes Space Helmet Cameras to a Verb

Nalarac:
You no longer enable the helmet camera when toggling helmet lights.

The verb is located in the <strike>IC</strike> **Object** tab.
The camera can be deactivated and reactivated with the verb after initial activation.
The user is set each time the camera is activated, so a new user can be scanned to the suit's camera.

Changes some lines to use to_chat

### DIFF
--- a/code/modules/clothing/spacesuits/spacesuits.dm
+++ b/code/modules/clothing/spacesuits/spacesuits.dm
@@ -30,23 +30,36 @@
 	brightness_on = 4
 	on = 0
 
-/obj/item/clothing/head/helmet/space/attack_self(mob/user)
+/obj/item/clothing/head/helmet/space/verb/toggle_camera()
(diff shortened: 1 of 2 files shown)


TheFurryFeline:
As a side note, I like this a LOT. Would make getting helmet lights turned on one click rather than the usual 2!

Anewbe:
I agree on the Object thought.
Have you tested full that the camera shuts off properly?

Nalarac:
Yes, the camera does shut off properly.

TheFurryFeline:
> Yes, the camera does shut off properly.
> 
> Also, this is not a thing for hardsuits. This is for all space suit helmets that have a camera inside of them (it just happens that we only have cameras in the rigs).

Yeah, I rechecked the code and it's weird. Go figure.

PR #6377, opened 2019-08-04
Fixes synth legs being broken causing feedback loop

Heroman3003:
Including sparking regardless of resting and never being able to get up.

Now resting properly prevents any sparks from generating because of broken synth legs as well.


Nalarac:
Looks like you have an extra ')' at the end

Heroman3003:
A copypaste error... Fix was a port from elsewhere, whoops.

PR #6379, opened 2019-08-05
Fixes Mech Cable Layer

Nalarac:
The mech cable layer will now properly lay down cables over the tiles it walks over.
Replaces the cable layer using the event system to work with procs instead.

Moved the cable layer from medical tools to just tools. To illustrate what was actually changed because of this, the changes before moving files are shown below:
image
image


Nalarac:
fuck I screwed it up again

PR #6380, opened 2019-08-05
Allows Construction of Serenity in Robotics

Nalarac:
Put Misty as author in changelog since I'm only removing the comments from #6208)

Notes for potential balancing decisions:

- Chassis costs 2 sheets of phoron (and some steel)
- All the standard gygax parts only cost cost steel/glass
- Instead of the gygax's diamond armor plating, it uses 5 plasteel sheets
- The "targetting" board costs the same as a standard board (glass and acid)


Mechoid:
Alright, I'll probably need to re-inspect the material costs per part but in order to get further testing I'm going to say it's kosher for now.

Issue #6394, opened 2019-08-08
Berserk Overlay blinds the player


FenodyreeAv:
#### Brief description of the issue
The berserk overlay is too intense. It's very hard to see anything under optimal conditions, and if your eyes are blurry (quite likely if you're using this), it's downright impossible to see anything.

Weather effects, meson scanners, material scanners, thermal scanners, if you're wearing any of these and berserk the overlays stack and blind you.

#### What you expected to happen
My senses to be sharpened by my impending rage!

#### What actually happened
I blindly fumbled around, and couldn't even see the spider to hit it.

#### Steps to reproduce
Go berserk!

#### Additional info:
 - **Server Revision**: Found using the "Show Server Revision" verb under the OOC tab.
 - **Anything else you may wish to add** (Location if it's a mapping issue, etc)

PR #6397, opened 2019-08-08
Blind rage doesn't mean actually make them blind.

Mechoid:
Berserk now has a less intense redshift.
fixes https://github.com/PolarisSS13/Polaris/issues/6394 Should theoretically be 'fixed', it's still red just not OH GOD red.

ater edit: mechoid you have to actually say that it fixes it or it won't


MisterLayne:
suggest referencing #6394 so that it gets closed when this is merged

PR #6398, opened 2019-08-08
Old - Adds custom DNA hash source for identical twins and other possible uses

Kyoshiku-git:
Old. Better implementation @ [https://github.com/PolarisSS13/Polaris/pull/6400](https://github.com/PolarisSS13/Polaris/pull/6400)

Adds config/example/custom_dna.txt _(to be copied to config/)_ for custom MD5 source for characters such as identical twins, to allow characters that should have matching DNA to actually have it, allowing for use of DNA locked items between characters, and another potential hurdle for detectives.

### DIFF
--- a/code/game/dna/dna2.dm
+++ b/code/game/dna/dna2.dm
@@ -359,6 +359,19 @@ var/global/list/datum/dna/gene/dna_genes[0]
 
 		if(length(unique_enzymes) != 32)
 			unique_enzymes = md5(character.real_name)
+
+			// Start custom DNA checking
+			for(var/line in file2list("config/custom_dna.txt"))
+				if(!length(line))
+					continue
+				if(copytext(line,1,2) == "#")
+					continue
+
+				var/list/List = splittext(line, "|")
+				if(List.len >= 2)
+					if(character.real_name == List[1])
+						testing("Custom DNA loaded - " + List[1] + " - " + List[2])
+						unique_enzymes = md5(List[2])	// End custom DNA checking
 	else
 		if(length(uni_identity) != 3*DNA_UI_LENGTH)
 			uni_identity = "00600200A00E0110148FC01300B0095BD7FD3F4"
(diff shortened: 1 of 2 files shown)


Atermonera:
Tested and working?

Kyoshiku-git:
Yes. Works as intended hosted locally, with assistance. Two characters can share the same DNA hash as generated by the code, and both enter a DNA locked mech. Characters not in the config/custom_dna.txt have their dna hash set as normal(MD5 hash of real_name)

Schnayy:
As interesting of an idea as this is, it's also a bit unfair. Custom items should not give mechanical advantages, which you cite in the form of DNA locked items being available and detectives not being able to determine who did what.

Kyoshiku-git:
But it makes sense from an in-character perspective. Plus it just gives detectives one less thing to check. Fingerprints still won't be identical. Identical twins _do_ have identical DNA.

Kyoshiku-git:
Plus two roboticists able to drive the same mech isn't that much of an advantage, considering they could just build two, and knowing security, they'd just brig them both anyways.

Atermonera:
Whether or not it makes sense in lore has no bearing on whether it makes sense mechanically, specifically with regards to balance. IMO, just because the instances where this would create a distinct mechanical advantage never come up does not excuse the fact that it creates a mechanical advantage at all.
In either case, with this current implementation, the people that have to handle this are the people with box access, which is a very small number, and so their approval for this is necessary if we are to entertain any intention of actually using the system ourselves

Kyoshiku-git:
I still don't see how there's any real 'advantage' to it. Everything is balanced by itself.

DNA locked items: You can use your twin's DNA locked mech, yes, but they can just as easily run off with yours.

Detectives and DNA: You're just as likely to get hauled in for something your twin did, and _both_ interrogated, regardless of if you did anything or not.

Overall, what little 'advantage' there is has it's own cons, and would create more interesting situations for RP.

Kyoshiku-git:
Could this one get a 'do not merge' label in the mean time? Leaving it open so the comments can still be seen, as much of the discussion still applies to the better implementation. Thank you.

_(Assuming the pull request just becomes hidden/inconvenient to see if I close it. If not, close away.)_

Atermonera:
As an update, this was slated to be discussed at the 18 August 2019 staff meeting, but mechoid was the only dev present, so staff benched it until the next meeting, so that more folks will be present to discuss.

Atermonera:
See the other pull request for staff decision as of 1 September 2019. @Kyoshiku-git unless you need/want this pr to be kept open, as I believe the other one is more up to date/useful to the changes requested, I'd like to close this

PR #6400, opened 2019-08-09
Adds custom DNA hash source - optional whitelist/character setup version

Kyoshiku-git:
Custom DNA hashes for players, via whitelist and character setup, with config/config.txt option to allow without whitelist. Useful for characters such as identical twins, to allow characters that should have matching DNA to actually have it.

Option for enabling and setting hash seed strings added to _Character Setup > General_ on per character basis. Like alien and language whitelist items, option is available to administrators regardless of whitelist entries.

whitelist requirement enabled via **/config/config.txt**
`## Uncomment to restrict non-admins from setting custom DNA hashing`
`USEDNAWHITELIST`

whitelist entries are added to **/config/alienwhitelist.txt**
`ckey - DNA`
Hashing not included in the '- ALL' option of alienwhitelist.

Hashes are generated off the user defined strings, with fixed prefix 'customdna:' to prevent duplication of non-custom hashes. Strings that match will yield identical DNA hashes _(Useful for twins, possibly other uses.)_

[https://i.gyazo.com/b7ed858b5665953a582ef7ea8347dc5b.png](https://i.gyazo.com/b7ed858b5665953a582ef7ea8347dc5b.png)

[https://i.gyazo.com/76f84747f16c761a5c92c62cc4879e1b.png](https://i.gyazo.com/76f84747f16c761a5c92c62cc4879e1b.png)

Tested and working.

I understand this one is likely to receive the same tags as the other, and that there's a good possibility of it not being used, but going back and doing it in a more reasonable and convenient way seemed like a good idea.

### DIFF
--- a/code/controllers/configuration.dm
+++ b/code/controllers/configuration.dm
@@ -95,6 +95,7 @@ var/list/gamemode_cache = list()
 	var/uneducated_mice = 0 //Set to 1 to prevent newly-spawned mice from understanding human speech
 
 	var/usealienwhitelist = 0
+	var/usednawhitelist = 0
 	var/limitalienplayers = 0
 	var/alien_to_human_ratio = 0.5
 	var/allow_extra_antags = 0
@@ -645,6 +646,9 @@ var/list/gamemode_cache = list()
 
 				if("usealienwhitelist")
 					usealienwhitelist = 1
+
+				if ("usednawhitelist")
+					config.usednawhitelist = 1
 
 				if("alien_player_ratio")
 					limitalienplayers = 1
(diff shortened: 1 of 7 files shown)


Atermonera:
As of 1 September 2019 staff meeting, staff say:
`We want the system, but using the custom item system, such that people don't need to edit a file on the server box directly, and can instead use the custom item repository and the systems in place for the use of such.`

As for a recommendation on how to go about it, formatting-wise I'd say do it how borg sprites are done.

Atermonera:
@Kyoshiku-git just checking in, any plans to work on this?

Atermonera:
No response in over a week, closing this. If you want to come back to this project, please let any dev know, we have to reopen it ourselves.

PR #6404, opened 2019-08-13
Makes airtight plastic flaps unmakeable

Heroman3003:
On dev request changed contents of PR. Now just removed the flaps recipe from the game.

Original PR descripton:
Can once again pass all living movs if they are resting.
Can no longer pass active mechs.
Are no longer constructible (map-in/spawn only). Original PR that restricted them was mostly aimed to address the ease of making what basically amounts to perfect airlock literally anywhere by using those, but this is a much better and smoother solution to that.

### DIFF
--- a/code/game/objects/structures/plasticflaps.dm
+++ b/code/game/objects/structures/plasticflaps.dm
@@ -8,7 +8,6 @@
 	layer = MOB_LAYER
 	plane = MOB_PLANE
 	explosion_resistance = 5
-	var/can_pass_lying = 1
 	var/list/mobs_can_pass = list(
 		/mob/living/bot,
 		/mob/living/simple_mob/slime/xenobio,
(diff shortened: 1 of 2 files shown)


Novacat:
I would personally split of plastic flaps. Have one version for mining that lets mobs pass through if resting. Have another version that is constructable that does not let mobs pass through.

Iroquois-Pliskin:
Imagine making a pr to nerf something that doesn't need nerfing, being shot down, and deciding the solution is to nerf them in a different way.

Iroquois-Pliskin:
Just revert them to how they were if you're going to nerf them then partially revert said nerf.

Atermonera:
If it's something that's going back and forth a bunch, then staff should probably properly establish what behaviour we want. I meant to mark this before the staff meeting last weekend, but I don't believe there were enough folks present to properly discuss it there anyways

Atermonera:
Staff was nonconclusive, forwarding to nonattending devs for further input

code/game/objects/structures/plasticflaps.dm
```diff
@@ -8,7 +8,6 @@
 	layer = MOB_LAYER
 	plane = MOB_PLANE
 	explosion_resistance = 5
-	var/can_pass_lying = 1
```
Atermonera: Sorry for the delay, but feedback from devs is that we're fine with the removal of the recipe, but want it to continue to block mobs from passing through.

Heroman3003: Changes applied

PR #6410, opened 2019-08-20
Promethean Organ Tweakening

Mechoid:
Tweaks / Additions:
- Prometheans now have a total of 5 organs. Removing any of the non-brain organs will cause a rapid increase in toxin damage, until active regeneration is used.
- The 'Heart', and four organ systems dedicated to the 4 damage type regens, each with a max buffer of 10.
- Prometheans can now take small amounts of damage in the 4 types in a short span of time, and not gain standard increases in pain from the regeneration of said damage, scaling with the amount of strain already present. After that, pain increases as normal. 
- Network strain does not decrease if regeneration is actively occuring.

- Active Regen now replaces removed internal organs.

Fixes:
- Organs now properly add themselves to the organs-by-name list.


Wickedtemp:
Injecting acid into a promethean was already a recipe for an insta-exploding mess of gibs and obscene burn damage. With the toxin spikes, it...might actually just straight up insta-death them.

PR #6411, opened 2019-08-20
Makes the Create Object menu not freeze the client for an eternity

Michiyamenotehifunana:
This PR is literally just a copy pasta of: https://github.com/Aurorastation/Aurora.3/pull/5442

It will make the Create Object menu instantly populate the objects list instantly instead of freezing the client. It works when you use the searchbar as well.

It's black magic to me.

I have no idea how or why it works.

It just does.


Atermonera:
Might be that it's not using JS list.add(), but instead making a big long string, by the looks of it, but I'm not actively familiar with JS, so I could be wrong

PR #6421, opened 2019-08-24
Adds a VR area to CentComm map

SubberTheFabulous:
Adds a VR area to CentComm level for use with VR sleepers.
Nothing crazy, just a thunderdome and a picnic area, like the one in the holodeck.


SubberTheFabulous:
I have no idea, why Travis hates my guts, so, closing this until I understand. 
Wew.

PR #6428, opened 2019-08-28
glove attack refactor

Screemonster:
In practical terms, the only change this PR makes for the player experience is that you have to use harm intent to punch people with the holographic boxing gloves now.

Removes a chunk from human attack_hand() that was performing an istype check to look for holographic boxing gloves every time a human was clicked on by another human (which overrode _everything_, even contact disease spreading and the like)

Adds the ability to associate a special unarmed_attack with a set of gloves so if people want they can make spiked knuckle dusters or power fists or taser gloves or whatever and keep the code for those things with the item instead of having to add code into attack_hand().

Adds the ability for unarmed attacks to override the normal attack sequence and perform some other proc instead.

Puts all the boxing glove stuff that was removed from attack_hand() into the above two places.

Slight change to the way attacks handle the existing glove damagetype - it no longer forces the damagetype to painloss for pulled punches where the attack is sharp or edge. Sharp stuff will cut you either way.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_attackhand.dm
+++ b/code/modules/mob/living/carbon/human/human_attackhand.dm
@@ -1,7 +1,7 @@
 /mob/living/carbon/human/proc/get_unarmed_attack(var/mob/living/carbon/human/target, var/hit_zone)
 	if(src.gloves)
 		var/obj/item/clothing/gloves/G = src.gloves
-		if(G.special_attack && G.special_attack.is_usable(src, target, hit_zone))
+		if(istype(G) && G.special_attack && G.special_attack.is_usable(src, target, hit_zone))
 			if(pulling_punches)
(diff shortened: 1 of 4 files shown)


code/modules/mob/living/carbon/human/human_attackhand.dm
```diff
@@ -1,4 +1,12 @@
 /mob/living/carbon/human/proc/get_unarmed_attack(var/mob/living/carbon/human/target, var/hit_zone)
+	if(src.gloves)
```
Anewbe: Might want to do a typecheck here for proper gloves items(?); there's at least one other item that can fit in that slot and it threw/throws runtimes on some other interactions.

Screemonster: Done, hopefully should catch it.

PR #6434, opened 2019-08-30
Adds hardhats to engineering lockers

Heroman3003:
Because those aren't just wardrobe items, but also technically safety equipment, they should probably go into the personal lockers too. Plus atmostech locker on its own lacked head covering to pair up with firesuit that it has.

Adds regular hardhat to engineer locker and red one to atmostech.


Heroman3003:
Reopening to try and see wtf is up with Travis

Atermonera:
Yeah sometimes travis just be like that

PR #6436, opened 2019-08-31
Fixed an issue with multiple boot/shoe overlays.

Cameron653:
Fixed an issue where boot/shoe overlays would magically be cleaned by putting in a knife and stepping on blood.

Before this, putting a knife into a boot would cut (remove) all overlays, along with anything that called update_icon on them, which included by stepping on blood/oil.

This meant that phoron contaminated shoes could be "cleaned" by stepping on blood, masking the fact they were still contaminated. This also meant that you could "clean" blood off by putting in a knife.

With this, it makes it so putting in a knife/stepping into blood would work as expected. (See: https://i.imgur.com/z0jL4pt.gifv)

Additionally, it makes it so if you step into multiple blood stains with the same pair of shoes, they will mix. (See: https://i.imgur.com/SWoXrOM.png This is normal blood + Skrell blood.)

Boots used to test this was the Workboots, and I also made sure to put on a pair on magboots and enable/disable them just to ensure that the light still worked on them.

Edit:
After fixing this, a bug was found where the overlays would build up and make cleaning boots a hassle. I fixed this by going around and fixing a few things around the code.
Things changed:

- When add_blood is called, it no longer immediately defaults to a red blood. Instead, it'll check if the object has had blood put onto it and, if so, it will leave the blood as it is. If it hasn't already been bloodied, it then defaults to the normal, stock red blood.

- Stepping on blood doesn't cut all overlays and remove bootknives anymore, so you can still have blood "cake up" on your shoes, but one clean will clean them all. (See: https://i.imgur.com/yfP1KEN.png). This actually shouldn't happen according to the checks in Cleanable/humans.dm, but... It does anyways, for some reason.

- Adjusted how add_blood works in objects/items.dm. Instead of checking for the blood_DNA before it applies the blood overlay and applies the color, it simply applies the color and the overlay. This shouldn't effect anything other than allowing things to have multiple layers of blood/oil/etc on it.

Did some additional testing to make sure everything is 100% A-OK with this, but from everything I've seen so far this should be good now.

### DIFF
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -388,7 +388,8 @@
 		blood_DNA = list()
 
 	was_bloodied = 1
-	blood_color = "#A10808"
+	if(!blood_color)
+		blood_color = "#A10808"
 	if(istype(M))
 		if (!istype(M.dna, /datum/dna))
 			M.dna = new /datum/dna(null)
--- a/code/game/objects/effects/decals/Cleanable/humans.dm
+++ b/code/game/objects/effects/decals/Cleanable/humans.dm
@@ -93,7 +93,6 @@ var/global/list/image/splatter_cache=list()
 				S.overlays += S.blood_overlay
 			if(S.blood_overlay && S.blood_overlay.color != basecolor)
 				S.blood_overlay.color = basecolor
-				S.overlays.Cut()
 				S.overlays += S.blood_overlay
 			S.blood_DNA |= blood_DNA.Copy()
 			perp.update_inv_shoes()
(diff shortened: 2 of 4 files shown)


Heroman3003:
An issue that will come from that is that overlays wont actually be removed when needed, like contamination overlay or blood prints. I suggest returning cutting overlays to update icon, and instead of updating icon when knife is inserted, just add the overlay on that step instead.

Cameron653:
I just woke up and was informed of what Heroman is talking about. I hadn't thought for a moment that fixing something like this would make them uncleanable.

Currently looking into how to fix that as well.

Cameron653:
Found a fix to blood not being able to be cleaned. Testing now.

PR #6440, opened 2019-09-02
fixes a stomach pump typo

chaoko99:
yeah uh It used to transfer reagents from all sources

### DIFF
--- a/code/game/machinery/Sleeper.dm
+++ b/code/game/machinery/Sleeper.dm
@@ -272,7 +272,7 @@
 				if(beaker.reagents.total_volume < beaker.reagents.maximum_volume)
 					var/pumped = 0
 					for(var/datum/reagent/x in occupant.ingested.reagent_list)
-						occupant.reagents.trans_to_obj(beaker, 3)
+						occupant.ingested.trans_to_obj(beaker, 3)
 						pumped++
 					if(ishuman(occupant))
 						occupant.ingested.trans_to_obj(beaker, pumped + 1)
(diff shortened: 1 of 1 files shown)


code/game/machinery/Sleeper.dm
```diff
@@ -272,7 +272,7 @@
 				if(beaker.reagents.total_volume < beaker.reagents.maximum_volume)
 					var/pumped = 0
 					for(var/datum/reagent/x in occupant.ingested.reagent_list)
-						occupant.reagents.trans_to_obj(beaker, 3)
+						occupant.ingested.trans_to_obj(beaker, 3)
 						pumped++
 					if(ishuman(occupant))
```
Cyantime: You can only put humans in sleepers, what's this doing here and why do humans get pumped twice?

chaoko99: I was copying how it works a few lines above.

Cyantime: The lines above take some of the patient's blood from the `vessel` reagent container, it doesn't pump their reagents again.

chaoko99: resolved

code/game/machinery/Sleeper.dm
```diff
@@ -272,10 +272,7 @@
 				if(beaker.reagents.total_volume < beaker.reagents.maximum_volume)
 					var/pumped = 0
```
Novacat: Variable defined but not used.

chaoko99: Used a few lines up

Cyantime: The var a few lines up is a different one with the same name, defined at L261

PR #6445, opened 2019-09-06
Fixes some lost drone laws not ever applying

Heroman3003:
P much.

### DIFF
--- a/code/modules/mob/living/silicon/robot/subtypes/lost_drone.dm
+++ b/code/modules/mob/living/silicon/robot/subtypes/lost_drone.dm
@@ -65,6 +65,7 @@
 
 					laws.add_supplied_law(9, "Safeguard [target_name]. Anyone threatening or attempting to harm [target_name] is no longer to \
 					be considered a crew member, and is a threat which must be neutralized.")
+					return laws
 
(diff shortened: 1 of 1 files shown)


TheFurryFeline:
This the thing that made the end of rounds with a Lost Drone fail to display a list of the random events that happened in a shift? That's an issue downstream at Chompers.

Heroman3003:
No idea about that, but I do believe this made the drone have missing lawsets outside specifically option for "empty lawset" diceroll

TheFurryFeline:
Lost Drone-547 (Played by: ) was unable to survive the rigors of being a stationbound synthetic without an AI. Its laws were:
DEBUG: Runtime in gameticker.dm,438: Cannot execute null.show laws(). [view] 
[2019-09-07T00:07:01] Runtime in gameticker.dm,438: Cannot execute null.show laws().
Just for reference since it happened again at the end of a shift, preventing the event list from displaying.

Heroman3003:
Then it might be that, yeah, since if laws failed to apply, they remained as null, which would be the cause of the runtime

TheFurryFeline:
Hmm, just tried it and it didn't fix the runtime there, ugh... Still got the same gameticker issue.

PR #6447, opened 2019-09-07
Orange Is The New Orange Part 2

TheFurryFeline:
Refactors prison jumpsuit slightly to not be a child of the orange jumpsuit, which in turn gets its own sprite. 

Left/right hand uniforms don't seem to be applying in-game when carrying the suits but that's not too bad of an issue. 

Uniform on the mob shows properly and works with sleeves rolled and uniform rolled down.


TheFurryFeline:
I fail to see why this is labelled as a bugfix when I'm only doing a sprite change to the orange jumpsuit while refactoring the prison jumpsuit to be its own item and not a child of the orange jumpsuit.

Atermonera:
PR desc seemed to indicate a bug was being fixed but I'm pretty sure that was mostly me misinterpreting words

TheFurryFeline:
> PR desc seemed to indicate a bug was being fixed but I'm pretty sure that was mostly me misinterpreting words

Yeah, I also added in sprites for the according carry-on thing for jumpsuits, which are right and left. But when tested, they didn't seem to show up for any of them. Imo, it's not that big of an issue, thus the comment. 

That said, I've reviewed my work and it seems that I neglected to change the admin commands filepaths and they had been left spawning in adjustable jumpsuit, not the prison one. I've fixed the oversight, so it's technically an oversight fix. Hope that clears it up.

SpadesNeil:
image

Feels like a fix to me, wherein previously the "orange" jumpsuit was identical to the prison jumpsuit when it shouldn't have been.

PR #6457, opened 2019-09-17
Bar and Kitchen Merge

Woodratt:
- Merge of the bar and the kitchen into one area.
- Updated the holodeck area to mesh with how it is set up

[Image](https://i.imgur.com/yupoW7n.png)


Woodratt:
Updated since apparently I forgot Pun Pun

PR #6467, opened 2019-09-25
Makes every single shot not create ~40 timers but 1

Heroman3003:
A timer per every single section of shot + the light sources... This unifies it all into one single deletion.


Wickedtemp:
Is there an expected effect this will have, or does it just make things more tidy code-wise

Novacat:
This will probably stop Virgo's timer subsystem from shitting the bed on a semi-regular basis.

PR #6471, opened 2019-10-01
Blind port of Bay Radiation Subsystem

Novacat:
Figured I would do my part to move everything over to subsystems.

Cursory tests suggest it works just fine.


Atermonera:
How does this differ from the current system, if I may ask?

Novacat:
There are some minor changes but otherwise it is exactly the same, unfortunately I could not find documentation on those changes, my guess is optimization?

chaoko99:
extremely major optimization and solid structures actually stopping radiation.

chaoko99:
I think it hooks into the material system in some way, but I can't remember.

Novacat:
Solid structures actually stopping radiation was a separate change that I did not port, though if it is desired I could attempt to seek it out and port it.

chaoko99:
It is desired. *VASTLY.*

Atermonera:
I mean the code looks mostly fine so if travis says it's ok then it probably won't set everything on fire? :+1:

PR #6494, opened 2019-10-24
Add Sounds to stuff! (Port from VOREStation)

Rykka-Stormheart:
The following critters have footstep sounds:
All Giant Spiders
~~Corrupt Hounds~~ N/A here, that's VORE Station only.

Suit Cycler now audibly PINGS when it's finished.
Analysis machine for xenoarch makes a soft plink when it plays the 'plink' text.
TBD: Air pumps quietly hum while they're running, and make an audible noise when they shutdown or lose power. Sound files are added, just not added the code for it yet.


parasitoide:
What's with all the unused sounds like the clockcult ones?

Rykka-Stormheart:
> What's with all the unused sounds like the clockcult ones?

Intended for use later, or added for possible use by someone. IE the airpump ones could be used if I knew how to make airpumps play a sound when starting, stopping, and running.

Mechoid:
You'll want to look at TEGs and the SM, they both have sound loop datums.

Rykka-Stormheart:
> You'll want to look at TEGs and the SM, they both have sound loop datums.

Found it, am giving it a shot now <3

Rykka-Stormheart:
Added Air Pump start/mid/stop, and a client toggle so you can disable the ambient hum.

code/modules/client/preferences_toggle_procs.dm
```diff
@@ -209,6 +209,21 @@
...
+/client/verb/toggle_air_pump_hum() 
+	set name = "Toggle Air Pump Noise" 
+	set category = "Preferences" 
+	set desc = "Toggles Air Pumps humming" 
+
+	var/pref_path = /datum/client_preference/air_pump_noise 
+
+	toggle_preference(pref_path) 
+
+	src << "You will [ (is_preference_enabled(pref_path)) ? "now" : "no longer"] hear air pumps hum, start, and stop."
```
Atermonera: `to_chat(src, "")`

Rykka-Stormheart: Corrected, thank you! <3

Rykka-Stormheart: Doing that made Travis throw an error?

Cerebulon:
I welded and/or switched off every vent and scrubber in the area I was in and the air flow noise continued in addition to a looping "fan stuck" noise so I'm not certain the "stop playing the air vent noises" proc is working properly in all cases.

PR #6495, opened 2019-10-25
Fixes non-planet zlevel weather runtime cascade

Verkister:
Fixes the potentially server killing runtime spam caused by humanmobs walking on "outdoors" tiles on zlevels outside the range of the planetary weather system.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_movement.dm
+++ b/code/modules/mob/living/carbon/human/human_movement.dm
@@ -140,7 +140,7 @@
 			. += turf_move_cost
 
 	// Wind makes it easier or harder to move, depending on if you're with or against the wind.
-	if(T.outdoors)
+	if(T.outdoors && (T.z <= SSplanets.z_to_planet.len))
 		var/datum/planet/P = SSplanets.z_to_planet[z]
 		if(P)
 			var/datum/weather_holder/WH = P.weather_holder


code/modules/mob/living/carbon/human/human_movement.dm
```diff
@@ -140,7 +140,7 @@
 			. += turf_move_cost
 
 	// Wind makes it easier or harder to move, depending on if you're with or against the wind.
-	if(T.outdoors)
```
Atermonera: What are the contents of SSplanets.z_to_planet? How does this prevent the checks against z levels that are lower than those used by planet-z's, such as our main station?

Verkister: It basically contains the expected_z_levels list defined for planet datums with added undefined nulls to fill the list position gaps between the zlevel numbers that are defined for the planet. (like if a planet datum expects zlevels 2 and 4 as planetary, the list would contain null, z2, another null, and z4)

The runtime part comes when the human movement code tries to find the planetary weather effects for a turf marked as "outdoors" (which certain turf types have by default) on every step, and if the zlevel is higher than any of the pre-defined planet zlevels, it will runtime trying to fetch a planet datum outside of the bounds of the list.

PR #6465, opened 2019-09-25
Medical Expansion

Mechoid:
Additions: 
 - Multiple new organs added. Humans and Skrell received spleens, all species expected to have a stomach and intestine organ have them.
  - Multiple augments added, currently only available in the Traitor / Mercenary uplinks as easy-to-install implants.
  - Anesthetic / Oxygen pumps added. A mobile stabilizer has been added. All three are available in Cargo.
     - Stabilizer acts as an oxygen pump, and provides a set pulse to individuals attached, corpses have a weaker pulse.
  - Medical MRE rations added to Cargo.
     - Similar to nutrient paste, arguably worse, contains trace amounts of tricordrazine and paracetamol. MRE pack contains a dessert and drink packet, unlike liquidfood. It's for patients, gotta have some morale.
  - Neural implant added to loadout, makes the brain of the user count as an assisted organ. Does not affect MMIs or their subtypes.
  - Bioprinters now unlock more organs upon being upgraded, once they pass the anomalous tier, they unlock quite probably illegal organs.
  - Three medical exosuit components have been added. Crisis and Hazmat response drones, and a mounted advanced medical analyzer.
  - Multiple new chemicals added, two used for upgrading bandage kits. [Y is distillery, N is not.]
     - [N]Calcium carbonate, an antacid used to stop vomiting.
       - ("oxygen" = 3, "calcium" = 1, "carbon" = 1)
     - [Y]Tricorlidaze, a topical tricordrazine. Used to upgrade bruise packs to trauma kits.
       - ("tricordrazine" = 1, "sterilizine" = 1, "foaming_agent" = 1) (T0C + 100, T0C + 120)
     - [Y]Necroxadone, an upgraded cryoxadone that can purge toxins from corpses, only when they are attached to a mobile stabilizer.
       - ("lichpowder" = 1, "cryoxadone" = 1, "carthatoline" = 1)CAT:("phoron" = 5) (T0C + 90, T0C + 95)
     - [Y]Spacomycaze, a topical antibiotic painkilling gel. Essentially, neosporin (?). Used for upgrading crude brute kits to bruise packs.
       - ("paracetamol" = 1, "spaceacillin" = 1, "foaming_agent" = 1) (T0C + 100, T0C + 120)
     - [Y]Lich Powder, an offshoot of zombie powder that does not cause paralysis.
       - ("zombiepowder" = 2, "leporazine" = 1) (T0C + 100, T0C + 150)
  - Brute-based medical stacks can be upgraded, up to advanced.
[...]

PR #6500, opened 2019-10-28
Loadout Implants Properly go in Mob, not on Floor

KasparoVy:
Hey! Posting this upstream to you kind folks from the OG https://github.com/VOREStation/VOREStation/pull/6121.
The issue was originally reported here at https://github.com/VOREStation/VOREStation/issues/3521 and was made even more apparent when the medical expansion in https://github.com/PolarisSS13/Polaris/pull/6465 was merged. So many awesome fluff implants, but they just didn't work out of loadouts.

I really puzzled over this one for some time a while ago but just couldn't figure it out, but thankfully this time around I had a better time.

-------------------

The main cause is the 'exploitable' check from `code/game/jobs/job_controller.dm` because the proc `amend_exploitable` from `code/modules/mob/mob.dm` spawns a duplicate of the implant at the mob's feet.
This was less apparent because for all the other implants (dud, neural, etc.) the `Initialize()` behaviours deleted them so they wouldn't be visible on the ground, so the indication they had the same issue was that they weren't present in the appropriate parent organ's `contents` or `implants` lists.

Tertiary issues were...
- The `implant_loadout` proc in `code/game/objects/items/weapons/implants/implant.dm` not inserting the implant in the correct location (was passing an organ instead of a zone define like BP_HEAD, which has a string value of "head".)
- There were unnecessary post-spawn behaviours for dud and neural fluff implants.

What now happens:
- The exploitable check now occurs in the `spawn_item` proc from `code/modules/client/preference_setup/loadout/loadout.dm`.
- The loadout implants are now inserted into the correct locations within a mob.
- There won't be implants created on the floor anymore.

-------------------

I verified this by spawning in with every implant, checking the parent organs (chest & head), referring to the `contents` & `implants` lists and seeing that for the head the brain was "assisted". No implants were on the floor.

### DIFF
--- a/code/game/jobs/job_controller.dm
+++ b/code/game/jobs/job_controller.dm
@@ -376,9 +376,6 @@ var/global/datum/controller/occupations/job_master
 							H << "<span class='warning'>Your current species, job or whitelist status does not permit you to spawn with [thing]!</span>"
 							continue
 
-						if(G.exploitable)
-							H.amend_exploitable(G.path)
-
 						if(G.slot == "implant")
 							var/obj/item/weapon/implant/I = G.spawn_item(H)
 							I.invisibility = 100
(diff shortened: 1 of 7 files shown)


Anewbe:
Praise be.

KasparoVy:
Lmfao Higgs Bugson, cheers

PR #6515, opened 2019-11-02
[WIP] Refactors tools to use TOOL_BEHAVIOR, combitools.

lolman360:
Refactors tools so that all you need for a tool to pass a specific isTool proc is a specific TOOL_BEHAVIOR.
As such:
 isTool procs check for TOOL_BEHAVIOR and are no longer overwritten by most tools
 Welders now use the doWeld proc instead of copypasting the same block of code everywhere. 
 Enables isWelder, isCableCoil and isMultitool.
 Powertools now switch behavior instead of creating new items.

Also, adds nu-combitools.
They're multiple tools combined into one device. Some run off cells. Stock combitools can be made at RnD, can be used as basic tools (screwdriver/wrench/crowbar/wirecutter) but are slightly slower than an individual tool (1.2 toolspeed). They also drain from a cell when active and on use.

Syndicate Combitools drain much less power, are slightly faster (0.8 toolsped) and can also be used as a welder, multitool, or simple knife (15 damage.)

Precursor/Alien combitools don't need a cell, and are as fast as powertools (0.5 toolspeed).

Balance suggestions for combitools are welcome. If they're not wanted I'm happy to remove them from this PR, but I'm throwing the idea out here for now.


Atermonera:
It's now been 4 months since the last activity upon this PR, I'm closing this for inactivity. @lolman360 if you have any plans on continuing this work, please contact the dev team so we can reopen this.

PR #6519, opened 2019-11-08
Fix Fruit Runtime

Mechoid:
On the tin, grown food items were initializing before their seed datum was assigned properly.


Atermonera:
pretty sure this is an initialization order issue moreso than a "make it sleep until it exists" issue. I.e., try making the seed controller initialize after the chemistry controller, then if that doesn't work more the chemistry controller before the object controller and see how much stuff *that* breaks.

PR #6520, opened 2019-11-08
Clean up dead code obsoleted by outfit system

Meghan-Rossi:
- job_access_type on ids is used literally nowhere.  Removed.
- spawnId is only called on people who already have ids, and does nothing because of it.  Removed.
- idtype is defined both for jobs and outfits, and for jobs it's only used in spawnid and to not give ais and cyborgs headsets.  Replaced with new has_headset variable
- equip_backpack and equip_survival on jobs are also useless, since outfits handle that.  Removed.
- spawnId put access on ids, but equip_id doesn't.  Made equip_id for job outfits put access on ids.
- With equip_id putting access on ids we don't need all the different subtypes of id.  Removed a lot of them.

### DIFF
--- a/code/datums/outfits/jobs/cargo.dm
+++ b/code/datums/outfits/jobs/cargo.dm
@@ -14,7 +14,7 @@
 /decl/hierarchy/outfit/job/cargo/cargo_tech
 	name = OUTFIT_JOB_NAME("Cargo technician")
 	uniform = /obj/item/clothing/under/rank/cargotech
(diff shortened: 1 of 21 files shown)


Atermonera:
Sorry for taking so long to actually get back to this, school was stressful and busy.
Has this been tested, and with what jobs was it tested?

Meghan-Rossi:
Tested it just now as Colony Director, Pilot, Cyborg, and Warden, seems to work fine.

PR #6526, opened 2019-11-13
Fixy-wixies the fucky-wucky I madey-wadey with toy swords.

lolman360:
Title. Toy swords now use new esword code! Cool and robust.


Schnayy:
Did somebody weave on the test mwap? uwu

lolman360:
i dont test my code, why would the test map be on?

PR #6533, opened 2019-11-21
More/Better Usagi Borg Sprites (Port)

timidvi:
Adds some new sprites which are similar to our Usagi sprites, but a bit nicer/sleeker/more detailed looking. Ported from another server which shall be named if required but otherwise not in case Polaris admins prefer not to "advertise" other servers on its platforms.

Image of sprite fronts:
https://gyazo.com/5b35cb1b6f9d436e61fa330a5cdbb501
Backs/sides can be seen from the dmi itself.

May also have missed some little details somewhere else in the codebase which would need to be edited - I probably wouldn't know, since I've neither worked much with BYOND as a dev tool, nor with Github. This is very much an amateur working with what little they know - please enlighten me as necessary!


Nalarac:
You forgot to add commas to the entries before your new Usagi II additions.

image

timidvi:
> You forgot to add commas to the entry before each of your new Usagi II additions.
> 
> image

Thank you for pointing that out, I've got to admit that was a bit silly even for a hopeless amateur~

Mechoid:
Closing/Re-opening to refresh the bots, since it seems they were fixed.

PR #6535, opened 2019-11-25
Fix Strange Rock Size

TheFurryFeline:
Title


TheFurryFeline:
Hang on, I know I got the CL file the wrong format. Working on it.

PR #6537, opened 2019-11-26
Inaccuracy Rebalance (NPC)

Mechoid:
Most mobs have a slight attack delay, and ranged mobs tend to have inaccuracy similar to a player. Automatic-firing enemies tend to have higher inaccuracy than single-fire, and mechanical tend to have higher accuracy than organic.

Base melee attack delay for simple mobs is now 2 ticks. 1 second is roughly 10 ticks. Viscerators have no delay, as their movement is inherently an attack.


MisterLayne:
Does this include combat drones?

Mechoid:
Due to their use of the weak laser and use as a telegraphed engagement I didn't change their accuracy specifically

PR #6538, opened 2019-11-26
Teshari tweakening

Mechoid:
Tweaks:
- Teshari have 65 max health from 50
- Teshari brute mod to 1.15 from 1.35 (Fragile, but not imminently vaporized by a brick. Unless someone throws it, then they might have a problem.)
- Teshari burn mod to 1.2 from 1.35 (Feathered, incredibly small mass to absorb heat, and high internal temperature pre-existing.)
- Universally, mobs that hide will un-hide automatically upon attacking, similar to breaking invisibility.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_helpers.dm
+++ b/code/modules/mob/living/carbon/human/human_helpers.dm
@@ -56,6 +56,8 @@
 
 //This is called when we want different types of 'cloaks' to stop working, e.g. when attacking.
 /mob/living/carbon/human/break_cloak()
+	..()
+
 	if(mind && mind.changeling) //Changeling visible camo
 		mind.changeling.cloaked = 0
 	if(istype(back, /obj/item/weapon/rig)) //Ninja cloak
--- a/code/modules/mob/living/carbon/human/species/station/seromi.dm
+++ b/code/modules/mob/living/carbon/human/species/station/seromi.dm
@@ -45,9 +45,9 @@
 	slowdown = -1
 	snow_movement = -2	// Ignores light snow
 	item_slowdown_mod = 2	// Tiny birds don't like heavy things
-	total_health = 50
-	brute_mod = 1.35
-	burn_mod =  1.35
+	total_health = 65
+	brute_mod = 1.15
+	burn_mod =  1.2
 	mob_size = MOB_SMALL
 	pass_flags = PASSTABLE
 	holder_type = /obj/item/weapon/holder/human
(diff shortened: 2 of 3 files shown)


listerla:
imo buffing health isn't needed if the modifiers are nerfed _and_ the simplemobs nerfed

they really oughta go down hard when they do go down, which can be hard in PVP given this still allows evasive Hide and High Speed Action

zekesulastin:
For context, the next weakest human mob to damage is tajaran, with 1.15/1.15 maluses.  With the changes 35 more health than a teshari still but just about the same resistance to things that are affected by damage thresholds (bone breaks etc).

Wickedtemp:
Do teshari still burn to death from getting a fever or was that fixed

listerla:
I feel like Teshari bones should probably break more easily than Taj bones and such, given they glide and presumably thus have hollow or at least light bones. Just a nitpick tho.

Schnayy:
We decided against the teshari health buff in the staff meeting, didn't we?

TheFurryFeline:
> Do teshari still burn to death from getting a fever or was that fixed

That's still a thing that's happened in old code on Chompers. Though since those days, PolarisMed has been changed a few times... Not sure, but should be easy to check with fever code.

Atermonera:
Experimental, we want to test #6537 first and see if that fixes the issues raised. This is DNM until such time as it has been merged for long enough for staff to decide on this at a staff meeting.

Novacat:
Species balance for Polaris has always had this weird overvaluation of speed bonuses and undervaluation of sheer durability, especially since the Tajara/Teshari speed bonuses were, in the grand scheme of things, not that huge.

This has made slow, tanky species like Unathi rather overpowered, especially since, when weighed down with gear, since Unathi not only get 125 health, 0.85 brute/burn modifiers, and higher blood amount, but also have 0.25 item slowdown which means, when weighed down with gear (Voidsuits and Dufflebags) they actually run faster than other species weighed down with such gear, since gear gives rather extreme slowdowns +1 for voidsuits/hardsuits, +1 for dufflebags, while Tajara only has -0.5 slowdown and Teshari have -1 slowdown)

I can see why things are they way they are on Polaris. The nature of Sif means that dufflebags and voidsuits are practically nonexistent, while Sif's expedition areas tend to be a lot bigger and more wide open with lots of snow, which all play to Teshari/Tajaras speed and snow bonuses.

On different maps where speed is not valued as highly, and snow slowdowns are practically nonexistent, and voidsuits/hardsuits become more necessary, the balance does not look as good.

While I do not have a good solution to this that would fit both Upstream and Downstream needs as, in the end, downstreams may need to rebalance their species for their maps, I do believe that both upstream and downstream could benefit from a refactoring of item-based slowdowns from intervals of 1 to intervals of 0.5, and halving slowdowns of all items with slowdown across the board.

Atermonera:
Reviewed at 02-02-2020 staff meeting, staff don't currently feel this is necessary. If at a later date people feel that teshari combat specifically needs to be tweaked further, we can discuss the matter then.

PR #6557, opened 2019-12-01
Stops bots from opening elevator doors

Heroman3003:
Title. According to elevator door code nothing is meant to normally open them but elevator. Of course, bot code is snowflakey trash and needs special exceptions, as usual.


DasIrrlicht:
Meaning that beepsky will no longer pry open the door if I hide in the elevator? Thanks.

PR #6562, opened 2019-12-10
Graffiti selection

PrismaticGynoid:
Lets you choose what type of graffiti or rune to draw when vandalizing the floor with crayons or markers. No more drawing and erasing a dozen times to get the exact graffiti you want.

### DIFF
--- a/code/game/objects/items/crayons.dm
+++ b/code/game/objects/items/crayons.dm
@@ -74,22 +74,22 @@
 		switch(drawtype)
 			if("letter")
 				drawtype = input("Choose the letter.", "Crayon scribbles") in list("a","b","c","d","e","f","g","h","i","j","k","l","m","n","o","p","q","r","s","t","u","v","w","x","y","z")
-				if(get_dist(target, user) > 1 || !(user.z == target.z))
+				if(get_dist(target, user) > 1 || !(user.z == target.z) || !drawtype)
 					return
 				to_chat(user, "You start drawing a letter on the [target.name].")
 			if("graffiti")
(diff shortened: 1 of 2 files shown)


code/game/objects/items/crayons.dm
```diff
@@ -78,8 +78,14 @@
 					return
 				to_chat(user, "You start drawing a letter on the [target.name].")
 			if("graffiti")
+				drawtype = input("Choose the graffiti.", "Crayon scribbles") in list("amyjon","face","matt","revolution","engie","guy","end","dwarf","uboa")
+				if(get_dist(target, user) > 1 || !(user.z == target.z))
```
Mechoid: Should probably have a `|| !drawtype` here so it will also return if no drawtype is input, such as by someone crashing.

code/game/objects/items/crayons.dm
```diff
@@ -78,8 +78,14 @@
 					return
 				to_chat(user, "You start drawing a letter on the [target.name].")
 			if("graffiti")
+				drawtype = input("Choose the graffiti.", "Crayon scribbles") in list("amyjon","face","matt","revolution","engie","guy","end","dwarf","uboa")
+				if(get_dist(target, user) > 1 || !(user.z == target.z))
+					return
 				to_chat(user, "You start drawing graffiti on the [target.name].")
 			if("rune")
+				drawtype = input("Choose the rune.", "Crayon scribbles") in list("rune1", "rune2", "rune3", "rune4", "rune5", "rune6")
+				if(get_dist(target, user) > 1 || !(user.z == target.z))
```
Mechoid: Ditto here

PR #6563, opened 2019-12-12
513 Compatability

Novacat:


Verkister:
Now that this hasn't been merged yet, there's one critical compatibility fix to be made in code/__defines/lighting.dm line 87, where the argument "border" needs to be renamed to "offset"

Without this the fake AO filter breaks and causes a hell of a runtime spam.

PR #6566, opened 2019-12-14
Fix Multitool Modes

Mechoid:
So apparently, I'm still an idiot. Fixes a bug I was confused about for a fucking year since I didn't understand the documentation for some asinine reason.

Multitools can properly store refs.

### DIFF
--- a/code/game/objects/items/devices/multitool.dm
+++ b/code/game/objects/items/devices/multitool.dm
@@ -51,10 +51,10 @@
 	return ..()
 
 /obj/item/device/multitool/proc/mode_switch(mob/living/user)
-	if(++mode_index > modes.len) mode_index = 1
+	if(mode_index + 1 > modes.len) mode_index = 1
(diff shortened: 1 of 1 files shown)


code/game/objects/items/devices/multitool.dm
```diff
@@ -51,10 +51,10 @@
 	return ..()
 
 /obj/item/device/multitool/proc/mode_switch(mob/living/user)
-	if(++mode_index > modes.len) mode_index = 1
+	if(mode_index + 1 > modes.len) mode_index = 1
```
Atermonera: The bug was originally caused by the double-increment because of the pre-increment. Basically when it was checking the old `if()` statement, it would increment `mode_index`, and then if that condition failed, it would increment it again.
i.e., just dropping the `else` clause and leaving the `if()`'s condition intact should also be a valid fix.

PR #6571, opened 2019-12-17
Fix Supply_conversion.dm runtime.

Mechoid:
On the tin.

### DIFF
--- a/code/modules/gamemaster/actions/supply_conversion.dm
+++ b/code/modules/gamemaster/actions/supply_conversion.dm
@@ -11,8 +11,11 @@
 
 	var/thaler_earned
 
-/datum/gm_action/nanotrasen_budget_allocation/set_up()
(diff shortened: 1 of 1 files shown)


code/modules/gamemaster/actions/supply_conversion.dm
```diff
@@ -11,8 +11,11 @@
 
 	var/thaler_earned
 
-/datum/gm_action/nanotrasen_budget_allocation/set_up()
+/datum/gm_action/nanotrasen_budget_allocation/Initialize()
```
TheFurryFeline: `code/modules/gamemaster/actions/supply_conversion.dm:14:error: Initialize: undefined proc`

Mechoid: [screaming] right I forgot datums are annoying.

PR #6588, opened 2019-12-26
Fixes chairs acting as handcuffs.

### LINKED ISSUE #5332: Certain devices cannot be used while sitting
Text-to-Speech device cannot be used while buckled to a chair
Able to be used normally, as it's a small device.
#### What actually happened
"You cannot activate the device in your state."
#### Steps to reproduce
Have a TTS device
Buckle self into chair
Try to use device
#### Additional info:
 - **Server Revision**: Server revision: master - 2018-06-03
493a497d25a32cc9cafe75b10757a2acaf597958 

Verkister:
Fixes some utterly redundant and ancient broken spaghetti that had simply been shrugged off for at least 7 years.

-Fixes chairs not allowing people to use hands. Leaves an exception for the xeno nest, but apparently even that is redundant due to the fact that the victim is _laid down_ on it. Same with all the beds, so this PR has no side effect of suddenly allowing self surgery and all that jazz.
-Also fixes chairs toggling the canmove var, which also was completely redundant, because a wheelchair specific var for allowing movement already exists and is utilized, and most chairs are anchored anyway. This was a fat issue because a criminally large amount of other unrelated mechanics have been using the canmove var as a type of STUN for some reason, including wack stuff such as the menu buttons for cartridge exclusive PDA features and much more. Also apparently most of the forementioned wack stuff already checks the actual incapacitation vars as well so canmove never really had business there in the first place.

Fixes #5332

### DIFF
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -289,7 +289,7 @@
 /mob/living/carbon/can_use_hands()
 	if(handcuffed)
 		return 0
-	if(buckled && ! istype(buckled, /obj/structure/bed/chair)) // buckling does not restrict hands
+	if(buckled && istype(buckled, /obj/structure/bed/nest)) // buckling does not restrict hands
 		return 0
 	return 1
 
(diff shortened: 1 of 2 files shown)


Nalarac:
This would resolve #5332 correct?

TheFurryFeline:
Yes, it would.

PR #6592, opened 2019-12-30
Xenoflora expansion take2

Mechoid:
I'm not 100% certain why Cerebul's commits are here too since that was tested on a different branch, but ???

Adds more traits for xenoflora plants.
Adds a unique xenoflorist locker which contains a biosuit and gas mask, and incredibly rarely a chainsaw for rapidly clearing plants.

Tweaks various plants to have a more reasonable amount of chems.


Atermonera:
Might not have swapped back to master before making the branch. Git defaults to splitting a new branch off of your current active branch, not sure if there's a command to change that

code/modules/reagents/reagent_containers/spray.dm
```diff
@@ -5,8 +5,6 @@
 	icon_state = "cleaner"
 	item_state = "cleaner"
 	flags = OPENCONTAINER|NOBLUDGEON
-	//TFF 24/12/19 - Let people print more spray bottles if needed.
-	matter = list("glass" = 300, DEFAULT_WALL_MATERIAL = 300)
```
Heroman3003: Was this intentional? This will make the spray bottles printable for free.

code/modules/reagents/dispenser/dispenser_presets.dm
```diff
@@ -139,7 +139,6 @@
 			/obj/item/weapon/reagent_containers/chem_disp_cartridge/lemon,
 			/obj/item/weapon/reagent_containers/chem_disp_cartridge/lime,
 			/obj/item/weapon/reagent_containers/chem_disp_cartridge/berry,
```
Heroman3003: Is it okay ending a list with a comma?

code/modules/reagents/reagent_containers/spray.dm
Mechoid: ..That was not. Been having some github issues recently and it's starting to be concerning.

Mechoid:
DNM until AFTER the GameMaster Tuning PR

PR #6595, opened 2020-01-01
adds gnomes.

Ketrai:
Much too complicated to add downstream, so I decided to just pull here. Perhaps it can be useful to you guys too.

- Adds a new plant type, gnomes.
- Grows relatively slowly.
- Can be planted anywhere.
- Ordered in as contraband (50) points
- Very nutritious, great for growing food away from station.
- Frankly adorable. Just watch them build their little gnome house!
- Self sufficient, gnomes gather their own food and water from their environment.

image

image


Schnayy:
This seems VERY silly.

Ketrai:
Isn't ss13 and memes what we live for?

Schnayy:
There's memes and then there's just... outlandishly silly.

DasIrrlicht:
Wold like a alternate version in that this is not-gnomes... And a special interaction for hitting it with a bat.

code/modules/hydroponics/seed_packets.dm
```diff
@@ -322,3 +322,6 @@ GLOBAL_LIST_BOILERPLATE(all_seed_packs, /obj/item/seeds)
 
 /obj/item/seeds/rose/blood
 	seed_type = "bloodrose"
+
+/obj/item/seeds/rose/gnomes
```
Heroman3003: Why is seed packet a subtype of rose seed packet?

Ketrai: I have no clue why, but it generates as one ingame. I didn't make it a subtype in the seed datum, so my only guess is that code is borked somewhere.

Woodratt:
As a in game thing that folks can normally get, no (raises too many questions, especially in regards to gnomes and lore and such). 

As adminspawn? 

I could see some uses for this and for amusement purposes every now and again.

So, yeah. Fine with this being a adminspawn only thing our side, but I'm just one admin.

Ketrai:
I don't have any problem with this being made adminspawn only yeah.

Ketrai:
Now adminspawn only as no one else has any other suggestions.

PR #6596, opened 2020-01-02
Informs ghosts when someone dies

GeneriedJenelle:
Informs ghosts when a living person dies. Shows a follow link as well as their location.


TheFurryFeline:
I get the feeling this may wind up spamming deadchat a ton of times considering that medical may have to defib someone multiple times before a patient is healed up properly. This is only one scenario where said spam can happen.

Atermonera:
If it becomes an issue, it shouldn't be too hard to make it a preference or something that people can toggle off.

TheFurryFeline:
Fair enough.

PR #6603, opened 2020-01-05
Add an audible alarm sound to SM Delamination

Rykka-Stormheart:
Does what it says. You cannot possibly miss SM delams now, even with a headset off. The sound isn't terribly loud, unless you're deepthroating an intercomm, because it uses the intercomms to play.

Port of https://github.com/VOREStation/VOREStation/pull/5980, since I realized it had never actually been ported. 

Enjoy your nightmares, engineering. <3


Rykka-Stormheart:
To note - for those interested, I have a trimmed/loop-able version of the audio file, but I'm not sure if anyone wants this alert sound on a loop. Though, panic-fixing the engine while alarms blare across the station is immersive-as-fuck...

PR #6606, opened 2020-01-05
Fixes clamp being used with wrong syntax in places

Heroman3003:
Broke some things, particularly breathloss. Fixes that.


Atermonera:
The syntax used was originally for a #define with (min, val, max) as the ordering, broken when byond made a native clamp function

Heroman3003:
Actually, nope. Even before 513, the defined function was "value, min, max". So those four were broken even before then.

PR #6612, opened 2020-01-09
Ctrl+Click now removes pens from PDAs

Shadow-Quill:
As per title. Tested and works.
Ctrl+Clicking on it to drag the PDA whilst it's on the ground will function as it would with anything else.


Atermonera:
Could it be made to use default functionality if a pen isn't present, or if it's not being held and the user's hands are full? Overwriting the primary mechanism for pulling (at least for myself, but I suspect most people that know of it use it) could lead to a fair deal of confusion, and possibly result in bug reports.

Shadow-Quill:
Unable to test at the moment, will do so when I get home.

Shadow-Quill:
@Atermonera Confirmed that it works as intended!

PR #6619, opened 2020-01-11
Can now (de)construct chem masters

Shadow-Quill:
Boards are printable at a circuit printer for `TECH_DATA = 3, TECH_MAGNET = 2`, and requires no components. Uses a machine frame.

If you have suggestions on where to put the board/what tech levels to use/parts to use, I'm happy to hear them.
Fixes Yawn-Wider/YWPolarisVore#347

### DIFF
--- a/code/game/objects/items/weapons/circuitboards/other.dm
+++ b/code/game/objects/items/weapons/circuitboards/other.dm
@@ -14,4 +14,6 @@
 	build_path = /obj/machinery/chem_master
 	board_type = new /datum/frame/frame_types/machine
 	origin_tech = list(TECH_DATA = 3, TECH_MAGNET = 2)
-	req_components = list()
+	req_components = list(
+							/obj/item/weapon/stock_parts/matter_bin = 2,
+							/obj/item/weapon/stock_parts/manipulator = 2)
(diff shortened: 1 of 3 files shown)


code/game/objects/items/weapons/circuitboards/other.dm
```diff
@@ -7,4 +7,11 @@
...
-	board_type = "other"
\ No newline at end of file
+	board_type = "other"
+
+/obj/item/weapon/circuitboard/chem_master
+	name = T_BOARD("ChemMaster 3000")
+	build_path = /obj/machinery/chem_master
+	board_type = new /datum/frame/frame_types/machine
+	origin_tech = list(TECH_DATA = 3, TECH_MAGNET = 2)
+	req_components = list()
```
TheFurryFeline: Honestly, I'd think that the ChemMaster would require at least tier 2 matter bins along with some other stuff. Maybe manipulators. Can't see a reason for capacitors or micro-lasers, so would just be the first two to balance it out.

MisterLayne: I'd suggest two matter bins and two manipulators.

PR #6622, opened 2020-01-11
Water resistance exists

Mechoid:
_fellas how do I oblitterate this cerebul thing_

Human get_water_protection() now checks for species resistance to water soaking.

Promethean regeneration now respects water soak. If you are in a resistant set of clothing I.E. voidsuit, explorer suit, or biosuit, you will take no or reduced damage over time.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -575,7 +575,6 @@ emp_act
 	return perm
 
 // This is for preventing harm by being covered in water, which only prometheans need to deal with.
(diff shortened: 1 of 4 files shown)


code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -577,15 +577,15 @@ emp_act
 // This is for preventing harm by being covered in water, which only prometheans need to deal with.
 // This is not actually used for now since the code for prometheans gets changed a lot.
```
Neerti: This line should get removed since it's now getting used now.

PR #6625, opened 2020-01-14
assets update pt. 1

clownidiot:
just what was agreed on so far.
all of these are ports from the world server and eris


clownidiot:
i'm gna do janitor.dmi & the towel in the next PR. otherwise, this is finished

Novacat:
I feel like the old bookshelf sprites would be great for a /wooden subtype of bookshelves.

Ketrai:
Whilst most of these are of higher quality, I'm kinda against mashing up the cabinet perspective for top-down perspective. I'd much rather see the updated textures applied to the items as is. Not only is it part of the classic SS13 feel, it also adds more depth to objects. Also the bookshelves feel way too lab-like for the current library aesthetic of wooden flooring people use. So don't replace those like ascian said. Lunchboxes suffer especially from this change in perspective, as they end up looking like flat shopping bags rather than boxes.

Personally not a fan of the first aid kits either. A lot of the colours used (especially the toxin kits) have low saturation (aka they're pale). Vibrant colours tend to work better to make them stand out, you don't want to have to go looking for important medical items because they blend in.

For everything I didn't specifically mention I'd say the better quality of sprite > loss of depth.

Ketrai:
I just honestly don't see the point in mixing the styles when most of the sprites we have show three sides. Unless you replace all of them, and get better walls... they're going to look weird and out of place.

clownidiot:
I am replacing the walls and shit like that. Also, your entire map is being remade by someone I know. This is to prepare for that.

Rand0mname:
I mean, we're happy to look at submissions for map edits but an entire remake? That's news to me.

Heroman3003:
Tbh I do not see any improvements other than Skeleton and Curtains.

Cerebulon:
Would prefer individual votes on each 'section' personally. Because as a whole there's too much stuff I don't like to port as a whole, but there's a handful of decent ones.
Generally I'm with Ketrai, most are too sterile looking and the head-on perspective is... Ugly.

Atermonera:
@clownidiot per discussion at the 19-Jan staff meeting, we'd like this to be split into multiple PRs by .dmi file, so that those sprites that we can quickly agree on can be merged, and those that require further discussion can be discussed in more distinct threads of conversation, if that's at all possible.

Ketrai:
Curtains, skeleton, racks, beds are a definite go in my opinion. The other stuff I'd only like to see  if the improved shading and colours is applied to the original perspective sprites. The medkits I would also rule out because they're hit and mis.

clownidiot:
yeah I'll do that later

Greenjoe12345:
if the curtain one could be changed so it's only the front curtain but with the new sprite, that'd be good

PR #6637, opened 2020-01-19
Augments Respect Beds, Robotic

Mechoid:
Augments that deploy items can no longer be used when strapped to a bed or other lying-restraint.

Augments properly act mechanical by default.

### DIFF
--- a/code/modules/organs/internal/augment.dm
+++ b/code/modules/organs/internal/augment.dm
@@ -7,6 +7,7 @@
 
 	icon_state = "cell_bay"
 
+	robotic = ORGAN_ROBOT
 	parent_organ = BP_TORSO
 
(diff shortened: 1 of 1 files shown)


code/modules/organs/internal/augment.dm
```diff
@@ -145,6 +146,12 @@
 
 	var/mob/living/carbon/human/M = src
 
+	if(buckled)
+		var/obj/Ob = buckled
+		if(Ob.buckle_lying)
+			to_chat(M, "<span class='notice'>You are unable to enable your augments when restrained.</span>")
```
Atermonera: `You cannot use your augments when restrained.`
`unable to enable` can be confusing

PR #6642, opened 2020-01-21
Port of Suit Cycler Tweak/Feature Change

Rykka-Stormheart:
This is a port of https://github.com/VOREStation/VOREStation/pull/6483 which is a port of https://github.com/CHOMPStation2/CHOMPStation2/pull/184 which is an update of https://github.com/CHOMPstation/CHOMPstation/pull/695

These apply the correct stats when placing a voidsuit in a cycler and switching from, say, engineering to atmos.

### DIFF
--- a/code/game/machinery/suit_storage_unit.dm
+++ b/code/game/machinery/suit_storage_unit.dm
@@ -994,12 +994,18 @@
 				helmet.name = "engineering voidsuit helmet"
 				helmet.icon_state = "rig0-engineering"
 				helmet.item_state = "rig0-engineering"
+				helmet.max_heat_protection_temperature = HELMET_MAX_HEAT_PROTECTION_TEMPERATURE
+				helmet.armor = list(melee = 40, bullet = 5, laser = 20, energy = 5, bomb = 35, bio = 100, rad = 80)
+				helmet.desc = "A special helmet designed for work in a hazardous, low-pressure environment. Has radiation shielding."
 			if(suit)
 				suit.name = "engineering voidsuit"
 				suit.icon_state = "rig-engineering"
 				suit.item_state = "rig-engineering"
 				suit.item_state_slots[slot_r_hand_str] = "eng_voidsuit"
 				suit.item_state_slots[slot_l_hand_str] = "eng_voidsuit"
+				suit.max_heat_protection_temperature = SPACE_SUIT_MAX_HEAT_PROTECTION_TEMPERATURE
+				suit.armor = list(melee = 40, bullet = 5, laser = 20, energy = 5, bomb = 35, bio = 100, rad = 80)
+				suit.desc = "A special suit that protects against hazardous, low pressure environments. Has radiation shielding."
 		if("Mining")
 			if(helmet)
 				helmet.name = "mining voidsuit helmet"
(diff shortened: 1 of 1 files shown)


Heroman3003:
Isn't the entire point of suit cycler's paintjob that it doesn't actually changes stats of the suit and just the appearance, so if you want a better spacesuit you need to actually get it rather than transform it? That always seemed like the intentional behaviour to me.

As in, this isn't as much a fix as it is a full on change to how the cycler works. And if thats the intent, then it would be better to actually change the type of the suit rather than the manual stat changes, considering that it will account for any stats outside armor category and any stat changes applied to specific suit types in the future.

Novacat:
I have no support for this particular change, for the following reasons:
- It makes it impossible for found special voidsuits, like the blood-red voidsuit and scorched voidsuits, to be used by non-human crew, since any attempts to cycle them would remove the stats
- I want to encourage people to cycle special voidsuits if they do find them, because it makes no sense for station crew to be walking around in syndie voidsuits. The current cycling mechanism is the best option considering that there is no way to properly recolor voidsuits.
- It is not consistent, the bloodred and scorched cycler remains cosmetic only. This is going to generate a lot of confusion
- If it were made consistent, then it would be too easy for station crew to acquire special voidsuits that are supposed to only be acquired in events and expeditions, since you can find emags in the trash

Novacat:
Alternatively, I feel that this should be a different machinery. Perhaps rename the old suit cycler to suit fitter, have it widely available in departments. Have this new piece of machinery be called Suit Cycler and have it exclusively in EVA?

Heroman3003:
I believe the reason its called 'Cycler' is because of its "Decontamination Cycle" feature, primarily?

Cyantime:
Have you read the description of the cycler at any time?
`"An industrial machine for painting and refitting voidsuits."`
It was never intended to change stats or give armor or anything, the sprite changes are purely cosmetic and the main function is allowing nonhumans to be able to wear space suits.
This isn't a "fix", it's a huge balance change.

Rykka-Stormheart:
> Have you read the description of the cycler at any time?
> `"An industrial machine for painting and refitting voidsuits."`
> It was never intended to change stats or give armor or anything, the sprite changes are purely cosmetic and the main function is allowing nonhumans to be able to wear space suits.
> This isn't a "fix", it's a huge balance change.

My apologies, information updated correctly.

I personally feel this is how the cycler should work, given it's dismantling the suit and rebuilding it. I simply ported it upstream with the "fix" label because it was restoring old Chompstation code. Sorry about the miscommunication.

Atermonera:
I recall this issue being a point of contention when the variant suits were first added / given special stats, but I will defer to any changed opinions among staff. You unfortunately missed the last staff meeting by a couple days, and the next one will be this Sunday, Feb 2

Heroman3003:
Its also supposedly rebuilding it from same parts and materials. Why would it change its protective qualities again?

Atermonera:
Discussed at the 02-02-2020 staff meeting, staff are generally opposed to this for reasons generally similar to those posted by others above. Leaving this open for a day or so before closing it, unless someone else does so first.

PR #6649, opened 2020-01-24
Fixes Bagel Poppy Recipe

Trexdude:
Such that crafting it using a microwave actually spawns a poppy bagel, instead of a raisin bagel.


Trexdude:
I noticed this bug in the VOREStation code, and found it was from here. I haven't downloaded my fork of the Polaris code and tested it in-game, but I have for the VOREStation fork, and I think it should work the similar.

Cerebulon:
Whoops my bad!

Trexdude:
Happens to the best of all of us.

PR #6653, opened 2020-01-28
Changes DBCore connection procs to check for config value SQL_ENABLED instead of ENABLE_STAT_TRACKING

zekesulastin:
Changes DBCore connection procs to check for existing config value SQL_ENABLED instead of ENABLE_STAT_TRACKING.  I think it makes a bit more sense to be able to control the old /tg/ stat logging system independently of being able to connect to the DB altogether.

(I haven't tested this locally because I am bad and have not set up a local MariaDB instance but it SHOULD work!  Maybe testing the PR directly in the dead of early morning...)


zekesulastin:
Bans and the player age job restriction thing worked fine in testing.

PR #6656, opened 2020-01-29
Prettifies closet contents list with (newly added) countable variant of english_list

Amunak:
## Summary
Contents of closets (as viewed by ghosts) are now less ugly and easier to read.

## Screenshots
Before (`english_list`):
dreamseeker_2020-01-29_00-42-46

After (`counting_english_list`):
dreamseeker_2020-01-30_01-22-38

Third option (`inline_counting_english_list`):
dreamseeker_2020-01-29_00-35-23

## Changes
- Adds a proc similar to `english_list` that counts items in lists and outputs the contents in actual (newline-separated) lists while also (optionally) displaying icons or specific determiners.
- Moves the ghost examine code of closets to `proc/examine` where it belongs.
- Ghosts now also see empty/full messages like characters standing
close do.
- The contents list for closets now shows number of items (if multiple)
and their icons (wherever possible), displaying it in a newline-separated list.
- The list no longer shows unnecessary determiners ("the" before every
single item).
- After some revisions, `english_list` remains untouched. Many places that use it would not support the new features anyway - some use it improperly/unnecessarily but others need to keep the list ordered, which the countable one cannot do.
- Also adds proc `inline_counting_english_list`, which is a "preset" for `counting_english_list` that is inline and looks more like `english_list`, but it supports icons and other features. It is not used anywhere yet.

## Additional notes
I tried really hard to make the new code work for the old use cases, but there were odd edge cases with some usages of `english_list` that broke logic (as, unexpectedly, this view proc is used for some code logic in cameras, weather and other places). Might revisit this later, fix those edge cases and make the new code used exclusively.

I might also look at other lists that would benefit from this; namely some stats/views where icons and/or counts would be helpful, or perhaps lists that use their own code but could use either of the newly added procs.

I also encourage people to use these procs in the future; they are highly configurable for (hopefully) any desired output imaginable, and it could still be expanded if need be. Please use named arguments to configure it though.

I would also like to discourage people from using these procs (including `english_list`) for stuff where native `jointext` would suffice - especially if it's for code logic and not for pretty printing.

Lastly, 🎉! Having fun after ~4 years.

### DIFF
--- a/code/_helpers/_lists.dm
+++ b/code/_helpers/_lists.dm
@@ -25,8 +25,8 @@
 		if(2) return "[input[1]][and_text][input[2]]"
 		else  return "[jointext(input, comma_text, 1, -1)][final_comma_text][and_text][input[input.len]]"
 
-//Returns a list that counts equal-ish items, outputting count and item names, optionally with icons and specific determiners
-/proc/counting_english_list(var/list/input, output_icons = TRUE, determiners = DET_NONE, nothing_text = "nothing", and_text = " and ", comma_text = ", ", final_comma_text = "")
+//Returns a newline-separated list that counts equal-ish items, outputting count and item names, optionally with icons and specific determiners
+/proc/counting_english_list(var/list/input, output_icons = TRUE, determiners = DET_NONE, nothing_text = "nothing", line_prefix = "\t", first_item_prefix = "\n", last_item_suffix = "\n", and_text = "\n", comma_text = "\n", final_comma_text = "")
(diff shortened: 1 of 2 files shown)


Amunak:
This essentially rewrites Baystation12/Baystation12#19436 (and the port here at fe9d686).

Amunak:
I have updated the description to reflect latest changes. This is probably in final stage unless there are issues (one debatable thing might be whether there should be an empty line after the list).

If anything I might add this to more lists. Do you know of any where it would fit?

PR #6666, opened 2020-02-03
Applies same restrictions as first aid to external synth repair

Heroman3003:
No longer can use nanopaste, cable or welders to repair synths through a spacesuit.


Heroman3003:
Reason this doesn't use coverage_check is because coverage_check is a proc of the surgery_step specifically. Wouldn't make much sense to use that way outside surgery, and pretty spaghett too, so here it still uses same exception that's copypasted in normal first aid.

Atermonera:
oooohhh, that's on applying gauze-type stuff, I see

PR #6667, opened 2020-02-04
Prevents surgery from being done while wearing a voidsuit

Heroman3003:
Other than hardsuit detachment one.

Ideally it should a special proc that prevents surgery from being done on an area covered by any non-medical clothing, IMO, but this is a start.

### DIFF
--- a/code/_onclick/item_attack.dm
+++ b/code/_onclick/item_attack.dm
@@ -44,8 +44,8 @@ avoid code duplication. This includes items that may sometimes act as a standard
 /mob/living/attackby(obj/item/I, mob/user, var/attack_modifier)
 	if(!ismob(user))
 		return 0
-	if(can_operate(src) && I.do_surgery(src,user))
-		if(I.can_do_surgery(src,user))
+	if(can_operate(src) && I.can_do_surgery(src,user))
+		if(I.do_surgery(src,user))
 			return 1
 		else
 			return 0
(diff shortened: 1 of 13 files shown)


Atermonera:
You could check if the target body part is covered, using the clothing flags? Though that might be too strict...

Heroman3003:
I was thinking of leaving specific clothing subtypes with a flag to allow for it regardless should I do it for ALL clothing. For example medical scrubs/surgery caps/medical masks would still allow surgery through them, as well as survival gear for species like Vox.

Heroman3003:
To clarify: this PR is more balancing of "surgery through voidsuit". If I'm gonna do it for all clothes it probably will be overriding this PR's changes, but it will be more of "how does it even make sense to surgery through the clothes".

Nalarac:
How will this impact the Zaddat, who can't remove their voidsuits in normal station environments?

Schnayy:
Zaddat were intended to not be able to get surgery through their suits w/out a special room, I believe. Working as intended.

Heroman3003:
Split it off. The new proc can later be modified should it be updated.

Cerebulon:
This appears to have entirely prevented surgery, voidsuit or no.

Heroman3003:
Fix for that aspect is already up. It appears a fix to way attackby logic works for surgery broke it due to one of procs involved being entirely nonfunctional (and yet still used for some god forsaken reason).

PR #6675, opened 2020-02-07
Logs graffiti and crayon drawings: Non-Cursed PR

GeneriedJenelle:
I will try to not do voo-doo somehow this time.

> Tells you when and where players graffiti in admin logs. Also tells you what letters they are drawing.


GeneriedJenelle:
> If you're going to make this a config thing, you'll also want to edit the example config.txt file so others can see what to enable/disable.

Now done. @TheFurryFeline

PR #6681, opened 2020-02-09
Warrants can only get authorized by people with HoS access now

Shadow-Quill:
### Port of https://github.com/CHOMPStation2/CHOMPStation2/pull/219

Only pretty much only the title.


Dunmeritude:
This is a very, VERY bad idea on a low-pop station that can go many rounds without seeing a Head of Security. Also makes the Warden's job even more useless. Restricting security's ability to (legally) do their job to when their boss is around is... hm. Not a good idea. 

In the lack of a HoS, this will just lead to officers outright ignoring the existence of warrants even more often.

Shadow-Quill:
> This is a very, VERY bad idea on a low-pop station that can go many rounds without seeing a Head of Security. Also makes the Warden's job even more useless. Restricting security's ability to (legally) do their job to when their boss is around is... hm. Not a good idea.
> 
> In the lack of a HoS, this will just lead to officers outright ignoring the existence of warrants even more often.

To your concerns, I give you SOP on warrants under code blue (above, warrants aren't necessary.)
`Searches of any kind may only be done with a signed warrant by the Head of Security or higher, or if found with visible contraband.`
So without a HoS or CD, you shouldn't even be serving warrants.

Atermonera:
That seems more like an issue with SOP assuming a higher pop to me.

Shadow-Quill:
@Atermonera anyone with security access can fill out the details with a tablet or modular computer, via the "Warrant Assistant" program. Anyone with a holowarrant device can choose from warrants, authorized or not. Officers can keep the device itself on them, without needing to swap between pre-set ones.

Example of a search warrant (It autofills itself, given details put into the program):
image

Atermonera:
Talked this over at the staff meeting (16 Feb 2020). The warrant system, as reported by our sec mains, hasn't been functional for a very long time, as warrants can't actually be printed. No one uses warrants because a) the system is clunky/nonfunctional and b) its usually pretty clear why sec needs to arrest/search someone and c) it's easier to adjust the alert level to permit the search/arrest anyways.
Given all this, no one was for this change, and there were several staff and security players against, so we will not be merging this change. We will be looking at adjusting the SOP, as that doesn't currently represent a practical standard course of action for us, especially with our fluctuating population.

Shadow-Quill:
It'll go downstream, then.

PR #6683, opened 2020-02-10
Implements blacklist for specific plant sprites for random seeds

Heroman3003:
For now just gnomes there, not being a plant at all.

Also makes gnome reagent, magic dust, excluded from the random seeds for 'adminspawn only' reason.


Atermonera:
travis

Heroman3003:
Initially made the list part of the plant controller and not global. Didn't test it after moving it, whoops.

PR #6686, opened 2020-02-10
[DNM] Emergency Lighting!

Shadow-Quill:
Upport of https://github.com/Yawn-Wider/YWPolarisVore/pull/564.

<hr>

DNM because a minor oversight was found, namely building fixtures auto-places a cell in them.

<hr>

How it works:

If the lighting circuit on the area's APC turns off (either automatically from loss of charge, or manually), light fixtures will use a new type of cell in their fixture (`/obj/item/weapon/cell/emergency_light`) that holds 120 watts.

The light fixture will use the cell's charge to show a dim, red light. Nowhere near enough to illuminate the halls, but enough to see by. Emergency power can last for about ten minutes, assuming a fully charged cell. The cell will charge once power is restored to the lighting circuit.

If the lights are turned off via a light switch, the emergency lights will not come on, and no power will be drawn from the cell. That is unless the power circuit goes out, at which point normal emergency operation will commence.

The cells will not recharge if the lightswitch is off (due to how it's coded, and I can't be bothered to work this out right now.) And the emergency light function can be disabled per apc. AIs/Cyborgs can disable the emergency light per light individually. (No more flickering each light individually, but whatever.)

There are two mapping lights that do not start with backup cells: `/obj/machinery/light/poi` and `/obj/machinery/light/small/poi`

Note: Light fixtures are not required to be built with a cell, it's an option. Emergency cells can be constructed in an autolathe for 20 units of glass.

No edit comments because will be ported upstream.

### DIFF
--- a/code/modules/power/apc.dm
+++ b/code/modules/power/apc.dm
@@ -104,6 +104,7 @@
 	var/beenhit = 0 // used for counting how many times it has been hit, used for Aliens at the moment
 	var/longtermpower = 10
 	var/datum/wires/apc/wires = null
+	var/emergency_lights = FALSE
 	var/update_state = -1
 	var/update_overlay = -1
 	var/is_critical = 0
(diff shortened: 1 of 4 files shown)


Leshana:
> No more flickering each light individually,

Nuuuuuuuu!

Shadow-Quill:
To be honest, I doubt light flickering is even really used.

Cyantime:
Move the flickering to alt-click or ctrl-click or something for AI/borgs, rather than flat removing it, subtly spooking people is Very Important for atmosphere.

PR #6691, opened 2020-02-13
Fixes runtime in frame.dm, line 509: Cannot read null.len

Atermonera:
tested, works

### DIFF
--- a/code/game/machinery/frame.dm
+++ b/code/game/machinery/frame.dm
@@ -231,6 +231,7 @@
 	icon_state = frame_type.get_icon_state(state)
 
 /obj/structure/frame/proc/check_components(mob/user as mob)
+	components = list()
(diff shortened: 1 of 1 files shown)


TheFurryFeline:
Ah, another one of those lists set to be null rather than a blank list. Of course.

code/game/machinery/frame.dm
```diff
@@ -231,7 +231,6 @@
 	icon_state = frame_type.get_icon_state(state)
 
 /obj/structure/frame/proc/check_components(mob/user as mob)
-	components = list()
```
Meghan-Rossi: This line is what empties the list of components when the frame is emptied of components.  Without it, emptying the frame of components teleports every component that has ever been in the frame to the location of the frame, even if you've taken those parts out and used them to build a different machine.

PR #6696, opened 2020-02-14
Replaces needless runtime spam with a sanity check

Verkister:
Replaces the multiple atom initialization runtime crash report with a simple sanity check.

The report in the runtime log is not really worth the trouble of choking the runtime logs with a 5 digit number of runtime errors in a matter of seconds every time someone loads up an area when just a simple sanity check can take care of the problem itself just fine.

### DIFF
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -68,7 +68,7 @@
 	if(QDELETED(src))
 		crash_with("GC: -- [type] had initialize() called after qdel() --")
 	if(initialized)
-		crash_with("Warning: [src]([type]) initialized multiple times!")
+		return
 	initialized = TRUE
 	return INITIALIZE_HINT_NORMAL
 


Leshana:
I think the purpose of that runtime log might be to encourage developers to fix the problem so it doesn't try to initialize twice.

Heroman3003:
This seems counter-productive... Double initialization is a bug and should be tracked down and prevented, not ignored.

Verkister:
That would make more sense with an active developer team and as a debug logging feature. As of now it just chokes up the logs with reports on a bug that has no effect whatsoever. Anyway I have now tracked down the area related part of the cause for double initialization (or more like recursive multiplied by ever-growing number of tiles -initialization) and that one might be a bit tricky because the map generator builds new areas tile-by-tile, spawning a new area tile to be initialized for each tile, but because the area is just a single ever-growing datum, each new tile just makes it initialize every tile of the area that has already been spawned and initialized.

PR #6700, opened 2020-02-17
Fix Air Pump loop sound stacking multiple times per pump.

Rykka-Stormheart:
Turns out that the # was so, so, so far off. It was acting as if it was 0.4 seconds per loop, for an 11 SECOND SOUND FILE.

My bad.


Leshana:
@Rykka-Stormheart Are you certain? At least in the polaris repo `sound/machines/air_pump/airpumpidle.ogg` looks like it is 7 seconds long, not 11.

PR #6701, opened 2020-02-17
Hide items in plant pots and plushies

GeneriedJenelle:
Also adds a thread and needle item. It can be useful, for future applications I suppose. On world we definitely have plans for it.

* You can see if an item is in a plant pot if you are next to it and examine it.
* You can hide tiny items in small plushies, and larger items in big plushies.
* Use a scalpel or any sharp knife to cut open a plushie, put an item inside (like candy, not bombs or anything) and sew it back up. 
* Small plushies and plant pots can only hold tiny items. Biggie plushies can hold bigger items.
* Use a needle and thread (buyable from gift vendor) to sew the plushies up when you're done. 
* For all your smuggling needs.
* Never trust plushies again.

image
image
image
image
image


GeneriedJenelle:
Code should be good, just gotta upload the sewing thread icon when I get home.

TheFurryFeline:
:eyes: Items in plushies? BOI! This'll be quite an interesting thing for those who like drug dealing RP stuff.

GeneriedJenelle:
Icon added. All done.

Mechoid:
Not having it for plushies makes sense, though I can see a purpose for having a borg-interaction be added for dumping / adding things from / to plant-pots via standard / gripper interaction respectively.

code/game/objects/structures/flora.dm
```diff
@@ -222,6 +222,44 @@
...
+	..()
+	if(in_range(user, src) && stored_item)
+		to_chat(user, "<i>You can see something in there...</i>")
+
+/obj/structure/flora/pottedplant/attackby(obj/item/I, mob/user)
+	if(stored_item)
+		to_chat(user, "<span class='notice'>[I] won't fit in. There already appears to be something in here...</span>")
+		return
+
+	if(I.w_class > ITEMSIZE_SMALL)
```
Atermonera: > ITEMSIZE_SMALL allows small items into the pots

GeneriedJenelle: Would you prefer if it was ITEMSIZE_TINY for balancing?

Atermonera: You noted they could only hold tiny items in the PR desc, was merely pointing out an inconsistency. I don't have any qualms either way.

GeneriedJenelle: Yeah I amended it, let's see how it goes. I hope tape recorders can fit.

PR #6704, opened 2020-02-17
Update and document GLOB macros

Leshana:
Adds comments and global typed lists to GLOB macros
 * Credit to tgstation/tgstation#38893 and tgstation/tgstation#44825
 * Document what the global var declaration macros do.
 * Add macros to declare typed global lists.

Ports avoiding overwriting raw globals with null during global init.
 * From tgstation/tgstation#31882
 * Previously globals declared without an initial value would get set to null when globals are initialized.  This might be fine, but typically if  a global isn't given an initial value at declaration its because other  (probably legacy) code intends to initialize it.  That code might run  before global variable initialization, and thus be overwritten with null  when global variables initialize.

*Summary*: Only one of the changed macros is actually used anywhere.  This PR is to fix things up and add documentation in preparation for starting to really use GLOB style globals.

### DIFF
--- a/code/__datastructures/globals.dm
+++ b/code/__datastructures/globals.dm
@@ -1,11 +1,15 @@
-//See controllers/globals.dm
+// See also controllers/globals.dm
+
+// Creates a global initializer with a given InitValue expression, do not use outside this file
 #define GLOBAL_MANAGED(X, InitValue)\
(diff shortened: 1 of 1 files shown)


Leshana:
Even tho this could not possibly break anything, I compiled and smoke tested initialization anyway.

PR #6710, opened 2020-02-19
Ports precision placement of items

Cerebulon:
You know how food and drink goes to the exact spot you place it on a table? Well now that applies to (almost) all items.

Gif from Aurora because my gyazo took a dump and wouldn't let me click while recording:
https://i.imgur.com/iKuFPzJ.gifv

_Minor tweaks:_
- Resets item x/y values on pickup
- Randomizes on-tile placement of thrown items
- Center of mass of many items adjusted to aid in placement
- Consolidated x/y placement code into a single proc randpixel_xy()
- Fixes potential runtime in floortile (re)placement

Ported from https://github.com/Aurorastation/Aurora.3/pull/6946


TheFurryFeline:
Damn, I was thinking about trying a crack at doing something like this for a specific thing downstream. This is brilliant, though!

code/game/turfs/simulated/floor_attackby.dm
```diff
@@ -156,4 +156,6 @@
 		return
 	if(flooring)
 		return
+	if(!istype(W))
```
Atermonera: It's cast initially to /obj/item/stack/tile, then this returns if it's not of /obj/item/weapon? 
Istype checks against the type of the var if you don't give it a type, such that 
> var/a/A = thing
> var/b/B = A

`istype(B) = 0`

GeneriedJenelle:
We have this code on world too, if you want to port it.

code/game/turfs/simulated/floor_attackby.dm
Cerebulon: It seems redundant but the code seems to work? So what, just nuke this particular if case?

Atermonera: Ok looking even more closely at the code I'm now confused how this works at all in the first place. `try_replace_tile()` is called on line 64 of this file, in `/turf/simulated/floor/attackby()`, in such case as `flooring` is non-null (Not plating) and a floor-tile stack is used as the weapon. That enters this proc, but... because there's an `if(flooring) return` check above... it just returns... but if that check wasn't there, it would just call `attackby()` on itself with the floor-tile stack as the weapon, which as far as I can understand should just recurse infinitely...
If there isn't flooring already, (i.e., it's just plating) then `attackby()` has its own code to actually build the new flooring.
flooring is null as plating confirmed by
> /turf/simulated/floor/is_plating()
> > return !flooring

I... think you should in theory be able to remove this entire proc, because the `try_deconstruct_tile()` is already handled in `attackby()`, and the only two places where this proc exists are its definition here and the single call in the `attackby()` definition for this same object.
However, that's rather out of scope for this PR, so since this code shouldn't ever do anything besides return, it's fine as-is for the moment.

... I _think_.

PR #6711, opened 2020-02-20
Revert "Fix runtime in helper.dm,92"

Atermonera:
Reverts PolarisSS13/Polaris#6706

That code running is grinding our server to a halt every round, this needs to be reverted while we investigate.
Evidence (from the profiler):

### DIFF
--- a/code/game/objects/items/devices/communicator/helper.dm
+++ b/code/game/objects/items/devices/communicator/helper.dm
@@ -89,8 +89,8 @@
 
 	// Cut out all but the youngest three
 	while(news.len > 3)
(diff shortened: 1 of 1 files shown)


Leshana:
Okay I see the problem.  While the old code was causing runtimes immediately because they were attempting to access `news[0]`  this one is the result of _another_ flaw in the original method.   Namely, that in byond, attempting to remove list A from list B actually removes the _contents_ of list A from list B.   
The only reason this has not been freezing the server for years along is that the runtime on line 92 saved it from reaching that far in the proc!

Atermonera:
Yep, I'm working on a fix now, that actually makes the newsfeed properly readable, since apparently it can't be used to look at specific channels. Been long enough since I wrote this stuff that I honestly have no idea what all I did while working on it, because it was part of a larger project

Issue #6531, opened 2019-11-21
Communicator Video Call Issue


TheFurryFeline:
#### Brief description of the issue
Attempting to start a video call with someone over a communicator with no visual restrictions results in your camera view switching to your call partner and then a second - give or take a few milliseconds - later, the camera view shifts right back to you!
#### What you expected to happen
Consistently being able to watch what your call partner is doing
#### What actually happened
Constant failure to maintain a connection via the camera change
#### Steps to reproduce
1. Have someone to call over your communicator
2. Enable video function and have no visual obstructions
3. Cry when you can't see what the other guy's doing without constantly clicking on the watch or communicator in your hand.
#### Additional info:
 - **Server Revision**: [a422d8b] 

Yes, I know I put in an old commit, but this is the latest Polaris code update that's being used on a live server version downstream.

PR #6713, opened 2020-02-20
Fixes communicator newsfeed app.

Atermonera:
Fixes non-terminal while loop in get_latest_news(), helper.dm
Newscaster posts now have a world.time-esque timestamp number, so they can be (almost) uniquely identified

[Tested](https://puu.sh/Fcwj6/f3cf2edd08.mp4) (.mp4)

"Latest news" posts are now sorted, in a sensible manner that most certainly does NOT use a while loop that doesn't terminate, murdering the entire server the moment a fourth news post is made from ANY system. And the post timestamps are from the first function I found in time.dm that seemed like a good fit, because there's so many of them at this point it's basically a matter of "I choose... _this_ one."

And since the fix that caused the server-crash was reverted, this isn't really a high priority, and speaking of which, I actually need to update this branch because I pulled from master before the revert so I could verify the issue.

### DIFF
--- a/code/game/objects/items/devices/communicator/UI.dm
+++ b/code/game/objects/items/devices/communicator/UI.dm
@@ -122,6 +122,8 @@
 	data["manifest"] = PDA_Manifest
 	data["feeds"] = compile_news()
 	data["latest_news"] = get_recent_news()
+	if(newsfeed_channel)
+		data["target_feed"] = data["feeds"][newsfeed_channel]
 	if(cartridge) // If there's a cartridge, we need to grab the information from it
 		data["cart_devices"] = cartridge.get_device_status()
 		data["cart_templates"] = cartridge.ui_templates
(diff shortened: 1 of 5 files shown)


code/game/objects/items/devices/communicator/UI.dm
```diff
@@ -280,6 +282,10 @@
 		var/obj/O = cartridge.internal_devices[text2num(href_list["toggle_device"])]
 		cartridge.active_devices ^= list(O) // Exclusive or, will toggle its presence
 
+	if(href_list["newsfeed"])
+		world << "newsfeed [href_list["newsfeed"]]"
```
Leshana: You probably did not intend to keep this.

TheFurryFeline:
Can we also see about getting #6531 fixed in some way?

Atermonera:
That's beyond the scope of this PR, which is specifically to fix the runtime originally fixed by #6706, and to fix the server-crash issue revealed by that fix and caused by a non-terminal while() loop due to fucky list semantics. I haven't ever looked at that side of communicator functions, so I can't make any binding promises.

TheFurryFeline:
> That's beyond the scope of this PR, which is specifically to fix the runtime originally fixed by #6706, and to fix the server-crash issue revealed by that fix and caused by a non-terminal while() loop due to fucky list semantics. I haven't ever looked at that side of communicator functions, so I can't make any binding promises.

Fair nuff. Just thought I'd bring it up since it's one of the parts of the communicator's functions. And you were here working on it.

PR #6714, opened 2020-02-20
Renames Make All Areas (Un)Powered secret

Heroman3003:
Into Power Failure Begin/End. Should reflect purpose of the secret much better. For restoring station power, Power All SMES should be used instead.


Atermonera:
Oh, oops, earlymerge.

PR #6715, opened 2020-02-20
Fixes Mint Apple Sparkle having a recipe, but no drink

LeighKnapenberger:
I don't know how github works


LeighKnapenberger:
I think I messed this up

LeighKnapenberger:
I think I get it now

PR #6718, opened 2020-02-21
Added two chemical spill PoIs + More Immersive PoI Names

Cerebulon:
Added two PoIs based on the recent chemical spill events:

Ruptured Canister
- Plains
- No loot
- Can be scanned for cataloguer points
- Mildly dangerous

Acrid Lake
- Wilderness
- Contains several canisters
- Random medkit, energy weapon and rigsuit
- Moderately dangerous

Also:
- Added cataloguer info for the Pascal-b and Decoupled Engine PoIs
- Renamed the 'front facing' names of many PoIs to be more descriptive, or at least not be numbered (e.g. "PoI - Epod1" is now "PoI - Crashed Escape Pod").
- Filler PoIs like snow patches with no notable features are no longer listed as "PoI - X" because they're literally nothing.

### DIFF
--- a/code/game/objects/items/poi_items.dm
+++ b/code/game/objects/items/poi_items.dm
@@ -3,10 +3,30 @@
 	icon = 'icons/obj/objects.dmi'
 	desc = "This is definitely something cool."
 
+/datum/category_item/catalogue/information/objects/pascalb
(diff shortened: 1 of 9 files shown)


TheFurryFeline:
I approve of these new POI names.

PR #6719, opened 2020-02-21
Adds an In-game Feedback System

Neerti:
Adds the ability to toggle on an in-game feedback system, which allows for anyone who qualifies, to write feedback about the server to the staff of the server.

Here's a video: [Demo](https://streamable.com/17qig)

The system has a few different settings that allow us or downstreams to tweak it.
Settings include whether to allow for people writing feedback to anonymize themselves (some extra setup also needs to be done server side for this), whether there should be a cooldown period between submissions, if there should be a player age restriction, and what topics should be available for users to choose from.

### Stuff for Players
If the system is toggled on, and you are not disqualified for whatever reason (e.g. you already submitted some feedback and you're still 'on cooldown'), you will see a new button in the center of your screen when in the lobby. Clicking it will open a window where you can write your thoughts, ideas, suggestions, complaints, gripes, or whatever, and submit it to the server.

### Stuff for Staff
There is a new admin verb called ``View Feedback``, which will open a window that shows the 50 most recent submissions. You can search for specific things like the submission ID, topics, words inside the content of the submission, and such.

### Stuff for Devs
This PR adds a few other things.
- It adds a nice href helper proc that makes adding href elements not need to copypaste from elsewhere due to having very specific formatting requirements.
- It also adds a 'managed browser' datum, which is like a regular browser object, but holds state, and the feedback form variant cleans itself up when its finished.
- Ports a macro from TG that makes View Variables be unable to see or do anything to a specific type. This is used on the feedback form object, to prevent snooping.
- Adds a few unit tests for this feedback system.
- This is also the first step into using a SQLite database. SQLite was used due to being vastly easier to test locally, compared to needing to set up a MySQL/MariaDB instance. In the future I'd like to be able to expand upon this system for things like persistent newscaster stories or statistics tracking. The possibility of replacing the 'real' database with this might also be considered in the distant future.


code/controllers/subsystems/sqlite.dm
```diff
@@ -0,0 +1,187 @@
...
+/datum/controller/subsystem/sqlite/proc/connect()
+	if(!config.sqlite_enabled)
+		return
+
+	if(!sqlite_db)
+		sqlite_db = new("data/sqlite/sqlite.db") // The path has to be hardcoded or BYOND silently fails.
+
+
+	if(!sqlite_db)
+		world.log << "Failed to load or create a SQLite database."
```
Atermonera: `to_world_log()`

code/controllers/subsystems/sqlite.dm
```diff
@@ -0,0 +1,187 @@
...
+		log_debug("ERROR: SQLite database is active in config but failed to load.")
+	else
+		world.log << "SQLite database connected."
```
Atermonera: `to_world_log()`

PR #6721, opened 2020-02-21
Updated corgi icon

listerla:
Ported from https://github.com/Baystation12/Baystation12/pull/25548

image


KasparoVy:
Damn that's adorable

PR #6728, opened 2020-02-23
Carbon mobs can be attacked again

Atermonera:
Data was being fed into the wrong parameter, resulting in garbage information being stuffed into `effective_force`, which was then evaluated to 0, resulting in no weapon attacks making it through.
Melee weapons, guns, and fists all tested, and now work. (Fists reported unaffected, guns likely were the same, tested for thoroughness)


SpadesNeil:
What the hell broke all this in the first place?

PR #6738, opened 2020-02-24
Fix Sundry Runtime Errors

Leshana:
* **Fix Runtime in seed.dm,102: Cannot read null.len**
	* Gotta check if the list is non-null before checking length! Tested by spawning plants.
* **Fix Runtime in filter.dm,91: Cannot execute null.set machine()**
	* Use `user` not `usr`.  Tested by spawning the machine, opening nanoui and walking away.
* **Fix Runtime in wires.dm,120: Cannot execute null.is wirecutter().**
	* You can click "Cut" with an empty hand!  Tested by doing that.
* **Fix Runtime in loadout.dm,277: bad index**
	* Runtime happens when someone has a saved preferences loadout containing an item that currently is expected to have gear_tweaks, but at the time that the savefile was last saved, did *not* have gear tweaks.  Thus the saved prefs have null metadata.
	* Also it makes no sense to check the length of a list *inside* a for loop iterating over that list..
	* Tested by manually setting up the situation with VV and then loading/saving/spawning.
* **Fix bad index and bad arg name runtimes in find_spawning.dm**
	* Code was assuming that item.origin_tech would be an initialized list. To be safe we use LAZYSET so it will work properly even if its null.
	* Other code was calling pickweight and passing it the options as named arguments.  They are supposed to be passed as a list.
	* Tested by spawning the /obj/item/weapon/archaeological_find and fixing things until runtimes stopped.

### DIFF
--- a/code/datums/wires/wires.dm
+++ b/code/datums/wires/wires.dm
@@ -117,7 +117,7 @@ var/list/wireColours = list("red", "blue", "green", "darkred", "orange", "brown"
 			var/obj/item/I = L.get_active_hand()
 			holder.add_hiddenprint(L)
 			if(href_list["cut"]) // Toggles the cut/mend status
-				if(I.is_wirecutter())
+				if(I?.is_wirecutter())
 					var/colour = href_list["cut"]
 					CutWireColour(colour)
 					playsound(holder, I.usesound, 20, 1)
--- a/code/ATMOSPHERICS/components/omni_devices/filter.dm
+++ b/code/ATMOSPHERICS/components/omni_devices/filter.dm
@@ -88,7 +88,7 @@
 	return 1
 
 /obj/machinery/atmospherics/omni/atmos_filter/ui_interact(mob/user, ui_key = "main", var/datum/nanoui/ui = null, var/force_open = 1)
-	usr.set_machine(src)
+	user.set_machine(src)
 
 	var/list/data = new()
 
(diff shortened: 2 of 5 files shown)


code/datums/wires/wires.dm
```diff
@@ -117,7 +117,7 @@ var/list/wireColours = list("red", "blue", "green", "darkred", "orange", "brown"
 			var/obj/item/I = L.get_active_hand()
 			holder.add_hiddenprint(L)
 			if(href_list["cut"]) // Toggles the cut/mend status
-				if(I.is_wirecutter())
+				if(I?.is_wirecutter())
```
Atermonera: Wait that's valid syntax?

Neerti: The byond docs say `This is used to access the procs and vars of a prototyped object. It is just like the . operator, but if the object is null, the access does not happen and there will be no runtime error.` and `When reading A?.B, it's equivalent to A && A.B except that A is only evalulated once, even if it's a complex expression like a proc call.`

Leshana: Null-safe member access operator. I was impressed that BYOND added it already, every language ought to add it.  In a situation like this,  doing  `I && I.is_wirecutter()` would have been just fine too.    But it really is cool in cases when getting an actual value rather than just being a boolean predicate, or when the left hand side is costly.

Atermonera: _cackling over the potential loss of characters typed and most likely usually exceedingly minor performance improvements_

PR #6742, opened 2020-02-24
 Better spriters for liquid dispensers

listerla:
Ported from https://github.com/ParadiseSS13/Paradise/pull/11728


code/game/machinery/ai_slipper.dm
```diff
@@ -24,9 +24,9 @@
 
 /obj/machinery/ai_slipper/update_icon()
 	if(stat & NOPOWER || stat & BROKEN)
-		icon_state = "motion0"
+		icon_state = "liquid_dispenser"
 	else
-		icon_state = disabled ? "motion0" : "motion3"
+		icon_state = disabled ? "liquid_dispenser" : "liquid_dispenser_on"
```
listerla: Gonna be straight up. I don't know what this does. If this would lead to unintended behavior, don't merge yet.

Neerti: The `?` operator is essentially an inline if statement, where the thing being evaluated is to the left of the ?, and what is returned from it is to the right of the ?, with the first thing being what happens if it's evaluated to true, and the second one is for when it is false.

In the example above, it'll make `icon_state` equal to `"liquid_dispenser"` if `disabled` is true, otherwise `icon_state` will be `"liquid_dispenser_on"`.

listerla: Thank you! That seems it works fine, then.

PR #6745, opened 2020-02-24
Removes Toy Story

Neerti:
Toy mice no longer breathe.


Atermonera:
Trying a thing to summon the icon diff

PR #6751, opened 2020-02-25
Ports tgstation bloodpack sprites

listerla:
image


Schnayy:
General practice is to not remove the old sprites if they're useable, and these ones weren't terrible. Could you please stick the old ones back in the dmi alongside the new ones?

PR #6753, opened 2020-02-25
Adds ecigs and vaping

listerla:
https://github.com/GeneriedJenelle/The-World-Server-Redux/pull/389


Cerebulon:
Nice. I kept meaning to do this but porting it makes way more sense actually

GeneriedJenelle:
Word of warning, those ecigs have been duds since forever - they need bay's SSobjs processing feature for the cigs to actually work. If you have that already, you might be better off porting theirs.

If you don't and manage it fix that for our version, pass it back to me. :)

Mechoid:
The various processing toggles for SSobj can be un-commented, as we do have SSobj.

PR #6755, opened 2020-02-26
Devs can now View Feedback

Neerti:
Apparently a lot less people have `R_ADMIN` than I thought.
Someday the permissions system needs to get redone.


Atermonera:
Some day, I will have buttons.

PR #6768, opened 2020-02-29
Salvageable machines

Arokha:
They're machines you can take apart with a crowbar to find some loot like computer components, things that can be recycled in autolathes, etc.

dreamseeker_2020-02-21_16-14-49


code/game/objects/items/trash_material.dm
```diff
@@ -0,0 +1,72 @@
...
+	)
+
+/obj/item/trash/material/circuit/Initialize()
+	. = ..()
+	icon_state = "circuit[rand(3)]"
+
+
+/obj/item/trash/material/device
+	name = "broken device"
+	desc = "A broken device that can be recycled in an autolathe."
```
Neerti: Not sure if this was meant to say protolathe, or if autolathes on virgo can use materaisl beyond steel and glass, but ours can't.

Arokha: Oh, was just copypaste from a Bay-based server. Maybe theirs can. Hmm.

Heroman3003: I believe protolathes themselves can't recycle any material here either. Its autolathe-only feature and its steel and glass only. Attempting to recycle something that doesn't have steel or glass will just fail, and attempting to recycle something that has more material than just steel or glass, will have that extra material lost.

Arokha: Well, it'll just work for the steel and glass parts then until someone upgrades the autolathes. Should be transparent to players in-game since as far as I'm aware there's no way to tell what materials are coded onto an item. Consider it motivation for someone to upgrade autolathe/protolathe code.

Mechoid: Autolathes were going to be the target of my material-expansion eventually, so this just pushes it forward.

PR #6773, opened 2020-03-01
Cutting rigs now retracts all components, to stop the dreaded pseudoprison.

### LINKED ISSUE #4544: Losing Power Bugs Rig Removal
Having a rigsuit deployed and losing power makes it impossible for an outside person to fully remove the suit. It doesn't matter if the suit is simply deployed or sealed (it can be cut but that doesn't solve the issue)
The suit to be capable of being removed even without power.
#### What actually happened
The back slot can be removed but all other pieces (head, gloves, shoes, exosuit) remain.
#### Steps to reproduce
1.) Equip rigsuit on back
2.) Deploy rigsuit
3.) Remove power cell to kill power (Having a dead power cell instead makes no difference)
4.) Attempt to remove rigsuit
#### Additional info:
 - **Server Revision**: master - 2018-01-14 cb22407e7cd99809f5eb79a55e9929c5d03c3a84 
The wearer of the rigsuit is still capable of manually removing all the pieces of the rigsuit by removing the back slot then removing all other pieces as if they were simply clothing (the clothing doesn't appear in the hand slot but they are moved and can then be dropped where they will disappear)

### LINKED ISSUE #6739: Rigsuit surgical removal is fucked all the way up
When you saw off a rigsuit and remove it, it only removes the control panel- not the associated rig parts, like chestpiece, boots, etc, which become irreparably stuck to the player they were on

for the rigsuit to come off in one piece

#### What actually happened
Molly Highlander became one with the AMI, permanently

#### Steps to reproduce
- Surgically saw off a RIG
- Attempt to remove
- watch as the control module comes off, but everything else remains on the player

#### Additional info:
 - **Server Revision**: master - 2020-02-23
2bea80895de462c4dcbe8796b37fdc4d1d800c33 

 - **Anything else you may wish to add** She was handcuffed if that has anything to do with it, and there was a pAI in the intellicard thingy jack

Mechoid:
Fix #6739
Fix #4544

Rigs now retract their components when cut off.

### DIFF
--- a/code/modules/clothing/spacesuits/rig/rig.dm
+++ b/code/modules/clothing/spacesuits/rig/rig.dm
@@ -234,19 +234,10 @@
 /obj/item/weapon/rig/proc/cut_suit()
 	offline = 2
 	canremove = 1
-	for(var/obj/item/piece in list(helmet,boots,gloves,chest))
-		if(!piece) continue
-		piece.icon_state = "[suit_state]"
-		if(airtight)
-			update_airtight(piece, 0) // Unseal
-		piece.canremove = 1
-		if(istype(loc, /mob/living/carbon/human))
-			var/mob/living/carbon/human/H = loc
-			H.drop_from_inventory(piece)
-		piece.forceMove(get_turf(src))
-		piece.dropped()
-		piece.canremove = 0
-		piece.forceMove(src)
+	toggle_piece("helmet", loc, ONLY_RETRACT, TRUE)
+	toggle_piece("gauntlets", loc, ONLY_RETRACT, TRUE)
+	toggle_piece("boots", loc, ONLY_RETRACT, TRUE)
+	toggle_piece("chest", loc, ONLY_RETRACT, TRUE)
 	update_icon(1)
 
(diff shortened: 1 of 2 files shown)


Nalarac:
So this would also fix #4544 as well as it's the same issue.

code/modules/clothing/spacesuits/rig/rig.dm
```diff
@@ -231,6 +231,24 @@
...
+	canremove = 1
+	for(var/obj/item/piece in list(helmet,boots,gloves,chest))
+		if(!piece) continue
+		piece.icon_state = "[suit_state]"
+		if(airtight)
+			update_airtight(piece, 0) // Unseal
+		piece.canremove = 1
+		if(istype(loc, /mob/living/carbon/human))
+			var/mob/living/carbon/human/H = loc
+			H.drop_from_inventory(piece)
```
Atermonera: You could instead add a `force` parameter to `/obj/item/weapon/rig/proc/toggle_piece()` (Line ~760), such that instead of having to do all this inventory management, you can simply tell the proc to ignore stuff like battery checks?

PR #6778, opened 2020-03-01
Let them finally eat cake

### LINKED ISSUE #6776: Chaos cake base layer impossible to make.
You cannot finish the first step to a chaos cake.

Delicious cake.

#### What actually happened
Burnt mess.

#### Steps to reproduce
Suffer for an hour getting the ingredients only to fail.

#### Additional info:

I believe it's because poisonberries do not have a kitchen tag.

### LINKED ISSUE #6777: Chaos cake has no sprite.
Chaos cake has no sprite.

To stack the cake layers.

#### What actually happened
Cake does a vanishing act.

#### Steps to reproduce
Place a layer onto the chaos cake. Watch it disappear right before your very eyes!

#### Additional info:

Was this even tested?

Schnayy:
Fixes #6777 and fixes #6776 

I only tested making the first layer and the last layer, as well as all the sprites, because if I go through and bake every layer of this cake right now I will cry.

### DIFF
--- a/code/modules/food/food/snacks.dm
+++ b/code/modules/food/food/snacks.dm
@@ -510,6 +510,25 @@
 		src.name = "Frosted Jelly Donut"
 		reagents.add_reagent("sprinkles", 2)
 
+/obj/item/weapon/reagent_containers/food/snacks/donut/poisonberry
+	name = "Jelly Donut"
+	desc = "You jelly?"
+	icon_state = "jdonut1"
+	filling_color = "#ED1169"
+	center_of_mass = list("x"=16, "y"=11)
+	nutriment_amt = 3
+
+/obj/item/weapon/reagent_containers/food/snacks/donut/poisonberry/Initialize()
+	. = ..()
+	reagents.add_reagent("sprinkles", 1)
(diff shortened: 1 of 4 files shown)


Heroman3003:
I don't really like idea of removing poisonberries being usable in normal berry recipies. That was a fun feature, IMO, similar to using poison apples in regular apple recipies.

Atermonera:
I mean, on the one hand, that should have been a consideration when the original author was first designing the cake. A simple fix would be to just add recipe datums for the recipes using berries that instead uses poisonberries.
I think the best fix would be to make the kitchen_tags a list object, since they're all strings. That would require tweaking the definitions of all the seed datums, and modifying the logic of the recipe processing to check against lists instead.

Schnayy:
Apologies if my code is crap but that should keep poisonberries an alternative. I also switched PB&J to use cherry jelly instead of berry juice, as I believe was originally intended.

TheFurryFeline:
Suggest also editing OP to be `fixes #6776` due to the way GitHub keywords work with closing issues.

PR #6781, opened 2020-03-01
Tweak Station-Railguns

Mechoid:
Station railgun tweaks as per Staff Meeting

### DIFF
--- a/code/modules/projectiles/guns/magnetic/magnetic_railgun.dm
+++ b/code/modules/projectiles/guns/magnetic/magnetic_railgun.dm
@@ -136,8 +136,8 @@
 	empty_sound = 'sound/weapons/smg_empty_alarm.ogg'
 
 	firemodes = list(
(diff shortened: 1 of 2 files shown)


FenodyreeAv:
As the only one who's actually used this thing, my thoughts:

As current, it's slightly worse than an egun until upgraded.
No weapon powercells hurt a lot, when you need sustained damage.
It's AP is too high, it's got the same AP as an Xray gun.
DNA lock is neat, but not that useful in practice. If I've dropped this thing I'm probably dead.

This change moves it from pretty decent, to basically useless unupgraded.
Breakdown of shots was 
2-lethal, 3 non-lethal on base.
4-lethal, 6 non-lethal on adv capacitor
6-lethal, 9 non-lethal on supercap.

This'd leave it firing 1 shot on lethal unupgraded, which is 30 damage, compared to the potential 200 on an energygun. This thing isn't even scary to a teshari.
2 non-lethal shots, is just enough to stun someone without armour, not enough to drop someone IN armour. So, useless in it's role, which is subduing armoured dude.

I'd say tweak how much capacitors improve it, and drop 20 points from the AP, rather than make it useless without upgrades.

PR #6786, opened 2020-03-03
Stabilizer stabilizes stably

Mechoid:
Stabilizer will slowly force the lungs to operate if they aren't dead, lowering them to just below the threshold for bruising, and lower losebreath.

Stabilizer will aid with oxyloss from damaged lungs and heart.


Atermonera:
Tested?

Mechoid:
Correctamundo

PR #6788, opened 2020-03-03
Antag Uplink Expansion

Mechoid:
Additions:
 - Energy Daggerpens (20): Disguised energy-knives, which do 15 searing on melee, or 30 when thrown.
 - Thieves gloves (30): Special gloves that allow you to peep in others' backpacks and belts, and plant items in their bags / pockets.
 - Buzzer Ring (30): Makes your first two punches electrically charged, first with 25 damage in a shock, then approximately 12.5 damage in the second. If the charge is over 90%, you can force-defib a corpse, even if it's a mindless one. Damage rules still apply, however time-of-death does not.
 - Resource Crate (60): A crate of uranium, phoron, gold, silver, osmium, and plasteel.
 - Exotic Plantlife Crate (20): A crate of numerous random seeds.
 - Spare Organ Crate (20): A crate of bioprinted organs.
 - Graviton goggles (15): A pair of combined meson/material goggles.
 - Integrated Circuit Printer (10): An upgraded circuit printer used to make integrated machine.
 - Flamethrower (60): A large, flame-based weapon.
 - 8 Concussion Grenades (30): A box of eight concussion grenades.
 - 4 Hunting Traps (30): A box of four hunting-traps, similar to those found in the explorer vendor.
 - 3 Virus Samples (40): A box of three unique virus samples.
 - Quickdraw Syringe Case (20): A case that can hold six syringes, and rapidly deploy them. Fits in your pocket.
 - Clotting Injector Case (20): A case like above which starts with 3 clotting med injectors instead.
 - Bonemed Injector Case (20): The same as above, but with bonemeds.

Tweaks:
 - Announcement costs lowered to be more equivalent.
 - Egun changed from 60 to 30 TC. It is nowhere near as powerful as a lascannon.
 - Exosuit rigged laser from 60 to 30 TC.
 - Xray gun from 85 to 60 TC.
 - Augments can now be used by everyone, as robot-specific ones will require FBP organ revamp.
 - Anti-Materiel Rifle can once again be used with thermals.
 - Augment guns use slightly less blood / system instability to charge, so it doesn't kill you dead in moments.

Fixes:
 - Energy Shields work again, and can be colored.


Wickedtemp:
Does the flamethrower work

Mechoid:
image

code/modules/projectiles/guns/magnetic/gasthrower.dm
```diff
@@ -0,0 +1,77 @@
...
+	Tank.air_contents.remove(moles_to_pull)
+
+/obj/item/weapon/gun/magnetic/gasthrower/show_ammo(var/mob/user)
+	..()
+
+	if(loaded)
+		var/obj/item/weapon/tank/T = loaded
+		to_chat(user, "<span class='notice'>\The [T]'s pressure meter shows: [T.air_contents.return_pressure()] kpa.</span>")
+
+		switch(check_ammo())
```
Atermonera: switching on a true/false value should probably just be an if(thing) else

Atermonera:
Have you tested each individual traitor toy?

Mechoid:
Second wave of testing revealed that the cases were in fact too small. Now fixt.

PR #6790, opened 2020-03-05
Water -> Wet

Arokha:
Allows using water turfs as water sources for buckets and the like. There's a var for what reagent type you want just in case there's like poison water or gross water or something who knows.

### DIFF
--- a/code/game/turfs/simulated/water.dm
+++ b/code/game/turfs/simulated/water.dm
@@ -16,6 +16,8 @@
 
 	var/depth = 1 // Higher numbers indicates deeper water.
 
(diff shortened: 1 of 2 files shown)


Cerebulon:
There is a poisoned water tile subtype, I guess this will require the adding of an actual poison water reagent rather than just having extra code on the tile that gives you tox on contact?

Arokha:
Ye, if you want to do that, that's how you'd do it. Just set the reagent_type on that turf to your new poison water reagent if you make one.

Arokha:
Changelogged

PR #6799, opened 2020-03-08
New book sprites + inhands

Cerebulon:
Changes book sprites to generally nicer ones (old ones retained where they were actually visually interesting), adds inhand sprites for books + manuals.

Sprites from https://github.com/Aurorastation/Aurora.3/pull/8395 but rearranged to keep colours consistent with what we currently have.

**NOTE:** The custom books repo will need to port the new custom_books.dmi
Custom books should also be given an appropriate item_state for inhands and related inhand icon pointer code.
It was also probably need left/right inhand files and submissions of custom book covers should (ideally) include inhand sprites or an equivalent inhand colour.


Heroman3003:
Why the colorchange on spesslaw tho? All the other books retained same colors (or at least color scheme).

Cerebulon:
I imagine it's to distinguish it from red corpregs? Which we use a totally different style of sprite (codex sprite) for anyway, but not including the new sprite for that one book would be a little odd.

PR #6803, opened 2020-03-08
Cooking adjustments

Schnayy:
The first of many PR's adjusting cooking. Does the following:

- Adds the reagent yeast.
- Bread recipe altered, no longer uses egg, now uses 5 units yeast.
- Croissants and baguettes now use 5 units yeast in addition to former ingredients.
- Sprites added for yeast, peanut butter, and mayonnaise condiments.
- Corn oil added to Booze-o-Mat.
- Yeast and corn oil added to kitchen vendor.
- Yeast added to kitchen supply crate in cargo.
- Fixes the boiled slime core sprite being invisible.

Before anyone can debate adding corn oil to the vendors: Corn oil has been ridiculously annoying to get ever since it was added. It is also used in a lot of things. It's very tiring to be a bartender with no botany and being unable to make the cool promethean drinks. If you're worried about it being used for nitroglycerin, that already calls for a chem dispenser, so it's whatever.

I will update the wiki recipes when the server is updated after this PR is merged.


code/modules/food/food/condiment.dm
```diff
@@ -390,3 +409,9 @@
 	. = ..()
 	reagents.add_reagent("flour", 30)
 	randpixel_xy()
+
+/obj/item/weapon/reagent_containers/food/condiment/yeast
+	name = "Yeast"
+
+/obj/item/weapon/reagent_containers/food/condiment/cornoil
```
Atermonera: These object definitions should be directly adjacent to their initialize calls

PR #6804, opened 2020-03-08
Big Bastard Incidental Sound Port

Cerebulon:
Ports an absolute shit ton of incidental sounds, but mostly sounds when items are dropped or thrown.

Dropped/Thrown item sounds are toggleable in preferences.

New/Added sounds for:
Vending machine products
Everything when dropped/thrown (Not placed)
Matches and cigarettes
Chem dispensers
Parcel wrapping
Pens and paper (Writing, burning, crumpling)
Books
Filing cabinets
Sinks
Accessing toolboxes, pill bottles and briefcases
Magboots activation
(More) holsters
Wheeling beds and chairs
Arrivals announcement
Stamps

Did NOT port radio sounds or other stuff that was quickly removed from original PRs. Also did not reorganize the entire sounds folders.

Primarily ported from: https://github.com/Aurorastation/Aurora.3/pull/6307
Followed by a dozen subsequent tweaks and fixes to that PR, including but not limited to:
https://github.com/Aurorastation/Aurora.3/pull/8131 (sinks)
https://github.com/Aurorastation/Aurora.3/pull/6533 (Just the sounds)
https://github.com/Aurorastation/Aurora.3/pull/4675
https://github.com/Aurorastation/Aurora.3/pull/6182
https://github.com/Aurorastation/Aurora.3/pull/6688
https://github.com/Aurorastation/Aurora.3/pull/6397

Will likely need tweaking/adding to as we find stuff which sounds weird or whatever. But it's been tested to me ability.


Cerebulon:
whats up with travis....

Cerebulon:
One point for review: I'm not 100% positive if the preference checks (in the inventory.dm and items.dm drop and throw item procs respectively) are correct. They work locally but I couldn't easily test if it's
If thrower/dropper has this preference off, do not play sound (which would be pointless)
Or
If a person has this preference off and somebody throws/drops item, do not play sound for that person (the intention of the pref)

sound/items/storage/toolbox.ogg
```diff
@@ -0,0 +1,1181 @@
+
+
+
+
+
+
+<!DOCTYPE html>
```
Atermonera: uuuuuuhhhhhhhh....

Cerebulon: Oh my god did i miss the download button with Save As and save the fucking github page as toolbox.ogg

code/game/objects/items/weapons/cigs_lighters.dm
```diff
@@ -225,6 +230,7 @@ CIGARETTE PACKETS ARE IN FANCY.DM
 			to_chat(H, "<span class='warning'>\The [blocked] is in the way!</span>")
 			return 1
 		to_chat(H, "<span class='notice'>You take a drag on your [name].</span>")
+		playsound(src, 'sound/items/cigs_lighters/inhale.ogg', 50, 0, -1)
```
Atermonera: inb4 being deafened by the sound of someone inhaling becomes a meme

Cerebulon: This sound only plays when somebody does the specific interaction action of "taking a drag" (Click self with cigarette in hand) so it's an active sound, not continuous.
Before testing I originally had it in the smoke() proc  so it was continuous while a cig was equipped and yes it did sound like a panicking asthmatic

Atermonera:
It looks like `playsound()` actually has a `preference` parameter that, when set, will make it check the specified pref for all players it tries to play the sound to.
`  
/proc/playsound(atom/source, soundin, vol as num, vary, extrarange as num, falloff, is_global, frequency = null, channel = 0, pressure_affected = TRUE, ignore_walls = TRUE, preference = null)`

Since it's the very last parameter, if you stick a `preference = SOUND_DROPPED` at the end of the parameters for the playsound calls, it should probably work?

`playsound()` and `playsound_local()` (Which actually checks the pref) are in https://github.com/PolarisSS13/Polaris/blob/5fb77b35ec2fd23b062ba3129339ff33a5794f60/code/game/sound.dm

Cerebulon:
Had to use the datum path rather than the Key but that seems to have worked.

PR #6805, opened 2020-03-09
Ports Inactivity SS from Bay

Novacat:
Remember that server-destroying timer bug that VORE keeps getting?

Turns out, it is related to the inactivity subsystem, that, due to having SS_BACKGROUND and SS_NO_TICK_CHECK flags, causes the master controller to break in a horrible way. One of the consequences of the master controller breaking is that the timer subsystem stops running.

I am relieved that this monster of a bug has finally been slain. Hopefully.

### DIFF
--- a/code/controllers/subsystems/inactivity.dm
+++ b/code/controllers/subsystems/inactivity.dm
@@ -1,42 +1,58 @@
 SUBSYSTEM_DEF(inactivity)
-	name = "AFK Kick"
-	wait = 600
-	flags = SS_BACKGROUND | SS_NO_TICK_CHECK
+	name = "Inactivity"
+	wait = 1 MINUTE
+	flags = SS_NO_INIT | SS_BACKGROUND
+	var/tmp/list/client_list
+	var/number_kicked = 0
 
-/datum/controller/subsystem/inactivity/fire()
-	if(config.kick_inactive)
-		for(var/i in GLOB.clients)
-			var/client/C = i
-			if(C.is_afk(config.kick_inactive MINUTES))
-				to_chat(C,"<span class='warning'>You have been inactive for more than [config.kick_inactive] minute\s and have been disconnected.</span>")
-				var/information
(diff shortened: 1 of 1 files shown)


MrStonedOne:
>I am relieved that this monster of a bug has finally been slain

The bug hasn't been fixed because there is still a bug in the mc.

I can fix the bug in the mc if you would explain what the bug is outside of a vague "the bug". What are the symptoms? What is or isn't happening that is incorrect?

Novacat:
Basically, what happens is that in the code snippet I posted above, which was designed to allocate 10% more resources to queue_priority_count to help along a subsystem with SS_NO_TICK_CHECK. However, a subsystem with SS_NO_TICK_CHECK and SS_BACKGROUND uses queue_priority_count_bg, which decreases when more resources are allocated to queue_priority_count.

what happens is that the subsystem gets starved of resources, spams the master controller with requests for more resources, which results in queue_priority_count getting a rediculously high number and queue_priority_count_bg getting a rediculously low negative number

Generally, this bug can be avoided by just not having any subsystems with SS_BACKGROUND and SS_NO_TICK_CHECK flags at the same time.

Arokha:
@MrStonedOne 
master.dm:439
```
			if (queue_node_flags & SS_NO_TICK_CHECK)
				if (queue_node.tick_usage > TICK_LIMIT_RUNNING - TICK_USAGE && ran_non_ticker)
					queue_node.queued_priority += queue_priority_count * 0.1
					queue_priority_count -= queue_node_priority
					queue_priority_count += queue_node.queued_priority
					current_tick_budget -= queue_node_priority
					queue_node = queue_node.queue_next
					continue
```

If a subsystem is both `SS_BACKGROUND` and `SS_NO_TICK_CHECK` it will fall into this branch and MC will incorrectly pretend that allocating more queue_priority_count will fix it, even though that does not affect SS_BACKGROUND subsystems. It will eventually starve all non-background subsystems in an effort to try and get this one background subsystem to run.

Issue #6806, opened 2020-03-09
Runtime: The cigarette attempted to add a reagent called 'menthol' which doesn't exist.


Neerti:
#### Brief description of the issue
A runtime error occurs on server initialization, quite frequently due to the large number of cigs mapped in.

#### What you expected to happen
No runtimes to happen before the round even starts.

#### What actually happened
```
[2020-03-09T09:15:03] Runtime in unsorted.dm,1315: The cigarette attempted to add a reagent called 'menthol' which doesn't exist. ()
  proc name: crash with (/proc/crash_with)
  src: null
  call stack:
  crash with("The cigarette attempted to add...")
  /datum/reagents (/datum/reagents): add reagent("menthol", 2, null, 0)
  the cigarette (/obj/item/clothing/mask/smokable/cigarette/menthol): Initialize(1)
  Atoms (/datum/controller/subsystem/atoms): InitAtom(the cigarette (/obj/item/clothing/mask/smokable/cigarette/menthol), /list (/list))
  the cigarette (/obj/item/clothing/mask/smokable/cigarette/menthol): New(1)
  the cigarette (/obj/item/clothing/mask/smokable/cigarette/menthol): New(the pack of Temperamento Menth... (/obj/item/weapon/storage/fancy/cigarettes/menthols))
  the cigarette (/obj/item/clothing/mask/smokable/cigarette/menthol): New(the pack of Temperamento Menth... (/obj/item/weapon/storage/fancy/cigarettes/menthols))
  the cigarette (/obj/item/clothing/mask/smokable/cigarette/menthol): New(the pack of Temperamento Menth... (/obj/item/weapon/storage/fancy/cigarettes/menthols), null)
  the cigarette (/obj/item/clothing/mask/smokable/cigarette/menthol): New(the pack of Temperamento Menth... (/obj/item/weapon/storage/fancy/cigarettes/menthols))
  the pack of Temperamento Menth... (/obj/item/weapon/storage/fancy/cigarettes/menthols): Initialize(1)
  the pack of Temperamento Menth... (/obj/item/weapon/storage/fancy/cigarettes/menthols): Initialize(1)
  Atoms (/datum/controller/subsystem/atoms): InitAtom(the pack of Temperamento Menth... (/obj/item/weapon/storage/fancy/cigarettes/menthols), /list (/list))
  the pack of Temperamento Menth... (/obj/item/weapon/storage/fancy/cigarettes/menthols): New(1)
[...]

PR #6808, opened 2020-03-09
Makes Travis fail PRs that runtime during init/unit testing.

Neerti:
Steals a bit of code from Bay, which will hopefully stop bugs like #6806 from happening again.
Also moves two vars to GLOB because why not.
Note that this PR should fail due to that issue linked above.


Neerti:
This should probably not be merged until #6815 is fixed.

Atermonera:
Travis has detected no further issues, is this good to be merged now?

Neerti:
Unless there's another stealth bug going around then it should be fine.

PR #6810, opened 2020-03-10
QoL: Hacking

Neerti:
Makes the hacking interface a lot nicer to use.

Video demonstration: https://streamable.com/br4vw

Changes:
- The `cut` and `pulse` buttons on the hacking window will now check all available hands instead of only checking the active hand. For most people that means you can have a multitool and wirecutters in both hands, and be able to hit both buttons without leaving the window. For borgs, it will check all three of their module slots. This feature is _octopus ready_.
- When the hacking window refreshes, and something has changed on the lower text area, the thing that changed will be bolded until the next refresh. This makes it very easy to see what your click just did.
- Hacking a secure door now opens a slightly longer window, so no scroll bar appears by default.
- Hacking an autolathe no longer requires you to hack it using the autolathe's main window, but instead opens the hacking window like everything else. No more having to scroll down every pulse.

Dev stuff:
- Adds a 'wire hint' object that keeps track of if it should output a bold line or not.
- Adds a proc to get a list of everything a mob is holding. Gives it to humans, borgs, and simplemobs with hands. Gives simple_mobs the ability to use the `is_holding_item_of_type()` proc, so now Ater can hack airlocks as a space goose with hands.
- Makes hacking window use `is_multitool()` instead of typechecking for a multitool.
- Overrides `is_multitool()` on the multitool to return `TRUE`.


Leshana:
It is unfortunate all those separate vars are needed, it would be nice if these could be in a list instead of separate vars to ease the tedium of null'ing all those vars individually!   Unfortunately the the wire constants are powers of two instead of indices, so oh well.  Doing log2() of each one is hardly pretty either.  Plus, the additional 8 datum instances for every airlock.
But the both hands thing sounds like it will be convenient.  We ought to get around to implementing all those is_tooltype() things someday.

PR #6812, opened 2020-03-11
Gives +MOD the traitor panel and colorable OOC.

Neerti:


Neerti:
Someday I should overhaul the permissions system to not depend on flags so we can have `all the permissions`.
But that day is not today.

Atermonera:
For those unaware, we tried removing +ADMIN and +BAN from event managers, only to find that they could no longer use the traitor panel. That's an important thing for them, so this is a fix for that.

PR #6813, opened 2020-03-11
Fixes the occupation screen.

Neerti:
Occupation screen is now generated using `SSjob` instead of iterating over a list of jobs in the hopefully soon to be defunct `job_controller`. This means the occupation screen code is aware of departments, and will group then in the correct order. It will also avoid splitting departments over two columns, if it can help it.
The department datums now have a list of 'primary jobs', which is just jobs that have their department as the first one in the jobs' list of departments. This makes it so no duplicates can happen.
Also swaps a few colors so the head role has the 'bright' color and the subordinate jobs have the 'dark' version of that color, for consistency.

It now looks like this
Example of new occupations screen


TheFurryFeline:
I approve of this new layout.

PR #6816, opened 2020-03-11
[Bugfix/Oversight] Response Team Channel Shortcut

TheFurryFeline:
Adds headset key shortcut to type in rather than having to resort to, well, :H to speak on the Response Team channel when examine shows the usage is `:-` which doesn't work squat.

Changelog Notes:

- Fixes lack of easily accessible Response Team shortcut to type in. Type .k or :k to speak on the channel.


TheFurryFeline:
Note to self, when there are non-English letters in code files, don't do web-edits.

PR #6819, opened 2020-03-12
Makes language keys case sensitive, adds unit tests for language conflicts, fixes some language conflicts

Meghan-Rossi:
*Adds unit tests to catch languages with the same name or language key
*Makes language keys case sensitive (Because we were running really low on free keys to use
*Fixes Chimpanzee and EAL both using the language key "6" by changing Chimpanzee's language key to "C"
*Fixes Mouse and Bird both using the language key "m" by changing Bird's language key to "B"
*Lets log_debug work during global list initialization by moving logging and config setup slightly early in initialization

Tested and seems to work fine.

### DIFF
--- a/code/_global_vars/lists/species.dm
+++ b/code/_global_vars/lists/species.dm
@@ -1,6 +1,8 @@
 //Languages/species/whitelist.
 GLOBAL_LIST_INIT(all_species, list())
 GLOBAL_LIST_INIT(all_languages, list())
+GLOBAL_LIST_INIT(language_name_conflicts, list())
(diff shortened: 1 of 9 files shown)


zekesulastin:
I’m kind of ambivalent about making them case sensitive (on our server at least), especially when it does not really feel like there are actually 26 (a-z) + 10 (1-0) + punctuation language keys (I know the Occurses event language uses “[“) actually being used right now.

Meghan-Rossi:
37 different language keys used, so yeah if we use punctuation we can get away with leaving them case insensitive.  Not sure that's better though.

PR #6820, opened 2020-03-12
Emergency lighting

Meghan-Rossi:
Fixed version of https://github.com/PolarisSS13/Polaris/pull/6686 .  Lets AIs still flicker lights and stops building lights from always putting cells in them.

### DIFF
--- a/code/modules/power/apc.dm
+++ b/code/modules/power/apc.dm
@@ -104,6 +104,7 @@
 	var/beenhit = 0 // used for counting how many times it has been hit, used for Aliens at the moment
 	var/longtermpower = 10
 	var/datum/wires/apc/wires = null
+	var/emergency_lights = FALSE
 	var/update_state = -1
 	var/update_overlay = -1
 	var/is_critical = 0
@@ -794,6 +795,7 @@
 		"gridCheck" = grid_check,
 		"coverLocked" = coverlocked,
 		"siliconUser" = issilicon(user) || isobserver(user), //I add observer here so admins can have more control, even if it makes 'siliconUser' seem inaccurate.
+		"emergencyLights" = !emergency_lights,
 
 		"powerChannels" = list(
 			list(
@@ -834,7 +836,7 @@
 	if (!ui)
 		// the ui does not exist, so we'll create a new() one
         // for a list of parameters and their descriptions see the code docs in \code\modules\nano\nanoui.dm
-		ui = new(user, src, ui_key, "apc.tmpl", "[area.name] - APC", 520, data["siliconUser"] ? 465 : 440)
+		ui = new(user, src, ui_key, "apc.tmpl", "[area.name] - APC", 520, data["siliconUser"] ? 490 : 465)
 		// when the ui is first opened this is the data it will use
 		ui.set_initial_data(data)
 		// open the new ui window
@@ -930,6 +932,14 @@
 		failure_timer = 0
 		update_icon()
 		update()
+
+	else if (href_list["emergency_lighting"])
+		emergency_lights = !emergency_lights
+		for(var/obj/machinery/light/L in area)
+			if(!initial(L.no_emergency)) //If there was an override set on creation, keep that override
+				L.no_emergency = emergency_lights
+				INVOKE_ASYNC(L, /obj/machinery/light/.proc/update, FALSE)
+			CHECK_TICK
 
 	else if (href_list["breaker"])
 		toggle_breaker()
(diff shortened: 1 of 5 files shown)


Shadow-Quill:
Thanks! 
Kindasortamaybe lost will to look at code for a while.

Leshana:
When I have time I'll see about a method for doing this that does not create a couple thousand additional cell objects.

Atermonera:
Two ideas re: not making thousands of cells:
1) Make them draw power from a master cell (or SMES, as the case may be) or
2) Bake the type of the cell and the cell's variables you actually care about (charge, max charge, etc) into the light, and track it on there. If something interesting (and comparatively rare) happens, like an emp, make a new cell for each affected light, update the cell with the vars the light tracked, apply the interesting thing to the cell, update the light's vars, then qdel the cell.

... 1 is probably actually the way to go now that I write it, whether you do it per z-level, per z-level group (i.e., using_map.station_z_levels), or for all of them ever, everywhere. (2 was quickly accepted as a horrible mess of spaghetti that coders 1 year down the line would reel in terror at when it broke)

Shadow-Quill:
I believe the reasoning for each light having a cell is so that you can have lights without cells roundstart, that'll never have emergency lighting (ie: pois), and so you can have sections of the station die at different times, but still run out of power in the same time.

Atermonera:
> SDQL query results: [ select[ /obj/machinery/light , [ world ] ] ]
> SDQL query completed: 1091 objects selected by path, and 
> SDQL query took 0.7 seconds to complete.

Having lights without cells would be a matter of a single boolean variable.
Your wording re: sections dying/going dark at different times is hard for me to read properly, but I think you could mitigate that by splitting the "master cell" object up like substations. If the APC can't give lights in one room, then the lights with the bool enabled in that room would instead pull power from the master cell. Once the master cell dies, anything in the department that can't get power from its APC will also go dark.
We currently see instances of particular rooms going dark when the engine hasn't been turned on, just by virtue of the variability of the ambient power draw across all rooms, and that would still occur, but if a master cell is out of juice, then the whole department goes dark all at once. If that makes sense.

code/modules/power/apc.dm
```diff
@@ -931,6 +933,14 @@
 		update_icon()
 		update()
 
+	else if (href_list["emergency_lighting"])
+		emergency_lights = !emergency_lights
+		for(var/obj/machinery/light/L in area)
+			if(!initial(L.no_emergency)) //If there was an override set on creation, keep that override
+				L.no_emergency = emergency_lights
+				INVOKE_ASYNC(L, /obj/machinery/light/.proc/update, FALSE)
```
Atermonera: `.proc`?

code/modules/power/cells/power_cells.dm
```diff
@@ -132,3 +132,16 @@
...
+	name = "miniature power cell"
+	desc = "A tiny power cell with a very low power capacity. Used in light fixtures to power them in the event of an outage."
+	maxcharge = 120 //Emergency lights use 0.2 W per tick, meaning ~10 minutes of emergency power from a cell
+	matter = list("glass" = 20)
+	w_class = ITEMSIZE_TINY
+
+/obj/item/weapon/cell/emergency_light/Initialize()
+	. = ..()
+	var/area/A = get_area(src)
+	if(!A.lightswitch || !A.light_power)
```
Atermonera: So only drain the initial charge if it neither has a light switch nor initial power? i.e., if it has a light switch but no power, or if it has power but no switch (or both, implicitly), give it cell charge?

code/modules/power/lighting.dm
```diff
@@ -61,9 +66,42 @@ var/global/list/light_type_cache = list()
...
 			to_chat(user, "The casing is closed.")
+	if(cell_connectors)
+		if(cell)
+			to_chat(user, "You see [cell] inside the casing.")
+		else
+			to_chat(user, "The casing has no power cell for backup power.")
+	else
+		to_chat(user, "<span class='danger'>This casing doesn't support power cells for backup power.</span>")
+
+/obj/machinery/light_construct/attack_hand(mob/user)
```
Atermonera: Shouldn't this have a ..() somewhere in case of construction steps being handled on /obj/machinery, or are all the other steps done solely in attackby()?

code/modules/power/apc.dm
Meghan-Rossi: That's right.  It's a proc typepath shortcut, there's a comment about them at the top of callback.dm

code/modules/power/cells/power_cells.dm
Meghan-Rossi: Not quite.  It's drain the initial charge if either the lights are switched off in that area or there's no power in the area's lighting power channel.  So give it charge only if the lights are on and powered.

code/modules/power/lighting.dm
Meghan-Rossi: /obj/machinery/attack_hand(user) does some checks on whether user can use the machine.  I've now updated /obj/machinery/light_construct/attack_hand(user) to call it.

PR #6821, opened 2020-03-12
Xenoarch fix

PrismaticGynoid:
Fixes the bug in xenoarchaeology where there was a significant chance an artifact would break or come out as a strange rock, even when you did everything correctly. Now that only happens if you did the excavation wrong.


Cerebulon:
OH my god, it was an INDENTATION ERROR?

TheFurryFeline:
Bruh. That's a big bruh moment.

PR #6823, opened 2020-03-13
Makes PDA beeps a toggable option.

lbnesquik:
image

I just like being a bit discrete about using my PDA sometimes.


Neerti:
Isn't there already a button to turn the ringer off?

TheFurryFeline:
There is. Wew.

TheFurryFeline:
But this is for the other thing, not the ringer.

lbnesquik:
@Neerti 
There is the ringer that comes when you have message, but there is also little beeps whenever you interact with the PDA's now. This allow you to disable that. #6823

TheFurryFeline:
#6801 is actually the PR

lbnesquik:
> 
> 
> #6801 is actually the PR

Yes thank you

PR #6824, opened 2020-03-13
Spreads out shock damage

Jamini:
All Damage taken on an electrocution from an object is currently taken to the defense zone actively used to trigger the object (commonly hands)
With increasing power values on the station overall and the addition of dismemberment from burn damage, this causes shocks to unprotected hands to effectively always remove the offending hand on the Polaris base map.

Changes 
- Electrocution on human mobs now deals four instances of damage. One at 25% to the hand, one at 50% to the body, and two more at 25% to random bodyparts.

Current Values

At round-start (so 300kw in system from solars, another 150 from the SMES to power the station) this would result in going from a roughly 40 damage burn to the hand (more for some species. Which, seemingly, is enough to often dismember it) to about 10 to the hand, 20 burn to the body, and two more 10 damage burns.


Wickedtemp:
Is this all sources of shock damage or just hacking/airlock/vendor related

Neerti:
It's inside of `/mob/living/carbon/electrocute_act()` so its all sources of zapping.

Jamini:
It is anything touched by Electrocute_act(). A quick search shows the following would be impacted

Shocks from Machinery or Cables
SMES deconstruction discharge (This targets the torso. So the target still would take 75% of the damage on the torso, and another 
Technomancer Lightning
Technomancer Instability
Tesla Ball Shocks
Energy Siphon
Bioelectric Genesis
Shock crate damage
Defib shock damage


The overall damage of these attacks remains the same, and other sources of burn (lasers, fires) would not call this proc.  Stun times of attacks also are not touched. (So if you get electrocuted by chain lightning, you will still be stunned just as long as before)

Jamini:
Also I cannot into math so I need to correct something.

PR #6825, opened 2020-03-13
Shock modification

Jamini:
All Damage taken on an electrocution from an object is currently taken to the defense zone actively used to trigger the object (commonly hands)
With increasing power values on the station overall and the addition of dismemberment from burn damage, this causes shocks to unprotected hands to effectively always remove the offending hand on the Polaris base map.

Changes

Electrocution on human mobs now deals four instances of damage. 

One at 20% to the hand, 
One at 40% to the body,
and two more at 20% to random bodyparts.

Current Values

At round-start (so 300kw in system from solars, another 150 from the SMES to power the station) this would result in going from a roughly 40 damage burn to the hand (more for some species. Which, seemingly, is enough to often dismember it) to about 10 to the hand, 20 burn to the body, and two more 10 damage burns.

---
This would impact anything calling the electrocute_act() proc in carbon.dm

This impacts: 

Shocks from Machinery or Cables
SMES deconstruction discharge (This targets the torso. So the target still would still take 60% damage to the torso regardless)
Technomancer Lightning
Technomancer Instability
Tesla Ball Shocks
Energy Siphon
Bioelectric Genesis
Shock crate damage
Defib shock damage

Total damage value remains the same, however the damage would be spread out on the target instead of centralized.

Edit: I can also into math


FenodyreeAv:
I like this, it means you're less likely to lose a hand under normal operations, and you CAN die from electrocution under extreme ones, like you should if you lick a 2 megawatt power cable.

Atermonera:
Tastes like fried tongue!

PR #6829, opened 2020-03-13
QoL: Projectiles

### LINKED ISSUE #6220: No log shown when shooting a mob

When shooting another mob with a weapon no log is shown in game to indicate you shot your target or 'narrow missed'

To see some sort of indication that I have hit my target with a laser or a ballistic gun

#### What actually happened

No log shown that I had hit the target

#### Steps to reproduce

Take two mobs, set one upright, spawn gun, shoot at the other mob.

#### Additional info:

Server revision: master - 2019-06-02
305ffcb7a776050b4cfb7ea4a9d67a8aa3eb0618 

Still shows attack logs for admins about shooting at targets. 

https://i.imgur.com/3H2DbMP.png

Neerti:
Makes firing projectiles of all kinds better by giving more feedback, both with visuals and sounds. Most of this was ported from TG, with all the sounds from there as well.

Video demo: https://streamable.com/tga6g
The screen shaking is pretty severe but changing that might be a balance thing for later.
Also the runtimes in the video are fixed, but hivebots tend to break a lot of stuff when they delete themselves. Making them not immediately do that would probably help but is outside of scope for this PR.

Changes:
 - Projectiles can have an 'impact' visual effect happen when they hit something. This effect works with the pixel projectile system, so you can see if your shots are making it across the corner of a wall or not.
   - Adds self-coloring impact effect for projectiles that use `light_color` or `color`.
- Adds ability for projectiles to play a sound when hitting a mob, and a different sound when hitting a non-mob, like a wall or girder. This gives some audio feedback on if you hitting someone/getting hit or not, although it might be hard to hear if your gun is loud.
- When a projectile 'narrowly misses' someone, a sound is played on the mob who almost got hit.
- Fixes #6220 , a bug that made it so the 'X was hit in the Y by the Z' message not appear most of the time.
- The person who got hit by a projectile now receives bigger text that they got shot. Everyone else will see the same message that the above bulletpoint now makes happen.
- Reorganizes the `playsound()` switch statement somewhat.
- Makes it so the message that says someone fired a gun is shown before someone getting hit with a hitscan projectile.
- Ports the two bat sound files since the attribution file mentioned it and I feel like if I don't bring all the files over and instead edit the attribution, then someone might port the sounds later, not realize that the file was changed, and the sounds would be unattributed and that would be a Bad Thing.


seahen:
This looks like it could work.

lbnesquik:
Really liking this! Nicely done!

Leshana:
This looks like it should work.

PR #6830, opened 2020-03-13
Make diagonal shuttle walls look less bad

Arokha:
Makes them adjust lighting and such to look nicer when over space since space is unlit, so it removes that weird 'dark corner' effect.

***NOTE***: The first commit in this (7c9cc38) might make space look weird under said corner turfs until you have ~~overmap~~ skybox.


Atermonera:
Since the icondiff isn't very informative, can you provide a screenshot of the new visual?
Current, for reference:
arrivals shuttle corner
pilot shuttle corner

Arokha:
You should probably just wait until you get skybox ported, and then You'll See What I Mean:tm:

Atermonera:
In that case, dnm until overmap (or whichever one is the skybox) is ported

Atermonera:
what they look like without this fix

PR #6725, opened 2020-02-22
Job description alt

Meghan-Rossi:
Fixed version of https://github.com/PolarisSS13/Polaris/pull/5489 .  
Including
* Resolved merge conflicts
* Spelling and grammar corrections
* Added missing alt-title datum for Research Director

PR #6839, opened 2020-03-14
Streamline Job Alt Titles

Leshana:
Continues PolarisSS13/Polaris#6725, implementing suggestions to reduce extra code.
- Remove the need for an alt title to be declared for the job's main title.
- Don't allocate a list if no alt titles are used, and use the indexes provided instead of looping to search the alt titles list.

All the existing preferences code except for two procs already expected the main title to be absent from the alt_titles list, so this is consistent with prior expectations.


Leshana:
Rebased to resolve conflicts.

PR #6840, opened 2020-03-15
QoL: Various Notifications

Neerti:
Ports some code and sounds from TG, concerned with letting the player know when specific important things have happened, which should make it harder to miss if they were alt-tabbed. Only the player that is being notified can hear these sounds.
These notifications also 'flash' the player's window, meaning their OS will make the tab holding the game turn yellow if they're alt-tabbed.

- Being defibbed, technomancer resurrected, or being put into the scanner while a ghost, will play a sound to the ghost, and give a message in the chat bar that someone's trying to revive them, along with a link in the message that'll put them in their body, should they click it.
- Getting an antagonist role will play a specific sound to the player. Some antags have their own sound, while others use a generic baddie sound. ERT, Loyalists, Traders, and Renegades use the generic goodie sound.
- A sound is now played to a ghost that is being offered a ghost role (e.g. talking sword, gravekeeper drone). Each ghost role can have their own sound, if desired.
- Getting a law change as a silicon mob will play a specific sound and display the added/removed law in the chat.


TheFurryFeline:
There's some times where someone getting defibbed shows to someone who's not even involved and never been spawned in in the first place for the round.

Neerti:
Not sure how that might happen since it does a mind object equality comparison, unless downstream servers do weird funky stuff to mind objects.

Is this good to go?

Atermonera:
Ater was murdered by very large PRs

PR #6851, opened 2020-03-15
BUREAUCRACY TIME

listerla:
Replaces the bulk of bureaucracy.dmi sprites with ones from https://github.com/discordia-space/CEV-Eris and https://github.com/Aurorastation/Aurora.3/

Some folders taken from https://github.com/VOREStation/VOREStation/pull/6625

Stamp sounds taken from https://github.com/Citadel-Station-13/Citadel-Station-13-RP/pull/470

Stamp sprites taken from https://github.com/Citadel-Station-13/Citadel-Station-13-RP/pull/462

If there are any complaints, I'm willing to remove or shuffle things around.


Cerebulon:
I already added stamp sounds, they just haven't been updated to the server yet: https://github.com/PolarisSS13/Polaris/pull/6804
You are on that version of the code though, but there should be some code in there which just uses a plain "play stamp.ogg" instead of your probably better 3-sound selection which you'll probably want to replace. Also the sounds should probably go with that one in the sound/bureaucracy folder in that case.
Also I believe some of these sprites were turned down previously? I think most should be fine but there was a disagreement about the filing cabinets specifically since for the mostpart we don't use the "head-on" perspective of some other servers.

listerla:
I'll remove the stamp sounds, thank you! My bad.

I was told that if we were gonna switch perspectives, I oughta do it DMI-by-DMI rather than all-at-once so if thats the case that seems like a catch-22.

Cerebulon:
In that case it should be fine. If there's a great riot I guess they can be subbed in, but there probably won't be.

Atermonera:
Feedback from the staff meeting, besides the paper plane issue which you elaborated on, the paperbin sprite doesn't seem to be much of an improvement to sprite quality, and per the perspective, loses the depth of the old sprite. Not interested in that sprite at present.

listerla:
@Atermonera Bins: Restored

listerla:
why has it decided to edit the .dme

PR #6852, opened 2020-03-16
Ports chemistry sprites from CEV-Eris

listerla:
Source: https://github.com/discordia-space/CEV-Eris


listerla:
ok i fucked this up within seconds

listerla:
I want to leave a note for why anyone asks what went so wrong that this simple change took half a dozen commits: I fucked up setting up DM on Linux and so half the info I need is invisible. I'll be working on fixing that before making my next PR.

Wickedtemp:
Can we get screenshots of the sprites?

listerla:
@Wickedtemp https://github.com/PolarisSS13/Polaris/pull/6852/checks?check_run_id=509900916

listerla:
@Heroman3003 I hate to ping, but could I ask which ones you don't like specifically? I'm willing to cut out or replace anything.

Heroman3003:
Honestly, I think all of them are worse. Beakers would be fine if not for darker color scheme that would clash with colors of reagents, but they are only one that is even 'okay', IMO.

Arokha:
I should mention that Bay has an all-4-direction version of the soda and booze dispensers in their chemical.dmi.

listerla:
I'll lift those when I have the chance, thanks.

SpadesNeil:
I like all of them and don't think you should change them for the complaints of a single individual.

Schnayy:
Sprites are nice but I cannot honestly understand why everyone wants to change to this perspective.

Arokha:
> Sprites are nice but I cannot honestly understand why everyone wants to change to this perspective.

I'm personally neither here nor there about the perspective but a lot of servers have gone with that perspective, which means newer sprites will be in that perspective. And a lot of the sprites are ooooOOOooollllld so replacing them will be nice.

Heroman3003:
On a more objective note, cutting off the bottom part of chem dispenser would be nice. Not because perspective or anything, but because both current and new sprites fail at representing the fact that its mechanically not a 'big' object, and is put on top of tables, can be walked over, etc.

SpadesNeil:
> On a more objective note, cutting off the bottom part of chem dispenser would be nice. Not because perspective or anything, but because both current and new sprites fail at representing the fact that its mechanically not a 'big' object, and is put on top of tables, can be walked over, etc.

This is a good point.

Leshana:
The new beakers are cool!   I think I prefer the old sprites for the chemistry machines tho.   If the new sprites were multi-directional it would be well worth it, (wouldn't it be nice to be able to have a chem dispenser facing WEST!?)  but the new ones make them all look like Apple IIe

Arokha:
Bay also has some sprites for mixer and dispenser that might be good but dunno if those are preferred. I think they're directional? Iono.

listerla:
Those ones are a large departure, so I'd avoided them because I figured they'd be extra-rejected.

Atermonera:
@Wickedtemp I'm late to the party here, but we have an icon-diff bot that's _supposed_ to run a check on all PRs that will render visually all of the .dmi changes made in that pr. Except sometimes it doesn't show up on some PRs and we have absolutely no idea what's causing it or how to fix it. You (and anyone else) can access it by clicking the `details` button on the check, and it'll give spoilers for each separate .dmi, showing both old and new sprites. It uses a white background so it's not the best for bright sprites, but in theory (haven't tried this yet myself) thanks to the now ubiquitous `inspect element` thing in any web browser worth its salt, that shouldn't be too hard to fix locally on a case-by-case basis.
Visual guide for accessing icon diff

Atermonera:
The sprites are nice, and while I personally hate the new (for other folks) wall perspective with a passion, stuff that sits on a table can have much more variable perspective, because the table itself (which is notably top-down) provides some perspective as well.

I am unsure how well the beakers will convey how full they are (since they get a new perspective, and it's suddenly weird to think about), and even recoloring the background of the white sprite (which I just tested and it does work), would it be possible for you to provide a screenshot of a beaker that actually has something in it?

Also, those side- and back-sprites are kinda vague about what the object actually is (which is a concern I share with schnayy about the new perspective), so I should note that there's a non-0 chance we'll look into a system of ignoring them, if not outright remove them later to force the use of the more informative front-sprite.

listerla:
img

PR #6853, opened 2020-03-16
Makes holsters and webbing get along.

Neerti:
Webbing and Holsters can now be worn at the same time. Requested at the staff meeting.
Also makes it so if you're only wearing one accessory on a piece of clothing, using the 'remove accessory' verb will immediately remove the only accessory instead of prompting you to choose an accessory.


Novacat:
At this point it might be better to refactor the whole accessory system (inc slots) to Baystation's, since a lot of the reason for the current system was to allow Holsters and Webbing to be exclusive from each other, and Bay moved away from bitmasks since they added a ton more accessory slots.

PR #6855, opened 2020-03-16
POI wall nonsense

Schnayy:
Fixes (???) the weird issue with the ice maze POI walls being incorrectly named and having no borders. The walls are now slightly darker than the cave tiles. Here is a picture.

image

I'm not a very good coder so this needs review. Once I get the okay, I'll be adding a new wall for POIs with it. Don't want to waste my time adding on to this if I did it really badly, and I also want to do a bit more testing.

Here is the new wall, picture is tiled 3x3
image

MAY, MAYBE, POSSIBLY fixes the weird bug with the cave borders. Not sure, don't know exactly how to replicate the bug so I can't test.


Novacat:
Does this also fix the impassable walls having the same name/desc as steel walls?

Schnayy:
Yes.

Schnayy:
Added the mossy walls. Figured out how to make the icons update to have borders when a neighboring drillable wall is drilled. Spawning the ice cave POI has the borders correctly placed.

PR #6857, opened 2020-03-16
Port computer and 'machine' frame icons from Bay

Arokha:
Kinda. I modified them. The console frames are fully directional so you can tell wtf way you're building them, and the machine frames, I added a little chevron to the front that points in their direction so you can tell.

Depends on https://github.com/PolarisSS13/Polaris/pull/6859 to look right

image


Cerebulon:
For computer frames we'd probably want the actual computer sprites in at the same time, otherwise they're probably going to look suddenly different when they're fully built.

Arokha:
Oh, do you not use those sprites? I thought you used the current Bay-originally-Eris style ones?

listerla:
Right now we use the old ass ones, aye.

Arokha:
Ok well gimme a minute then.

Arokha:
Ok, https://github.com/PolarisSS13/Polaris/pull/6859

Atermonera:
staff input mostly as a corollary to 6859, since this is generally dependent upon that to make sense.

PR #6859, opened 2020-03-17
Port Bay computer icons

Arokha:
They're better. They have directions which I'm pretty sure just means they're better. Also 3:4 view.

Originally from CEV-Eris


Schnayy:
Pretty sprites, though I wouldn't say directionals make something better when it's something you probably always want the screen facing forward for. Would still want them facing south personally.

Atermonera:
The sprites are definitely nice, but I personally feel like they're too... monolithic, kinda? Like, they feel like the sort of block thing that you'd see in a shuttle or ship, where they're built out of the wall. Which is also a little jarring when the side sprite is clearly displaced from what in most cases would be a wall directly behind it, so it's just this giant slab of metal sitting freely.

Marked for staff input to get an official consensus from staff, I'll likely be plugging this in the discord to gather player feedback as well.

Cerebulon:
I like that about them, it's more sci-fi-y than just weirdly wide computers.

Arokha:
They're a lot more versatile than the 'existing' consoles, given the directions, which I think is a crucial feature.

zekesulastin:
I just like that they all have different keyboards/interfaces, going all in on that particular part of the sci-fi computer aesthetic :)

Woodratt:
Bit blocky, but I like them

Arokha:
Sooooooooo

Cerebulon:
While I think most people are for it, this might end up getting delayed until the staff meeting NEXT sunday just by nature of bad timing. Hard to say.

Arokha:
Fix'd.

PR #6865, opened 2020-03-17
Tweak Fire, Flamethrowers, and Water.

### LINKED ISSUE #6864: Gas thrower issues
Gasthrower fires 3 projectiles, that each split into 7.
Total of 21 projectiles dealing 2 burn on impact, 3 agony and 20 burn over time.

On base components this weapon can fire 4 times.

Flames shouldn't do hallos, because otherwise this is an instastun.
Fire extinguishers don't put it out, nor can you resist.
Flamethrower should be a huge item, not a bulky item so you can't fit 3 of them in a backpack.

To be able to stop-drop-roll, or be fire extinguisher'd, not to be insta stunned.

Not to have a dude walk up next to you and pull an entire flamethrower, complete with gas tank outa his bag.

Mechoid:
Fixes #6864 
Water now properly extinguishes again (especially regarding fire modifiers), and won't duplicate itself. Flamethrowers tweaked again to be properly sane, but dangerous. Again.

Fixes:
- Extinguishers work on normal fire -and- modifier fire.
- Gasthrower properly uses stack based fire modifiers which extinguish over time faster when you have no fire_stacks.
- Resisting will also attempt to extinguish modifier fire.

Tweaks:
- Water now calls water_act on touch_mob, rather than directly touching fire stacks. Related to fix above.
- Gasthrower has slowdown of 1 when held due to being a volatile weapon, and does not do agony.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_resist.dm
+++ b/code/modules/mob/living/carbon/human/human_resist.dm
@@ -1,6 +1,6 @@
 /mob/living/carbon/human/process_resist()
 	//drop && roll
-	if(on_fire && !buckled)
+	if((on_fire || has_modifier_of_type(/datum/modifier/fire)) && !buckled)
 		adjust_fire_stacks(-1.2)
 		Weaken(3)
 		spin(32,2)
@@ -9,7 +9,7 @@
 			"<span class='notice'>You stop, drop, and roll!</span>"
 			)
 		sleep(30)
-		if(fire_stacks <= 0)
+		if(fire_stacks <= 0 && !(has_modifier_of_type(/datum/modifier/fire)))
 			visible_message(
 				"<span class='danger'>[src] has successfully extinguished themselves!</span>",
 				"<span class='notice'>You extinguish yourself.</span>"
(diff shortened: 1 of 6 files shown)


Wickedtemp:
For another pr regarding water and mobs, but.. watersmoke slimekilling when

code/modules/mob/living/carbon/human/human_resist.dm
```diff
@@ -9,7 +9,7 @@
 			"<span class='notice'>You stop, drop, and roll!</span>"
 			)
 		sleep(30)
-		if(fire_stacks <= 0)
+		if(fire_stacks <= 0 && !(has_modifier_of_type(/datum/modifier/fire)))
 			visible_message(
 				"<span class='danger'>[src] has successfully extinguished themselves!</span>",
 				"<span class='notice'>You extinguish yourself.</span>"
```
Atermonera: Will this double up with any messages from the modifiers expiring?

Mechoid: Unfortunately, yes. Though I'm not certain how to resolve that.

Atermonera: You could add a boolean variable at the top of the if that remembers if they had any fire modifiers, and if they did, skip this message? i.e.,
> if(conditions)
> > var/had_mod = has_modifier_of_type(/datum/modifier/fire)
> > [...]
> > if(fire_stacks <= 0 && !had_mod && !(has_modifier of type())

PR #6867, opened 2020-03-17
Teshari sprite fixes

Greenjoe12345:
fixes teshari torso sprites, hand sprites and arm sprites
seems that they were never designed for teshari to lose limbs?


Greenjoe12345:
temp closing this, think it'll be implemented alongside teshari prosthetics

PR #6870, opened 2020-03-18
Makes AI give up bashing if nothing to bash

Arokha:
When trying to move somewhere, the AI may just run into a wall and stare at it forever, trying to bash things. That's not very useful, so, after 5 attempts to bash with no results, we'll just give up.

### DIFF
--- a/code/modules/ai/ai_holder_combat.dm
+++ b/code/modules/ai/ai_holder_combat.dm
@@ -70,21 +70,22 @@
 /datum/ai_holder/proc/melee_attack(atom/A)
 	pre_melee_attack(A)
 	. = holder.IAttack(A)
-	if(.)
+	world.log << "melee_attack: [.]"
+	if(. == ATTACK_SUCCESSFUL)
 		post_melee_attack(A)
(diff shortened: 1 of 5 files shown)


code/modules/ai/ai_holder_combat.dm
```diff
@@ -70,21 +70,22 @@
 /datum/ai_holder/proc/melee_attack(atom/A)
 	pre_melee_attack(A)
 	. = holder.IAttack(A)
-	if(.)
```
Neerti: Not sure if you meant to leave this here, but that could spam the log a lot.

code/modules/ai/ai_holder_combat.dm
```diff
@@ -70,21 +70,22 @@
...
+	world.log << "melee_attack: [.]"
```
Neerti: Not sure if you meant to leave this here, but that could spam the log a lot.

Arokha: Whoops, removed

PR #6871, opened 2020-03-18
Remove stray 'return'

Arokha:
Looks like a stray return snuck in here. No reason to stop there.
~~Also removes the bizarre HTML-encoded ASCII comas.~~


Arokha:
That is a byond bug.

http://www.byond.com/forum/post/2553300

Arokha:
I'll just remove the coma-removing part for now.

PR #6874, opened 2020-03-19
Fixes Manifest Bug

Novacat:
Cargo jobs no longer show under civilian. Civilian jobs returned to their proper place.


Neerti:
Copypasta caught me again :cry:

PR #6875, opened 2020-03-20
[Oversight] Yay, another typo fixy

TheFurryFeline:
Changelog Notes: 

- Fixes typo with loading items into service borg's tray module.

- Fixes typo with attack_generic visible_message for when mobs attack doors. "smashes into the the Cargo Office!" -> "smashes into the Cargo Office!"

- Also fixing attacktext tenses and spelling issues for simple_mobs

### DIFF
--- a/code/modules/mob/living/simple_mob/subtypes/animal/sif/duck.dm
+++ b/code/modules/mob/living/simple_mob/subtypes/animal/sif/duck.dm
@@ -36,7 +36,7 @@
 	melee_damage_upper = 10
 	base_attack_cooldown = 1 SECOND
 	attack_edge = 1		// Razor-edged wings, and 'claws' made for digging through ice.
-	attacktext = list("nipped", "bitten", "cut", "clawed")
+	attacktext = list("nipped", "bit", "cut", "clawed")
 
 	say_list_type = /datum/say_list/duck
 	ai_holder_type = /datum/ai_holder/simple_mob/retaliate/cooperative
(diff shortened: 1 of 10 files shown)


code/modules/mob/living/simple_mob/subtypes/animal/sif/duck.dm
```diff
@@ -36,7 +36,7 @@
 	melee_damage_upper = 10
 	base_attack_cooldown = 1 SECOND
 	attack_edge = 1		// Razor-edged wings, and 'claws' made for digging through ice.
-	attacktext = list("nipped", "bit", "cut", "clawed")
+	attacktext = list("nipped", "bitten", "cut", "clawed")
```
Atermonera: attacktext gets fed into attack_generic() via a pick, and the following is a sample implementation
 > /obj/structure/barricade/attack_generic(var/mob/user, var/damage, var/attack_verb)
 > >visible_message("<span class='danger'>[user] [attack_verb] the [src]!</span>")

correcting to bitten will result in (approximately) `"crystal-feather duck bitten [thing]"`

TheFurryFeline: So this file is the only thing I should change or are there any others? Sorry for delay to response to this review. Bashing my head against another minor issue regarding bike kickstands message output.

Atermonera: All instances of "bit" that you changed to "bitten" will look fine when the mobs are smashing a person, but will look wrong when smashing objects, because attack_generic() uses different phrasing for each.
https://github.com/PolarisSS13/Polaris/blob/c8b32418ac3af408aa588066aa21440feaf752c0/code/modules/mob/living/carbon/human/human_attackhand.dm#L320 is where it lives for humans, simple_mobs and borgs don't appear to have a uniform attack_generic() proc

TheFurryFeline: Alright, then. Reverted that portion of changes.

PR #6876, opened 2020-03-20
Adds PDA-lawyer-old to pda_holo.dmi

Novacat:


Neerti:
The icon-diff bot fell over but I found a way to make it run again and now it's happy.

Atermonera:
Pray tell, what fell magicks did you invoke?

PR #6880, opened 2020-03-21
Teshari prosthetics

Greenjoe12345:
Adds two sets of Teshari prosthetics, unbranded, and Cenilimi Cybernetics

unbranded is unlocked via research, while cenilimi is ordered on the cargo console
a 'bald' hairstyle is added for FBP teshari to use
a sprite fix for teshari torsos and far hand are also included, to avoid having holes in teshari that have prosthetic limbs
some minor fixes for the teshari nurse cap and eyepatch are also included


Schnayy:
Preferably put the description in the description, not just the commit. Easier to read.

code/modules/organs/robolimbs.dm
```diff
@@ -98,6 +98,14 @@ var/const/standard_monitor_styles = "blank=ipc_blank;\
 	icon = 'icons/mob/human_races/cyberlimbs/unbranded/unbranded_unathi.dmi'
 	unavailable_to_build = 1
 
+/datum/robolimb/unbranded_teshari
+	company = "Unbranded - Teshari"
+	species_cannot_use = list(SPECIES_UNATHI, SPECIES_PROMETHEAN, SPECIES_DIONA, SPECIES_HUMAN, SPECIES_VOX, SPECIES_HUMAN_VATBORN, SPECIES_TAJ, SPECIES_SKRELL, SPECIES_ZADDAT)
+	suggested_species = SPECIES_TESHARI
+	desc = "A simple robotic limb with a small, raptor-like design. Seems rather stiff."
+	icon = 'icons/mob/human_races/cyberlimbs/unbranded/unbranded_teshari.dmi'
+	unavailable_to_build = 0
```
Schnayy: The other species unbranded are set to 1, is this intentional?

Greenjoe12345: it is, that's a fix to it not showing in the prosthetic fab even after using the disk

Atermonera: I believe the other two are set to be unavailable to be built because they're identified in the base unbranded limb's datum as alternatives, so the code will nominally use that instead? Would have to ask mechoid for specifics though

PR #6882, opened 2020-03-21
Machinery: Always use update_use_power()

Leshana:
### Always call update_use_power() instead of directly setting the use_power var.
This function has existed but lots of code doesn't use it.  Using it is step one in the project to improve machinery power processing efficiency.
### Define constants for use_power
We've been using bare numbers for long enough here.  Lets define constants and use them.
Note: I had some debate over naming the constants.  I settled for a prefix of the variable name itself followed by the names given for the three states in the original source comments.
```c
#define USE_POWER_OFF    0	// No continuous power use
#define USE_POWER_IDLE   1	// Machine is using power at its idle power level
#define USE_POWER_ACTIVE 2	// Machine is using power at its active power level
```


TheFurryFeline:
Does this make the body scanner console also require power? There's a thing far downstream where you're able to view medical status even if the area's power is offline.

Leshana:
> Does this make the body scanner console also require power? There's a thing far downstream where you're able to view medical status even if the area's power is offline.

This should change absolutely nothing about gameplay.   50% was just replacing numbers with constants, and the other 50% was replacing  `use_power = new_value` with  a call to the `update_use_power()` proc, which (for now) simply does `use_power = new_value`

The purpose of the change is to that (in upcoming projects) we can have code react to use_power being changed.  The proc in fact already existed, its just that most existing code didn't use it.

PR #6886, opened 2020-03-22
Inactivity subsystem kicks ghosts and admins again

Atermonera:


Neerti:
I think it'd be better to restore the old behavior and have it kick everyone who's afk, since this change makes it go from `kick everyone who isn't staff or a ghost` to `kick everyone who isn't staff`, when I feel it should be `kick everyone`.

Arokha:
Would be appreciated if you could just put the check into a proc on that subsystem so it could be easily overridden by downstreams. /proc/is_kickworthy or something probably with a better name.

code/controllers/subsystems/inactivity.dm
```diff
@@ -34,6 +33,9 @@ SUBSYSTEM_DEF(inactivity)
 					if(job)
 						information = " while [job]."
 
+				else if(isobserver(C.mob))
+					information = "while a ghost."
```
Neerti: Might need a space before the `while`, to be consistent with the other messages.

PR #6898, opened 2020-03-24
Prometheans can be slipped again

### LINKED ISSUE #6708: Prometheans cannot be disarmed due to no-slip tag
Prometheans cannot be shoved over, because shove checks for the if the mob can slip, and does not take into account footwear.

A human to shove a prommie over.

#### What actually happened
Not that.

#### Steps to reproduce
Attempt to disarm a prommie.

#### Additional info:

Atermonera:
Fixes #6708 
tested

### DIFF
--- a/code/modules/mob/living/carbon/human/human_attackhand.dm
+++ b/code/modules/mob/living/carbon/human/human_attackhand.dm
@@ -283,7 +283,7 @@
 			// We ARE wearing shoes OR
 			// We as a species CAN be slipped when barefoot
 			// And also 1 in 4 because rngesus
-			if(shoes || !(species.flags & NO_SLIP) && randn <= 25)
+			if((shoes || !(species.flags & NO_SLIP)) && randn <= 25)
 				var/armor_check = run_armor_check(affecting, "melee")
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/human/human_attackhand.dm
```diff
@@ -280,7 +280,10 @@
 
 			var/randn = rand(1, 100)
 			last_push_time = world.time
-			if(!(species.flags & NO_SLIP) && randn <= 25)
+			// We ARE wearing shoes OR
+			// We as a species CAN be slipped when barefoot
+			// And also 1 in 4 because rngesus
+			if(shoes || !(species.flags & NO_SLIP) && randn <= 25)
```
Neerti: I might be reading this wrong, but it looks like this will make it so if you're wearing shoes, you'll get disarmed 100% of the time, and if you're not wearing shoes, it will be 25% of the time?

Atermonera: ... That might explain why I got the disarm on the first try in testing.

PR #6900, opened 2020-03-25
Converted old supply_controller to SSsupply

Leshana:
- Removed the old processController process and refactored it into a subsystem
- Updated references from supply-controller to SSsupply
- Changed from 1.5 points per 30 seconds to 1 point per 20 seconds.  Same rate of 3 per min, but avoids fractional points.
- Aside from that, should be invisible to players because no other behaviors will be changed.


code/__defines/subsystems.dm
```diff
@@ -75,6 +75,7 @@ var/global/list/runlevel_flags = list(RUNLEVEL_LOBBY, RUNLEVEL_SETUP, RUNLEVEL_G
 // Subsystem fire priority, from lowest to highest priority
 // If the subsystem isn't listed here it's either DEFAULT or PROCESS (if it's a processing subsystem child)
 #define FIRE_PRIORITY_SHUTTLES		5
+#define FIRE_PRIORITY_SUPPLY		5
```
TheFurryFeline: This intentional to be both 5?

code/game/machinery/status_display.dm
```diff
@@ -220,7 +220,7 @@
 	return "[add_zero(num2text((timeleft / 60) % 60),2)]:[add_zero(num2text(timeleft % 60), 2)]"
 
 /obj/machinery/status_display/proc/get_supply_shuttle_timer()
-	var/datum/shuttle/autodock/ferry/supply/shuttle = supply_controller.shuttle
+	var/datum/shuttle/autodock/ferry/supply/shuttle = SSsupply.shuttle
```
TheFurryFeline: Is this going to make the display work like it used to? As in show how long is left before supply shuttle gets to the station, as happened before the original refactor of the supply system into using NanoUI for its console.

Leshana: No, the only behaviors this PR will change is the frequency of supply points (Still 3/minute, but distributed as 1 every 20s instead of 1.5 every 30s)
The issue you are referring to is due to a bug in the supply computer.  I could fix that at the same time but wasn't sure if it would be confusing to include in this PR?

code/__defines/subsystems.dm
Leshana: There is no particular reason why they need to be different.  Many subsystems all share the default priority of 50.    
I set it as 5 because it is an *extremely* quick fire() proc, and it doesn't really matter if it gets delayed even several seconds since its SS_KEEP_TIMING.     I like multiples of 5 or 10 to be pretty, but it could be fine as 4.  Would that be less troubling?

code/game/machinery/status_display.dm
TheFurryFeline: Ah, alright. Thanks for your answer. :> And I'd think that'd have to go in a different PR, yeah.

code/__defines/subsystems.dm
TheFurryFeline: Got'cha. Bit unfamiliar with this type of code so whatever you think is fine.

code/game/machinery/status_display.dm
Atermonera: Can you make a bug report for that? Because just looking at the code below here I can't see a reason why it shouldn't work.

Leshana: @Atermonera Issue PolarisSS13/Polaris#6910 added.

PR #6902, opened 2020-03-25
Convert alarm_manager into SSalarm subsystem.

Leshana:
Ported from Baystation's SSalarm.
- Switch to using MC_TICK_CHECK and replace all references.
- Optimize stat panel display by caching the active alarms.
- No player-visible changes, behavior is the same as before.


code/controllers/subsystems/alarm.dm
```diff
@@ -0,0 +1,45 @@
+
+// We manually initialize the alarm handlers instead of looping over all existing types
+// to make it possible to write: camera_alarm.triggerAlarm() rather than SSalarm.managers[datum/alarm_handler/camera].triggerAlarm() or a variant thereof.
+/var/global/datum/alarm_handler/atmosphere/atmosphere_alarm	= new()
```
Atermonera: Should these not live on GLOB?

Leshana: They probably should.   I just didn't touch those lines at all, leaving them exactly as they were, so it wasn't changed.    
Annoyingly github diff doesn't detect that I simply moved code/controllers/Processes/alarm.dm to code/controllers/subsystems/alarm.dm and then edited; instead making it look like I added it.

PR #6903, opened 2020-03-25
Creates test branch for VChat

Atermonera:
Updated to master so we don't lose a few weeks' worth of PRs that confuses everyone when we actually test this.


Atermonera:
Wait why are there two icon diffs and map diffs

PR #6909, opened 2020-03-27
Fix stupid lung runtime

Arokha:
Occasionally runtimes while doing Travis because the human it's testing suffocating in space on gets deleted and the lungs lose their owner but still try to make them gasp for air.


TheFurryFeline:
Uhhhhhhhhhh https://github.com/PolarisSS13/Polaris/pull/6906

Arokha:
O

PR #6914, opened 2020-03-28
Fixes examine panel icons

Atermonera:
The deconstruction step icons are actually images, which vchat presently doesn't like, because images aren't icons and the internal representation of graphical objects is actually very complex and poorly documented on account of being, y'know, internal. But mostly those aren't going to the chat, so why in the heck do they need to be treated as such
tested

### DIFF
--- a/interface/skin.dmf
+++ b/interface/skin.dmf
@@ -1423,7 +1423,6 @@ window "infowindow"
 		is-default = true
 		saved-params = ""
 		highlight-color = #00aa00
-		allow-html = true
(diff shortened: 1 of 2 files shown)


interface/skin.dmf
```diff
@@ -1423,7 +1423,6 @@ window "infowindow"
 		is-default = true
 		saved-params = ""
 		highlight-color = #00aa00
-		allow-html = true
```
Neerti: Does this break the colored text for the different types of examines?

Leshana: Oddly enough, no.   They were colored before this was added a few days ago, and still work without it.

PR #6920, opened 2020-03-29
QoL: Small Medical Things

Neerti:
The event system project took up a lot of my time so here's a QoL PR to make up for it.
Changes:
- Chem Masters and Chem Dispensers now eject beakers into your hands instead of on top of the machine, if a free hand is available. (Nalarac)
- Cryoxadone beakers are now labeled as such. (Nalarac)
- Suit sensor computers and programs will no longer show that Thought is bad at naming PoIs like `rock3` and instead the computer will read `Unknown Area`, in the same way as GPS devices. The Z number has also been replaced with a textual indicator, again in the same way as GPSes.
- Sleeper consoles now face the sleepers, at last.

This PR would've had another thing but the code for that is expected to conflict with current PRs up quite a lot so I'm gonna put it into a separate PR so this can get in unhindered.

### DIFF
--- a/code/datums/repositories/crew.dm
+++ b/code/datums/repositories/crew.dm
@@ -48,10 +48,10 @@ var/global/datum/repository/crew/crew_repository = new()
 
 				if(C.sensor_mode >= SUIT_SENSOR_TRACKING)
 					var/area/A = get_area(H)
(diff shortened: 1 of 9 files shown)


TheFurryFeline:
"Smell Medical Things" 🤔

PR #6927, opened 2020-03-30
Adjust turret target finding process

Arokha:
Woah there turrets I'm just not ready for the kind of commitment that lets you call get_mob() 1.2 million times every 10 seconds.

### DIFF
--- a/code/game/machinery/portable_turret.dm
+++ b/code/game/machinery/portable_turret.dm
@@ -705,7 +705,7 @@
 	for(var/turf/T in oview(world.view, src))
 		seenturfs += T
 
-	for(var/mob in player_list)
+	for(var/mob in living_mob_list)
 		var/mob/M = mob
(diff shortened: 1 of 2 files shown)


code/game/machinery/portable_turret.dm
```diff
@@ -701,8 +701,21 @@
 	var/list/targets = list()			//list of primary targets
 	var/list/secondarytargets = list()	//targets that are least important
 
+	var/list/seenturfs = list()
+	for(var/turf/T in oview(world.view, src))
+		seenturfs += T
+
+	for(var/mob in player_list)
```
Atermonera, line 708: player_list precludes anti-carp turrets (or any other npc, as the case may be) unless the carp is being actively controlled by a client

Arokha: changed to living_mob_list

PR #6934, opened 2020-03-31
Adjust two things that loop forever

Arokha:
Map effect intervals and pinpointer

We have those things you wanted now. Also bumped check time to 5s because 3s seems a lot for the repeated player list iteration. Not horrible but still.

### DIFF
--- a/code/game/gamemodes/nuclear/pinpointer.dm
+++ b/code/game/gamemodes/nuclear/pinpointer.dm
@@ -213,7 +213,7 @@
 	if(bomb_set)	//If the bomb is set, lead to the shuttle
 		mode = 1	//Ensures worklocation() continues to work
 		playsound(loc, 'sound/machines/twobeep.ogg', 50, 1)	//Plays a beep
-		visible_message("Shuttle Locator active.")			//Lets the mob holding it know that the mode has changed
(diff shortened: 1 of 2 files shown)


code/game/gamemodes/nuclear/pinpointer.dm
```diff
@@ -150,9 +146,10 @@
 		if("Other Signature")
 			mode = 2
 			switch(alert("Search for item signature or DNA fragment?" , "Signature Mode Select" , "" , "Item" , "DNA"))
+
 				if("Item")
 					var/datum/objective/steal/itemlist
-					itemlist = itemlist // To supress a 'variable defined but not used' error.
+					itemlist = itemlist
 					var/targetitem = input("Select item to search for.", "Item Mode Select","") as null|anything in itemlist.possible_items
```
Atermonera, line 153: Is itemlist actually initialized to anything when it's used here?

Leshana: No.  `itemlist` is in fat a global static variable that needs no init.  The self-assignment is just as the comment mentioned: a workaround for a byond bug to suppress DreamMaker issuing a spurious warning.

PR #6935, opened 2020-03-31
Remove emergencyShuttle as a process controller

Leshana:
- The `emergencyShuttle` process controller (/datum/controller/process/emergencyShuttle) was just a wrapper around the true workhorse `emergency_shuttle` anyway.
- The `emergency_shuttle` (/datum/emergency_shuttle_controller) is the _actual_ controller and it only needs to process during the tiny fraction of the round when the shuttle is getting ready to launch.  Therefore we leave it non-processing initially, and just register/deregister with `SSprocessing` when it actually needs to.

Four down, two to go.


Leshana:
Rebased to fix merge conflicts.

PR #6937, opened 2020-03-31
Adds airlock setup helpers

Arokha:
Search for map_helper in DreamMaker and put those objects on your airlock machinery. No need to set ID or frequency on anything except the controller, which should at least have a unique `id_tag` set. The frequency everyone seems to have decided on is 1380 so maybe set that too. No need to edit anything on the doors/sensors/pumps.

image


Woodratt:
<3

code/game/machinery/embedded_controller/mapping_helpers.dm
```diff
@@ -0,0 +1,176 @@
...
+	var/my_controller_type = /obj/machinery/embedded_controller/radio/airlock
+	//The device we're setting up
+	var/my_device
+	var/my_device_type
+	//Most things have a radio tag of some sort that needs adjusting
+	var/tag_addon
+
+/obj/effect/map_helper/airlock/Initialize()
+	..()
+	my_controller = get_controller(get_area(src))
```
Atermonera, line 26: Can these be mapped in with particular values (As airlocks are mapped right now [A process of which I am generally unaware, so this may be nonsense]) such that if someone were to, for some reason, try and stuff five different airlocks right next to each other, so that the helpers in one aren't seeing the controller in the other?

Arokha: They find the nearest one a few lines below, so you can have several in the same area without worrying. Tether has like 8 in one area at the dock.

PR #6941, opened 2020-04-01
Fix some parse errors

spookydonut:
Someone reported issues with langserver pluging for VSC and the `;` on the defines was the cause.

I also fixed the other errors that showed up.


Shadow-Quill:
\o/

Atermonera:
inb4 it reveals an infinite while loop like my oopsie did

PR #6947, opened 2020-04-02
Fixes supply shuttle display timer

### LINKED ISSUE #6910: Supply displays don't show supply shuttle status.
Supply displays (/obj/machinery/status_display/supply_display) are supposed to show supply shuttle ETA/status etc.

When supply console is used to summon the supply shuttle, for nearby supply displays to show the ETA.

#### What actually happened
No change to displays.

#### Steps to reproduce
1) Use supply computer in cargo to summon the shuttle.

#### Additional info:
The cause of this issue is that the supply displays are not in `mode = STATUS_DISPLAY_CUSTOM` because the post_signal() in supply computer that sent it the signal to set the mode was removed:  https://github.com/PolarisSS13/Polaris/pull/5522/files#diff-b2eac717a5ce819af9d498cb84126b14

Atermonera:
Fixes #6910 
tested

### DIFF
--- a/code/modules/shuttles/shuttle_supply.dm
+++ b/code/modules/shuttles/shuttle_supply.dm
@@ -13,6 +13,13 @@
 	if(isnull(location))
 		return
 
+	// Set the supply shuttle displays to read out the ETA
+	var/datum/signal/S = new()
+	S.source = src
+	S.data = list("command" = "supply")
+	var/datum/radio_frequency/F = radio_controller.return_frequency(1435)
+	F.post_signal(src, S)
+
 	//it would be cool to play a sound here
 	moving_status = SHUTTLE_WARMUP
 	spawn(warmup_time*10)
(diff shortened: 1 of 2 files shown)


Leshana:
I believe these will still get reset back to the same display as status displays whenever someone updates status displays and will not switch back to shuttle since the shuttle computer is still not sending the update signal as described in the issue.

code/modules/shuttles/shuttle_supply.dm
```diff
@@ -77,5 +77,8 @@
 
 //returns the ETA in minutes
 /datum/shuttle/autodock/ferry/supply/proc/eta_minutes()
-	var/ticksleft = arrive_time - world.time
-	return round(ticksleft/600,1)
+	return round((arrive_time - world.time) / (1 MINUTE)) // Floor, so it's an actual timer
```
Leshana, line 80: Won't this make all the PDAs, computers, etc which call this proc report "0" minutes remaining when there are still 59 seconds until arrival?

Atermonera:
[Tested with radio magic](https://puu.sh/FsZeD/7473f2f44c.mp4)

PR #6950, opened 2020-04-02
Improves space edge-of-map behavior and add mapping tools

Arokha:
- Maps no longer move you when you're 7 away from the edge to avoid you seeing the edge. Instead, now the edge doesn't look like anything special, and you move when you touch it, because:
- Space turfs at the edge of maps stretch out their sprites into the void beyond the map borders to make it appear like space continues in that direction (only for the equivalent of 7 turfs worth of distance though).

For mappers:
- Fake stairs, to allow you to pretend to have multi-z but not, as they just teleport you between themselves. `/obj/structure/fake_stairs`. They behave like normal stairs and connect when you set `_stair_tag` to something identical on two stairs. Read the file for more info.
- You can map in `/turf/space/internal_edge` to 'force' space edges in the middle of your maps. This can allow you to divide up a single z-level for a 'fake multiz ship'. You can put the pieces very close together, even directly on the other side of a 2-thick internal_edge wall facing each other, and one side can't see the other, as it'll just look like space in that direction.


code/controllers/subsystems/skybox.dm
```diff
@@ -4,12 +4,13 @@ SUBSYSTEM_DEF(skybox)
 	name = "Space skybox"
 	init_order = INIT_ORDER_SKYBOX
 	flags = SS_NO_FIRE
-	var/list/skybox_cache = list()
+	var/static/list/skybox_cache = list()
```
Atermonera, line 7: Will this prevent it from being updated when events trigger skybox changes?

Neerti: `static` is an alias for `global` in BYOND iirc

PR #6970, opened 2020-04-06
Fix 2 bugs

Arokha:
- Unique per-zlevel skyboxes didn't work
- Infinite lagloop in map_effect/interval


Leshana:
This ports VOREStation/VOREStation#7159

PR #6957, opened 2020-04-03
Security Voidsuit Overhaul

KillianKirilenko:
Downstream ports ahoy! This PR includes updates to a subset of security voidsuits that were incomplete, but are no longer so. It includes some mechanical/stat changes that aim to clear up confusion/lack of clarity and make them worth using as alternatives to the standard sec voidsuit in certain situations.

Includes the following mechanical changes:
* Crowd Control voidsuits (dark grey-black with minor red highlights and yellow visors) now have the highest melee defense values, making them actually suited to crowd control, instead of carbon copies of regular suits. Their supply crates cost a tiny bit more now. They may also need a small reduction in bullet/laser defense values to offset their melee bonus.
* Sec EVA Riot voidsuits (what a mouthful) are now simply called Sec EVA voidsuits, and exchange most of their melee defense for improved bullet/laser/energy/bomb/radiation protection, under the logic that small, high-velocity bits of metal and sudden bursts of energy/radiation are bigger threats during EVA than some chud smacking you upside the head with a toolbox out of nowhere.
* Security (& emagged) Suit Cyclers can now cycle to the Sec EVA style, meaning they are no longer "best stats but humans only".

The Sec EVA set now includes full UI sprites (except for headlamp versions):
image

Sprites are a mix of my own work and some odds and ends grabbed from Paradise.

PR #6971, opened 2020-04-06
Suit Cyclers: Faster, Smarter, Better

KillianKirilenko:
A mostly-finished refactor of suit cycler code that vastly increases the ease of maintaining existing options and adding new ones by querying helmets/suits for their name/description/etc. values via use of the `initial()` proc, instead of requiring they all be manually set in the ssu.dm itself. No need to mirror/duplicate changes across multiple files!

~~Will interfere with #6957 as that adds a new entry to cyclers.~~
6957 cycler option added in this PR.

~~Missing some minor polish features (that, on further inspection, might be difficult to pull off) but otherwise fully functional.~~
`Initial()` can't pull the contents of lists, so there's nothing I can do for inhand icons without rebloating the proc or doing some silly workarounds. I figure unupdated inhands are a small price to pay for the whole system being easier to maintain though.

### DIFF
--- a/code/game/machinery/suit_storage_unit.dm
+++ b/code/game/machinery/suit_storage_unit.dm
@@ -586,7 +586,7 @@
 	var/electrified = 0
 
 	//Departments that the cycler can paint suits to look like.
-	var/list/departments = list("Engineering","Mining","Medical","Security","Atmos","HAZMAT","Construction","Biohazard","Emergency Medical Response","Crowd Control","Sec EVA")
+	var/list/departments = list("Engineering","Mining","Medical","Security","Atmos","HAZMAT","Construction","Biohazard","Emergency Medical Response","Crowd Control","Security EVA")
 	//Species that the suits can be configured to fit.
 	var/list/species = list(SPECIES_HUMAN,SPECIES_SKRELL,SPECIES_UNATHI,SPECIES_TAJ, SPECIES_TESHARI)
 
@@ -628,7 +628,7 @@
 	name = "Security suit cycler"
 	model_text = "Security"
 	req_access = list(access_security)
-	departments = list("Security","Crowd Control","Sec EVA")
+	departments = list("Security","Crowd Control","Security EVA")
 
 /obj/machinery/suit_cycler/medical
 	name = "Medical suit cycler"
@@ -766,7 +766,7 @@
 
 	//Clear the access reqs, disable the safeties, and open up all paintjobs.
 	to_chat(user, "<span class='danger'>You run the sequencer across the interface, corrupting the operating protocols.</span>")
-	departments = list("Engineering","Mining","Medical","Security","Atmos","HAZMAT","Construction","Biohazard","Crowd Control","Sec EVA","Emergency Medical Response","^%###^%$", "Charring")
+	departments = list("Engineering","Mining","Medical","Security","Atmos","HAZMAT","Construction","Biohazard","Crowd Control","Security EVA","Emergency Medical Response","^%###^%$", "Charring")
 	species = list(SPECIES_HUMAN,SPECIES_TAJ,SPECIES_SKRELL,SPECIES_UNATHI, SPECIES_TESHARI)
 
 	emagged = 1
@@ -980,8 +980,7 @@
 
 //There HAS to be a less bloated way to do this. TODO: some kind of table/icon name coding? ~Z
 //hold onto your hat, this sumbitch just got streamlined -KK
-/obj/machinery/suit_cycler/proc/apply_paintjob(var/obj/item/clothing/head/helmet/parent_helmet as obj,var/obj/item/clothing/suit/space/parent_suit as obj)
-	
+/obj/machinery/suit_cycler/proc/apply_paintjob(var/obj/item/clothing/head/helmet/parent_helmet as obj,var/obj/item/clothing/suit/space/parent_suit as obj)	
 	if(!target_species || !target_department)
 		return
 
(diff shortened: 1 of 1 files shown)


code/game/machinery/suit_storage_unit.dm
```diff
@@ -586,7 +586,7 @@
 	var/electrified = 0
 
 	//Departments that the cycler can paint suits to look like.
-	var/list/departments = list("Engineering","Mining","Medical","Security","Atmos","HAZMAT","Construction","Biohazard","Emergency Medical Response","Crowd Control")
+	var/list/departments = list("Engineering","Mining","Medical","Security","Atmos","HAZMAT","Construction","Biohazard","Emergency Medical Response","Crowd Control","Sec EVA")
```
Atermonera, line 589: "Sec EVA" should probably use a proper name, since any suits that can go in here are EVA suits, and "sec", while a useful abbreviation when trying to call out for help (esp when typing), is not the sort of thing you'd expect to appear on a menu selection. (I also personally object to "Atmos" for the same reason, but "Atmospherics" may well be long enough that it makes the UI clunkier)

Might it also be possible to generate this list at runtime, using initial() checks on typesof() // subtypesof() to fetch the names and whether they should be available? Ideas off the top of my head for limiting availability are either using a boolean flag, or, better yet, making station voidsuits subtypes of a base station voidsuit type, and using that as the progenitor for the list of types to pull from.

code/game/machinery/suit_storage_unit.dm
```diff
@@ -990,182 +991,69 @@
...
+			parent_suit = /obj/item/clothing/suit/space/void/pilot/alt
+	
+	//look at this! isn't it beautiful? -KK
+	if(helmet)
+		helmet.name = "refitted [initial(parent_helmet.name)]"
+		helmet.desc = "refitted [initial(parent_helmet.desc)]"
+		helmet.icon_state = initial(parent_helmet.icon_state)
+		helmet.item_state = initial(parent_helmet.item_state)
+		helmet.light_overlay = initial(parent_helmet.light_overlay)
+		//TODO: figure out how to grab the item state slots too, they're a bit harder
```
Atermonera, line 1052: Might be easiest to just rename all of those sprites to actually reflect the modern suit names (having not looked at them at all and thus being unaware of what's peculiar about them beyond having unusual names)

code/game/machinery/suit_storage_unit.dm
KillianKirilenko: Yeah, I just used Sec EVA at first because I wasn't sure if setting it to "Security EVA" would short out and return to "Security" settings. I need to get around to updating this part now that I've run some more tests.

I've considered something like runtime list generation, but honestly I'm not entirely sure it's worth the effort*. More importantly, it's a bit above my level of expertise; this whole thing is more or a less an amateur fiddling at things and being pleasantly surprised when an idea they had actually worked.

*Although one could reasonably argue that whoever made this thing in the first place clearly didn't put in much effort themselves if they didn't think to use initial()...

code/game/machinery/suit_storage_unit.dm
KillianKirilenko: Not sure what you're suggesting here exactly?

code/game/machinery/suit_storage_unit.dm
KillianKirilenko: Also, with a little hindsight: the main thing I wanted to avoid with the Security EVA suits is using edgy buzzwords like "Tactical" or "Elite".

That job is reserved for the ERT and Security attracts enough easily-inflated egos already. Problem is "Riot" is basically synonymous with "Crowd Control", and it doesn't make sense to call them "Surplus" (as the comments imply they are) when they're *better* than standard-issue suits (unless the intent is to mean 'military surplus', but in that case there are better terms I think).

code/game/machinery/suit_storage_unit.dm
Atermonera: In the .dmi files, renaming the hand sprites it's using to make finding them from the voidsuits more feasible/sensible.

KillianKirilenko: Oh! It's not that *finding* them is difficult, I know exactly where to look; it's that `Initial()` just can't copy from lists *period*. If you try, all you get is a blank entry (not even `null`) so the suits turn invisible whilst holding them.

I've implemented a workaround in the latest update that spawns a copy of the suit to be copied from, copies the inhands from that, and then immediately deletes it. It's a little wasteful, but it works well enough.

code/game/machinery/suit_storage_unit.dm
```diff
@@ -979,8 +979,8 @@
 	return
 
 //There HAS to be a less bloated way to do this. TODO: some kind of table/icon name coding? ~Z
-/obj/machinery/suit_cycler/proc/apply_paintjob()
-
+//hold onto your hat, this sumbitch just got streamlined -KK
+/obj/machinery/suit_cycler/proc/apply_paintjob(var/obj/item/clothing/head/helmet/parent_helmet as obj,var/obj/item/clothing/suit/space/parent_suit as obj)
```
Leshana, line 983: Question:  Looks like the two parameter values passed in ignored and overwritten in the switch statement below yes?

KillianKirilenko: Incorporated into the latest update with a minor adjustment.

KillianKirilenko:
Right. Barring any final requests for changes this one's done. I even snuck in a tweak so that it uses Atmospherics in the display rather than just Atmos.

PR #6977, opened 2020-04-06
Remove non-global references to SSgameMaster and SSevent_ticker

Leshana:
### Reasoning
- I'd say it is best practice to avoid holding onto references to the subsystem instances aside from the global variables
  - As an example: if Master/Failsafe decided to restart/recover a subsystem, it fixes the global variable, but anyone out there holding a reference to the old one in their own local variables would A) prevent it from GCing and B) reference the old one that isn't getting fired anymore!
- The reference to GM was entirely unused by any event or meta-event.  (events actually didn't even store the reference being passed to their New)  so safe to remove.
-  /datum/game_master would work fine using the global refs.  Personally I also was a bit confused by the variable `ticker` referring to SSgame_master not SSgame_ticker, so its safe to remove too.

### DIFF
--- a/code/controllers/subsystems/events2.dm
+++ b/code/controllers/subsystems/events2.dm
@@ -23,7 +23,7 @@ SUBSYSTEM_DEF(event_ticker)
 // Starts an event, independent of the GM system.
 // This means it will always run, and won't affect the GM system in any way, e.g. not putting the event off limits after one use.
 /datum/controller/subsystem/event_ticker/proc/start_event(event_type)
-	var/datum/event2/event/E = new event_type(src)
+	var/datum/event2/event/E = new event_type()
 	E.execute()
 	event_started(E)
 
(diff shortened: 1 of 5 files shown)


Atermonera:
Just to confirm, will this require us to switch off of the legacy event system and onto neerti's latest rewrite? current status

Neerti:
This PR shouldn't require that at all, since it just cleans up some code in regards to references always pointing to the correct instance of subsystem, in the sense of if the `SSGameMaster` system fell over for some reason and got reset by some process (I assume the MC), the events themselves won't hold a reference to the broken `SSGameMaster` and instead will point to the current `SSGameMaster`.

Leshana:
Indeed Neerti is correct, this requires nothing special.   This PR is just a slight cleanup of the new rewrite, in exactly the sense that Neerti mentioned.

PR #6980, opened 2020-04-07
Adds pref to toggle vchat without 60s wait for linux users

Atermonera:
Tested, reloading vchat is enough to affect the change in a single round, reconnecting with the preference disabled brings the wait for vchat to give up down from 60s for linux to users to ~5s. 

## DISCLAIMER
If you have configured vchat options off of defaults and play with this preference, any issues that arise will not be supported, because vchat clearly works for you and you should just accept the change. This change is primarily intended as a QoL for our linux playerbase, who do not have access to chat at all for the first 60s of a connection while vchat tries to give them resources.
As an example of what can happen, I disabled it with dark mode enabled and after reloading (And reconnecting!) I got this:
this

Also this absolutely will conflict with the other preference pr I made this morning, either of these can theoretically be merged first, but I highly suspect that people will much rather this be merged as soon as possible.


TheFurryFeline:
Does this also work for non-Linux users? I know some people may want to default back to the old chat logging ways seeing that there's no way to code that continuous logging that can't be added to VChat

Atermonera:
It was tested on windows, as I don't have immediate access to a linux machine. The preference system doesn't really care what you're running, since it's basically just a bunch of boolean toggles for various parts of code.

And you can just export the logs, and that'll grab the whole session.

PR #6981, opened 2020-04-07
Port followup misc fixes

Leshana:
After a post-overmap event sweep of necessary backported PRs that hadn't been merged to Polaris, these are the ones that I found that were needed. 

- Cause the maploader to log an error if dmm file contains an undefined type.
  - This will greatly ease debugging as otherwise the only error you might get is "Runtime in reader.dm,370: list index out of bounds", and even then only if the turf has no objects.
- Make blood etc decals vanish when overmap ships fly away.
- CHECK_TICK in xenoarch setup for the sake of largermaps.


Leshana:
Edit: Rebased onto alternative fix for carp migration, added in others while we are at it. This now ports fixes from : VOREStation/VOREStation#7056, VOREStation/VOREStation#7145, and VOREStation/VOREStation#7159

Leshana:
Edit: Rebased onto alternative fix for carp migration, added in others while we are at it. This now ports fixes from : VOREStation/VOREStation#7056 and VOREStation/VOREStation#7145

PR #6990, opened 2020-04-11
Refactor telecomms and consoles to operate on logical zlevels

Arokha:
General
--------------
using_map.get_map_levels refactored to return logically 'connected' zlevels for the purpose of communications.
**When using the overmap:**
- `get_map_levels` returns zlevels based on the provided z-level's overmap sector, or multiz if you are not in an overmap sector
- `get_map_levels` with `long_range = TRUE` returns any zlevel in any sector on your current overmap turf (so ships in orbit too, etc)
- `get_map_levels` with `om_range` set to anything greater than -1 will use that range on the overmap and return any zlevels in that range

**When not using the overmap:**
- `get_map_levels` returns the station zlevels if the provided zlevel is in `station_levels`, otherwise it will return your multiz levels
- `get_map_levels` returns the contact levels if `long_range = TRUE` and the z is in `contact_levels`
- `om_range` is ignored when not using the overmap

Telecomms
--------------
- Broadcasters and Receivers by default have an om_range of 0 (so everything on their overmap turf) or without the overmap, everything in contact_levels on your map. They can have their power boosted (exponentially increasing power draw) to boost their range to a max of a 9x9 grid on the overmap centered on them.
- Relays have an om_range of -1 (so everything in their overmap sector) is relayed. Like now, they have infinite 'link' range to the tcomms devices they're multitooled to.
- All-In-One machines are actually All-In-One machines not a weird machine for antags. The /antag subtype is for antags and has the current functionality in it. You can plop an AIO machine on any zlevel and it will operate as a full telecomms network (assuming it's programmed to listen to any frequencies) for everything in om_range 0 (so everything on the overmap turf, good for ships in orbit).
- Bluespace radio is a backpack that can be linked directly to receivers and transmitters and can communicate over any distance. It's a port from vorestation. Basically it's a one-person connection to a specific telecomms network from anywhere. You can link it to an AIO machine as well. I didn't code in a means to link them in-round because eh. You can make subtypes that have autolinking capability though so they just link themselves when they spawn or are mapped in.

Consoles
--------------
- Consoles and modular computer software operates on the same basis as the above, on 'logically connected z-levels'. They all have the om_range -1 (and not `long_range` flag) with the exception of suit sensors and cameras, because I like the idea of Aliens-style things going on with monitoring sensors and helmet cameras from orbit.
- Cameras can be made to be always visible with the aptly named `always_visible` var. The 'entertainment' preset has this flag, so it is visible from every console that has the Entertainment network. You can set this on other things if you want, like helmet cams or whatever, I'ono!
- Atmos alerts, power alerts, RCON, etc all operate on the above, so there's no need to try and hide APC alarms and other things on POIs unless you'd rather explorers not be able to bring a power monitor console WITH them and see them.

Announcements
--------------
Announcements can have a zlevel specified for some of them, and they will operate on the above basis (with `long_range = TRUE`). Your announcement system is not fully radio, a lot of the things are just to_world messages, and I didn't modify those. If you want to copypaste our announcements.dm you're welcome to, and convert them all to colored radio messages or whatever but I didn't do that for you because I don't know what you want.


code/modules/alarm/atmosphere_alarm.dm
```diff
@@ -1,16 +1,22 @@
...
 		if(A.max_severity() > 1)
 			major_alarms.Add(A)
 	return major_alarms
 
-/datum/alarm_handler/atmosphere/minor_alarms()
+/datum/alarm_handler/atmosphere/minor_alarms(var/z)
 	var/list/minor_alarms = new()
+	var/list/map_levels = using_map.get_map_levels(z)
 	for(var/datum/alarm/A in visible_alarms())
+		if(z && (z && !(A.origin?.z in map_levels)))
```
Atermonera, line 18: `z && z && [...]`?

Arokha:
Merg'd.

Arokha:
Merged again

Atermonera:
Hey so remember how you gave use `announce_channel` in the last pr touching this stuff but didn't use it anywhere there so we made a PR to remove it because it was making compiler warnings for everyone?
it needs to go back in

PR #6991, opened 2020-04-11
Allow 'offmap spawn' jobs to be isolated from station employees

Arokha:
Offmap Spawn Jobs
-----------
This PR adds the ability to make a job `offmap_spawn = TRUE`, which isolates them from every station thing I could think of. They won't display on PDA manifests, won't have visible money accounts in the accounts DB console, won't have visible employee or other records, and it plays with the telecomms refactor PR I made to only push join announcements to the zlevel they joined on.

It also allows you to specify what channel join announcements should be done on, on the radio, so you can confine them to particular channels if you want. Common remains the default. It also confines the captain whistle to the zlevels they joined on.

Doing this because downstream we're adding other spaceships people can join on, and we don't want to let the entire world know when someone joined or cryo'd.


code/modules/mob/new_player/new_player.dm
```diff
@@ -355,8 +355,13 @@
 
 	//Find our spawning point.
 	var/list/join_props = job_master.LateSpawn(client, rank)
+	
+	if(!join_props)
+		return
+
 	var/turf/T = join_props["turf"]
 	var/join_message = join_props["msg"]
+	var/announce_channel = join_props["channel"] || "Common"
```
Atermonera, line 364: Is there something in here that's supposed to be using this? Because every time the code compiles it's throwing a warning that this variable isn't used.

Neerti: It's probably a thing used downstream that got left in here.
Might be a good idea to make Travis fail on warnings as well somehow.

PR #6992, opened 2020-04-12
Dual-wield Penalties and Urist Onmob Sprites

Novacat:
- Reenables Dual Wielding penalties on most heavy guns.
- Adds wielded sprites for all weapons with dual-wield penalties.
- Ports a ton of onmob gun sprites from Urist McStation
- Removes mention of requires_two_hands in code since that is not used anymore.


Cerebulon:
I'm still pretty heavily opposed to adding more real-world 20th/21st century guns because it doesn't really make a lick of sense from an IC perspective, but I guess this is just sprites and the sprites could be used as the basis for other stuff so... shrugs.

PR #6994, opened 2020-04-13
[PORT] Refactor the gameticker controller into SSticker

Leshana:
**Port of VOREStation/VOREStation#7150**

- Convert the gameticker controller into a subsystem: SSticker
  - Instead of independently spawn'd while/sleep loops, it stores its state and lets the MC fire it.
  - Convert relative path indentation procs to absolute path as per modern style standards.
  - Break apart the inner loops into separate procs so you can see what is actually going on.
  - Cleanup auto-voting so it actually obeys configurations better.
  - It now shows up in MC tab, along with stats about what phase it is in.
- Revamp Start Now admin verb to actually wait until init is done before starting.
- Lobby stat panel to show time remaining on lobby votes and if the server is done initializing.
- Remove the obsolete "processSchedler" controller that wrapped the gameticker.
Based on a blend of /tg and baystation's solutions.

lobby stat panel

#### Github PR Diff Issue
**NOTE**: Github does not show the diff very well by default!  It does not detect the rename of `code/game/gamemodes/gameticker.dm` -> `code/controllers/subsystems/ticker.dm` and thus shows them as deletion and addition of the entire files.  Only the startup/shutdown looping functionality was re-written, the support functions below are left alone, simply converted from relative pathing to absolute pathing.


code/controllers/subsystems/ticker.dm
```diff
@@ -0,0 +1,564 @@
...
+	var/end_game_state = END_GAME_NOT_OVER	// Track where we are ending game/round
+	var/restart_timeleft				// Time remaining until restart in desiseconds
+	var/last_restart_notify				// world.time of last restart warning.
+	var/delay_end = FALSE               // If set, the round will not restart on its own.
+
+	var/login_music						// music played in pregame lobby
+
+	var/list/datum/mind/minds = list()	// The people in the game. Used for objective tracking.
+
+	// TODO - I am sure there is a better place these can go.
```
Atermonera, line 35: GLOB? Or perhaps the job controller?

code/controllers/subsystems/ticker.dm
```diff
@@ -0,0 +1,564 @@
...
+		if(GAME_STATE_SETTING_UP)
+			setup_tick()
+		if(GAME_STATE_PLAYING)
+			playing_tick()
+		if(GAME_STATE_FINISHED)
+			post_game_tick()
+
+/datum/controller/subsystem/ticker/proc/pregame_welcome()
+	to_chat(world, "<B><FONT color='blue'>Welcome to the pregame lobby!</FONT></B>")
+	to_chat(world, "Please set up your character and select ready. The round will start in [pregame_timeleft] seconds.")
```
Atermonera, line 85: Should these just be `to_world()`?
Could they also use a `<span class='filter_system'>` (Or some other filter tag)?

code/controllers/subsystems/ticker.dm
```diff
@@ -0,0 +1,564 @@
...
+// Returns 0 if failed to pick a mode, otherwise 1
+/datum/controller/subsystem/ticker/proc/setup_choose_gamemode()
+	//Create and announce mode
+	if(master_mode == "secret")
+		src.hide_mode = TRUE
+
+	var/list/runnable_modes = config.get_runnable_modes()
+	if((master_mode == "random") || (master_mode == "secret"))
+		if(!runnable_modes.len)
+			to_chat(world, "<B>Unable to choose playable game mode.</B> Reverting to pregame lobby.")
```
Atermonera, line 132: `to_world()`? Filtering tags? Leaving this as the last comment for this because there's several other to_chat()s below here and leaving a review with 20 comments complaining about the same thing in the same file feels tedious and a little rude

code/modules/admin/admin.dm
```diff
@@ -868,21 +868,26 @@ proc/admin_notice(var/message, var/rights)
...
-	set desc="Start the round RIGHT NOW"
+	set desc="Start the round ASAP"
 	set name="Start Now"
-	if(!ticker)
-		alert("Unable to start the game as it is not set up.")
+
+	if(!check_rights(R_SERVER|R_EVENT))
+		return
+	if(SSticker.current_state > GAME_STATE_PREGAME)
+		to_chat(usr, "<font color='red'>Error: Start Now: Game has already started.</font>")
```
Atermonera, line 877: filter tags needed in here too

code/controllers/subsystems/ticker.dm
Leshana: These to_chats are just taken verbatim as they were when the code was moved around in this file.  But I don't mind changing them to to_world here, its good to do.

Leshana: I converted all the `to_chat(world, ...)` to   `to_world(...)` instead.   I also added wrapping <span> tags, except for the end of round declaration.  That one has particular formatting baked into it and probably needs a specialist's attention if you want to revamp it.   Its just end of round anyway.

code/controllers/subsystems/ticker.dm
Leshana: Either would be good.  In fact there are several vars throughout the system which ought to be re-oraganized onto GLOB or the job controller.  But I think that is best left to those PRs so this is focused on SSticker, leaving the vars that were on ticker before to be on ticker still.

code/controllers/subsystems/ticker.dm
Atermonera: The `filter_` tags should be working now, so just applying one of those will work. I added them specifically to be able to precisely control which filter catches what message without tampering with the message styling

PR #6999, opened 2020-04-15
Update Human HUD Generation to Fix Internals Graphical Bug

KasparoVy:
Fix an issue whereby when reconnecting to the server after disconnecting/losing connection your internals HUD element will always be off, even if you're actually still breathing from a tank.
Now it checks & shows the correct state so you don't panic after lagging out IN SPACE AAAAAHHHH


KasparoVy:
Man that PR number. I was SO CLOSE

PR #7005, opened 2020-04-15
Changes shuttle landing soon warning

Arokha:
Also here have some goonsprites


Atermonera:
What's the licensing on the sprites?

Atermonera:
Also the shuttles wall sprites may yet be viable but we are against the other wall/door sprites, full stop. This PR will not be  The other sprites will be reviewed, subject to license compatibility.

Atermonera:
license conversation

Atermonera:
After discussing this with the rest of the development team, we will not be merging this due to the following reasons, in no particular order:
- The wall sprites use a perspective that several members have a distinct and noted dislike for,
- The window sprites don't present any sort of graphical improvement over our current window sprites, and would look out of place when used with any of our existing sprites,
- The shuttle landing indicator is actually obfuscating some of the information that the current indicator presents (Time to landing, namely), and is less immediately visually apparent that you need to move,
- The proposed technomancer mark sprite wouldn't require such a high degree of work to create ourselves as to present a challenge in doing so, and Neerti is presently planning on making one.

Arokha:
I find it strange those are your points, when the first two of those are not used in anything at all, and have nothing to do with this PR. You've stated in the past "sure, unused sprites are good to have, just include them" so if this is a policy change from that then that's interesting to know.

I should mention the shuttle landing indicator currently does not have any special connection to the timing of shuttles landing. It simply plays the animation, which lasts 4.9 seconds (and is actually not the time it takes shuttles to land. It disappears before the shuttles land and there's usually about a 1 second gap where there's no warning and no shuttle.

If you change the time it takes shuttles to land, or the point at which you warn players that there's a shuttle landing there, it will not be accurate. I would think for that reason the warning being abstract from the landing time would be desirable unless someone wants to create an entirely programmatic way of creating the icon in-round for the warning.

I also spoke to Neerti about this and showed them the icon days ago when I created it, it's a bit unusual that didn't come up during our conversation.

PR #7006, opened 2020-04-15
Ports some power.dmi sprites from Bay

Arokha:
Perspective matchy, and also some anchored/unanchored sprites for TEG that have nice cold/hot overlays. Using tg's cablecoils instead because *sooOOOoomeee*one doesn't like them.


Cerebulon:
I'm absolutely baffled by the SMES sprite which is essentially a redraw of the existing sprite which already matched the "new" perspective... Except they made it worse in every way somehow.
And I guess the upper part of the new junction box/grid checker sprites is supposed to be like the 'top surface' but they just look like they have no perspective whatsoever and are flat wall-mounted panels or something.
TEG is very pretty though. And more "clickable" wires is definitely definitely good.

Arokha:
The SMES angle was very different from the others. Yes, it was a 3/4ths perspective sprite, but if placed next to any of the other 3/4ths perspective sprites, such as the breaker, or any of the other pre-existing ones, it becomes apparent that the perspective points are different and it's like you're looking at it from a much lower angle than everything else in the game.

Neerti:
I personally like all the new power sprites. It's too bad the new wire sprites got changed to TG's instead of Bay's but oh well, I suppose the invisible pixels will help with clicking regardless.

listerla:
I'll note I do prefer Bay's cables as well, if that could be used by any chance?

PR #7010, opened 2020-04-17
Ports over power-related enhancements and tweaks i coded for Bay a while back

atlantiscze:
- Please fasten your seatbelts, this will be a long one as it joins few older Bay PRs into one.

1. Ports https://github.com/Baystation12/Baystation12/pull/12626
- SMES units may now be damaged and destroyed. Charged SMES units fail quite violently. Damage can be fixed by welding tool.
- PSUs completely refactored, ghetto variant no longer exists.
- Cell rack PSUs now can be considered a hybrid between large battery charger and a SMES. They actually use the cells to store power (so you can hot-swap the cells to get more juice simply via the UI without deconstructing the whole machine), but in comparison to SMES have poor throughput and capacity in general (cells are simply too small). They are also somewhat limited in configuration options (lacks the precision electronics of a SMES). Better matter bin lets you put in more cells, up to 9.
- Cell rack PSU also has a mode that allows charge-balancing all inserted batteries (moves energy around so each battery has the same charge %)

2. Ports https://github.com/Baystation12/Baystation12/pull/11977
- SMES units now have full load balancing capability, getting rid of that annoying "One SMES charges at full, other SMES gets nothing" problem. If insufficient power is available on input, all inputting SMESes will now charge at same percentage. If more SMESes power a single output, they will all output equal percentage of their setting.
- This appears to have a pleasant side effect of fixing the issue where SMESes could starve APCs of energy. SMESes are ALWAYS last to input power on a powernet.
- This also appears to have fixed weird values displayed in SMES output/input load readings in the UI. By weird values i mostly mean inputs/outputs actually higher than the SMES is set to have.

3. Ports https://github.com/Baystation12/Baystation12/pull/18137
- SMES units (and subtypes, therefore effectively also PSUs from previous entry) can now have more than one terminal. This effectively allows a setup where two (or more) sources feed a single SMES, which then feeds the output. SMESes can not exceed input setting even with multiple terminals.
- Typical example of use in practice would be SMES that runs something important (for example an AI, telecomms, pick whatever you want). It could have one input from the power grid, and second input from a PACMAN generator set up nearby as a backup. Before the generator would have to be wired into main grid, therefore main grid would siphon off power from it. Now the generator can be separate and dedicated for whatever use you want.

4. Extra
- Adds cable coil recoloring using multitool (Bay went the path of having dedicated cable painter device but that really feels as too much of a hassle having to carry another device for this - lets just say the paint is electrically reactive in some way).
- Ports debug power items (replacement of fractal reactor and magical SMESes). This is adminspawn only. Contains generator that produces infinite power, dummy load that acts as a power sink, and another dummy load that drains energy through an APC. All of these also provide detailed info about attached powernet on examine.


atlantiscze:
Going to leave this hanging around for about a day before marking as ready for review, in case some feedback shows up for the feature part.

atlantiscze:
Allright. Did some final testing and fixing, everything seems to work as expected. Also included additional little feature that allows changing of cable coil colour.

code/modules/power/smes.dm
```diff
@@ -113,69 +116,87 @@
 /obj/machinery/power/smes/proc/chargedisplay()
 	return round(5.5*charge/(capacity ? capacity : 5e6))
 
+/obj/machinery/power/smes/proc/input_power(var/percentage)
```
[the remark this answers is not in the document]
Leshana: Consider an even simpler scenario actually, a single SMES with two terminals connected to the *same* powernet.  Even tho input_power() is called only once, it still draws from both terminals. 
I actually tested this, setting input limiter to 100, but it drew 200 per cycle anyway.   (You *might* say this is intended behavior when you double-terminal a single network, but the same thing will happen when you do it to two won't it?)

html/changelogs/atlantiscze-power.yml
```diff
@@ -0,0 +1,37 @@
...
+delete-after: True
+
+# Any changes you've made.  See valid prefix list above.
+# INDENT WITH TWO SPACES.  NOT TABS.  SPACES.
+# SCREW THIS UP AND IT WON'T WORK.
+# Also, all entries are changed into a single [] after a master changelog generation. Just remove the brackets when you add new entries.
+# Please surround your changes in  double quotes ("), as certain characters otherwise screws up compiling. The quotes will not show up in the changelog.
+changes: 
+  - rscadd: "Added a changelog editing system that should cause fewer conflicts and more accurate timestamps."
+  - rscdel: "Killed innocent kittens."
```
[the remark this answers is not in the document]
atlantiscze: Huh weird, it would seem that my changes to that file actually never saved. Will rectify.

code/modules/power/smes.dm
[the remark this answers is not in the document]
atlantiscze: input_power is called by the powernet, so percentage is as well determined by the powernet. However the fact that SMESes can exceed set input rating is a bug which i have fixed now.

Did a quite thorough test now with updated version (will commit in a moment), multiple inputs never exceeded SMES rating. Inputs are however not load balanced (so if first input covers all input requirement, other one won't provide any power) as that would require quite wide refactors.

code/modules/power/smes.dm
```diff
@@ -116,20 +118,18 @@
 /obj/machinery/power/smes/proc/chargedisplay()
 	return round(5.5*charge/(capacity ? capacity : 5e6))
 
-/obj/machinery/power/smes/proc/input_power(var/percentage)
+/obj/machinery/power/smes/proc/input_power(var/percentage, var/obj/machinery/power/terminal/term)
```
Leshana, line 121: That'll do it!  Only drawing from the specific terminal (combined with looping over terminals instead of SMES) will ensure the `percentage` parameter is applied only to the powernet that generated it.

code/modules/power/smes.dm
```diff
@@ -423,13 +446,27 @@
 	update_icon()
 	..()
 
-/obj/machinery/power/smes/magical
```
Leshana, line 426: Please don't remove `/obj/machinery/power/smes/magical`, they're used in several away mission maps.   Also would appreciate if  `outputting = 1` was set on the magical ones.

code/modules/power/smes_construction.dm
```diff
@@ -84,6 +84,9 @@
 /obj/machinery/power/smes/buildable/Destroy()
 	qdel(wires)
 	wires = null
+	for(var/obj/machinery/power/terminal/T in terminals)
+		T.master = null
+	terminals = null
```
Leshana, line 89: I believe this needs to move up to parent type (`/obj/machinery/power/smes`) otherwise batteryracks won't clear their terminal's `master` when destroyed.
image

code/modules/power/smes.dm
```diff
@@ -262,17 +302,25 @@
...
 		building_terminal = 1
-		var/turf/tempTDir = terminal.loc
+		var/obj/machinery/power/terminal/term
+		for(var/obj/machinery/power/terminal/T in get_turf(user))
+			if(T.master == src)
+				term = T
+				break
+		if(!term)
+			to_chat(user, "<span class='warning'>There is no terminal on this tile.</span>")
+			return 0
```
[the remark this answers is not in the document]
atlantiscze: For the second occurence, building_terminal = 0 is already set on line 327, therefore it is already set at line 329 and doesn't have to be set again. The first point is valid, though. Will fix.

code/modules/power/smes.dm
atlantiscze: As written on discord, magical SMESes felt kind of deprecated for quite some time (at least i tried to get rid of them back on Bay) as they are one giant hack. I included newest version of customisable power sources for debugging in this PR, which i believe would be a suitable replacement. See /obj/machinery/power/debug_items/infinite_generator - for actual map use that may be encountered by players show_extended_information should be set to FALSE.

Neerti:
It's not a super big deal but the test map will need to be updated after this PR, if it's not done in this PR, to have its fractal reactor path point to the new one and thus be able to be compiled.

atlantiscze:
Done

code/modules/power/batteryrack.dm
```diff
@@ -1,234 +1,313 @@
...
+
+	var/cellcount = 0
+	var/charge_level = between(0, round(Percentage() / 12), 7)
+
+
+	overlays += "charge[charge_level]"
+
+	for(var/obj/item/weapon/cell/C in internal_cells)
+		cellcount++
+		overlays += "cell[cellcount]"
```
Atermonera, line 76: Should this be checking against PSU_MAXCELLS?

code/modules/power/batteryrack.dm
```diff
@@ -1,234 +1,313 @@
...
+			return
+	// We're still here, so it means the least charged cell was full OR we don't care about equalising the charge. Give power to other cells instead.
+	for(var/obj/item/weapon/cell/C in internal_cells)
+		amount -= C.give(amount)
+		// No more power to input so return.
+		if(!amount)
+			return
+
+
+/obj/machinery/power/smes/batteryrack/remove_charge(var/amount)
```
Atermonera, line 128: Should this return some feedback on whether it was able to remove the desired amount? Or does it check that it has enough charge to remove before calling this?

code/modules/power/batteryrack.dm
```diff
@@ -1,234 +1,313 @@
...
+/obj/machinery/power/smes/batteryrack/attackby(var/obj/item/weapon/W as obj, var/mob/user as mob)
+	if(!..())
+		return 0
+	if(default_deconstruction_crowbar(user, W))
+		return
+	if(default_part_replacement(user, W))
+		return
+	if(istype(W, /obj/item/weapon/cell)) // ID Card, try to insert it.
+		if(insert_cell(W, user))
+			to_chat(user, "You insert \the [W] into \the [src].")
```
Atermonera, line 264: filter tags in here, please wrap the message in `<span class='filter_[category]'> [...]</span>`

code/modules/power/debug_items.dm
```diff
@@ -0,0 +1,65 @@
...
+	var/show_extended_information = 1	// Set to 0 to disable extra information on examining (for example, when used on admin events)
+
+/obj/machinery/power/debug_items/examine(mob/user)
+	. = ..()
+	if(show_extended_information)
+		show_info(user)
+
+/obj/machinery/power/debug_items/proc/show_info(var/mob/user)
+	if(!powernet)
+		to_chat(user, "This device is not connected to a powernet")
```
Atermonera, line 15: filter tags needed in this file

code/modules/power/power.dm
```diff
@@ -37,6 +35,8 @@
 /obj/machinery/power/proc/add_avail(var/amount)
 	if(powernet)
 		powernet.newavail += amount
+		return 1
+	return 0
```
Atermonera, line 39: These should probably be TRUE and FALSE?

code/modules/power/smes.dm
```diff
@@ -228,23 +267,26 @@
...
-	if (!open_hatch)
+	if (!panel_open)
 		to_chat(user, "<span class='warning'>You need to open access hatch on [src] first!</span>")
 		return 0
 
-	if(istype(W, /obj/item/stack/cable_coil) && !terminal && !building_terminal)
+	if(istype(W, /obj/item/weapon/weldingtool))
+		var/obj/item/weapon/weldingtool/WT = W
+		if(!WT.isOn())
+			to_chat(user, "Turn on \the [WT] first!")
```
Atermonera, line 280: filter tags

code/modules/power/smes.dm
```diff
@@ -342,10 +387,11 @@
 	if( href_list["cmode"] )
 		inputting(!input_attempt)
 		update_icon()
-
+		return 1
```
Atermonera, line 390: Should these be TRUE/FALSE?

code/modules/power/smes.dm
```diff
@@ -412,6 +423,21 @@
 	if(!output_attempt)
 		outputting = 0
 
+/obj/machinery/power/smes/take_damage(var/amount)
+	amount = max(0, round(amount))
+	damage += amount
+	if(damage > maxdamage)
+		visible_message("<span class='danger'>\The [src] explodes in large rain of sparks and smoke!</span>")
```
Atermonera, line 430: `in a large shower of`, I think. It's definitely missing an article, and although I haven't personally ever heard "rain of sparks" used over "shower" they're synonyms, so I don't think it particularly matters which one is used.

code/modules/power/smes.dm
```diff
@@ -423,13 +449,27 @@
...
+/obj/machinery/power/smes/ex_act(var/severity)
+	// Two strong explosions will destroy a SMES.
+	// Given the SMES creates another explosion on it's destruction it sounds fairly reasonable.
+	take_damage(250 / severity)
 
-/obj/machinery/power/smes/magical/process()
-	charge = 5000000
+/obj/machinery/power/smes/examine(var/mob/user)
 	..()
+	user << "The service hatch is [panel_open ? "open" : "closed"]."
```
Atermonera, line 463: `to_chat()`, filter tag

atlantiscze:
Sorry for the few days delay, had some personal matters that had priority. I'll get to implementing feedback today.

code/modules/power/batteryrack.dm
atlantiscze: Unnecessary, as PSU_MAXCELLS is also being checked when trying to insert a new cell. Therefore it is not possible to exceed 9 cells even with T4/T5 components.

code/modules/power/batteryrack.dm
atlantiscze: Double-checked this, it does check there is enough charge before this is called.

PR #7013, opened 2020-04-18
Caps burning human temperature

Novacat:
Basically makes it so that a burning human no longer infinitely increases in temperature.

This is also being combined with reductions to heat damage level downstream to make high temperatures less super lethal as they currently are on this codebase. Those changes can be ported as well if desired.

### DIFF
--- a/code/modules/mob/living/carbon/human/life.dm
+++ b/code/modules/mob/living/carbon/human/life.dm
@@ -22,6 +22,7 @@
 #define COLD_GAS_DAMAGE_LEVEL_3 3 //Amount of damage applied when the current breath's temperature passes the 120K point
 
 #define RADIATION_SPEED_COEFFICIENT 0.1
+#define HUMAN_COMBUSTION_TEMP 524 //524k is the sustained combustion temperature of human fat
 
 /mob/living/carbon/human
 	var/oxygen_alert = 0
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/life.dm
```diff
@@ -1802,7 +1803,15 @@
...
-		bodytemperature += (BODYTEMP_HEATING_MAX + (fire_stacks * 15)) * (1-thermal_protection)
+		var/fire_temp_add = (BODYTEMP_HEATING_MAX + (fire_stacks * 15)) * (1-thermal_protection)
+		//This is to prevent humans from heating up indefinitely. A human being on fire (fat burns at 250C) can't magically
+		//   increase your body temperature beyond 250C, but it's possible something else (atmos) has heated us up beyond it,
+		//   so don't worry about the firestacks at that point. Really, we should be cooling the room down, because it has
+		//   to expend energy to heat our body up! But let's not worry about that.
+		if((bodytemperature + fire_temp_add) > HUMAN_COMBUSTION_TEMP)
+			return
+
+		bodytemperature += fire_temp_add
```
Mechoid, line 1814: Should this not just set it to `HUMAN_COMBUSTION_TEMP` rather than prevent a temperature increase altogether?

Atermonera: i.e., `bodytemperature = min(bodytemperature + fire_temp_add, HUMAN_COMBUSTION_TEMP)`

Arokha: This suggestion doesn't make any sense. It means that if someone has been heated up by a room to 2000k and then catches on fire, their body temperature will lower to ~500k.

Arokha:
To clarify since there seems to be a misunderstanding, this is the proc that handles humans having fire_stacks. Their body temperature can be raised other ways, via the environment or other things, above the temperature of human combustion, which makes sense.

Capping 'the temperature at which human flesh burns at' is the point of the PR, because previously if someone had fire_stacks, they would continually rise in temperature, indefinitely, even if the room was T20C. They would eventually reach stellar temperatures just because their skin was on fire.

PR #7021, opened 2020-04-20
Fixes Large Escape Pod #2

### LINKED ISSUE #7019: Large Escape Pods Don't Escape
post shuttle departure
The escape pod to bring these weary survivors to safety
#### What actually happened
Trapped forever with the alien horrors
#### Steps to reproduce
Emergency shuttle, then watch which pods leave.
#### Additional info:

Atermonera:
Fixes #7019 
Tested, not sure why the medical shuttle didn't leave, may have been related to the fact that the doors were on space turfs. If it comes up again, I can take another stab at poking it.


Nalarac:
So does this not fix Large Escape Pod #1?

Atermonera:
I was unable to reproduce the issue with escape pod 1 when I retested it on the current server the following round. My testing procedure was

1. Debug-Controller -> LEGACY/emergency_shuttle_controller
2. Mark object
3. VV the target escape pod from the `escape_pods` list
4. VV the target pod's `arming_controller`
5. Proc-call on the arming controller `arm`, no arguments
6. Proc-call on the target pod `launch`, with the emergency shuttle controller as the first and only argument through the marked datum
7. Wait a few minutes for the shuttle engine warmup sound to play, then observe if the shuttle jumped when that finished or not.

Large escape pod 2 didn't move, on account of its landmark being mapped to the wrong side of the area (Or so I surmise, it was failing the check_collision() as determined by the addition of some print statements)
Further information on the issue that was noticed with pod 1 may be found in the debug log, because attempt_move() actually dumps most all of its failure modes, but I don't actually have access to those V.V

PR #7023, opened 2020-04-21
Roiling Mold Blob Fix/Tweak

Mechoid:
Roiling Mold blob projectiles fly in shorter arcs. Fix projectile targeting recursion. Does not entirely fix the blob's window-targeting, as that will require a PR of its own due to possibly larger tweaks to pixel projectiles being needed in order to fix it, and it is not an inherent issue of the blob.


Atermonera:
For those curious, our investigation into the arc projectile targeting through windows found that the trace projectiles were always returning null, regardless of if they hit something or not, and figuring out how to extract a useful value from process_hitscan in projectile code isn't particularly trivial and may have consequences for other systems, like turrets.

PR #7024, opened 2020-04-21
Converts Shutoff valve monitoring into a MC program

atlantiscze:
- Existing consoles remain for now, only refactored to use a shared nano_module. Included by default on engineering consoles, downloadable by anyone with basic engineering access.
- Tidies up the UI a bit, prevents manual toggling of valves when they are in Auto mode (as it would immediately toggle back)

### DIFF
--- a/code/game/machinery/computer/shutoff_monitor.dm
+++ b/code/game/machinery/computer/shutoff_monitor.dm
@@ -5,66 +5,27 @@
 	icon_screen = "power:0"
 	light_color = "#a97faa"
 	circuit = /obj/item/weapon/circuitboard/shutoff_monitor
+	var/datum/nano_module/shutoff_monitor/monitor
 
-/obj/machinery/computer/shutoff_monitor/attack_hand(var/mob/user)
+/obj/machinery/computer/shutoff_monitor/New()
+	..()
+	monitor = new(src)
+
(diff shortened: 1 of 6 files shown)


code/game/machinery/computer/shutoff_monitor.dm
```diff
@@ -5,66 +5,27 @@
 	icon_screen = "power:0"
 	light_color = "#a97faa"
 	circuit = /obj/item/weapon/circuitboard/shutoff_monitor
+	var/datum/nano_module/shutoff_monitor/monitor
 
-/obj/machinery/computer/shutoff_monitor/attack_hand(var/mob/user)
+/obj/machinery/computer/shutoff_monitor/New()
```
Atermonera, line 10: Does this need to go in New(), or could it go in initialize(), and would it runtime if you tried to use this before the nanoui controller has initialized?

atlantiscze: To be honest, i am not certain here. Usually when refactoring things into MC programs i outright remove the old computer type. I didn't want to go this path in this particular PR as it requires map changes.

Other computers with NanoModules in them appear to use New(), therefore i mirrored that assuming it is for some reason.

PR #7026, opened 2020-04-21
More machinery power update_var() proc usage

Leshana:
- Adds the update_idle_power_usage() and update_active_power_usage() procs for the respective vars.
- Switches all places modifying those vars directly to call the procs instead.
- This will let us react to the change appropriately, paving the way towards static area power.
- Adds update_power_channel proc for the sake of completeness, but no machines actually modify it so far.
- Also fixes some spots needing to call update_use_power() that were either missed or added since the last PR.

This continuation of PolarisSS13/Polaris#6882 changes _absolutely nothing_ behavior wise; all the procs do is set the variable.   What this does do is paves the way for the goal of "static" area power usage by letting us write code to react to changes in power consumption.

### DIFF
--- a/code/game/machinery/machinery.dm
+++ b/code/game/machinery/machinery.dm
@@ -196,6 +196,18 @@ Class Procs:
 /obj/machinery/proc/update_use_power(var/new_use_power)
 	use_power = new_use_power
 
+// Sets the power_channel var
+/obj/machinery/proc/update_power_channel(var/new_channel)
+	power_channel = new_channel
+
+// Sets the idle_power_usage var
+/obj/machinery/proc/update_idle_power_usage(var/new_power_usage)
+	idle_power_usage = new_power_usage
+
+// Sets the active_power_usage var
+/obj/machinery/proc/update_active_power_usage(var/new_power_usage)
+	active_power_usage = new_power_usage
+
 /obj/machinery/proc/auto_use_power()
 	if(!powered(power_channel))
 		return 0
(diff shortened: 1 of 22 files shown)


code/game/machinery/machinery.dm
```diff
@@ -196,6 +196,18 @@ Class Procs:
...
+	power_channel = new_channel
+
+// Sets the idle_power_usage var
+/obj/machinery/proc/update_idle_power_usage(var/new_power_usage)
+	idle_power_usage = new_power_usage
+
+// Sets the active_power_usage var
+/obj/machinery/proc/update_active_power_usage(var/new_power_usage)
+	active_power_usage = new_power_usage
+
```
atlantiscze, line 210: I'd advise using something like this (please double-check this, didn't actually compile it), to ensure varedit modifications work correctly. While it does nothing at this point, if you eventually want to implement static area power this ensures VV modifications call these setters instead of setting these vars directly.
```suggestion

/obj/machinery/proc/vv_edit_var(var/var_name, var/new_value)
	if(var_name == "power_channel")
		update_power_channel(new_value)
		return TRUE
	else if (var_name == "idle_power_usage")
		update_idle_power_usage(new_value)
		return TRUE
	else if (var_name == "active_power_usage")
		update_active_power_usage(new_value)
		return TRUE
	return ..()
```

Leshana: That is a good idea! I'll add it here.

PR #7042, opened 2020-04-25
Add "icon sheet" asset type

Leshana:
This asset class bundles entire dmi files worth of icons together into a single CSS file using inline data urls.  Much faster than sending dozens of files.
Also changed it so we don't need a hard coded list of all types of asset daturms.


KasparoVy:
Wow. That's hardcore

PR #7045, opened 2020-04-26
Removes limits on eating, centralizes nutrition changing.

Neerti:
This was requested ages ago but nobody ever got around to doing it. I got asked to do this now, and so here we are.

You can now eat as much food as you want. This will probably not cause problems balance-wise since the nutriment reagent already is metabolized at a higher than average rate, but it can be increased further if it becomes a problem.


Atermonera:
There's no cake reaction so hooray will have to do as an ice cream cone or smth

code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Toxins.dm
```diff
@@ -19,7 +19,7 @@
 		if(alien == IS_SLIME)
 			removed *= 0.25 // Results in half the standard tox as normal. Prometheans are 'Small' for flaps.
 			if(dose >= 10)
-				M.nutrition += strength * removed //Body has to deal with the massive influx of toxins, rather than try using them to repair.
+				M.adjust_nutrition(removed * removed) // Body has to deal with the massive influx of toxins, rather than try using them to repair.
```
Mechoid, line 22: Wayward removed instead of strength.

code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Toxins.dm
```diff
@@ -396,7 +396,7 @@
 
 /datum/reagent/toxin/sifslurry/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
 	if(alien == IS_DIONA) // Symbiotic bacteria.
-		M.nutrition += strength * removed
+		M.adjust_nutrition(removed * removed)
```
Mechoid, line 399: As above, so below.

Issue #6984, opened 2020-04-09
New Event System Observations (System 3)


Neerti:
A list of things I will need to tweak before System 3 goes live forever. This will be edited as more things are found.
- [x] Airlock shock and door crush events might need to be balanced better, potentially requiring medical to be present before they can be chosen.
- [x] Engineering window damage event needs to have its weight divided by how many times it's been ran.
- [x] Sif loses gravity during the gravity failure event.
- [x] Xenobio jailbreak event also targets xenoflora isolation because it inherits from xenobio area for some reason. (`/area/rnd/xenobiology/xenoflora_isolation)`
- [x] Should add a message to engineering radio for when jailbreak event ends, so they know when to turn APCs back on.
- [x] Gravity failure seems to scale too heavily with even small amounts of pop, hitting 200+ weight with five or so players.
- [x] Airlock zap/door crush events have a positive weight value when no engineering exists. This should probably not be the case.
- [ ] Spacevine event needs a minimum area size check to avoid spawning outside the escape pods.
- [x] Spacevine event might need a lower weight.
- [x] Maint drones are considered synthetics for the purposes of events like ion laws, which should probably not be the case.
- [x] Make sure the grid check scaling checks for TEGs that are actually running and not just all TEGs that exist.
- [ ] Consider having some events check for if an AI or borgs of a specific module for the purposes of if an event should be runnable.

PR #7046, opened 2020-04-26
Tweaks and fixes for the new new event system.

Neerti:
Resolves a lot of the things noted in #6984, and fixes all the bugs noted there.
Changes:
- Prison break event now tells engineering when they can turn the APCs back on.
- Xenobio prison break should stop breaking a part of xenoflora as well.
- Xenobio prison break won't touch blast doors now, and as such it won't breach xenobio, since getting slimed and ZAS'd is just a bit too unfair.
- The dangerous airlock failure events require medical to be present to be able to fire.
- Gravity failure no longer makes Sif lose gravity.
- Gravity failure lasts for less time, but still a lot longer than the current duration.
- Maint drones no longer make ion laws more likely by existing.
- TEGs that aren't hooked up won't contribute towards causing a grid check.

### DIFF
--- a/code/game/area/areas.dm
+++ b/code/game/area/areas.dm
@@ -47,7 +47,7 @@
 
 	luminosity = !(dynamic_lighting)
 	icon_state = ""
-	
+
 	return INITIALIZE_HINT_LATELOAD // Areas tradiationally are initialized AFTER other atoms.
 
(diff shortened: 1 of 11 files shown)


Atermonera:
On the PoI test: errors

PR #7049, opened 2020-04-26
Overmap Consoles now check access

Novacat:
Also slightly tweaks 513 compatibility to be a little better.


Atermonera:
What access is it checking, specifically?

Novacat:
It works the same way as doors, it checks req_access and req_one_access vars and returns true if those vars are defined and matches the access on the ID the user is wearing.

PR #7051, opened 2020-04-27
Tweak to mining scanner

InterroLouis:
Basically what I'm doing here is adding a new category to the hand held mining scanner. Precious Gems. It's just diamonds. I'm removing diamonds from precious metals, and adding it to its own category, so it's easier to differentiate.


Mechoid:
Well, shit, I guess I'm adding jewlering.

PR #7054, opened 2020-04-27
Technomancer 2.0

Neerti:
Technomancer code is several years old by now, and it shows.
More text will be here when I'm closer to finishing this, just putting it up so people can :eyes: it if they want to.


Neerti:
I'm gonna close this and go take care of some more pressing code needs, since a lot of stuff is starting to fall apart due to a lack of general maintenance.

After that's sorted, I'll open smaller PRs with specific components and prerequsite subsystems (like modifier armor values) first, so it should be easier to review and is probably better verses giant Ater-liver-destroying monolithic PRs.

PR #7055, opened 2020-04-27
Updates machines to use default_apply_parts()

Arokha:
Look Gordon, parts!


code/modules/power/pacman2.dm
```diff
@@ -30,16 +30,9 @@
...
-		..()
-		component_parts = list()
-		component_parts += new /obj/item/weapon/stock_parts/matter_bin(src)
-		component_parts += new /obj/item/weapon/stock_parts/micro_laser(src)
-		component_parts += new /obj/item/stack/cable_coil(src)
-		component_parts += new /obj/item/stack/cable_coil(src)
-		component_parts += new /obj/item/weapon/stock_parts/capacitor(src)
-		component_parts += new board_path(src)
-		RefreshParts()
+	Initialize()
```
Atermonera, line 33: pathing

code/modules/power/smes_construction.dm
```diff
@@ -41,23 +41,23 @@
 
 // These are used on individual outposts as backup should power line be cut, or engineering outpost lost power.
 // 1M Charge, 150K I/O
-/obj/machinery/power/smes/buildable/outpost_substation/New()
-	..(0)
+/obj/machinery/power/smes/buildable/outpost_substation/Initialize()
+	. = ..()
 	component_parts += new /obj/item/weapon/smes_coil/weak(src)
```
Atermonera, line 46: Can these not use default_add_parts?

Leshana: In this case its not quite that simple, it require creating special circuit board types for these mapped-only smes types.  Best way might be to adopt something along the lines of the `additional_spawn_components` Baystation has, but that sort of revamp would be its own project I think.

code/modules/power/pacman2.dm
Leshana: For sure. But I can also totally understand keeping that separate given how it would make the diff be the entire file.

Arokha: Ye, that's why I did it. I think pathing fixes are generally their own thing. I definitely think it should be fixed, but maybe soooomeone should just go through and fix ALL the relative pathing at once.

Atermonera: The problem with that, even moreso than replacing << with to_chat, is that there isn't a particularly easy or simple way of finding _all_ instances

PR #7056, opened 2020-04-27
Adds some hyperlink functionality

the-og-gear:
Used some regex and JS .replace function. Enjoy.

### DIFF
--- a/code/modules/vchat/js/vchat.js
+++ b/code/modules/vchat/js/vchat.js
@@ -579,7 +579,7 @@ function start_vue() {
 				}
 
 				newmessage.content = newmessage.content.replace(
-					/(\b(https?|ftp|file):\/\/[\-A-Z0-9+&@#\/%?=~_|!:,.;]*[\-A-Z09+&@#\/%=~_|])/img, //Honestly good luck with this regex ~Gear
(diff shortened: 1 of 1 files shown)


the-og-gear:
image
Collects links as expected, and gives a valid tag for when you click. Only thing I haven't done is modify any CSS because honestly I'll just make it look horrible.

code/modules/vchat/js/vchat.js
```diff
@@ -578,6 +578,10 @@ function start_vue() {
 					}
 				}
 
+				newmessage.content = newmessage.content.replace(
+					/(\b(https?|ftp|file):\/\/[\-A-Z0-9+&@#\/%?=~_|!:,.;]*[\-A-Z09+&@#\/%=~_|])/img, //Honestly good luck with this regex ~Gear
```
Arokha, line 582: Seems good but allowing `file://` links seems inappropriate given that people could link people to .bat or .exe files on their own computer or do other silly things that might be undesirable in the very old version of IE that byond is using. Maybe just `https?` given that ftp is silly.

the-og-gear: Oop, fixed.

Atermonera:
absolute regex

PR #7057, opened 2020-04-27
Allows robots to search loot piles

atlantiscze:


code/game/objects/structures/loot_piles.dm
```diff
@@ -40,6 +40,10 @@ Loot piles can be depleted, if loot_depleted is turned on.  Note that players wh
 	var/list/uncommon_loot = list()	// Uncommon is actually maybe some useful items, usually the reason someone bothers looking inside.
 	var/list/rare_loot = list()		// Rare is really powerful, or at least unique items.
 
+/obj/structure/loot_pile/attack_ai(var/mob/user)
+	if(isrobot(user) && Adjacent(user))
+		attack_hand(user)
```
Mechoid, line 45: This should probably instead `return attack_hand(user)`

PR #7058, opened 2020-04-27
Allows backpacks to be carried on most hardsuits

atlantiscze:
- Backpacks can be carried on suit storage slot of those hardsuits that occupy back slot (breacher excluded).
- Now, i understand this may be a somewhat controversial change. Feel free to close this PR if you believe it does not fit. Just opening this as i've received this suggestion from another player, and i in fact tried proposing the same thing back at Bay when hardsuits as we know them were added.
- Reformatted some of the allowed lists while i was at it. Makes future modifications somewhat less likely to hit merge conflicts with each entry on its own line.

My balance perspective (long rant):
Hardsuits, as-is, fill very specific roles. In general, unless you absolutely need one of the modules or some specific function (for example, 100% RAD resist combined with EVA capability of CE suit) you don't want to use them, as they require you to trade-in your back slot (which is what this PR aims to change), therefore significantly reducing carrying capacity. Of course, this has its workarounds - usually dragging crates around, or carrying your backpack in hand. This provides minimal actual restriction (you can still carry tons of stuff), but acts as quite major inconvenience and annoyance for the player (you have to juggle with items, keep placing stuff on the ground, etc - generally just clicking 10x more). In total, the only real effect is that it slows you down a bit due to all the clicking required.

Therefore i believe this change is not as massive buff as it might seem in first place. Sure, it lets you carry more things on yourself, but you could also do that easily with a crate or by holding backpack in hand. It just makes doing so more convenient and comfortable to the player.

### DIFF
--- a/code/modules/clothing/spacesuits/rig/suits/alien.dm
+++ b/code/modules/clothing/spacesuits/rig/suits/alien.dm
@@ -51,7 +51,13 @@
 	item_flags = THICKMATERIAL
 	siemens_coefficient = 0.2
 	offline_slowdown = 5
-	allowed = list(/obj/item/weapon/gun,/obj/item/device/flashlight,/obj/item/weapon/tank,/obj/item/device/suit_cooling_unit)
+	allowed = list(
+		/obj/item/weapon/gun,
+		/obj/item/device/flashlight,
+		/obj/item/weapon/tank,
+		/obj/item/device/suit_cooling_unit,
+		/obj/item/weapon/storage
+		)
 
 	air_type = /obj/item/weapon/tank/vox
 
--- a/code/modules/clothing/spacesuits/rig/suits/combat.dm
+++ b/code/modules/clothing/spacesuits/rig/suits/combat.dm
@@ -12,7 +12,14 @@
 	offline_vision_restriction = 1
 
 	helm_type = /obj/item/clothing/head/helmet/space/rig/combat
-	allowed = list(/obj/item/weapon/gun,/obj/item/device/flashlight,/obj/item/weapon/tank,/obj/item/device/suit_cooling_unit,/obj/item/weapon/melee/baton)
+	allowed = list(
+		/obj/item/weapon/gun,
+		/obj/item/device/flashlight,
+		/obj/item/weapon/tank,
+		/obj/item/device/suit_cooling_unit,
+		/obj/item/weapon/melee/baton,
+		/obj/item/weapon/storage
+		)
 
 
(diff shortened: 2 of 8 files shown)


zekesulastin:
What I remember from the last time this was really pushed was that restrictions were still wanted but something like a half/limited-capacity backpack for suit storage might be nice.

The AMI rig at least needs something like this so that it's useful for something; tbh the EVA and CE rigs aren't terrible to use without a backpack (but I might be used to it); haven't used the rescue rig at all to have an opinion one way or the other.

Woodratt:
Last time it came up, Zeke is correct. There was hesitation to allow one to carry a full sized backpack with a rig, but folks seemed alright with a half sized backpack or satchel of sorts. The issue at the time was a lack of sprites and motivation to code such objects.

Novacat:
Coming from a downstream developer who has had suit-mounted backpacks for a while, I found it did not adversely affect balance a whole bunch. Most people here still prefer voidsuits due to their simplicity and ease of acquisition. So I fully support this change. On top of that, Polaris recently added belt-mounted rigs which allow a backpack to be used with them. So that proverbial bridge has been crossed already.

woody41:
Nice pull request. It is very helpfull. Regarding the half-size backpack: I think this is not justified by real life. Imagine the hard suit with backpack - it is possible when the hardsuit is deployed to body, the character's back is free, so suit storage should be able to hold the backpack when it's deployed. (Anyway suit storage can hold oxygen tank now) Or another idea, when you can insert a lot of modules to hardsuit, what about to be able to insert backpack as module - like a internal storage? I agree with this pull request. It can be a part of station which makes a difference from another servers.

KillianKirilenko:
Not particularly gamebreaking, but if anyone's really hung up on the \~balance\~ I think a simple way to resolve it would be to make the various bag types mechanically unique as opposed to almost all cosmetically interchangeable.

Right now satchels, messenger bags, and backpacks are all identical; vary their slowdown based on their actual load %, then adjust their capacity/max item size/max slowdown to suit. Satchels would be the smallest/lightest, then messenger bags, then backpacks, then duffels remain as the biggest option. BoHs would be unaffected because transdimensional shenanigans or however they do their thing, obviously.

Plus this would resolve silliness like tucking an entire 3-4ft long pump shotgun in a modestly-sized hip/waist satchel.

Atermonera:
We'll see if I actually remember to bring this up at a staff meeting in a few months time, but based on folks' feedback here I don't anticipate issues

PR #7059, opened 2020-04-28
Supermatter delamination tweaks

atlantiscze:
- In short, delamination changed to more or less mirror effects it has on Bay.
- All of the new delamination effects can be easily tweaked in the future using set of defines.
- In general, aim is to make supermatter delamination a more or less "recoverable" event with enough materials and manpower. Before it was just a mangled irrepairable mess of scuttled steel.
- This brings a LOT of performance improvements and runtime error fixes (grav pull of anchored items which were never meant to move tends to cause runtimes)

Detailed changes:
- Supermatter (and also S5/S6 singularity in the event someone manages to build one, they share the proc) no longer pull anchored items. This brings a MASSIVE reduction in delamination lag (from few second freezes to quite smooth) and prevents things from breaking (i saw dozens of various runtimes when delamination occured as it forcibly dragged things around that weren't supposed to be dragged).
- Delamination crashes the station's power grid, shutting down all APCs for few minutes. Critical areas get shut down only for fraction of that time (we don't want to one-hit the AI). This includes healthcare facilities, which are likely to be very important given the other effects.
- Delamination irradiates and briefly weakens all mobs on the station (not only on the same Z level). I have verified that this is enough to kill an unprotected human if no kind of medical care (at least dylovene, ideally something against rads) is administered.
- Delamination will break a bunch of lights. Nuisance, but easily repairable. Adds to the atmosphere.
- Delamination breaks part of the solar arrays, reducing their output by roughly 60%. This is way less than required to run the station. Puts more emphasis on delamination being engineering-centric emergency, with power rationing being a necessity until solars can be fixed, or other power source secured. Rewards engineering if they managed to stockpile some energy in the SMESes around the station.
- Delamination's explosion is quite weaker than it used to be. It should still be enough to pretty much destroy the engine room (see attached screenshots below), and cause some minor damage in adjacent rooms. It should still, however, be repairable with enough resources and some time.
- Supermatter no longer has additional 99% flat modifier to miss a station Z level when ejected. Emergency core ejection is a last resort measure. It does have its risks. With other changes in this PR, delamination on outer hull is significantly less dangerous than it would be before. It still has various secondary risks, however, such as the radiation or temporary power outage, but with considerably mitigated damage (explosion in space, mostly on outer grilles is generally weaker)

Screenshots:
https://i.imgur.com/K7nBd8a.png    (Standing directly where the core was. Engine room has heavy structural damage and all pipes/machinery are pretty much devastated)
https://i.imgur.com/SGBdTRo.png     (Standing in engine control room. Damage nearby is quite minor (occassional snapped wire or wall panel), but quite widespread).
https://i.imgur.com/JqzF4pW.png     (Solar arrays are heavily damaged. They still provide /some/ power but only a fraction of the original)

Going to leave this open for a day or two before marking as ready for feedback regarding the mechanics.

### DIFF
--- a/code/modules/power/singularity/act.dm
+++ b/code/modules/power/singularity/act.dm
@@ -47,31 +47,24 @@
 		return 2
 
 /obj/singularity_pull(S, current_size)
-	if(simulated)
-		if(anchored)
-			if(current_size >= STAGE_FIVE)
-				step_towards(src, S)
-		else
-			step_towards(src, S)
+	set waitfor = 0
+
+	if(anchored)
+		return
+
+	sleep(0) //this is needed or multiple items will be thrown sequentially and not simultaneously
+	if(current_size >= STAGE_FOUR)
+		step_towards(src,S)
+		sleep(1)
+		step_towards(src,S)
+	else if(current_size > STAGE_ONE)
+		step_towards(src,S)
+	else ..()
 
 /obj/effect/beam/singularity_pull()
 	return
 
 /obj/effect/overlay/singularity_pull()
-	return
-
-/obj/item/singularity_pull(S, current_size)
-	spawn(0) //this is needed or multiple items will be thrown sequentially and not simultaneously
-		if(current_size >= STAGE_FOUR)
-			//throw_at(S, 14, 3)
-			step_towards(src,S)
-			sleep(1)
-			step_towards(src,S)
-		else if(current_size > STAGE_ONE)
-			step_towards(src,S)
-		else ..()
-
-/obj/machinery/atmospherics/pipe/singularity_pull()
 	return
 
 /obj/machinery/power/supermatter/shard/singularity_act()
@@ -112,30 +105,6 @@
 				O.singularity_act(src, current_size)
 	ChangeTurf(get_base_turf_by_area(src))
 	return 2
-
-/turf/simulated/floor/singularity_pull(S, current_size)
-	if(flooring && current_size >= STAGE_THREE)
-		if(prob(current_size / 2))
-			var/leave_tile = TRUE
-			if(broken || burnt || flooring.flags & TURF_IS_FRAGILE)
-				leave_tile = FALSE
-			playsound(src, 'sound/items/crowbar.ogg', 50, 1)
-			make_plating(leave_tile)
-
-/turf/simulated/wall/singularity_pull(S, current_size)
-
-	if(!reinf_material)
-		if(current_size >= STAGE_FIVE)
-			if(prob(75))
-				dismantle_wall()
-			return
-		if(current_size == STAGE_FOUR)
-			if(prob(30))
-				dismantle_wall()
-	else
-		if(current_size >= STAGE_FIVE)
-			if(prob(30))
-				dismantle_wall()
 
 /turf/space/singularity_act()
 	return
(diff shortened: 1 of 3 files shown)


Greenjoe12345:
eh, I kinda like a delam absolutely nuking engineering, and the implosion of walls and such before the explosion is cool too, if that could be fixed to not cause runtimes somehow

atlantiscze:
I'm afraid, that is not feasible. I've actually been thinking about keeping this in one way or another, but decided against it. Not only the gravitation pull creates massive lag (both with SM and singularity), it also breaks the rule of "Anchored object [anchored = TRUE] should not move". A lot of things are designed with this rule in mind, and when the game suddenly breaks it (arguably this proc is the only one which does it), weird things happen that cause flood of runtimes from everywhere. Power, atmospherics, machinery, all these things are pretty sensitive to their position and expect to not be moved without their standard construction/deconstruction being used.

Now, sure, there is a different solution, but it has a large _but_. It would be possible to go through all various problematic objects nearby, and ensuring they implement Move() or similar proc that ensures movement updates the object's state (such as cable connection or whatever). This may be tempting but i strongly advise against it. Even now, delamination is an extremely laggy incident. If this was implemented, for example only for cables (rebuilding powernets is a performance nightmare), the sheer amount of calculations which would have to be performed by the server would rise a lot, likely _multiplying_ the duration for which game lags.

So in general this PR aims to offset this in some way. Direct destruction was reduced (without gravity pull the damage looks considerably more repairable), while indirect effects were added. I'm open to more ideas what interesting and possibly anomalous effects it might have. Effects like this also have an advantage of creating things to do for other departments than engineering (which is likely pretty screwed at this point). Right now this is mostly focused at Medical (Triage of heavily irradiated patients), the janitor/utility staff (Replacement of about half the lights on the station) and Cargo (Coordinating with engineering, if anyone survived there, to provide repair materials and equipment). The general idea, however, is that while delam should be a heavy impact event with very high danger, it should not be _guaranteed_ round-ender if the crew manages to actually muster a response and deal with it. Some effects are immediate (radiation) while other creep up and stab you from behind if you ignore them (solar damage, once storages around the station are discharged most power fails)

woody41:
I don't care about runtime errors (I am not byond developer), but I agree with this, because it is possible to do repairs in engineering and fix everything. I saw onec what happend when the supermatter exploded on Polaris and it is almost unfixable. This can be repaired, however this is why we play this game or no? We play Engineering to repair the station and keep them in operative mode, not to state "Call the shuttle".

Cerebulon:
Personally I'd prefer if at the very least the direct explosion radius wasn't so... Piddly small. If the moving anchored objects (Which honestly I didn't mind the lag for because it looked gnarly as hell but I can see the potential need for a change) is out then I think it should at least cause a similar level of fuckery to engineering/cargo.
Does it still scale with power output/temp at least?

atlantiscze:
I've added a change that scales explosion strength depending on supermatter power level. It is, once again, easily tweakable in terms of actual thresholds if we need to fine-tune actual values later.

Right now i've went with following values (depending on energy level, which is equal to EER displayed in the monitor program).
For reference - old explosion power was 8.

Below 200 - Unchanged, explosion of average strength 8.
Between 200 and 2000 - Scales linearly, between 8 and 16, depending on power level
Above 2000 - Capped at explosion strength 16 (double of what it used to be)

Once again, for reference: Energy level of 200 is just about enough to run the station with small reserve (without solars at all). Energy level of 2000 is just about the maximum you can push it to with three emitters.

Atermonera:
Re: shutting down power for a bit, would it be possible to mitigate that with substations? Either being on substations when the blast hits not affecting connected subnets, or being able to toggle the breaker to trip an earlier restart? This has more balance considerations than "it affects everything but these things that're important to keeping people in the round are arbitrarily less affected", but I also feel like making it more interactive could be interesting, at least.

Re: global radiation, about how long did it take for the unprotected human to die? Just so we can get a sense of what the timeframe would be for medical to find and stab _everyone_ with a syringe.

atlantiscze:
Almost point blank (few tiles outside of explosion radius) with 0 protection it took approximately ten minutes.

As for power outage, it can be manually mitigated by clicking a reset button in the APC (assuming you can unlock the APC)

PR #7064, opened 2020-04-30
Adds preference to control multilingual behaviour

### LINKED ISSUE #7040: Gibberish + multilanguages = no longer able to use hyphens in a sentence.
The 'hyphen + letter' trigger for speaking in gibberish causes the same thing when used mid-sentence, when hyphenating words.

`Bejel Salin says, "How are-you."`

#### What actually happened
`Bejel Salin says, "How are [gibberish string]."`

#### Steps to reproduce
Use a hyphen in a sentence with certain letters surrounding it - for instance, "Test-test" does not trigger it.

Atermonera:
The modes are
* Default, which functions exactly as the parser works now
* Extra space, which only makes a new language piece when there's an extra space following the language key, so matching the pattern "[delimiter][key] "
* Extra delimiter, which only makes a new language piece when there's a valid language prefix following the language key, so matching the pattern "[delimiter][key][delimiter]"
* Disabled, which will only look at the first 2 characters after trimming radio prefixes to find any language keys, and lacking those will use the default language, for the entire message.

This should hopefully be more amenable to people with characters that stutter.

Examine mode preference should persist across reconnections to a single round, but it still appears to reset on server restart.
Still looking into how to properly get these to go into the savefile
Fixes #7040 
Tested, and the preference performs as noted above for examine pref, where if I closed my client and reconnected to the server, it would persist, but if I closed and reopened the server, it would be reset. Unfortunately, I can't simply use the /datum/client_preference stuff because that is, as far as I can tell, explicitly a binary toggle, and getting these multi-mode preferences to work with that would require a not-insignificant refactor (Not to mention having to figure out how in the heck they're actually manipulating the preferences in and out of savefiles).

### DIFF
--- a/html/changelogs/atermonera_multilingual_pref.yml
+++ b/html/changelogs/atermonera_multilingual_pref.yml
@@ -1,4 +1,4 @@
-author: Shadow Quill
+author: Atermonera
 delete-after: True
 
 # Any changes you've made.  See valid prefix list above.
--- a/html/changelogs/atermonera_multilingual_pref.yml
+++ b/html/changelogs/atermonera_multilingual_pref.yml
@@ -1,0 +1,11 @@
+author: Shadow Quill
+delete-after: True
+
+# Any changes you've made.  See valid prefix list above.
+# INDENT WITH TWO SPACES.  NOT TABS.  SPACES.
+# SCREW THIS UP AND IT WON'T WORK.
+# Also, all entries are changed into a single [] after a master changelog generation. Just remove the brackets when you add new entries.
+# Please surround your changes in  double quotes ("), as certain characters otherwise screws up compiling. The quotes will not show up in the changelog.
+changes: 
+  - rscadd: "Added a preference to control multilingual parsing behaviour, with a few different modes. Should hopefully be less punishing to people who stutter and use hyphens as a language key."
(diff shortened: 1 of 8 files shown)


html/changelogs/atermonera_multilingual_pref.yml
```diff
@@ -0,0 +1,11 @@
+author: Shadow Quill
```
Neerti, line 1: Not sure if this is intentional or not.

polaris.dme
```diff
@@ -2932,7 +2932,7 @@
 #include "code\ZAS\Zone.dm"
 #include "interface\interface.dm"
 #include "interface\skin.dmf"
-#include "maps\southern_cross\southern_cross.dm"
+#include "maps\example\example.dm"
```
Neerti, line 2935: Looks like a minor oopsie, assuming this isn't a WIP PR. Surprised Travis didn't fail due to this.

TheFurryFeline: Interesting indeed. 
```## UNIT_TEST: --------------- AI Satellite(/area/aisat) lacks an Air scrubber.
## UNIT_TEST: --------------- AI Satellite(/area/aisat) lacks an Air vent.
## UNIT_TEST: !! FAILURE !! [MAP: Area Test APC / Scrubbers / Vents Z level 1]: [1/1]Some areas lacked APCs, Air Scrubbers, or Air vents.
```

PR #7080, opened 2020-05-03
Add UAVs

Arokha:
Adds UAVs, printable at a mechfab
They're controllable via NTnet software that you can download and fly them around with.
The further they are, the slower they get, and the worse the picture quality is.
They can hear speech/see emotes, and use a power cell to power themselves.
They can take damage.


code/modules/modular_computers/file_system/programs/generic/uav.dm
```diff
@@ -0,0 +1,266 @@
+/obj/item/modular_computer
+	var/list/paired_uavs //Weakrefs, don't worry about it!
```
atlantiscze, line 2: This should definitely not be a computer-level variable. It should be on nano module level or, at most, at program level.

An easy alternative way to allow a nano module to connect to UAVs (since nano module is not a physical object you can click) would be to allow pairing with UAVs on directly adjacent tiles through the UI instead of requiring a click with computer inhand. That way, you open the program, move the UAV next to whatever computer you are running it on, and press a button in the UI to connect them. It also has a nice bonus of letting you control UAVs from devices which can not be picked up.

Arokha: I mean, that's great and all, but they would get lost every time you end that program.

atlantiscze: Fair point, in that case, i'd recommend going with program-level variable.

Arokha: Don't those also get destroyed when you end the program?

atlantiscze: Nope, nano modules are destroyed when program is ended, programs themselves are not. They are permanent (specifically, they are bound to the hard drive component).

code/game/objects/items/uav.dm
```diff
@@ -0,0 +1,348 @@
...
+		return ..()
+
+	var/list/options = list(
+		"Pick Up" = radial_pickup,
+		"(Dis)Assemble" = radial_wrench,
+		"Toggle Power" = radial_power,
+		"Pairing Mode" = radial_pair)
+	var/choice = show_radial_menu(user, src, options, require_near = !issilicon(user))
+	
+	switch(choice)
```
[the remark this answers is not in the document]
Arokha: That's what the require_near is for. In this case, any non-AI has to have adjacency.

code/game/objects/items/uav.dm
```diff
@@ -0,0 +1,348 @@
+#define UAV_OFF 0
+#define UAV_ON 1
+#define UAV_PAIRING 2
+#define UAV_PACKED 3
+
+/obj/item/device/uav
+	name = "recon skimmer"
+	desc = "A semi-portable reconisance drone that folds into a backpack-sized carrying case."
```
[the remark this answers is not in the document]
Arokha: Fixed

code/game/machinery/doors/door.dm
```diff
@@ -111,7 +111,13 @@
 			return
 		else
 			bumpopen(M)
-
+	// VOREStation Add - UAVs open public doors
```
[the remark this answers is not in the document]
Arokha: Removed

Arokha:
Rebased to fix conflict

PR #7084, opened 2020-05-04
Make human speech bubbles more accurate

Arokha:
Organics with a synthetic voicebox get a synthetic speech bubble. Lifelike synths don't, until they get damaged.


Arokha:
Nevermind Novacat beat me to it

PR #7087, opened 2020-05-04
Port SSnightshift from Paradise

Arokha:
Dims all the lights at night, based on what your map datum returns from get_nightmode(), which checks get_night, and get_zlevel_time(). You should customize those if you don't want to base it on whatever planet z1 is, otherwise it will use the fictional 'station time'.

### DIFF
--- a/code/controllers/subsystems/nightshift.dm
+++ b/code/controllers/subsystems/nightshift.dm
@@ -20,7 +20,7 @@ SUBSYSTEM_DEF(nightshift)
 	return ..()
 
 /datum/controller/subsystem/nightshift/fire(resumed = FALSE)
-	if(round_duration_in_ticks < nightshift_first_check)
+	if(round_duration_in_ds < nightshift_first_check)
 		return
 	check_nightshift()
 
--- a/code/controllers/subsystems/nightshift.dm
+++ b/code/controllers/subsystems/nightshift.dm
@@ -1,0 +1,62 @@
+SUBSYSTEM_DEF(nightshift)
+	name = "Night Shift"
+	init_order = INIT_ORDER_NIGHTSHIFT
+	priority = FIRE_PRIORITY_NIGHTSHIFT
+	wait = 60 SECONDS
+	flags = SS_NO_TICK_CHECK
+
+	var/nightshift_active = FALSE
+	var/nightshift_first_check = 30 SECONDS
+
+	var/high_security_mode = FALSE
+
+/datum/controller/subsystem/nightshift/Initialize()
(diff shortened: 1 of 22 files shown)


code/controllers/subsystems/nightshift.dm
```diff
@@ -0,0 +1,62 @@
...
+	if(nightshift_active != night_time)
+		update_nightshift(night_time, announcing)
+
+/datum/controller/subsystem/nightshift/proc/update_nightshift(active, announce = TRUE)
+	nightshift_active = active
+	if(announce)
+		if(active)
+			announce("Good evening, crew. To reduce power consumption and stimulate the circadian rhythms of some species, all of the lights aboard the station have been dimmed for the night.")
+		else
+			announce("Good morning, crew. As it is now day time, all of the lights aboard the station have been restored to their former brightness.")
```
Atermonera, line 58: `standard`, rather than `former`? It's a lot more likely for a round to pass from night to day by virtue of rng picking a predawn time than it is for a round to last long enough that both transitions could be seen

Arokha: Depends on your station/planet. We have 6 hour shifts on a planet with 3 hour days, so it transitions every 1.5 hours. up to 4 times.

The announcement is a copypaste from Paradise either way. If you want me to change it, I can.

Arokha:
> Would it be possible to get an admin button to set/unset nightshift on all apcs for adminbus?

Would be easy to just call it on SSnightshift or SDQL it on all the APCs, but I can make a verb too if you want. Do you?

Atermonera:
Using the controller or sdql should be sufficient. Since you're still pushing commits, is this ready for merge?

Arokha:
Short version: Yup, it's fine.

Long version: It had some bugs on servers with *no* planets defined at all, and there was a preexisting bug in that the stat panel 'station time' was off by the timezone offset between the server and GMT due to whoever wrote that's oversight, so I fixed those things too. It wouldn't affect nightshift but you might look at the stat panel's "Station Time" and be like "why is it night, it's 8am!" or whatever because the time displayed there was wrong.

Arokha:
I was gonna include this in a separate PR but here's another for fun. More controls on APCs for nightshift stuff that a player PR'd to us. So there's a toggle between 'disabled', 'automatic', and 'enabled' on each APC.

Cerebulon:
This is already merged now but I'd like to say we do have the gentle "chime" alert noise that I think other servers use for this, it's notice3.ogg and is currently only used for... Something which doesn't work (Arrivals announcements sound)

PR #7095, opened 2020-05-05
Add cooldown to pen clicks

Arokha:
Fixes https://github.com/VOREStation/VOREStation/issues/7000

### DIFF
--- a/code/modules/paperwork/pen.dm
+++ b/code/modules/paperwork/pen.dm
@@ -26,6 +26,9 @@
 	drop_sound = 'sound/items/drop/accessory.ogg'
 
 /obj/item/weapon/pen/attack_self(var/mob/user)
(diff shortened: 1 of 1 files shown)


Atermonera:
travis

Arokha:
Uses code from the move/click refactor PR

Arokha:
Can you re-blep travis now that the other PR is done?

Arokha:
You can restart travis on this now and it should compile fine now that move/click code is in.

Also you can call it bikeshed but it brings the server to it's knees if someone AHK's it, because there's no cooldown at all.

Atermonera:
It's bikeshed because I recognize it as a fix that's necessary, but one which will enrage a fair few people that treasure the ability to spamclick pens without doing it to the point it kills the server.

PR #7099, opened 2020-05-05
Rewrite airlock electronics

Arokha:
- Airlock electronics do a better job trying to find your ID. You don't need it in your hand to swipe your ID, it can read it from your ID slot, etc.
- Configuration access can be bypassed via emag (not on secure airlock electronics).
- Even bypassed, if you don't have engineer access, you can't set accesses other than the ones on your current ID (presumably those aren't actually stored on the device? I dunno.)
- Converts the access you need to be able to apply 'any' access into a separate var. Defaults to access_engine (so no change, but we need this for offmap spawns being able to build their own airlocks).
- If you unlock it (either via emag or because you have an ID in req_one_access), and you don't have an access in the apply_any_access list, you can apply whatever accesses your own ID card has, but not 'every station access' like before.

### DIFF
--- a/code/game/machinery/doors/airlock_electronics.dm
+++ b/code/game/machinery/doors/airlock_electronics.dm
@@ -64,7 +64,7 @@
 	if (usr.stat || usr.restrained() || (!ishuman(usr) && !istype(usr,/mob/living/silicon)))
 		return
 	if (href_list["close"])
-		usr << browse(null, "window=airlock")
+		usr << browse(null, "window=airlock_electronics")
 		return
 
 	if (href_list["login"])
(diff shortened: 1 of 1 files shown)


code/game/machinery/doors/airlock_electronics.dm
```diff
@@ -8,106 +6,148 @@
...
+
+	user << browse(t1, "window=airlock_electronics")
+	onclose(user, "airlock")
+
+/obj/item/weapon/airlock_electronics/Topic(href, href_list)
+	..()
+	if (usr.stat || usr.restrained() || (!ishuman(usr) && !istype(usr,/mob/living/silicon)))
+		return
+	if (href_list["close"])
+		usr << browse(null, "window=airlock")
```
Atermonera, line 67: Is this supposed to be `"window=airlock_electronics"`, as used a few lines above?

Arokha: Yyyyes. That bug predates this PR though, but yes it should be that. I'll fix it.

PR #7008, opened 2020-04-16
Add ianshirt sprite for Teshari

Razgriz1032:
Adds a teshari sprite for the ianshirt. Sprite made by Skits from CHOMPStation.
https://github.com/PolarisSS13/Polaris/blob/a8776af30b0c98ed867c7a78e8d422230b23c703/code/modules/clothing/suits/miscellaneous.dm#L245

image

PR #7102, opened 2020-05-07
Ian shirt for Teshari attempt 2

Razgriz1032:
This is attempt number 2 because a conflict wasn't resolving no matter what I tried.

Old attempt: https://github.com/PolarisSS13/Polaris/pull/7008

Adds a teshari sprite for the ianshirt. Sprite made by Skits from CHOMPStation.
https://github.com/PolarisSS13/Polaris/blob/master/code/modules/clothing/suits/miscellaneous.dm#L245

image


Razgriz1032:
Right, so this passes map diff renders @Atermonera

PR #7110, opened 2020-05-08
adds logging variable to enable shuttle log spam

### LINKED ISSUE #7048: [SUGGESTION] Shuttle Debug Logs Galore
Okay, so, it's been a heck of a problem with every shuttle move popping up on DEBUG logs because of its defines in https://github.com/PolarisSS13/Polaris/blob/master/code/__defines/shuttle.dm#L25
```
#ifndef DEBUG_SHUTTLES
	#define log_shuttle(M)
#else
	#define log_shuttle(M) log_debug("[M]")
#endif
```
DEBUG not to clutter up chat constantly with every shuttle move, which is an issue far downstream on Chompstation and I assume the same's the case here because you also have Southern Cross with the automated shuttle transports to and from Sif.
#### What actually happened
BLAH, SO MUCH LOG CLUTTERING THINGS UP
#### Steps to reproduce
Be Dev, turn DEBUG logs on, be constantly spammed about things like 

> DEBUG: Shuttle 2 moving to Skies of Sif (211,52). Areas are The Hangar Two 
DEBUG: perform_shuttle_move() current=Shuttle 2 Transit (239,19) destination=Skies of Sif (211,52) 
DEBUG: Shuttle 2 moving to Shuttle 2 Sky Transit (237,52). Areas are The Hangar Two 
DEBUG: Translating the Hangar Two x2
DEBUG: perform_shuttle_move() current=Skies of Sif (211,52) destination=Shuttle 2 Sky Transit (237,52) 
DEBUG: Shuttle 2 moving to Outpost East Pad (132,203). Areas are The Hangar Two 
DEBUG: Translating the Hangar Two 
DEBUG: perform_shuttle_move() current=Shuttle 2 Sky Transit (237,52) destination=Outpost East Pad (132,203

#### Additional info:
 - I'd love this to be refactored into something else so it doesn't clutter a whole lot of things up. Maybe make it add to log_world or something. Or make a new system up and have it log to that instead of DEBUG

Atermonera:
Fixes #7048
If you want to debug a particular shuttle you just turn the variable on with VV, or change the code to force it on when the shuttle is assembled at roundstart.

### DIFF
--- a/code/modules/shuttles/shuttle.dm
+++ b/code/modules/shuttles/shuttle.dm
@@ -30,6 +30,8 @@
 
 	var/tmp/depart_time = 0 //Similar to above, set when the shuttle leaves when long jumping. Used for progress bars.
 
+	var/debug_logging = FALSE // If set to true, the shuttle will start broadcasting its debug messages to admins
+
 	// Future Thoughts: Baystation put "docking" stuff in a subtype, leaving base type pure and free of docking stuff. Is this best?
 
 /datum/shuttle/New(_name, var/obj/effect/shuttle_landmark/initial_location)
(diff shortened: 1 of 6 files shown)


TheFurryFeline:
Thank you for this. <3

PR #7114, opened 2020-05-08
Fixes #649

### LINKED ISSUE #649: Trying to delete my communicator or ID deleted my mob.
I have no idea how to reproduce this.
- Right-click delete on a communicator on the ground with my mob and my ID card: deleted my mob.
- Right-click delete on the communicator with my resulting ghost + ID on the turf: deleted my ghost.
- Moved off the turf, right-click and delete the ID or the communicator: deleted my ghost.
- Right-click deleted a random wheelchair: deleted the wheelchair.

Atermonera:
[Tested](https://puu.sh/FHLVd/fd1597fc03.gif) (The communicator was deleted shortly after the end of this because apparently I guess the gif recorder removes the last few seconds, where the regular recorder doesn't)
screenshot
Fixes #649

### DIFF
--- a/code/modules/admin/admin_tools.dm
+++ b/code/modules/admin/admin_tools.dm
@@ -9,18 +9,18 @@
 	if(M.mind)
 		dat += "<b>Current Antag?:</b> [(M.mind.special_role)?"Yes":"No"]<br>"
 	dat += "<br><b>Note:</b> This is arranged from earliest to latest. <br><br>"
-	
+
 
 	if(!isemptylist(M.attack_log))
-		dat += "<fieldset style='border: 2px solid white; display: inline'>"	
+		dat += "<fieldset style='border: 2px solid white; display: inline'>"
 		for(var/l in M.attack_log)
 			dat += "[l]<br>"
-		
+
 		dat += "</fieldset>"
-		
+
 	else
 		dat += "<i>No attack logs found for [M].</i>"
-			
+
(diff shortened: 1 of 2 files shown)


Atermonera:
The issue appears to have been how the atom's name was being expanded and parsed, that is, when I typed delete I to the command line and supplied it with a partial name, then selected the communicator from the drop-down, it would escape the apostrophe with a \. If I then pressed enter, it would confirm to delete the mob, but if I manually removed the slash, it would confirm to delete the communicator. This behaviour was replicated with a double-quote in the name as well, but it was only reproducible if the substring prior to one of those two characters was a valid match for the name of another atom. For the mob named John Doe, `John Doe's communicator` would match the mob, but `J'ohn Doe's communicator` would not.

While I was playing with the fix, sending the autocomplete validation thing into a function and printing it would yield "[name].[memory address]", which I assume to mean that it's actually evaluating the whole name and matching it correctly to a ref instead of matching the substring. But I still don't really understand why that is.

If I returned `list(A)`, as was used in some of the handful of examples of the usage of this feature, it wouldn't actually match anything, but providing it with view() appears to work just fine, despite view() being the same list it was originally trying to match against

KasparoVy:
Checked the mirror and was relieved to see higgs bugson on this. What a wierd issue! Great catch

MistakeNot4892:
Nice fix. Was not expecting to see this in my notifs.

PR #7115, opened 2020-05-08
Add some ctrl/alt vehicle helper clicks

Arokha:
Bike: Ctrl is engine, alt is kickstand
Train/Quad: Ctrl is engine, alt is key grab

### DIFF
--- a/code/modules/vehicles/bike.dm
+++ b/code/modules/vehicles/bike.dm
@@ -54,6 +54,12 @@
 			return
 	..()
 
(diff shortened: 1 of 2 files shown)


Atermonera:
Could you please add that usage to the description_info? Even though people don't read the examine tab, it's a lot easier to tell them they should read it when they complain about usage not being recorded anywhere.

Arokha:
Ok, added!

PR #7116, opened 2020-05-08
When looking at machines, don't allow mesons/material/etc

Arokha:
Don't allow mesons through machine views


Atermonera:
Will this require resetting mesons if you stop looking through the machine?

Arokha:
Nop

PR #7119, opened 2020-05-08
Fixes mob AI not getting properly disabled under player control

### LINKED ISSUE #7073: Possessing simple mobs doesn't disable their AI
While performing actions of entropy as a goose, a detective chanced to pick a fight with me. Now, I know full well that a goose is not about to win in a fight against human mobs with tools, so my reaction was to run away, and steal again another day.
To my surprise, instead of respecting only my keyboard commands, the goose beelined for the poor detective to try and bite at them, and would do so at a moment's notice should the detective get too close to me until the ai_holder was swapped for a passive one, and even then it would wander aimlessly because wander was set to true, when it should have remained motionless without movement commands from me.

Goose to behave like a good little possessed Chaos Extraordinaire

#### What actually happened
A brand new combat style?

#### Additional info:

Verkister:
Something made the things that used to disable mob AI just make the mob take a nap instead so here's a thing that actually disables it again.
Fixes #7073

### DIFF
--- a/code/controllers/subsystems/ai.dm
+++ b/code/controllers/subsystems/ai.dm
@@ -26,6 +26,8 @@ SUBSYSTEM_DEF(ai)
 		--currentrun.len
 		if(!A || QDELETED(A) || A.busy) // Doesn't exist or won't exist soon or not doing it this tick
 			continue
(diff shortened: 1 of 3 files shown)


Verkister:
okay there's a better way

PR #7121, opened 2020-05-08
Door shock() always has a cooldown

### LINKED ISSUE #877: No click delay on electrified doors!
Your armor softens the blow!
You feel a mild shock course through your body.
You feel a mild shock course through your body.
You feel a mild shock course through your body.
Opal fires the laser carbine!
You feel a mild shock course through your body.
You feel a mild shock course through your body.

I was treated to this!

Atermonera:
Fixes #877 
The issue was that ..() only returned TRUE if the mob was stunned, i.e., if they actually took damage
In theory this fix could be abused by multiple people, i.e., if one person were to trigger the shock, another could interact with it safely in the brief period before it resets. In practice, I expect this will require timing too precise to reliably achieve, but if it does become an issue then I suppose a fix for that would be to store the has_shocked in a list, with an entry for each mob that interacts with the door while it's electrified

### DIFF
--- a/code/game/machinery/doors/airlock.dm
+++ b/code/game/machinery/doors/airlock.dm
@@ -637,13 +637,10 @@ About the new airlock wires panel:
 		return 0
 	if(hasShocked)
 		return 0	//Already shocked someone recently?
-	if(..())
-		hasShocked = 1
-		sleep(10)
+	. = ..()
+	hasShocked = 1
+	spawn(10)
 		hasShocked = 0
-		return 1
-	else
-		return 0
 
 
 /obj/machinery/door/airlock/update_icon()


code/game/machinery/doors/airlock.dm
```diff
@@ -637,13 +637,10 @@ About the new airlock wires panel:
 		return 0
 	if(hasShocked)
 		return 0	//Already shocked someone recently?
-	if(..())
-		hasShocked = 1
-		sleep(10)
+	. = ..()
+	hasShocked = 1
+	spawn(10)
 		hasShocked = 0
```
Mechoid, line 643: Should this not default to return `TRUE`?

Atermonera: It will return whether or not the door actually shocked someone. I tried using setClickCooldown, but that didn't appear to be processing correctly (I set it to user.get_click_cooldown() [sic], and could spam click it for lots of sparks when it should only have been processing one click every 0.8 seconds, similar behaviour for giving it 10 directly)

PR #7132, opened 2020-05-09
Prevents simplemob spawning on random turfs, pending a better system.

Mechoid:


Atermonera:
Specifically comments out the code that's used for it. While nice in theory, spawning mobs on a per-turf basis is a very bad idea, because the law of large numbers guarantees that no matter how small you make the percentile chance, there's still going to be hundreds of mobs. On the gala round today, we rolled particularly poorly and the game was so laggy as to be unplayable until I Del-All()'d all /mob/living/simple_mobs, a process which took a significant amount of time to actually finish.

A better implementation would choose a number of each type of mob to spawn after poi's are placed, and then choose where to spawn them, such that they aren't placed on poi's and could theoretically be placed in groups/clusters.

PR #7133, opened 2020-05-09
Upstream Port of Footstep Sounds for multiple tiles.

Rykka-Stormheart:
Shuttle tiles, default floors are still "regular".
Lava, sand, water, grass, beach, mining turfs all have sounds playing now for each footstep. <3

A small balance note - water tiles slow you down by 4, and 8 for deep/ocean tiles. I'll remove these and put them in a separate PR if requested.

Massive credit to @Arokha for helping me figure out the floor system and sitting through getting it put together. <3 <3 <3

Earlyport of https://github.com/VOREStation/VOREStation/pull/7737


Greenjoe12345:
This appears to be copying up a Vorestation edit too, also snow has hull.ogg sounds?

Rykka-Stormheart:
> This appears to be copying up a Vorestation edit too, also snow has hull.ogg sounds?

It has a vorestation edit in regards to the animal_chance, I can remove that, I'll have to anyhow to resolve conflicts.

Snow does not have hull.ogg, it's github reducing the file preview.

Rykka-Stormheart:
Fixed the VOREStation edit.

code/game/turfs/simulated/outdoors/grass.dm
```diff
@@ -63,10 +65,6 @@ var/list/grass_types = list(
 	. = ..()
 
 /turf/simulated/floor/outdoors/grass/Initialize()
-	if(prob(50))
-		icon_state = "[initial(icon_state)]2"
```
Atermonera, line 67: Was this code not doing anything or..?

Rykka-Stormheart: It's been replaced by the has_base_range system, which automatically picks a turf determined by the name in files.

Same effect, just does it in the flooring define using the pre-existing system instead of individually in the turf code.

Rykka-Stormheart: Refer to flooring.dm, line 31: https://github.com/PolarisSS13/Polaris/pull/7133/files#diff-539c840e82878c1f430addffa6474fb2R31

Rykka-Stormheart:
> I don't think I see anything editting the frequency, so just to confirm, this is just updating the sounds that play/adding new ones, yes?
> 
> And obligatory request for pertinent licensing/source information for the sounds.

Regarding the licensing, I was told all these sounds came from open-source libraries and public access stuff like freesound.org iirc. I'll check with Chompers for specifics, but all of these are open-source to my knowledge.

And this is basically making turfs **actually** play the sounds they've had. They've had these sounds for... years, I can look up the gitblame, but basically, these have been there for ages. Just that nobody ever actually linked the turfs to their defines, meaning the footstep sounds were never called and were always set to the base_footstep_sounds, which is plating.

PR #7135, opened 2020-05-10
Port Hover Inventory and Body Indicators

Rykka-Stormheart:
The body zone selector now indicates which body part you are about to select when hovered over, and the hover inventory indicator basically shows where you are trying to put an item, and then shows an item ghost in red or green, giving visual feedback as to if that item will go in the slot or not.

Gifs of this in action:
 
Body Selection parts.

 
Inventory Icon Overlays!

Credit goes to @ShadowLarkens for assistance with port.

Port of https://github.com/VOREStation/VOREStation/pull/7748

### DIFF
--- a/code/_onclick/hud/screen_objects.dm
+++ b/code/_onclick/hud/screen_objects.dm
@@ -31,7 +31,36 @@
 
 /obj/screen/inventory
 	var/slot_id	//The indentifier for the slot. It has nothing to do with ID cards.
+	var/list/object_overlays = list() // Required for inventory/screen overlays.
 
+/obj/screen/inventory/MouseEntered()
+	..()
+	add_overlays()
+
+/obj/screen/inventory/MouseExited()
+	..()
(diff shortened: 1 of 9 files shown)


code/_onclick/hud/screen_objects.dm
```diff
@@ -31,7 +31,36 @@
...
+		if(!holding || user.get_equipped_item(slot_id))
+			return
+
+		var/image/item_overlay = image(holding)
+		item_overlay.alpha = 92
+
+		if(!holding.mob_can_equip(user, slot_id, disable_warning = TRUE))
+			item_overlay.color = "#ff0000"
+		else
+			item_overlay.color = "#00ff00"
```
Atermonera, line 60: inb4 long-winded philosophical argument over whether the aesthetic value of `ff0000` going before `00ff00` outweighs the aesthetic damage of inverting the output of the if statement to permit that ordering

PR #7136, opened 2020-05-10
Add AI Picture-in-Picture Mode!

Rykka-Stormheart:
This adds a system for picture-in-picture UI windows using
vis_contents. Essentially, it allows you to make UI windows that show an area of turfs. It also refactors how cameranet visibility works.

Currently, this is implemented on AIs. They gain two new UI buttons - "Enter Multicam Mode", and "Create Multicam". When they go into Multicam Mode, they see a background of animated binary numbers, and they are allowed to create an infinite amount of these picture in picture windows, which subsequently creates an AI Eye for each one. They are able to control each AI eye individually, by first clicking on the PIP window to select it as "active" and then using the normal arrow key controls. The PIP windows can be freely resized and moved around the area.

You can control everything inside these PIP windows EXACTLY the same as you can using a traditional AI Eye, as demonstrated below.

For admins, there is a config option to disable PIP entirely - simply set `var/multicam_allowed = TRUE` to FALSE if you wish to disable it from being used. (Please be reasonable.) <3

You can see an example of how this works here:
https://i.imgur.com/TCND8W9.mp4

AI Multicam functionality.

Do note that if the lightbulb in the AI core is busted, the multicam room is dark(er) than it would be, but you can still see your camera windows just fine. (I'll probably fix this later.) It only affects the "matrix" backdrop, the PIP windows are still fine.

This has been runtime-tested with the latest `master` revision and produces 0 runtimes, and has no noticeable impact on server CPU usage.

This is the Polarisport! Port of https://github.com/VOREStation/VOREStation/pull/7752


code/modules/mob/living/silicon/ai/login.dm
```diff
@@ -1,10 +1,12 @@
-/mob/living/silicon/ai/Login()	//ThisIsDumb(TM) TODO: tidy this up �_� ~Carn
+/mob/living/silicon/ai/Login()	//ThisIsDumb(TM) TODO: tidy this up �_� ~Carn
```
Atermonera, line 1: Wooo character encoding! (Doesn't need to be fixed)

code/modules/mob/mob.dm
```diff
@@ -209,7 +209,7 @@
 			else
 				client.perspective = EYE_PERSPECTIVE
 				client.eye = loc
-	return
+		return 1
```
Atermonera, line 212: TRUE?

code/_onclick/hud/ai.dm
```diff
@@ -1,3 +1,25 @@
+/obj/screen/ai/multicam 
+	name = "Multicamera Mode"
+	icon = 'icons/mob/screen_ai.dmi'
+	icon_state = "multicam"
+ 
+/obj/screen/ai/multicam/Click() 
+    if(..()) 
+        return 
+    var/mob/living/silicon/ai/AI = usr 
+    AI.toggle_multicam()
```
Atermonera, line 10: Are there verbs that can be used interchangeably with these UI buttons?

code/modules/mob/mob.dm
Rykka-Stormheart: What's the difference, functionally?

Am still new to this, sorry! <3

code/_onclick/hud/ai.dm
Rykka-Stormheart: Uh, as far as I'm aware there are no verbs, it's just buttons? I can check, but I don't think there's verbs for it, same with the rest of the AI UI bits.

code/modules/mob/mob.dm
Atermonera: TRUE/FALSE are for values that are treated like booleans, that is, you only care if they're strictly non-0 or 0. Actual numbers are used when you actually use the numeric output, like when returning an error code (the closest we come to that is `mob.stat`), or the value of a particular field. A function `isConscious()` would probably be expected to return TRUE/FALSE, but a function `getHealth()` is more likely to give proper numeric output.
To further the mob.stat example, there's lots of places in code where you care chiefly if a mob is conscious (`!stat == TRUE`), but you can also look at the value of stat numerically to tell whether the mob is unconscious or dead (`stat == UNCONSCIOUS` and `stat == DEAD`, respectively).

code/_onclick/hud/ai.dm
Atermonera: There was a great deal of trouble when kelenius first introduced the ui buttons and simultaneously removed the verbs corresponding to those buttons, and then vehemently refused to add the verbs back. As far as I can tell, making these functions verbs really just involves either setting the procs these buttons are calling as verbs, or otherwise making a verb that calls the same procs in the same manner.

In the interest of maintaining uniform usability, where UI buttons augment what's already available via verbs, I would like verbs that achieve the same effect as the buttons. Some people use trackpads, or are otherwise heavily inconvenienced by trying to use buttons where being able to use their keyboard would be quicker and simpler.

code/modules/mob/mob.dm
Rykka-Stormheart: This is done.

code/_onclick/hud/ai.dm
Rykka-Stormheart: I don't really know how to make verbs for it, I'm sorry.

Rykka-Stormheart:
> And then in `/code/modules/mob/living/silicon/ai/ai.dm`, you need to add the paths of both these procs to `ai_verbs_default`, at the top of the file.

This has been done!

PR #7141, opened 2020-05-11
["Feature"] Crate Breaching w/ Tools

KillianKirilenko:
For a long time, I sat thinking "hey, why can't you cut open secure crates with regular tools?"

So I decided, on a whim, to make it so you can!

This PR adds two vars to secure crates, `can_breach` and `breach_time`. The former prevents you from cutting through the locking mechanism entirely: the latter determines the time taken to breach a crate (in deciseconds). The default time is 10 seconds for most secure crates, or 20 for weapons/gear/phoron. Abandoned crates cannot be cut open.

Welding tools and surgical saws can be used to cut open crates; saws take 50% longer but don't require fuel. Better tools cut faster, as one would expect.

Breaching this way is ugly though; it leaves visible evidence on the crate's examine description, it takes a fair bit of time, and it makes noise when starting. If I could, I'd make it throw in more cutting noises, but I'm not sure how I'd do that exactly.

Using a tool on a crate won't try to cut the lock if it's already open (you'll just put your tool in it instead, like normal) or already unlocked.

### DIFF
--- a/code/game/objects/structures/crates_lockers/crates.dm
+++ b/code/game/objects/structures/crates_lockers/crates.dm
@@ -134,6 +134,8 @@
 	var/emag = "securecrateemag"
 	var/broken = 0
 	var/locked = 1
+	var/can_breach = TRUE	//if false, the crate can't be cut open with welders or saws
+	var/breach_time = 100	//time to cut open, in deciseconds (why the fuck does byond use deciseconds?)
 
 /obj/structure/closet/crate/secure/New()
 	..()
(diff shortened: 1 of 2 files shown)


Greenjoe12345:
Will this work with ID locked  closets too?

KillianKirilenko:
Just crates for now. Closets can be welded shut, so if I was going to add it to them as well it'd require a little extra fiddling I don't feel like messing with right now.

zekesulastin:
... I hate to be a killjoy, but letting cargo open any and every crate with tools from their roundstart autolathe (or tool storage) sounds like an absolute balancing nightmare.

KillianKirilenko:
Oh, I don't doubt it. It was just a passing idea that I wanted to take a crack at coding up with amateur-level skills. If it gets rejected I won't be particularly upset. But I'm also potentially down for tinkering with it further if anyone has good suggestions/input that aren't too feature-creep-y.

Greenjoe12345:
Maybe you have to make a special saw or something? Like, get the default circular saw then modify it to be able to cut open crates. Some method that would make it not possible for cargo to easily get at roundstart.

Arokha:
I was thinking about porting tool mods from Eris, maybe those would apply.

Atermonera:
Right, results from staff meeting, as written this completely invalidates the fact that secure crates are supposed to keep things out of the hands of Urist McCargotech, because the tools involved are printable from an autolathe, which cargotechs have access to.

In theory... having some new way to bust open crates would be interesting, but exceptionally difficult to balance against the existing methods of "acquire gun". Making a lesser material restriction with greater time restriction... just makes it easier than present, but if you make it harder to get the tools for than to get a gun... why would anyone use the new method?

If you're really keen on this idea, you're welcome to continue brainstorming ideas that might meet this balance criteria, but as it stands this PR will not be merged.

KillianKirilenko:
Yeah, I understand the concerns. Like I said, it's entirely possible to just... arbitrarily make certain crate types unbreachable with tools, but I feel like that takes half the point out of it.

Honestly, I got nothin' right now, both in terms of ideas and energy to implement said nonexistant ideas. I'm just gonna close the PR, and if I figure out anything good maybe I'll dust it off again. If anyone else thinks they can make their own better implementation they're welcome to take a shot.

PR #7146, opened 2020-05-11
Small overhaul of floor icon generation

Arokha:
Did this on a map I'm making, thought I'd port it


code/game/turfs/simulated/floor_icon.dm
```diff
@@ -1,4 +1,4 @@
-GLOBAL_LIST_EMPTY(flooring_cache)
+var/list/flooring_cache = list()
```
Atermonera, line 1: Performance, or just a "why would anyone possibly want to try editing something in a cache"?

code/game/turfs/simulated/floor_icon.dm
```diff
@@ -79,11 +78,10 @@ var/image/no_ceiling_image = null
 		icon = 'icons/turf/flooring/plating.dmi'
 		icon_state = "dmg[rand(1,4)]"
 	else if(flooring)
-		var/rand_key = rand(0,2)
 		if(!isnull(broken) && (flooring.flags & TURF_CAN_BREAK))
-			add_overlay(get_flooring_overlay("[flooring.icon_base]-broken-[rand_key]","broken[rand_key]"))
+			add_overlay(get_flooring_overlay("[flooring.icon_base]-broken-[broken]","broken[broken]")) // VOREStation Edit - Eris overlays
```
Atermonera, line 82: Wayward Vorestation edit flags

code/game/turfs/simulated/floor_icon.dm
Arokha: Performance will be the same, though in general I don't see a reason for cache stuff to be in GLOB. I can put it back if preferred.

code/game/turfs/simulated/floor_icon.dm
Arokha: Fixy'd.

PR #7172, opened 2020-05-14
Hybrid tools buffs.

### LINKED ISSUE #7169: Hybrid Tools are worse than useless
Hybrid tools do not work at range in several cases. Including windows and walls, the two main things they're used for.

The radiation makes them uselss unless you're in a 100% radproof suit. There are only 2 of these in the game.

They do not perform the action when they malfunction. This happens often enough that you miss a step and waste time.

When the screwdriver malfunctions, it does not act as a screwdriver when interacting with things like autolathes. This means the Autolathe will eat your super expensive high tech screwdriver.

The most high tech expensive precursotech tools available to be better than basic toolbox stuff.

#### What actually happened
Got mad after I lost my screwdriver and murdered myself.

#### Steps to reproduce
Attempt to use Hybrid tools.

#### Additional info:
 - **Server Revision**: Server revision: master - 2020-05-12
eb8ad5791793b697492e9d6dd6382ac900f369c9

lolman360:
Hybrid tools are generally worse than power tools, and their main advantage, reach, isn't even checked properly half the time. I plan on revisiting the tools refactor and making these work as intended in future, but now this is a solution to make hybrid tools not a really good assassination tool (they irradiate a lot and you can spam-screwdriver/crowbar an object such as a circuit assembly or vendor to irradiate the area a lot.)

Changes:
-Hybrid welder itemsize bulky > normal (so it fits in toolbelts.) However, hybrid welder is still really super bright, as it's a straight upgrade from purple welder otherwise. 
-All other hybrid tools no longer irradiate. They retain their toolspeeds and sizes.

Closes #7169

### DIFF
--- a/code/game/objects/items/weapons/tools/screwdriver.dm
+++ b/code/game/objects/items/weapons/tools/screwdriver.dm
@@ -107,14 +107,6 @@
 	random_color = FALSE
 	reach = 2
 
-/obj/item/weapon/tool/screwdriver/hybrid/is_screwdriver()
-	if(prob(10))
-		var/turf/T = get_turf(src)
-		SSradiation.radiate(get_turf(src), 5)
-		T.visible_message("<span class='alien'>\The [src] shudders!</span>")
-		return FALSE
-	return TRUE
-
 /obj/item/weapon/tool/screwdriver/cyborg
 	name = "powered screwdriver"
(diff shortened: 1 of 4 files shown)


FenodyreeAv:
Hero! \o/

PR #7173, opened 2020-05-15
Refactors obj/item/weapon and /obj/item/device to /obj/item

lolman360:
Title. There's one minor functional difference between obj/item and /item/weapon, and that's a single proc which I've moved to obj/item. There is literally no difference between obj/item and obj/item/device. 

This will screw up persistence systems that rely on typepaths (if we have them, I know downstreams do.)

inspired by https://github.com/Citadel-Station-13/Citadel-Station-13-RP/pull/1811

Changes:
folds /obj/item/device and /obj/item/weapon into /obj/item. There is only /obj/item now.


Atermonera:
Ok so there's a few reasons why this won't be merged. Gonna put that out there straight and simple.

1. I am a single person, who would be expected to read twenty three _thousand_ lines. And I make it a point to read every single line of every single pull request that I merge, whether I'm just glancing at them to make sure the code looks valid, or I'm actually expending the effort to fully understand what the code is trying to do. That in itself would necessitate _weeks_ of effort, as it took for some of the other major pull requests, like simple mobs, during which we would either instate a code freeze to prevent conflicts from accumulating, or literally every single pull request we merge would generate conflicts.
2. As I stated on discord, the _mechanical_ differences between `/obj/item/weapon` and `/obj/item/device` have largely been lost to time. That said, the _semantic_ differences are still entirely relevant. Something that is a device generally has some complex operation, and something that is a weapon generally is used primarily for hitting things. I would actually be much more amenable to further expanding the paths to better encapsulate the different types of actions various things can be used for (steel sheets and other building materials aren't so much weapons as things used to _make_ weapons, among other things).
3. The semantic difference between the different categories is actually incredibly helpful for mapping. When you edit a map in dreammaker, the IDE generates an object tree, which is a hierarchical representation of the paths of all atoms that could be mapped in. At present, that tree is actually very messy _because_ there are relatively few category/template classes, and mostly functional ones. A template class here describes a specific path whose sole function is to exist as a single node in the tree that can be expanded to reveal subtypes that you're actually interested in mapping in. `/obj/item/device` and `/obj/item/weapon` are both template classes of this sort. To remove them would actually make mapping much more difficult and tedious, because _all_ `/weapon`s and `/device`s would be moved into the same level of the tree, so when you expand `/obj/item`, you'd be presented with an enormous array of things to sift through to find the particular thing you're actually trying to map in.

PR #7176, opened 2020-05-16
Custom RIGs function properly.

Mechoid:
Debug icons for custom item kits. RIG kits actually work again, how did I fuck that up so badly?

Adds a debugging RIG kit, and missing dummy custom obj dmi. All present custom DMIs will be overwritten when updating, by the proper ones on the separate repo.


Mechoid:
Medium priority'd since 2 players have made applications for RIG kits.

PR #7182, opened 2020-05-16
Add serial comma

Arokha:
Use of the comma is consistent with the conventional practice of the region.
It matches the spoken cadence of sentences better.
It can resolve ambiguity.
Its use is consistent with other means of separating items in a list (for example, when semicolons are used to separate items, one is always included before the last item).
Its omission may suggest a stronger connection between the last two items in a series than actually exists.
Its use can "prevent any misreading that the last item is part of the preceding one".

I eagerly await my bikeshed label.

### DIFF
--- a/code/_helpers/_lists.dm
+++ b/code/_helpers/_lists.dm
@@ -16,7 +16,7 @@
  */
 
 //Returns a list in plain english as a string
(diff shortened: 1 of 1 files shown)


Neerti:
>Not calling it the Oxford Comma.

Arokha:
Serial comma is the *TeChNiCaL NaMe*

Atermonera:
You say bikeshed like the few people here that actually care about grammar like being wrong. :V

PR #7184, opened 2020-05-17
adds some coats for teshari + fancy clothes

Greenjoe12345:
one white and recolorable, one black with recolorable stripes.

adds some fancy tesh clothes to the loadout that were previously just icons (suitable for Tesh HOPs (and captains downstream I guess)


KasparoVy:
Those coats look sick!

PR #7185, opened 2020-05-17
New implanter and camera sprites

Arokha:
Now it actually looks like a polaroid camera


Atermonera:
Was `needle_thread` moved somewhere else?

Arokha:
No, and that's super weird that our items.dmi doesn't have that, yet hasn't merge conflicted. :thinking:

Well, fixed.

PR #7189, opened 2020-05-17
New ore chunk sprites, ore box sprite

Arokha:
The items, like on the ground, after being mined, not the ones shown on the rocks before you mine. These have shading, and suffer less from the 'coal is a solid black blob' effect.


MistakeNot4892:
I don't know how relevant this is generally to your setup but uranium does not come out of the ground glowing green.

Arokha:
Uranite is just grey, yeh. Pretty much all the real versions of the ingame ores run the gamut of "grey" to "light grey" to "kind of reddish grey", with the exception of gold. Though this is a videogame, so. Having things be identifiable at a glance is cool and I think that trumps realism in this case.

PR #7190, opened 2020-05-17
New janicart sprites/features

Arokha:
Separates bucket from cart
Drag bucket onto cart to put it on
Alt-click with mop to attach mob
Alt-click with reagent container to dump into bucket without putting reagent container into trash


Atermonera:
Please add the new usage to the extended description.

Arokha:
Added description info

PR #7194, opened 2020-05-19
Stops hiding italicised messages from observers

Cyantime:


TheFurryFeline:
So THIS was why I could never hear whispers anymore. Good grief!

code/modules/mob/hear_say.dm
```diff
@@ -91,8 +91,6 @@
 
 	var/track = null
 	if(isobserver(src))
-		if(italics && is_preference_enabled(/datum/client_preference/ghost_radio))
```
Atermonera, line 94: Have you tested this, and does it show ghosts both the radio message and the whispered message like this preference was intended to avoid?

Cyantime: I've tested it, and it does show both. I don't consider this a downside or problem.

PR #7197, opened 2020-05-19
Increases automute detection interval to 1s

Atermonera:
This is intended to catch the recent bout of ooc spammers  that've been touring the ss13 community trying to smear the World server the past few weeks, Unfortunately, the system appears to be in working order, which means they were posting slow enough to avoid it. We do not at present have any better ideas to combat such an issue, and cultists will need to be careful not to mute themselves by clicking on runes too frequently.

But for the fact that cultists still have to click on 10 different runes, I'd consider caching recent messages to try and catch people repeating themselves several times, but as it stands that seems most likely to catch cultists.

### DIFF
--- a/code/modules/admin/verbs/randomverbs.dm
+++ b/code/modules/admin/verbs/randomverbs.dm
@@ -164,6 +164,7 @@
 proc/cmd_admin_mute(mob/M as mob, mute_type, automute = 0)
 	if(automute)
 		if(!config.automute_on)
+			to_world("automute off")
 			return
 	else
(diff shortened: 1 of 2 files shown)


MistakeNot4892:
You could possibly have cult speech suspend the mute checking by throwing a flag onto the speaker for the duration of the speech.

Atermonera:
Been a week, haven't seen any signs of the spammers so I guess cassie's fix was sufficient deterrent//they ran out of accounts

PR #7204, opened 2020-05-19
Performance improvements

Arokha:
I'd suggest reviewing this one commit at a time, starting with oldest.

- Converts unzooming (which was previously done by having every item check to see if they're zoomed every time the mob containing them moved) to just be a listener that's added on zoom.
- Optimize the get_mobs_in_radio_range proc to be more performant
- Remove the old PROXMOVE flag, which was made in The Time Before
- Remove the on_loc_moved proc entirely, replace with stateful things
- Refactor space moves out of turf/Entered. Ghost flying around don't need to be triggering this stuff.
- Move openspace updates to listeners that openspace turfs apply below them.


Arokha:
Travis is done but forgot to tell github. Maybe restart just one of the latter two builds?

code/modules/mob/living/living.dm
```diff
@@ -774,26 +774,54 @@ default behaviour is:
...
+		var/locthen = loc
+		spawn(5)
+			if(!anchored && !pulledby && loc == locthen)
+				var/stepdir = inertia_dir ? inertia_dir : last_move
+				if(!stepdir)
+					return
+				var/turf/T = get_step(src, stepdir)
+				if(!T)
+					return
+				Move(T, stepdir, 5)
```
Atermonera, line 819: Will this end up recursively calling inertial_drift() until it stops drifting?

Arokha: Well, I'm not sure I'd consider it 'recursively'. I guess there's a chain of things that happens that does come back to here until they're done drifting.
Move() into a turf that qualifies for inertial drift -> interial_drift() check to be sure, then 5 seconds later spawn a Move() -> check again to see if it still qualifies for drifting, repeat

S'the same as it was before, pretty much.

PR #7208, opened 2020-05-19
Added wooden bucket inhands and let buckets be equipped to the head.

MistakeNot4892:


MistakeNot4892:
Incompatible with existing helmet icons, oops.

PR #7211, opened 2020-05-20
Talking, and other bugs

Arokha:
There were a couple bugs I noticed while testing a species downstream.
- Brains (like the mob held inside posibrains) couldn't speak because it would runtime as they had no default language
- No speech bubbles for brains (you can't see their mobs, so that does make sense and isn't really a bug so much as is just lame)

This fixes that and tries to make it more robust to runtimes by at least giving people gibberish talk as a failsafe rather than making their language null.

Also adds speech bubbles when hiding in things, with -50 alpha per layer you're hidden in.

### DIFF
--- a/code/modules/mob/language/language.dm
+++ b/code/modules/mob/language/language.dm
@@ -270,7 +270,7 @@
 			if (species_language)
 				set_default_language(GLOB.all_languages[species_language])
 			else
-				set_default_language(null)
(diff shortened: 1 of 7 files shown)


Atermonera:
travis

Atermonera:
Travis got drunk and fell asleep on the subway en route to informing github it's done

PR #7214, opened 2020-05-21
Adds vchat export chatlog verb

Atermonera:
Verb is located in OOC tab, should function exactly as the export chatlog button in vchat's settings page.
Verb is added when vchat successfully loads, removed when it's reloaded.
`to_world()` tweaked to save messages to all clients, so stuff like vote messages are saved correctly.
Changing the buffer size (how many messages are visible) sends a message to byond to update a variable in the chatOutput datum, so when you reconnect, it'll only dump as many messages as will fill up your buffer, rather than dumping all of them (30+s delay for people who've been active on long rounds). The export chatlog function in vchat's settings page is still implemented because I'm lazy, but it can no longer be expected to gather all messages, only those since you first connected. The verb is now required to get an entire round's connection.

The temp file that's generated to compile the css and messages presently tries to be deleted after 10 seconds, and 1 minute after that. The first delay was found to be necessary in some measure through testing because ftp is asynchronous, and while the client decides where and how to save the log, it'll move on to trying to delete the temp, only to discover it's locked as in-use by the OS (Or at least I presume that's what's happening).

- [ ] Switch to using callbacks to attempt to delete the file until it successfully deletes?

[tested](https://puu.sh/FMVRk/bfafca8fed.mp4) log export works. Buffer length tested and confirmed to also update correctly on the byond side.

### DIFF
--- a/code/modules/client/client procs.dm
+++ b/code/modules/client/client procs.dm
@@ -405,18 +405,20 @@ client/verb/character_setup()
 	if(src.chatOutputLoadedAt > (world.time - 10 SECONDS))
 		alert(src, "You can only try to reload VChat every 10 seconds at most.")
 		return
-	
+
+	verbs -= /client/proc/vchat_export_log
+
 	//Log, disable
 	log_debug("[key_name(src)] reloaded VChat.")
 	winset(src, null, "outputwindow.htmloutput.is-visible=false;outputwindow.oldoutput.is-visible=false;outputwindow.chatloadlabel.is-visible=true")
-	
+
 	//The hard way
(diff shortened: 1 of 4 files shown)


Cyantime:
This introduces some real issues with the logging system. Since it's a byond verb, it's totally nonfunctional if the server isn't alive. If the round restarts during a play session, it won't save anything from the previous round. In the event of a person reconnecting during a round and the server crashing after, only the settings menu save will work, and because the round history isn't sent to the client buffer any more, everything before the player reconnecting is permanently lost.

PR #7215, opened 2020-05-21
Streamline (Facial) Hair & Marking Character Customization, Port color_square to Clean Prefcode

KasparoVy:
Further improves on #5429

Adds the ability to quickly cycle through previous or next facial or hair styles. The buttons are behind the style so you don't have to move your mouse because the name of one style is longer than the other.

If you click forward while on the last style, it'll wrap around to the first one.
If you click backward while on the first style, it'll wrap around to the last one.

For markings, you can now pick a marking and inject it ahead of another so long as it isn't the one you picked, or the one directly behind in the list order.
Additionally this cleans up prefcode by aligning markings in a table, as well as ports `color_square()` from Paradise to which you can pass r,g,b values or a single hex string to generate an inline colour square. This was necessary to correct an alignment issue with in-line tables nested within table rows.
e.g. `"[color_square(red, blue, green)]"` or `"[color_square(hex = hex_color)]"`

polaris ss13 marking and facial hair customization QOL update 20200530

### DIFF
--- a/code/modules/client/preference_setup/general/03_body.dm
+++ b/code/modules/client/preference_setup/general/03_body.dm
@@ -321,33 +321,33 @@ var/global/list/valid_bloodtypes = list("A+", "A-", "B+", "B-", "AB+", "AB-", "O
 
 	. += "<b>Hair</b><br>"
 	if(has_flag(mob_species, HAS_HAIR_COLOR))
-		. += "<a href='?src=\ref[src];hair_color=1'>Change Color</a> <font face='fixedsys' size='3' color='#[num2hex(pref.r_hair, 2)][num2hex(pref.g_hair, 2)][num2hex(pref.b_hair, 2)]'><table style='display:inline;' bgcolor='#[num2hex(pref.r_hair, 2)][num2hex(pref.g_hair, 2)][num2hex(pref.b_hair, 2)]'><tr><td>__</td></tr></table></font> "
+		. += "<a href='?src=\ref[src];hair_color=1'>Change Color</a> [color_square(pref.r_hair, pref.g_hair, pref.b_hair)] "
 	. += " Style: <a href='?src=\ref[src];hair_style_left=[pref.h_style]'><</a> <a href='?src=\ref[src];hair_style_right=[pref.h_style]''>></a> <a href='?src=\ref[src];hair_style=1'>[pref.h_style]</a><br>"
 
 	. += "<br><b>Facial</b><br>"
 	if(has_flag(mob_species, HAS_HAIR_COLOR))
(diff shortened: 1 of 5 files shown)


KasparoVy:
@Atermonera Thanks for the critique on #5429! Is there a symbol you would prefer in place of 'mv' (for move) when changing marking location?

MistakeNot4892:
Could I suggest adding some table tags to the markings list? With this many buttons it's beginning to look strange when misaligned.

KasparoVy:
Good thinking, thanks Zuh

Atermonera:
;tldr I think "mv" is fine.

In theory you could use `#` to represent the "assign position number", but since the positions aren't numbered, (and thinking about this system in such a manner isn't intuitive even for me, rearranging it casts it more as a linked list than an array), I think leaving it as `mv` is an appropriate abbreviation that clearly and concisely implies the function of the button, to move the connected marking's position within the list. Or at least, I can't see any characters on my keyboard that more emphatically say "adjust the numerical position by displacement _n_" without rooting through unicode, where the up/down carats are comparably intuitive for the function of "adjust position by 1".

code/modules/client/preference_setup/general/03_body.dm
```diff
@@ -322,12 +322,12 @@ var/global/list/valid_bloodtypes = list("A+", "A-", "B+", "B-", "AB+", "AB-", "O
 	. += "<b>Hair</b><br>"
 	if(has_flag(mob_species, HAS_HAIR_COLOR))
 		. += "<a href='?src=\ref[src];hair_color=1'>Change Color</a> <font face='fixedsys' size='3' color='#[num2hex(pref.r_hair, 2)][num2hex(pref.g_hair, 2)][num2hex(pref.b_hair, 2)]'><table style='display:inline;' bgcolor='#[num2hex(pref.r_hair, 2)][num2hex(pref.g_hair, 2)][num2hex(pref.b_hair, 2)]'><tr><td>__</td></tr></table></font> "
-	. += " Style: <a href='?src=\ref[src];hair_style=1'>[pref.h_style]</a><br>"
+	. += " Style: <a href='?src=\ref[src];hair_style_left=[pref.h_style]'><</a> <a href='?src=\ref[src];hair_style_right=[pref.h_style]''>></a> <a href='?src=\ref[src];hair_style=1'>[pref.h_style]</a><br>"
```
Atermonera, line 325: It might be good to leave a comment on the function of the extra `<`/`>` characters here in case they confuse other people, it took me a minute to realize those're the link's text rather than misplaced tag opening/closings

KasparoVy:
Thanks @Atermonera 
I've figured out the table business & ported color_square from Paradise and it looks fine. Just implementing it all over preferences & committing w/ screenshots

KasparoVy:
Alright we good works like a charm

KasparoVy:
Good to go unless any further requests, thanks for the review guys

code/modules/client/preference_setup/general/03_body.dm
KasparoVy: Good thinking

KasparoVy:
There we are, all commented and aligned, thanks for the review guys

code/modules/client/preference_setup/general/03_body.dm
KasparoVy: Good call, thanks -- oversight on my part

MistakeNot4892:
That looks nice, cheers.

PR #7217, opened 2020-05-21
Shapeshifter color selection knows default/current

Arokha:
QOL for shapeshifters with color choice


Arokha:
Apparently this usage of num2text was added in 513.1514, which is 000.0001 versions higher than current Travis. What's your server run?

Atermonera:
We're on 1513 as well. What're you folks running on?

Arokha:
Ye, we moved to 513.1520. Seems to work fine.

Shadow-Quill:
Stale PR @Arokha

PR #7219, opened 2020-05-21
Fix sinks washing things at a distance

Arokha:
Sleep replaced with do_after

### DIFF
--- a/code/game/objects/structures/watercloset.dm
+++ b/code/game/objects/structures/watercloset.dm
@@ -387,10 +387,11 @@
 	playsound(loc, 'sound/effects/sink_long.ogg', 75, 1)
 
 	busy = 1
-	sleep(40)
+	if(!do_after(user, 40, src))
+		busy = 0
+		to_chat(usr, "<span class='notice'>You stop washing your hands.</span>")
+		return
 	busy = 0
-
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/watercloset.dm
```diff
@@ -443,13 +444,12 @@
...
-	sleep(40)
+	if(!do_after(user, 40, src))
+		busy = 0
+		to_chat(usr, "<span class='notice'>You stop washing \the [I].</span>")
+		return
 	busy = 0
 
-	if(user.loc != location) return				//User has moved
-	if(!I) return 								//Item's been destroyed while washing
-	if(user.get_active_hand() != I) return		//Person has switched hands or the item in their hands
```
MistakeNot4892, line 451: This seems like an unnecessary removal.

Atermonera: Those're all handled by the do_after(), if the state changes between the start and end of it, it'll fail, hitting the to_chat()/return above

Arokha: ^

MistakeNot4892: Realized I was thinking of old surgery where you can swap hands, don't mind me.

PR #7225, opened 2020-05-22
Increase the amounts of SNACKS and cigarettes in vending machines.

lbnesquik:
Essentially double the number of items of vice you can purchase from those.
Besides, we don't have vending machine refills in the code unlike TG.

Upsides: No more running out of snacks at the very popular machines.
Downsides: None because it isn't a balance issue.


Atermonera:
I'm... pretty sure we have the code for it, possibly just no supply packs to get refills.

PR #7237, opened 2020-05-24
Fixes runtimes related to mmi_holder and defib_timer

Atermonera:
Fixes runtime in life.dm,1831: undefined proc or verb /obj/item/organ/internal/mmi_holder/robot/tick defib timer().
Fixes runtime in defib.dm,511: undefined variable /obj/item/organ/internal/mmi_holder/robot/var/defib_timer


polaris.dme
```diff
@@ -1420,8 +1420,8 @@
 #include "code\modules\ai\ai_holder_targeting.dm"
 #include "code\modules\ai\interfaces.dm"
 #include "code\modules\ai\say_list.dm"
-#include "code\modules\ai\ai_holder_subtypes\simple_mob_ai.dm"
-#include "code\modules\ai\ai_holder_subtypes\slime_xenobio_ai.dm"
+#include "code\modules\ai\aI_holder_subtypes\simple_mob_ai.dm"
```
Mechoid, line 1423: This won't explodify folders right?

TheFurryFeline:
This reminds me of an odd bug where you can be ghosting without having spawned in for a shift, when suddenly your screen fills with the purple(?) text of being revived when a doc is defibbing someone. Does this fix that or is that thing long since resolved?

polaris.dme
Atermonera: Apparently the folders are already named with a capital I?
evidence

PR #7240, opened 2020-05-25
Fix doublewhispering

Cyantime:
Fixes ghosts being given two messages when in view of someone who whispers.


Atermonera:
_glances at #7194 _

PR #7241, opened 2020-05-25
Adds proc to replace default cells with high cap on machinery

### LINKED ISSUE #7157: Machine Power Cells Nerfed
Machines that have power cells as part of their components have been nerfed from a default high-capacity power cell to a regular power cell (10k to 1k). This makes cyborg rechargers awful for even cyborgs without upgraded power cells and makes replacing the mining drill head cells a requirement before using them.
To have the base-line standard high-capacity power cell.
#### What actually happened
The mining drill head shut off very quickly after using up the small amount of power that a regular power cells have. I was also unable to fully charge my default (7500 power rating) cyborg cell without running the cyborg charger out of power.
#### Steps to reproduce
1a.) Use mining drill
1b.) Be cyborg and recharge
#### Additional info:
 - **Server Revision**: master - 2020-05-12 eb8ad5791793b697492e9d6dd6382ac900f369c9

Atermonera:
The usage of `component_parts` to contain the cell is something that I think should be expanded to include all other machinery, because the cell in such cases is a component part. Unfortunately, I don't have the energy to go through all 346 instances of `/obj/item/weapon/cell` again to figure out where else I should refactor that usage.
Fixes #7157 without requiring map changes

### DIFF
--- a/code/game/machinery/machinery.dm
+++ b/code/game/machinery/machinery.dm
@@ -334,6 +334,14 @@ Class Procs:
 	CB.apply_default_parts(src)
 	RefreshParts()
 
+/obj/machinery/proc/default_use_hicell()
+	var/obj/item/weapon/cell/C = locate(C) in component_parts
+	if(C)
+		component_parts -= C
+		qdel(C)
+		C = new /obj/item/weapon/cell/high(src)
+		component_parts += C
+
 /obj/machinery/proc/default_part_replacement(var/mob/user, var/obj/item/weapon/storage/part_replacer/R)
 	if(!istype(R))
 		return 0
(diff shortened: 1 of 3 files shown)


TheFurryFeline:
This will render Aro's #7216 unnecessary.

Atermonera:
Yes. In my mind, this is a better fix, since it extends to other machines that have cells as components (at present just borg rechargers) more readily, and doesn't require map changes to correct such issues (which, because the map merger has a known habit of eating changes and being unreliable, could present delays to other people planning map changes.

TheFurryFeline:
It does look better to me, as well. I was saying it just for the record about whether there was another PR open to fix issues with machines and power cells.

Atermonera:
The proc is also placed such that it can be easily overridden for machines that require something else

PR #7249, opened 2020-05-28
Species Vars Extension

KillianKirilenko:
"Inspired" by some downstream work, this extends the species-level vars and adds a couple of new features that can be used for creating new species and traits.

New vars as follows:
`chem_strength_heal` - multiplies the effects of the vast majority of healing-related reagents created through chemistry
`chem_strength_pain` - multiplies the effects (duration?) of painkillers
`chem_strength_tox` - multiplies the effects and threshold values of toxins
`chemOD_threshold` - multiplies the overdose threshold for reagents; lower = easier overdosing. pairs with the existing chemOD_mod
`alcohol_mod` - multiplies alcohol's effective strength when processing
`emp_sensitivity` - enables vulnerability to EMPs for non-synths; bitflag based
`emp_dmg_mod` - Multiplier to all EMP damage sustained by the mob, if it's EMP-sensitive. applies *after* severity adjustments.
`emp_stun_mod` - Multiplier to all EMP disorient/etc. sustained by the mob, if it's EMP-sensitive. applies *after* severity adjustments.

Example use cases include;
Alcohol/painkiller tolerance traits
Drug burnout/tolerance traits
EMP sensitivity for races with 'exotic' senses or capabilities that extend into the electromagnetic spectrum (dionaea, vox?)

Sensitivity settings include the following effects;
- `EMP_PAIN`: inflicts a variable amount of pain, on par with or exceeding a shotgun beanbag round at high intensity
- `EMP_BLIND`: blinds the victim for a short duration, even if they have flash-proof eyewear
- `EMP_DEAFEN`: deafens the victim, up to the same duration as a flashbang at point blank, incurs ear damage as well
- `EMP_CONFUSE`: confuses the victim, causing their inputs to be scrambled.
- `EMP_WEAKEN`: causes the victim to collapse, briefly, dropping whatever they were holding.
- `EMP_[TYPE]_DMG`: where TYPE = BRUTE/BURN/TOX/OXY: causes the specified damage type
All effects scale with the EMP's `severity`; a high severity inflicts more pain, blinds/deafens/etc. for longer, and causes more damage.
- _Any_ level of sensitivity will also cause emote reactions, with the chance of a reaction increasing with severity.

No values/effects are final, input and thoughts appreciated.

I haven't tested what emp_sensitivity does on mobs that also have prosthetics yet (I'll do some VV-testing soon) but I assume that no naturally-emp-sensitive critters would also be capable of having prosthetics so ideally it shouldn't be a problem. Basic testing shows that the sensitivity itself works fine.

Also, I should emphasise; this is purely a set of vars that can be hooked into. It incorporates no changes to existing species.


Greenjoe12345:
Could be interesting applying this to Teshari, since they're apprantly meant to be lightweights when it comes to drinking alcohol

KillianKirilenko:
..not sure what the fuck is going on with the carbon.dm diffs. 🤔

Atermonera:
Appears to be conflicting with the changes to handcuffing from the screen alerts pr

KillianKirilenko:
Right, ought to be settled now. I think I must've forgotten to update my branch to master before porting the stuff across and that made git super mad for whatever reason.

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -96,6 +96,59 @@
...
 
+//EMP vulnerability for non-synth carbons. could be useful for diona, vox, or others
+//the species' emp_sensitivity var needs to be greater than 0 for this to proc, and it defaults to 0 - shouldn't stack with prosthetics/fbps in most cases
+//higher sensitivity values incur additional effects, starting with confusion/blinding/knockdown and ending with increasing amounts of damage
+//the degree of damage and duration of effects can be tweaked up or down based on the species emp_mod and emp_stun_mod vars (default 1) on top of tuning the random ranges
+/mob/living/carbon/emp_act(severity)
+	if(species.emp_sensitivity > 0) //receive basic effects regardless of sensitivity, if we're even the tiniest bit sensitive
+		to_chat(src, "<span class='danger'><B>*BZZZT*</B></span>")
+		switch(severity)
+			if(1) //highest intensity, absolutely sucks
```
Atermonera, line 107: magic numbers should be #defines, for clarity
And possibly use bitflags since these seem to control different factors.

code/modules/reagents/Chemistry-Reagents.dm
```diff
@@ -125,7 +125,7 @@
 				affect_ingest(M, alien, removed * ingest_abs_mult)
 			if(CHEM_TOUCH)
 				affect_touch(M, alien, removed)
-	if(overdose && (volume > overdose) && (active_metab.metabolism_class != CHEM_TOUCH && !can_overdose_touch))
+	if(overdose && (volume > overdose * M.species.chemOD_threshold) && (active_metab.metabolism_class != CHEM_TOUCH && !can_overdose_touch))
```
Atermonera, line 128: What if species is null? All of these changes to reagents can runtime, because you're just given a carbon mob, without any guarantee as to the value of its variables

KillianKirilenko: Only cases of null species I can see are;
1) brains - should be fixed by having them inherit species from the mob they're spawned in, or else default to random/human
2) xeno larvae - all other xenos have their own species, so larvae should too

code/modules/mob/living/carbon/carbon.dm
KillianKirilenko: I'd contemplated bitflags or something like how species flags are handled (such as NO_POISON/NO_SCAN), but they're not something I've really messed with before so I was hesitant to start messing with things I didn't understand.

As for defines; I assume you're referring to the damage values specifically, and possibly effect durations?

Atermonera: The numbers you're switching on. They don't mean anything to anyone else if nothing is written down (hence the phrase "magic numbers", their meaning is indecipherable i.e., magical, on account of the possibilities that the switching variable is computed, rather than stored somewhere, or that the selfsame variable is used elsewhere to different effect), so they're particularly likely to be misused by others who don't understand what precisely they mean. If you were to replace those with #defines, then you can elaborate on the meaning of the numbers in the name of the define, such that someone else can glance at the code and realize, "Ok, they're switching between these different modes of operation for this specific thing". Actually properly reading what it's switching on, this would work a lot like the event severity defines (`EVENT_LEVEL_MUNDANE`, `EVENT_LEVEL_MODERATE `, and `EVENT_LEVEL_MAJOR`), which pretty clearly tell you what the value is in reference to the involved systems. It's an event, and its severity is of some relative measurement.

Bitflags are pretty straightforward, so long as you remember to use logical AND/OR (&/|) instead of boolean (or short-circuited) AND/OR (&&/||). Basically, in a 32-bit integer variable (byond is notably 24-bit), you have 32 individual bits that can be either TRUE or FALSE. If you assign a particular meaning to each individual bit, then you can condense a set of up to 32 toggleable things into a single int, instead of requiring 32 separate boolean variables to represent each individual toggle, or trying to use complex switching logic to encapsulate various combinations of the different toggles.

A simple example of why they're useful is in this file, where you have a few if-else if blocks, where you've got something like `if(species.emp_sensistivity == 3 || species.emp_sensitivity >= 5) [toxloss]`. That there's multiple integer values that represent "You take toxloss from emp" is a consequence of trying to store the different states in consecutive integers instead of as bitflags, where you'd instead get something on the order of `if(species.emp_sensitivity & EMP_GIVES_TOXLOSS) [toxloss]`. I feel I should again emphasize the single &, which is performing a logical AND of each individual bit (`0b10 & 0b11 = 0b10`) rather than performing the boolean and of "X is non-zero AND Y is non-zero". Especially when these internal values are stored as magic numbers, which you only explain in a comment next to the declaration of the base variable, rather than defines, this logic is pretty hard for someone else to understand what exactly the numbers mean, especially when you have to use the complex behaviour of "these multiple sets of possible values each trigger this one thing", as opposed to "if this bit is set, do this this"

code/modules/reagents/Chemistry-Reagents.dm
Atermonera: My point is rather that you shouldn't assume something will be non-null because errors can occur, whether by coders or by admins messing with things they don't fully understand, so if it's possible to prevent a runtime by spotting it before the code is run, then it's generally a good idea to include some sort of check to nip the issue in the bud. In this case, I'm pretty sure just using the ?. operator on species (so, `M?.species.chemOD_threshold`) would prevent it from trying to index a variable on species if it's null, thereby preventing the runtime.

code/modules/mob/living/carbon/carbon.dm
KillianKirilenko: Yeah, it was a bit of a patch-fix when I wanted to make it a bit more flexible instead of a linear progression. The whole thing is honestly kind of messy, but I'm not sure how better to handle it unless I invert the order in a tidier fashion, such as checking weaknesses *then* applying effects multiplied by some value based off the emp severity. 🤔

Guess I'll start digging and see what I can whip up!

code/modules/reagents/Chemistry-Reagents.dm
KillianKirilenko: Hadn't heard of that trick before, so I'll work it in and see what BYOND thinks.

What about the similar uses of M.species on lines 74+75 where it isn't being shunted along to the /human/ subtype?

code/modules/mob/living/carbon/carbon.dm
KillianKirilenko: ..hm, you know, that might actually work. EMP intensity working in *reverse order* means that I think I can subtract the severity from a base duration/damage and then round up to zero if the result would be negative.

One thing at a time though.

code/modules/reagents/Chemistry-Reagents.dm
Atermonera: Those're also a risk of being runtimes, that I've never seen them runtime is in part a testament to how reliably species is non-null. Still bad practice, but since it's not directly related to your work or a procedural shift we're trying to implement on a large scale (like to_chat() used to be, or enforcing absolute pathing) I don't feel a need to demand that it be fixed.

KillianKirilenko: Fixed my instance to `M?` and left the others intact, then. 👌

code/modules/mob/living/carbon/carbon.dm
KillianKirilenko: Refactored the entire thing to integrate severity in a cleaner fashion, and introduced a little more randomness to tiers. Also added support for pain (how did I overlook that?!), seperated out the stun types, and the other two damage types.

Basically the goal I'm going for here is to make this as flexible as possible, to support whatever Fun Stuff folks want to do with it.

KillianKirilenko:
Updated initial message with a bunch more details on what you can do if/when this is merged.

Would appreciate thoughts on effect duration/balance/etc. at this point, I think I've covered the major bases of possible effects.

Issue #7259, opened 2020-05-28
missing rotated stuff


Greenjoe12345:
Here's some things I noticed were not rotated in @Cerebulon 's rotating PR

the HOP console should be facing where Ian is sat, as that's where it would be used from, I imagine
![image](https://user-images.githubusercontent.com/33647525/83204419-b2087180-a143-11ea-8ad0-e299def1c411.png)

the brig garden vending machines
![image](https://user-images.githubusercontent.com/33647525/83204448-c187ba80-a143-11ea-8252-cf0046d44a90.png)

and the water cooler in the brig
![image](https://user-images.githubusercontent.com/33647525/83204707-67d3c000-a144-11ea-957e-5b62ed1685e8.png)

the console in the brig air control room
![image](https://user-images.githubusercontent.com/33647525/83204466-d06e6d00-a143-11ea-907d-65a0ac614d90.png)

the console frame in hanger 2 control room
![image](https://user-images.githubusercontent.com/33647525/83204481-de23f280-a143-11ea-92a8-e2d2b56ae032.png)

the console in the large first aid room on the first deck, central ring
![image](https://user-images.githubusercontent.com/33647525/83204521-f562e000-a143-11ea-9b6a-283370f4605f.png)

the cloning computer? if that's not bugged like teleporters
![image](https://user-images.githubusercontent.com/33647525/83204568-0d3a6400-a144-11ea-8912-0fd7fa76a091.png)

soft drink dispenser in the engi breakroom
![image](https://user-images.githubusercontent.com/33647525/83205016-458e7200-a145-11ea-946e-6ba554baa8c3.png)

drinks dispensers  (the tap things in the bar and cafe) also seem to lack the ability to be rotated via rightclick like the vending machines can

PR #7261, opened 2020-05-29
Another Lesser Rotation

Cerebulon:
Fixes for #7259 
Rotates missed crap, allows chem dispensers to be rotated when unfastened since they have directionals and constructing them might fuck up otherwise.
Can't fix the cloning computer, it's coded by cthulhu.


Greenjoe12345:
there's another thing, the soda dispenser in the security breakroom
image

PR #7152, opened 2020-05-13
[Project] Vintage Voidsuits

KillianKirilenko:
Rifling through the files the other day, I found the Rugged Zaddat shroud, and was like "man, this is a really neat design", and it's apparently based on old human voidsuits? So here we go! A set of refurbished voidsuits for use in events or space exploration POIs.

Five variants (crew, medic, marine, engineer, officer), four* species, four facings, two UI states. That makes 160 states.
*Well, four for you guys. Five for downstreams! ~~and 40 more states. RIP my wrist.~~

talonsuits_v2

- [x] Items & Cycler Entries coded
- [x] Human UI sprites imported
- [x] Human Mob sprites imported
- [x] Skrell UI sprites imported
- [x] Skrell Mob sprites imported
- [x] Tajaran UI sprites imported
- [x] Tajaran Mob sprites imported
- [x] Unathi UI sprites imported
- [x] Unathi Mob sprites imported
- [x] Final testing
- [x] Final polish & item coding

PR #7266, opened 2020-05-31
Vintage Voidsuits Expansion

KillianKirilenko:
Expands on #7152 by adding three new suit variants (and an alt helmet variant for one of those suit variants). Demogif;

talonsuits-tier2v3

Pictured are the new Pilot (Royal Blue, *a la* NASA flight suits), Scientist (Purple), and Mercenary (Red). The pilot is ultralight with poor protection, the scientist has awful physical defense but is the only non-engineer model with 100% rad resist, and the mercenary variant is almost as good as crimson/syndicate voidsuits but incurs additional slowdown.

The Medical and Research variants can also use the alternate "bubble" helmet style or the standard "closed" style, and several more cyclers have been added (plus the base cycler has been tweaked) for ease of use.

Preview gif now includes all states! Woo.

Also expands the range of items that can be put into suit storages for a few of them, and cleans up a couple of duplicate icons. Finally, it removes the vintage options from emagged cyclers (due to lack of tesh sprites).

Not sure why it included all the old commits but *whatever, github*.


Atermonera:
Probably because I squash-merged, so the older commits aren't seen in the commit history

PR #7268, opened 2020-06-01
Ports Bay's lockers/cabinets/crates/etc

Arokha:
We talked about it in the #code channel. They're dynamically generated and don't require a spriter to poke at them to change how they look or make new ones, you just specify what decals you want on them and such. If you don't like how they look, poke into the closet appearances file and you can alter the colors/decals/etc.

Most look pretty much the same, though. Also crates have all 4 direction sprites, so when you're pushing/pulling them or whatever they can rotate. Closets don't though, alas. Maybe someday.


code/game/objects/structures/crates_lockers/__closets.dm
```diff
@@ -44,23 +47,30 @@
 		starts_with = null
 
 	if(!opened)		// if closed, any item at the crate's loc is put in the contents
+		if(istype(loc, /mob/living)) return //VOREStation Edit - No collecting mob organs if spawned inside mob
```
Atermonera, line 50: misplaced tag

PR #7274, opened 2020-06-03
Adds get_cell to drill heads so you can use an inducer on them

Arokha:


Atermonera:
What is an inducer?

Mechoid:
Sitting for a while. An Inducer is IIRC a standalone object that transfers power, IE, the integrated circuit power transfer.

PR #7282, opened 2020-06-07
Add the Aurora Caelus Event!

Rykka-Stormheart:
Adds the Aurora Caelus, a rare event that occurs on the station space z-levels~!

During this event a .ogg will play to all clients, and you will see the outside appear almost like a storm - consider it a fun little break and a rare event for a moment of calm.

The event announces as soon as it fires, and assuming time dilation isn't horrible, starts 120 seconds after (60 "ticks" of the event subsystem being fired by the MC, which happens every 2 IRL seconds, affected by time dilation).

Current event weight (frequency) can be tweaked according to admin liking, and it's currently a MUNDANE level event. This can be changed to MODERATE or otherwise as determined by admins, or given a higher weight - during testing, stuff like PDA spam or news or etc very often went over it and it took giving it a lot of weight for it to fire - but it is supposed to be a "rare" cosmic thing, not something that happens every shift.

This is the upstream port of https://github.com/VOREStation/VOREStation/pull/8208 with the new synthesized AI voice removed - if you'd like it added, please refer to the VOREStation PR for what it sounds like, and then reply!


Heroman3003:
Suggested on downstream and also suggesting here. The music event plays should probably be affected by ambiance global preference.

Rykka-Stormheart:
> Suggested on downstream and also suggesting here. The music event plays should probably be affected by ambiance global preference.

I still disagree with that. If it was something that played multiple times, perhaps, but it plays once and it's a relatively rare event. I'd rather not people miss out on it over the handful of people that complain "but I have to hear the sound" - there's a simple solution: don't be on space z's right as it starts, or walk like... Three rooms and you stop the sound from playing.

Atermonera:
If people actually complain about it, then we can reconsider adding a pref later.

PR #7283, opened 2020-06-08
Generalizes supply points per material sheet to the material datum.

MistakeNot4892:


KasparoVy:
Good thinking and use of `?.`, conflict b/c other PR got merged so this'll need to be updated with the new material point values

PR #7284, opened 2020-06-08
Fixes cyborg rechargers not recharging

Cyantime:


Atermonera:
#7290 though this is actually probably a cleaner fix.

Cyantime:
Fixed in some form by another PR, good enough

PR #7287, opened 2020-06-10
Make gloves half as expensive.

lbnesquik:
They are not very important items, nor very expensive to make even now, especially compared to a dress or jumpsuit.
Exception made for the surgical gloves & armored equipment.


Novacat:
You forgot to also make the latex gloves cost 2 points as well.

lbnesquik:
AGH

lbnesquik:
Fixed that.

PR #7289, opened 2020-06-12
Give rock turfs radiation blocking capabilities.

Cyantime:
Right now they act as if they were completely empty, which is baffling.


lbnesquik:
Radiation shielded bunker when

PR #7292, opened 2020-06-14
Fixes prommies passively chonking it up

Verkister:
Fixes prommies passively gaining nutrition on every tick to reach the maxcap within minutes by just standing around regardless if they're even standing on dirt or even standing on a turf at all.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/prometheans.dm
+++ b/code/modules/mob/living/carbon/human/species/station/prometheans.dm
@@ -185,7 +185,7 @@ var/datum/species/shapeshifter/promethean/prometheans
 				var/turf/simulated/S = T
 				if(T.clean_blood())
 					H.adjust_nutrition(rand(10, 20))
-				if(S.dirt > 0)
+				if(S.dirt > 50)
 					S.dirt = 0
 					H.adjust_nutrition(rand(10, 20))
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/species/station/prometheans.dm
```diff
@@ -179,13 +179,15 @@ var/datum/species/shapeshifter/promethean/prometheans
...
+				if(O.clean_blood())
+					H.adjust_nutrition(rand(5, 15))
 			if (istype(T, /turf/simulated))
 				var/turf/simulated/S = T
-				T.clean_blood()
-				S.dirt = 0
-				H.adjust_nutrition(rand(10, 20))
+				if(T.clean_blood())
+					H.adjust_nutrition(rand(10, 20))
+				if(S.dirt > 50)
```
MistakeNot4892, line 188: Just an idle suggestion:
````
if(S.dirt)
  var/nutrition = min(S.dirt, 50) // or some other desirable cap
  S.dirt -= nutrition
  H.adjust_nutrition(max(round(nutrition/20), 1) 
````
This would make amount of dirt determine how much it feeds them, and would stop them eating 500u of dirt in one tick.

PR #7300, opened 2020-06-21
Streamlined Voidsuits Upgrade

KillianKirilenko:
A little feature-sprite expansion combo for the Streamlined Medical Voidsuit, based on ideas I've been playing with downstream (specifically [VS8138](https://github.com/VOREStation/VOREStation/pull/8138) and [VS8308](https://github.com/VOREStation/VOREStation/pull/8308)) so that it can now auto-adapt to most species when worn whilst also having unique sprites for said species *and* without adding the style to medical cyclers. Includes tweaks to cyclers to stop it from being refitted (and thus breaking horribly).

Sprites bashed together by me. I'm not 100% satisfied with them but they're sitting around my 'good enough' threshold.

image


Atermonera:
Sorry for the late response. Teshari probably should require refitting the suit, on account of having such a different body shape.

KillianKirilenko:
Mmn... let me experiment a little. I have an idea, but it might require adjusting cycler code a little more.

KillianKirilenko:
Didn't feel like trying to wrangle a special case just for one fancy suit, so I've instead made a subtype (which can be ordered seperately) that only fits teshari but can be manually cycled. The standard ("humanoid") model retains its adaptive perk but cannot be worn by tesh.

PR #7306, opened 2020-06-24
Add straffing to mechs. This took my entire morning goddamnit.

lbnesquik:
ou now have a straffing verb. Very convenient to shoot while firing back.
Thanks to @drexample even if i didn't use all that much of the original code i stole.

### DIFF
--- a/code/game/mecha/mecha.dm
+++ b/code/game/mecha/mecha.dm
@@ -100,7 +100,7 @@
 	var/static/image/radial_image_lighttoggle = image(icon = 'icons/mob/radial.dmi', icon_state = "radial_light")
 	var/static/image/radial_image_statpanel = image(icon = 'icons/mob/radial.dmi', icon_state = "radial_examine2")
 
-	var/straffing = 0
+	var/strafing = 0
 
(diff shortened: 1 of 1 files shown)


lbnesquik:
I have tested and retested each part btw and they work just fine.

Atermonera:
(It's spelled `strafing`, one `f`. We shall use the other one to pay respects)

lbnesquik:
Spelling corrected. Even if two f's look better.

code/game/mecha/mecha.dm
```diff
@@ -508,9 +509,16 @@
 				break
 		if(result)
 			move_result = mechstep(direction)
+
 	//Turning
-	else if(src.dir != direction)
-		move_result = mechturn(direction)
+
+	else if(src.dir!=direction && strafing)
```
Mechoid, line 515: `else if(src.dir!=direction && strafing)`
Should still be
`else if(src.dir != direction)`
The first && strafing means it will skip over, and not allow it to turn, even when not strafing.

lbnesquik: @Mechoid Thanks, that should be corrected now.

PR #7315, opened 2020-06-30
Add clickable action buttons from /TG/, rework a lot of backend and other improvements.

lbnesquik:
- Gygax overload, Durand defence mode, phazon phase, marauder smoke, zoom and thrusters have been made into modular things. Just put the 'smoke_possible' var to get smoke on your custom thing for example. This should work for custom map spawns, but that hasn't been properly tested yet.
- Move most abilities that aren't for default mechs to 'Mecha_actions.dm' for a more reasonable and not so huge file. 
  Marauder flight to be moved shortly.
- Lot of verbs have been broken into procs such as toggle_lights() now being a verb refering to lights() for easier work later on.
- Delete some movement code duplication with the gygax, gorilla, durand and phazon.
- Changed the smoke system to be more up to date, inspiration taken from the cooking machines.
- Added a noise for the internal air tank usage.
- Improved on the ripley speedbooster module code.
- Even added COMMENTS

image


lbnesquik:
The biggest stick still left in this JUNK is the actual thrusters code, being that there is still a lot of things to pick apart and test out, which will come in a later PR. Not to slow down anyone urges  regarding this one, given everything but that part works.

PR #7316, opened 2020-07-01
Add Mechs damage & penetration thresholds along with other improvements.

lbnesquik:
If you do not enough damage, it will simply bounce off consistently. Something that does 1 damage to a person would be so ridiculously nothing to a metal, possibly combat mech that it shouldn't achieve anything.
- If you throw, shoot or wack at a mech, unless the damage is decent, it will pretty much do nothing but cosmetic damage. Aka 0.

Additionally, armor reduces the effectiveness of attacks without a minimum of armor piercing.
- A flat hit with an hammer might be damaging to a squishy human, but it would probably not be all that effective against armored outerskin. The damage will be reduced to 2/3 or x0.66 to remark on this.

The values of the thresholds are modular for each mech and can be integrated in certain processes, like the Durand mode for example. The actual reduction value for failing to penetrate is also modular. 

-----------------------------------------------------
Other, lesser changes which are bundled in because Sourcetree can be a pain:
- Made the list for armor vertical instead of horizontal for easier read. Added bio & radioactive checks.
- Indented some comments to be easier to read.
- Added a couple comments of my own.

### DIFF
--- a/code/game/mecha/mecha.dm
+++ b/code/game/mecha/mecha.dm
@@ -907,8 +907,8 @@
 		src.log_append_to_last("Armor saved.")
 
 	else if(W.force < damage_minimum)	//Is your attack too PATHETIC to do anything. 3 damage to a person shouldn't do anything to a mech.
-		src.occupant_message("<span class='notice'>\The [A] bounces off the armor.</span>")
-		src.visible_message("\The [A] bounces off \the [src] armor")
+		src.occupant_message("<span class='notice'>\The [W] bounces off the armor.</span>")
+		src.visible_message("\The [W] bounces off \the [src] armor")
 		return
 
 	else if(W.armor_penetration < minimum_penetration)	//If you don't have enough pen, you won't do full damage
--- a/code/game/mecha/mecha.dm
+++ b/code/game/mecha/mecha.dm
@@ -16,21 +16,21 @@
 	name = "Mecha"
 	desc = "Exosuit"
 	icon = 'icons/mecha/mecha.dmi'
-	density = 1 //Dense. To raise the heat.
-	opacity = 1 ///opaque. Menacing.
-	anchored = 1 //no pulling around.
-	unacidable = 1 //and no deleting hoomans inside
-	layer = MOB_LAYER //icon draw layer
-	infra_luminosity = 15 //byond implementation is bugged.
-	var/initial_icon = null //Mech type for resetting icon. Only used for reskinning kits (see custom items)
+	density = 1							//Dense. To raise the heat.
+	opacity = 1							///opaque. Menacing.
+	anchored = 1						//no pulling around.
+	unacidable = 1						//and no deleting hoomans inside
+	layer = MOB_LAYER					//icon draw layer
+	infra_luminosity = 15				//byond implementation is bugged.
+	var/initial_icon = null				//Mech type for resetting icon. Only used for reskinning kits (see custom items)
 	var/can_move = 1
 	var/mob/living/carbon/occupant = null
-	var/step_in = 10 //make a step in step_in/10 sec.
-	var/dir_in = 2//What direction will the mech face when entered/powered on? Defaults to South.
+	var/step_in = 10					//make a step in step_in/10 sec.
+	var/dir_in = 2						//What direction will the mech face when entered/powered on? Defaults to South.
 	var/step_energy_drain = 10
-	var/health = 300 //health is health
-	var/maxhealth = 300 //maxhealth is maxhealth.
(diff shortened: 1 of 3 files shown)


code/game/mecha/mecha.dm
```diff
@@ -726,9 +741,27 @@
 		if(O.throwforce)
 
 			var/pass_damage = O.throwforce
+			var/pass_damage_reduc_mod
+			if(pass_damage <= damage_minimum)//Too little to go through.
+				src.occupant_message("<span class='notice'>\The [A] doesn't dents \the [src] paint.</span>")
+				src.visible_message("\The [A] doesn't dents \the [src] armor")
```
Mechoid, line 747: It might be best to have these messages match the X bounces off the Y's Armor message.

code/game/mecha/mecha.dm
```diff
@@ -761,9 +794,25 @@
 			ignore_threshold = 1
 
 		var/pass_damage = Proj.damage
+		var/pass_damage_reduc_mod
 		for(var/obj/item/mecha_parts/mecha_equipment/ME in equipment)
 			pass_damage = ME.handle_projectile_contact(Proj, pass_damage)
 
+		if(pass_damage < damage_minimum)//too pathetic to really damage you.
+			src.occupant_message("<span class='notice'>\The [Proj] doesn't dents \the [src] paint.</span>")
+			src.visible_message("\The [Proj] doesn't dents \the [src] armor")
```
Mechoid, line 803: As above, so below.

code/game/mecha/mecha.dm
```diff
@@ -849,19 +899,30 @@
...
 		src.log_append_to_last("Armor saved.")
-/*
-		for (var/mob/V in viewers(src))
-			if(V.client && !(V.blinded))
-				V.show_message("The [W] bounces off [src.name] armor.", 1)
-*/
+
+	else if(W.force < damage_minimum)	//Is your attack too PATHETIC to do anything. 3 damage to a person shouldn't do anything to a mech.
+		src.occupant_message("<span class='notice'>\The [W] doesn't dents \the [src] paint.</span>")
+		src.visible_message("\The [W] doesn't dents \the [src] armor")
```
Mechoid, line 911: As above, so below.

code/game/mecha/mecha.dm
```diff
@@ -2018,26 +2079,46 @@
...
+	if(prob(src.deflect_chance))//Deflected
 		src.log_append_to_last("Armor saved.")
-		playsound(src, 'sound/weapons/slash.ogg', 50, 1, -1)
 		src.occupant_message("<span class='notice'>\The [user]'s attack is stopped by the armor.</span>")
 		visible_message("<span class='notice'>\The [user] rebounds off [src.name]'s armor!</span>")
 		user.attack_log += text("\[[time_stamp()]\] <font color='red'>attacked [src.name]</font>")
+		playsound(src, 'sound/weapons/slash.ogg', 50, 1, -1)
+
+	else if(damage < damage_minimum)//Pathetic damage levels just don't harm MECH.
+		src.occupant_message("<span class='notice'>\The [user]'s doesn't dent \the [src] paint.</span>")
```
Mechoid, line 2102: As above, so below.

code/game/mecha/mecha.dm
lbnesquik: I figured it was better to have something distinct, to not have extra confusion. I can change those if you confirm you'd prefer that.

Mechoid: 'X bouncing off Y' is a better informant than the present, since the message currently already informs attackers that they did no damage.

code/game/mecha/mecha.dm
lbnesquik: Changed.

code/game/mecha/mecha.dm
lbnesquik: Changed. 
For below, it actually does a fair amount of damage, so the deflect would be misleading.

code/game/mecha/mecha.dm
```diff
@@ -743,8 +743,8 @@
 			var/pass_damage = O.throwforce
 			var/pass_damage_reduc_mod
 			if(pass_damage <= damage_minimum)//Too little to go through.
-				src.occupant_message("<span class='notice'>\The [A] doesn't dents \the [src] paint.</span>")
-				src.visible_message("\The [A] doesn't dents \the [src] armor")
+			src.occupant_message("<span class='notice'>\The [A] bounces off the armor.</span>")
```
Mechoid, line 746: Un-indented.

PR #7317, opened 2020-07-02
TGUI-V3

ShadowLarkens:
From https://github.com/ParadiseSS13/Paradise/pull/13310

> 
> ## What Does This PR Do
> 
> This PR implements TGUI-V3. This is mostly a direct port, but has been tweaked to allow it to co-exist alongside nanoUI with no connection between the two. This means that nanoUI behaves completley independantly from TGUI, and thus can be removed fairly easily when needed. This PR just adds in the framework for TGUI, as well as two interfaces (Disposal units + Crew monitors) to keep bloat down. If this is merged, I will make more PRs down the road to get all the UIs converted to TGUI.
> 
> TGUI-V3 also includes a development server which allows you to create UIs ingame **IN REAL TIME**. No more recompiling or even reloading the server. Save the file in your IDE and watch your changes instantly appear in the UI ingame.
> ## Why It's Good For The Game
> 
> TGUI is FAST, has more control options, is easier to develop for (explain above), is actually documented unlike nano, and has more features. It also allows new TG ports for objects they have which use this UI.
> 
> Responsiveness:
> 79617615-d06c5d80-80ff-11ea-87d0-b7a727938638
> 
> Rendering speed:
> image
> 
> More controls (Plus you can change numerical values without opening another popup):
> image
> [Gif Link](https://cdn.discordapp.com/attachments/145700319819464704/700674876528132156/RuwonGCX4U.gif)
> 
> image
> [Gif Link](https://cdn.discordapp.com/attachments/211813686296313856/700073005421690900/t8YQgfyOH5.gif)
> 
> More UI elements (The graph at the top):
> image
> ## Images of changes
> 
> Disposal UI
> image
> 
> Crew Monitor UI (AI + Console + Handheld Monitor)
> image

Note that the crew monitor UI is a little different, since I added multi-z support. image
~~It's not an amazing replacement for NanoMap yet, but it definitely could be improved with pan/zoom support. I didn't want to spend too much time focusing on it before contributing TGUI. If this will be too much of a feature loss, considering that this new version is used for Cyborgs and the AI, let me know and I can just switch them back to using the nano version until I get around to improving this crew monitor.~~ Probably good enough now.


ShadowLarkens:
Update, I've gone ahead and made the crew console map work just as well as the old one. Mostly. It doesn't have mousewheel support because weird IE stuff, and it doesn't have the fancy rank icons or what not, but it's most of the way there. I'm happy with it now, at least.

image

ShadowLarkens:
Polished it a little with @AffectedArc07's help. Calling this done enough.

Neerti:
I've done some testing with this on Linux through WINE. Not sure if you're able to fix any of the issues encountered but figured I should post my findings here anyways.

# Disposal chute
Works surprisingly well, with the only issue being the icons disappearing, which NanoUI seems to handle just fine, but the text and buttons still function so its just a cosmetic problem.
Disposal chute UI on Linux

# Suit sensors.
Data view works fine. Has the icon problem so presumably everything will have it, which isn't too big of a deal.
Map view is concerning, however, since it both seems to not really work, but also that if I drag the map and then wait a few seconds, the window freezes, taking my client down with it. Leaving the map window open and not touching the map doesn't do that.

The nanoUI map doesn't appear to work in WINE (but doesn't freeze), so the map not working isn't a step backwards I suppose. I am worried about the freezing, however, since tracking down what exactly is causing that will probably be difficult, and I'm a bit worried that some future UI that does something important and isn't just a 'nice to have' might induce a freeze.

Here's how it looks when not touching anything.
Suit sensor UI on Linux

Atermonera:
Fair warning, this will take me a while to review.

ShadowLarkens:
Leaving myself a note here more than anything so I can find and fix it easier later, these will cause merge conflicts
https://github.com/PolarisSS13/Polaris/pull/7326 and https://github.com/VOREStation/VOREStation/pull/8370 (if it gets backported here)

ShadowLarkens:
` Map view is concerning, however, since it both seems to not really work, but also that if I drag the map and then wait a few seconds, the window freezes, taking my client down with it. Leaving the map window open and not touching the map doesn't do that.` As it turns out, this happens on Windows too. Investigating.

ShadowLarkens:
Nanomaps now mostly matches the implementation here https://github.com/ParadiseSS13/Paradise/pull/13766

Neerti:
Tested again with the changes, and I'm happy to report that the freezing problem seems to be gone now.
Sadly the map still doesn't quite work, but the current NanoUI map also doesn't work so it's not a big deal for me since nothing really changes feature-wise, yet this will probably end up being an upgrade over the long term.

Probably not worth the effort into trying to fix whatever might be making it act weird, but just in case you want to, here's a short video of [how it looks on my end](https://streamable.com/t20zbb)

AffectedArc07:
Did a brief skim over the code, looks fine to me

Atermonera:
Live test didn't uncover anything terrible

PR #7329, opened 2020-07-09
Exosuit Modular Internals, The Squeakening

Mechoid:
Additions:
 - Exosuits now have internal components that determine functionality and speed.
 - - Armor: First layer of defense, determines incoming damage values similar to old armor. Has a health pool that takes up to half of the incoming damage, before passing it to the Hull. Crippling will stop the exosuit from being able to insulate from high or low temperatures.
 - - Hull: Second layer of defense, acts as a health buffer for other components and the chassis. Can absorb 50% to 100% of the damage passing the armor. Excess damage has a chance to flow to components, based on their 'size' in comparison to the body. Remaining damage will pass to the chassis. Crippling it will cause an internal fire.
 - - Actuator: Determines strafing and movement speed. Crippling will cause loss of movement control.
 - - Electrics: Determines cell charge use efficiency. Crippling will cause a short circuit.
 - - Life Support: Determines reliability of exosuit atmosphere control.  Crippling will cause a tank breach.

 - Robot analyzer will display exosuit information.
 - Cloaking can be an innate Exosuit capability, similar to Defense Mode or Overload.
 - Phazons gain the ability to Cloak innately, as their advantage over the Janus.

 - Flechette shotgun shells added to autolathes.
 - EMP shotgun shells added to R&D.
 - AP ammunition for 45 pistols and 9mm SMGs added to autolathes, requiring advanced components and hacking.
 - Hollowpoint ammunition for 45 pistols added to autolathes.
 - Autolathes can accept Plasteel and Plastic.

Tweaks:
 - Many exosuit combat components now slow down exosuits they are attached to, due to the increase in Exosuit durability.
 - Exosuit health lowered across the board, as Armor and Hull components provide their own health pools.
 - Durand components generally increased in cost as it stands, Experimentally, at a high point of durability in exchange for speed.
 - R&D builds stun and emp shotgun shells in boxes of 8.
 - Ripleys are blast-resistant. Explosives are not guaranteed to destroy exosuits, instead dealing a _massive_ amount of damage scaling with the exosuit chassis healthpool.


code/game/objects/items/weapons/storage/boxes.dm
```diff
@@ -198,6 +198,16 @@
...
+/obj/item/weapon/storage/box/flechetteshells
+	name = "box of shotgun flechettes"
+	desc = "It has a picture of a gun and several warning symbols on the front.<br>WARNING: Live ammunition. Misuse may result in serious injury or death."
+	icon_state = "lethalslug_box"
+	item_state_slots = list(slot_r_hand_str = "syringe_kit", slot_l_hand_str = "syringe_kit")
+	starts_with = list(/obj/item/ammo_casing/a12g/flechette = 8)
+
+/obj/item/weapon/storage/box/flechetteshells/large
+	starts_with = list(/obj/item/ammo_casing/a12g/flechette = 16)
+
```
lbnesquik, line 210: What's all the above about relative to mechs?

lbnesquik:
I am pretty interested in how this might turn out to play, and the side tweaks are nice!

----------------------------

> Many exosuit combat components now slow down exosuits they are attached to, due to the increase in Exosuit durability.

I am personally against this. Slowdowns are ass and just makes using mechs overall less fun. If the goal is making them more usable and interesting, this is not the best.

And especially for some of those modules, the flamethrower guns are already terrible enough, given they fire incendiary bullets instead of straight jets of flame unlike CM tanks. The solaris is actually a fair bit weaker for energy, DPS and ease of use with its only upside being slight armor penetration. Same for the scattershot. I can see it for the mortar to some degree.

DerBlitzkrieg:
Slowdown is a no for me, I've seen too many mech fights where not being able to outrun the enemy leads to death. The slowdown alone is what makes the gygax a better mech than a durand in my opinion, as a durand can be easily overwhelmed with a lot of frankly weak guns, as it has no ability to escape on its own. Everything else seems interesting, not sure exactly how this will affect mech durability but it's worth a test.

code/game/objects/items/weapons/storage/boxes.dm
Mechoid: It's to help counter Heavy Armor on Exos, yes.

Mechoid:
As always, none of the slowdown values are set in stone, they're what I think are good for live testing values to be reconsidered later. Durands in particular are incredibly strong, unless you have EMPs or high explosives, as their default armor is immune to the basic weapons Security has (Standard ammunition, las carbines), hence the tangential updates in other areas (AP ammunition, special ammunition). The slowdown from weaponry and defensive modules is to incentivize not loading an exosuit like a Gygax or said Durand with as many weapons as it can carry.

Exosuit slowdown isn't in increments matching mob slowdown, either, it is in 10ths of a second, while mob slowdown is in quarters IIRC. So 1 slowdown for a mob is 2.5 for an exosuit.

Lascannons are particularly horrifying, given the general lack of armor usually encountered by people using exosuits in game.

Internal damages are mostly susceptible to EMP, given they take only a portion of incoming damage that passes armor or hull, if the pilot is particularly unlucky. Internal components require replacement, unless you have Nanopaste to repair them. Their faults can be 'patched', and will not recur until they take damage again, if they are crippled, however.

lbnesquik:
It is worth nothing most mechs are HEAVILY PVE, especially downstream so that's where our feelings lie, not so much on PVP like teegee might.

code/game/mecha/components/_component.dm
```diff
@@ -0,0 +1,155 @@
...
+	origin_tech = list(TECH_DATA = 2, TECH_ENGINEERING = 2)
+
+	var/component_type = null
+
+	var/obj/mecha/chassis = null
+	var/start_damaged = FALSE
+
+	var/emp_resistance = 0	// Amount of emp 'levels' removed.
+
+	var/required_type = null	// List, if it exists. Exosuits meant to use the component.
```
Atermonera, line 16: If the variable is meant to hold a list (and nothing else), it should be cast as a list

code/game/mecha/components/_component.dm
```diff
@@ -0,0 +1,155 @@
...
+	if(integrity < integ_limit)
+		var/int_percent = round(integrity / integ_limit, 0.1)
+
+		return int_percent
+
+	return 1
+
+// Attach/Detach code.
+
+/obj/item/mecha_parts/component/proc/attach(var/obj/mecha/target, var/mob/living/user)
```
Atermonera, line 94: Shouldn't required_type be checked here?

code/game/mecha/components/armor.dm
```diff
@@ -0,0 +1,238 @@
...
+
+	internal_damage_flag = MECHA_INT_TEMP_CONTROL
+
+	step_delay = 1
+
+	var/deflect_chance = 10
+	var/list/damage_absorption = list(
+									"brute"=0.8,
+									"fire"=1.2,
+									"bullet"=0.9,
```
Atermonera, line 28: Great googly indentation, batman! (I don't think 8 extra indents are really necessary, but you should vertically align the numbers)

code/game/mecha/mecha.dm
```diff
@@ -1082,6 +1312,20 @@
 			else
 				to_chat(user, "You were unable to attach [W] to [src]")
 		return
+
+	if(istype(W, /obj/item/mecha_parts/component) && state == 4)
```
Atermonera, line 1316: magic numbers

code/modules/research/mechfab_designs.dm
```diff
@@ -1038,4 +1038,151 @@
...
+/datum/design/item/mechfab/exointernal/mining_armor
+	name = "Armor Plate (Blast)"
+	category = "Exosuit Internals"
+	id = "exo_int_armor_blast"
+	req_tech = list(TECH_MATERIAL = 4, TECH_ENGINEERING = 4)
+	materials = list(DEFAULT_WALL_MATERIAL = 20000, MAT_PLASTEEL = 10000)
+	build_path = /obj/item/mecha_parts/component/armor/mining
+
+/datum/design/item/mechfab/exointernal/gygax_armor
+	name = "Armor Plate (Marshal)"
```
Atermonera, line 1082: Is there anything explaining what these different armor types do and which mechs they're actually designed for/based on?

code/modules/research/mechfab_designs.dm
```diff
@@ -1038,4 +1038,151 @@
...
+	id = "exo_int_hull_light"
+	req_tech = list(TECH_MATERIAL = 3, TECH_ENGINEERING = 4)
+	materials = list(DEFAULT_WALL_MATERIAL = 5000, MAT_PLASTIC = 3000)
+	build_path = /obj/item/mecha_parts/component/hull/lightweight
+
+/datum/design/item/mechfab/exointernal/stan_gas
+	name = "Life-Support (Standard)"
+	category = "Exosuit Internals"
+	id = "exo_int_lifesup_standard"
+	req_tech = list(TECH_MATERIAL = 2, TECH_ENGINEERING = 2)
```
Atermonera, line 1146: So now robotics _requires_ research to make _any_ mechs whatsoever?

code/modules/research/mechfab_designs.dm
Mechoid: Marshal is in reference to the Police / Corp Sec exosuit, Gygax. Military is the Durand.

code/modules/research/mechfab_designs.dm
Mechoid: Negatory, exosuits start with their 'default' components by.. default. Research is needed to produce -extra- components.

code/game/mecha/components/_component.dm
Mechoid: required_type is the same as the Mech equip required_type, wherein the part can specify var changes or other unique effects if it is attached to a specific exosuit. For instance, so the Durand doesn't get way more slowdown from its armor, but a Ripley does.

code/game/mecha/components/armor.dm
Mechoid: This shows up as far more indented here on Github, it's in-line with the list( component on DM

code/datums/autolathe/autolathe.dm
```diff
@@ -71,6 +71,7 @@ var/datum/category_collection/autolathe/autolathe_recipes
 	var/is_stack // Creates multiple of an item if applied to non-stack items
 	var/max_stack
 	var/no_scale
+	var/man_rating = 0
```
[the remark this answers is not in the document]
Mechoid: This is the manipulator rating used in RefreshParts, it determines the 'tier' of items produceable by the Autolathe. Only the few example items use it so far.

code/game/mecha/components/_component.dm
Mechoid: I don't know how I missed this.

PR #7334, opened 2020-07-12
Reorganized Library Computer Menu

Cerebulon:
More clearly labelled the new, functional archive and the old, dodgy archive, and put the former in the proper place in the list instead of the bottom.

Code-wise this was just switching their places in the switch statements.


Greenjoe12345:
Oh, while you're looking at that, I think the old external archive kinda breaks the computer
It got stuck on the external archive menu

Cerebulon:
I haven't encountered that issue, and the code indicates that the External (now 'Deprecated') Archive should have a "back" button which takes you back to the main menu?
Has anybody else had that problem?

Greenjoe12345:
There was a back button, or main menu button
Issue is the button did not work

PR #7335, opened 2020-07-13
Attempts to fix blurry icon scaling

Cerebulon:
Actually two fixes from different fixes downstream. Sets the skin default to nearest-neighbor ('distort'), and also adds the KEEP_TOGETHER flag to atoms for scale purposes.

### DIFF
--- a/code/game/atoms_movable.dm
+++ b/code/game/atoms_movable.dm
@@ -1,6 +1,6 @@
 /atom/movable
 	layer = OBJ_LAYER
-	appearance_flags = TILE_BOUND|PIXEL_SCALE
+	appearance_flags = TILE_BOUND|PIXEL_SCALE|KEEP_TOGETHER
 	glide_size = 8
 	var/last_move = null //The direction the atom last moved
 	var/anchored = 0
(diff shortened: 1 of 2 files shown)


KillianKirilenko:
I'm gonna be honest, nearest neighbor scaling looks pretty bad if you have stretch-to-fit and your resolution takes icons too far away from 32x32.

I'm testing out player-choice scaling based off a top menu dropdown.

Cerebulon:
I can't imagine using stretch-to-fit that's pretty messed up

KillianKirilenko:
Ignoring 32x32 (for obvious reasons), the only other real option at my resolution is 48x48... at which nearest neighbor still looks inferior to point sampling. At 64x64 things like the ingame UI run off the edge of the screen, so not really an option.

Cerebulon:
We don't have 48x48 scaling, which yes I can see being made a little fucked. But you can't really expect us to have accounted for client options which we do not have when fixing what was on our end a complete Yes/No for having hideous blurry sprites.
It would have been nice if you'd fixed the issue on your end which only exists on your end, without kinkshaming our very sexy fetish.

PR #7337, opened 2020-07-13
Persistent Trash/Dirt/Graffiti/Notices

Cerebulon:
Based on/ported from World Server https://github.com/GeneriedJenelle/The-World-Server-Redux/pull/412

Ports cross-round persistence of:
Cleanable decals (dirt, oil, vomit, drawings etc.). Blood and gibs are converted to a generic greenish smudge for immersion reasons.
Noticeboard contents.
Stickynotes (New, orderable from cargo) and taped paper.
Carved graffiti (New, carvable on suitably dense - but not too dense - walls and floors with any 'sharp' object, repairable with a welder)

Decals will be deleted if they persist for too many rounds, and graffiti will deteriorate over many rounds before being deleted.

Certain areas are flagged to be non-persistent. This includes:
Space
The station roof
Outdoors on Sif (including the mountain/caves)
PoIs
Shuttles
The elevator
The holodeck
Medical trauma center and operating theatres (The most sterile place onboard shouldn't be a filth pit, and that place gets DIRTY)
Delivery Office + Disposals (If trash is here, it's probably worth deleting rather than piling up 10 rounds of crap and lagging)
Asteroids/other z-level crap, including Centcom/Antag start areas.
Head offices with pets, for sanity reasons.

This is currently a config option, enabled by default. If an admin is expecting a particularly filthy round, (e.g. an event round), it can be toggled off under the Server tab with the "Toggle persistent data" verb, ideally early in the round.
Persistent items/decals will only save when persistence is enabled at the moment they are created or moved, but can be cleaned ('deleted' from the data lists) any time.

Also makes cigbutts a trash item so they're persistent and can appear as random trash.

This MAY require changing how random maint trash is spawned so it doesn't pile up too hard, but that's to be seen.

Apologies in advance to Ater for the 56 files of crap.


elgeonmb:
the rd's office is going to be just disgusting now lmao

code/controllers/configuration.dm
```diff
@@ -232,6 +232,8 @@ var/list/gamemode_cache = list()
 	var/dooc_allowed = 1
 	var/dsay_allowed = 1
 
+	var/persistence_enabled = 1
```
Mechoid, line 235: Shouldn't this be 0 by default, if it is a toggle enable?

Cerebulon: The idea is it's a toggle DISable. If it's disabled by default, shit won't load at roundstart so enabling it after that would be pointless and just create a new 'blank' save for the round only including stuff from that round (and overwriting the previous 'save files') which then in turn wouldn't be loaded for next time.

PR #7340, opened 2020-07-14
Quick fix to stop mobs in nullspace runtiming the mobs and ai subsystems

Rykka-Stormheart:
Note: Theoretically an error still could happen if the following happens in exactly this order:

1. Subsystem starts firing, then sleeps.
2. A new z-level is allocated.
3. A mob that previously existed prior to step 1 but was not processed during step 1 is moved to the new z-level.

But the error would clear itself after one cycle.

(This bug would cause mobs and AI subsytems to stop processing anything at all)

Upstream port of https://github.com/VOREStation/VOREStation/pull/8412


Rykka-Stormheart:
Closing this, your subsystem *currently* is not vulnerable to this failure, if `living_players_by_zlevel` is ported, then this fix will need to be applied.

PR #7345, opened 2020-07-17
custom_sprites.dm small improvement and better commenting. 

LunarFleet:
Added better instructions in the form of comments at the bottom of the file, for those who read it in the future. Changed the - in splittext to | because many borg playeres use - in their names.


Mechoid:
This will cause problems for the current configs on the custom repo, unless you also make a PR to swap that over.

cadyn:
> This will cause problems for the current configs on the custom repo, unless you also make a PR to swap that over.
https://github.com/PolarisSS13/Custom-Items/pull/163
Giving this gal a break since she's not even a dev and has had enough trouble getting this through already. Done and done.

PR #7354, opened 2020-07-19
Implements Status Indicators

Neerti:
This PR adds the ability to have certain status effects such as stuns or confusion have a visual component, by having the ability to have icons show up over mobs.

Here's a video of it in action: https://streamable.com/rdp6a9

Some notes:
This is not tied to the med/sechuds at all whatsoever. The video shows the medhud bar to show that the icons don't conflict with it.
A lot of effort was put in to make sure that the icons would always be over the mob, even for edge cases like 64x64 sprites or scaled up sprites from things like modifiers or just a `adjust_scale()` call. Thus this should work just fine for big xenos.
Currently this PR adds icons and code to have all three stun types, blindness, confusion, and sleeping. The system is nice and generic so it can be hooked into things like Modifiers later down the road.
This also adds a player pref to disable seeing status indicators, if you don't like them for whatever reason.
The icons are, atm 16x16. Someone offered to make me 8x8 sprites but I'm skeptical that they can be distinguishable between each other during a high stress event, but the system's easy to adjust for 8x8 sprites, only needing to tweak a few `#DEFINE`s.


DasIrrlicht:
On one hand, sounds and looks neat...
On the other... Not sure this should be visible to other people.
I _have_ faked most, if not all, of this conditions for RPs sake.
And there are at least a few characters I know who are 'blind' or 'blind when cloned', who would naturally not have the blind indicator because they Roleplay it, instead of having the actual condition mechanically applied.
In a simelar vein, a Antag can very well ty to lure someone into a fake sense of security by trying to play on someones believes, fakeing to be still stunned even if not and so on...

So while I can not deny that this are neat indicators, I do think they should be limited to the own player to see, or maybe ghosts or simelar, since there are many cases where a character has a 'condition' in RP that is not actually mechanically there...

There are a few indicators I believe that should be universally visable, however: Handcuffed, for a example, and its counterpart Legcuffed. SSD, being something of big mechanical importance people should comfort around when RPing, would also deserve its own thing besides sleeping, given, if sleeping should be visible or not is something I am on edge, since faking to sleep is a very rare case where people usually go with throwing themself around in bed _trying_ to sleep.

Neerti:
The indicators being visible to other people is very much intended, and I'd argue very important, for multiple reasons I will list below.

- It lets certain kinds of mobs besides the human type, such as `/mob/living/simple_mob` (stuff like spiders) or `/mob/living/silicon/robot` (borgs) be able to visually show that they're stunned, or otherwise incapacitated. Being shown only to the mob itself makes it so others can't see that the giant spider who slammed into a wall is stunned.
- It allows us to bring back usage of the `Stun()` and `AdjustStunned()` effect[1], as previously there was no way for the person suffering from it that they can't move until they try to and find they can't (and more often than not think that they're lagging as a result). Other people also can't know if they're not able to move, and since that kind of stun leaves you still standing and holding weapons, people tended to assume they were still capable of fighting at that moment and usually end up getting dunked. Being able to see that the mob currently can't attack or move requires that other people be able to see the icon.
- It improves feedback between the player and the game, since the results are shown as soon as the mob gets stunned, instead of wondering if they or the server are having lag issues. While the feedback remains intact if the player's mob gets stunned, the feedback from a player stunning another mob (player or otherwise) is gone if other people can't see the icons.

[1] In the code, there are three kinds of stuns, which are all rather poorly named and misleading. There's Weaken, Stun, and Paralyze. Weaken makes you fall over, and is the most common kind of stun and probably the kind you think of when talking about stuns. Stun just makes you unable to act but you still stand, arguably the weakest form of stun, and used to be used in places like flashbangs, but has been phased out (#4922)  since it tended to get people killed by security while they were stunned but sec thought that they were unaffected. That change probably wouldn't have been needed if status indicators existed before that. Finally, Paralyze makes you fall down and it makes you blind as well, and is used for stuff like being pain-critted, and is arguably the most 'severe' form of stun.

PR #7356, opened 2020-07-20
Port Medibot Tipping!

Rykka-Stormheart:
This PR lets you tip over medbots which prevents them from moving or healing, and they'll slowly get more and more distressed and desperate asking for help from people around them. After a few minutes (currently moments, but can be tweaked), if no one helps right them, they'll manage to right themselves up and snitch on whoever tipped them over to medical so the psychologist can keep an eye on them.


Rykka-Stormheart:
This is clearly high-priority and will crash the server if not merged, merge ASAP.

PR #7357, opened 2020-07-20
Pressing Q now properly updates borg modules

Shadow-Quill:
Previously using the Q hotkey to drop modules from the toolbar as a borg would not update to have the most recently dropped modules, until you pull a fresh module out/hit the store button/close&open the module inventory

Tested, works as I'd expect. Hopefully. I'm sleepy, feel free to scree at me with anything I can fix.

### DIFF
--- a/code/modules/mob/living/silicon/robot/inventory.dm
+++ b/code/modules/mob/living/silicon/robot/inventory.dm
@@ -52,7 +52,7 @@
 		module_state_3 = null
 		inv3.icon_state = "inv3"
 	updateicon()
-	usr.hud_used.update_robot_modules_display()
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/silicon/robot/inventory.dm
```diff
@@ -52,6 +52,7 @@
 		module_state_3 = null
 		inv3.icon_state = "inv3"
 	updateicon()
+	usr.hud_used.update_robot_modules_display()
```
Mechoid, line 55: May be a good idea to test without the `usr.`, to avoid anything wonky involving things like EMPs.

PR #7361, opened 2020-07-22
Overdose changes + General chem mods

ExabyteDiscord:
## What Does This PR Do
Changes a few chemicals normal effects, and adds a few unique effects to overdoses. Overdose toxin damage has been completely changed to be more of a threat than it's previous damage. Allows health/body scanner to detect overdoses

## Why It's Good For The Game
 Overdoses previously were quite frankly negligable, meaning that there's hardly any reason to worry about overdosing in the first place, which has been changed entirely.

## Images of changes

image
image

Upstream port of https://github.com/VOREStation/VOREStation/pull/8442

### DIFF
--- a/code/game/machinery/adv_med.dm
+++ b/code/game/machinery/adv_med.dm
@@ -306,7 +306,7 @@
 			var/reagentData[0]
 			if(H.reagents.reagent_list.len >= 1)
 				for(var/datum/reagent/R in H.reagents.reagent_list)
-					reagentData[++reagentData.len] = list("name" = R.name, "amount" = R.volume)
+					reagentData[++reagentData.len] = list("name" = R.name, "amount" = R.volume, "overdose" = ((R.overdose && R.volume > R.overdose) ? "Overdose" : ""))
 			else
 				reagentData = null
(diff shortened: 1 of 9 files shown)


Atermonera:
Staff input: The numbers look like in theory they're manageable with dylo (who even makes carthatoline), but we'll review the results in a few weeks and tweak as needed. Making this more threatening is all well and good, but it's important to note that one of the main ways of treating IB is by ODing on bicard, and if this would make it untenable to do so, then we're back to IB->surgery, which is undesirable.

Wickedtemp:
As far as bicaridine-overdose for internal bleeding goes, all it'd really mean is that you'd have to manage the toxin damage a little more carefully, use myelamine, or...actually operate. Of the four available treatment options, bicaridine-overdose is the second worst way to go about it, behind myelamine, but ahead of amputation.

If the overdose damage *is* capped at 3.6 toxin per tick, that's entirely manageable. Bicaridine overdose remains as it is - a potentially dangerous method of healing internal bleeding when other more reliable and safe options aren't available.

PR #7364, opened 2020-07-24
Rag Coding Fix

Draycu:
- A quick one file fix for a rag coding issue related to face covering helmets not providing protection to smothering should rags be filled with chloral hydrate.  

- Basic face masks will not protect against the intent, but items like welding masks and space suit helmets will.

ragmaskdetection

This is a fix for an issue brought up on a server downstream, pulled here to hopefully prevent any future conflicts arising should this file be modified in the future.

### DIFF
--- a/code/modules/detectivework/tools/rag.dm
+++ b/code/modules/detectivework/tools/rag.dm
@@ -105,29 +105,39 @@
 			A.clean_blood()
 
 /obj/item/weapon/reagent_containers/glass/rag/attack(atom/target as obj|turf|area, mob/user as mob , flag)
-	if(isliving(target))
-		var/mob/living/M = target
-		if(on_fire)
-			user.visible_message("<span class='danger'>\The [user] hits [target] with [src]!</span>",)
-			user.do_attack_animation(src)
-			M.IgniteMob()
-		else if(reagents.total_volume)
-			if(user.zone_sel.selecting == O_MOUTH)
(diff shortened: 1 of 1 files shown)


Neerti:
This might be Github lying to me but it really looks like you used spaces for indentation instead of tabs.
I'm surprised it compiled if that's what happened.

Draycu:
You weren't wrong, heh!  Turns out something weird happened and spaces didn't cause an issue during compiling.  Hopefully everything should be fixed properly now.

code/modules/detectivework/tools/rag.dm
```diff
@@ -105,29 +105,42 @@
...
+					//it's inhaled, so... maybe CHEM_BLOOD doesn't make a whole lot of sense but it's the best we can do for now
+					reagents.trans_to_mob(target, amount_per_transfer_from_this, CHEM_BLOOD)
+					update_name()
+				else
+					to_chat(user, "<span class='warning'>You can't smother this creature.</span>")
+					return
 			else
-				wipe_down(target, user)
+				to_chat(user, "<span class='warning'>You can't smother this creature.</span>")
+				return
```
Atermonera, line 135: All these extra returns are redundant, since the code would normally just walk out the end of these if blocks and right into the end of the function

Draycu: Made some adjustments, thank you for the suggestion there~!

PR #7366, opened 2020-07-24
Ambiance and Sound Updates

Rykka-Stormheart:
All living mobs will now have a check run every 30 seconds or so to replay the current ambience of the area they're in, if they've not had it played in the last minute. Said check is a 35% chance of playing the ambience. Good for if you're standing around in one area.

Computers + turrets got new sounds, turrets have a deploy, retract, and rotate sound. Arcade machines now play sound effects depending on what step you're on and what you're doing.

Replaced maintenance ambience with different ambience from Paradise/Aurora, and moved our maintenance to foreboding, and moved old foreboding to it's own folder.

Upstream port of https://github.com/VOREStation/VOREStation/pull/8462


code/modules/mob/living/life.dm
```diff
@@ -88,6 +91,12 @@
 /mob/living/proc/handle_stomach()
 	return
 
+/mob/living/proc/handle_ambience() // If you're in an ambient area and have not moved out of it for x time, we're going to play ambience again to you, to help break up the silence.
+	if(world.time >= (lastareachange + 30 SECONDS)) // Every 30 seconds, we're going to run a 35% chance to play ambience.
```
Atermonera, line 95: Is the 35% in play_ambience()?
Does play_ambience() update lastareachange, or will this fire every life tick after the first 30 seconds?

Rykka-Stormheart: If the prob(35) in the play_ambience() fires, it will update the last-played - so while yes this fires every life tick afterwards until you move to another /area/, you realistically cannot get ambience played to you any longer than once per 60 seconds, and even then, it's a 35% chance to have the ambience played.

Atermonera: So yes, the 35% is in play_ambience()

listerla:
This maint sound literally includes background noises of cars. Can we revert that change, at the least? Not very immersive tracks IMO. Nevermind the jukeboxes and such that've broken for a lot of people.

Rykka-Stormheart:
> This maint sound literally includes background noises of cars. Can we revert that change, at the least? Not very immersive tracks IMO.

Which maintenance tracks?

listerla:
Not sure which one. Heard it ingame, there was the faint noise of cars driving overhead.

Rykka-Stormheart:
> Not sure which one. Heard it ingame, there was the faint noise of cars driving overhead.

If you can locate the raw files in /ambience/maintenance and tell me which one, I can look into it, but none of them have cars driving overhead?

Rykka-Stormheart:
Yeah, I had someone else listen and there's nothing that's "car driving overhead". There's faint rumbling in maintenance2.ogg but that's easily chalked up to station vibrations.

PR #7369, opened 2020-07-25
Department and Round Goals

Mechoid:
Implements a tiered method for Ibnesquik's #5720.
Baseline system for the sorting of and checking of Departmental goals.


Mechoid:
Moved to standard PR as this will likely generate more ideas for usable goals if implemented for people to work off of themselves.

PR #7372, opened 2020-07-26
Medibot Interaction Hotfix

Rykka-Stormheart:
Always check that you're passing the arguments a proc needs.

In this case, I was calling interact(), but not giving it a mob to interact with. Whoops.

Upstream of https://github.com/VOREStation/VOREStation/pull/8482


Rykka-Stormheart:
Closing this because mechoid already had one up

PR #7376, opened 2020-07-26
Attempt to fix mech and fighter movement issues within space.

lbnesquik:
Here is the issue in question: https://github.com/VOREStation/VOREStation/issues/8438

### DIFF
--- a/code/game/mecha/mecha.dm
+++ b/code/game/mecha/mecha.dm
@@ -582,8 +582,6 @@
 			last_message = world.time
 		return 0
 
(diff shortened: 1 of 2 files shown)


code/game/mecha/mecha.dm
```diff
@@ -582,6 +582,17 @@
 			last_message = world.time
 		return 0
 
+	if(!thrusters && src.pr_inertial_movement.active())	//Trying to move when you are drifting already while lacking thrusters will result in nothing.
```
Mechoid, line 585: This looks to be just a copy of the later check.

PR #7377, opened 2020-07-26
Species Vars Extension Pt3: Blood!

KillianKirilenko:
Time to roll up your sleeves, folks! 💉

This offloads constant blood levels to species level vars that allow greater control of thresholds for various symptoms. Want a species that gets woozy earlier but takes damage later? Can do. Want a trait that just makes you more vulnerable to blood loss in general? Go ahead.

Lightly tested to avoid carbons dropping dead on the spot, which did actually happen at one point because I forgot to tweak a couple of comparisons. I might have missed a couple of things though, so I'd appreciate a second or third pair of eyes to check the math changeovers.

Also tweaks the feedback from low blood levels for clarity/flavor.

### DIFF
--- a/code/game/machinery/iv_drip.dm
+++ b/code/game/machinery/iv_drip.dm
@@ -128,7 +128,7 @@
 				return
 
 			// If the human is losing too much blood, beep.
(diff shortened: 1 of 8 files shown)


lbnesquik:
This is nice.

PR #7378, opened 2020-07-26
Think with Portals

Neerti:
Implements a new `/map_effect`, portals, which allow a mapper to visually link two places together, and go between them by walking through. This can allow for strange PoIs that twist and turn in apparently physically impossible ways.

Portals aren't perfect, their limitations have been noted in the code file. Perhaps some of the flaws can be worked out in the future.

This PR adds a 'spatial anomaly' cave PoI that uses several portals to make it work. It doesn't really do a whole lot else and its mostly to give an example of how to use portals. The loot isn't amazing and it's not that dangerous, its mostly there to show off what you can do with maps now.

This also replaces the step_trigger teleporters that are used to let people go between the plains, mountains, and wilderness, with portals, so that it visually looks nicer.

Some pictures of portals in action.
(I moved the lights after this shot away from each other)
Portal of wilderness gate

The left and right side are actually the same place.
Portals used in the PoI

What that place looks like in the map editor. Was tweaked slightly since this image was taken.
Portals in map maker

Video of how portals mostly look when crossing them. The actual crossing isn't as smooth anymore since the method used at the time of recording had a tendency to make you be unable to move forever if you crossed enough. Perhaps someday we can have fake gliding. (Also the place in the video was used only as a demo, its not in this PR.)
[Video of portals](https://streamable.com/3qfs4q)


MisterLayne:
Euclidean time.

SplinterGP:
quick question, if per example someone moves to the other side of it that the portal is showing, you can see them ?, what is the limit of things that show on it ?

Neerti:
As far as I know, all objects the mob could normally see will be seen through the other side. It works using similar functionality as the AI's picture in picture mode that was introduced recently.

lbnesquik:
This is really cool. Hope we get people using it!

PR #7380, opened 2020-07-27
Port Blast Door Crushing!

Rykka-Stormheart:
Adds blast door/shutter crushing. Does 40 damage for blast doors and 10 for shutters.

Upstream port of https://github.com/VOREStation/VOREStation/pull/8498

### DIFF
--- a/code/game/machinery/doors/airlock.dm
+++ b/code/game/machinery/doors/airlock.dm
@@ -1085,6 +1085,7 @@ About the new airlock wires panel:
 	SetWeakened(5)
 	var/turf/T = get_turf(src)
 	T.add_blood(src)
+	return 1
 
 /mob/living/carbon/airlock_crush(var/crush_damage)
 	. = ..()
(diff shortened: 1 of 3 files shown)


Woodratt:
40 damage is a bit hrm, would be more on board with it if it shoved you out of the way rather than getting nom stunlocked by a blastdoor.

Another staffer suggested the idea of there being a warning of some sort, maybe a siren or a sound of 'heavy machinery' before the doors closed, so it gave you some warning to jump out of the way. But that's probably harder to code.

Rykka-Stormheart:
> 40 damage is a bit hrm, would be more on board with it if it shoved you out of the way rather than getting nom stunlocked by a blastdoor.
> 
> Another staffer suggested the idea of there being a warning of some sort, maybe a siren or a sound of 'heavy machinery' a couple seconds before the doors closed, so it gave you some warning to jump out of the way. But that's probably harder to code.

Blast doors do not deal damage until they fully close. You have ~around~ 1 second or two to get out of the door if you see it start to close.

See

Atermonera:
Staff input: This is fine as-is, but we'll probably be looking into a) fiddling with the delay (tending towards longer) and/or b) making some kind of visual warning, sorta like how when you crowbar shut firelocks, they scrunch down a little at the start, then fully close after the do_after()

PR #7382, opened 2020-07-28
Supermatter Causality Destabilization Field

Shadow-Quill:
In short, this prevents the supermatter from immediately exploding when integrity reaches zero.
Instead, it activates a 30 second causality destabilization so that engineers either have;

A) A chance to get stability above 0% (in which case the timer resets and field disables.)
B) Run away screaming during the timer, and when the timer reaches 0, the SM will explode as normal.

The SM stabilization field announcements are conveyed over Common comms, so that everyone knows Stuff's Going Down(TM).


Cyantime:
Why make the supermatter even safer? It already loudly announces its status and gives a rough timer that way, what do engineers need with more leeway for failures/sabotage.

Shadow-Quill:
I had forgotten that Polaris had antags, to be honest. (Or at least, commonish?) Never actually played on Polaris, though I can move this down a stream if not accepted here.

zekesulastin:
> Why make the supermatter even safer? It already loudly announces its status and gives a rough timer that way, what do engineers need with more leeway for failures/sabotage.

It's possible to set up a delamination that goes off basically the instant you get a warning, but I've only seen it rarely in years (I'm fairly certain that would proc the max size explosion in the new regime, though!).  I feel inclined to agree that in most situations the current delay should be enough to fix the problem or gtfo.

Atermonera:
Staff input: This covers an extremely rare edge case where the SM delams normally in a very short span of time, and staff don't feel that that is an edge case that needs to be covered, particularly in a manner that makes normal delamination even easier to recover from

Issue #7291, opened 2020-06-14
Robot high-temperature warnings do not show.


martin509:
#### Brief description of the issue
The warning icon that appears when you get too hot as a robot does not show.
#### What you expected to happen
A warning icon like a red thermometer or something when you overheat in a suit.
#### What actually happened
No icon appeared, and I did not realize I had not put my suit cooler on until substantial damage occured.
#### Steps to reproduce
Put on a suit as a robot with no suit cooler, and wait.

PR #7384, opened 2020-07-28
Fix alerts not being thrown for body temps being too out of range

Rykka-Stormheart:
Specifically for body temperatures to high/cold.


Neerti:
Hopefully this fixes #7291

Rykka-Stormheart:
> Hopefully this fixes #7291

Can confirm it does, that was the original intent. Robots/synths use these temp bits, they don't handle breathing. The original temp alert only handled breathing specifically.

PR #7391, opened 2020-07-30
Adds tg-style fast(ish) magazine restocking.

ForFoxSake:
This PR adds tg-style magazine restocking. For the longest time tgstation has had the ability to instantly load a magazine from a pile of bullets lying on the ground.
Personally I feel like this is long over-due for being ported to virtually every branch of modern SS13, individually clicking 7 or more 2x6 pixel sprites to then click on the magazine is a long and tedious process that's fun for no one.

Now, on tgstation and it's derived code-bases the process takes place in a single game tick, which I figured would be undesirable here. To deal with that potential issue I've added a slight delay to the process of adding each bullet to the magazine. It's a small delay, that as far as I can tell, will always be faster than a player actually picking the bullet up. (I'm not gunna say for certain though, maybe someone out there can click things _really really fast_.)
But that should at least make it feel like there's an action taking place, and not that the process is being glossed over.

I've explicitly tested the code with the standard .45 pistol magazine, the .45 speedloader and the top loading 9mm magazine, to ensure that incorrect ammo types can't be snuck into magazines with this method, and that magazines can't be overloaded with this method.  As the code is located in /obj/item/ammo_magazine, almost all ammo containers will function in the same way, with the exception of shotgun shell boxes, since those are regular boxes and not ammo containers.

I am open to suggestions or requests involving this PR, but be aware that I don't check github very often, so please ping me in the discord if my attention is needed.

### DIFF
--- a/code/modules/projectiles/ammunition.dm
+++ b/code/modules/projectiles/ammunition.dm
@@ -59,7 +59,7 @@
 			if(box.caliber == bullet.caliber && bullet.BB)
 				if (boolets < 1)
 					to_chat(user, "<span class='notice'>You start collecting shells.</span>") // Say it here so it doesn't get said if we don't find anything useful.
-				if(do_after(usr,5,box))
+				if(do_after(user,5,box))
 					if(box.stored_ammo.len >= box.max_ammo) // Double check because these can change during the wait.
 						break
 					if(bullet.loc != floor)
--- a/code/modules/projectiles/ammunition.dm
+++ b/code/modules/projectiles/ammunition.dm
@@ -29,8 +29,8 @@
 	set_dir(pick(cardinal)) //spin spent casings
 	update_icon()
 
-/obj/item/ammo_casing/attackby(obj/item/weapon/W as obj, mob/user as mob)
-	if(W.is_screwdriver())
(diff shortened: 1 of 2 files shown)


code/modules/projectiles/ammunition.dm
```diff
@@ -45,6 +45,39 @@
...
+		box.reloading = TRUE
+		var/boolets = 0
+		var/turf/floor = loc
+		for(var/obj/item/ammo_casing/bullet in floor)
+			if(box.stored_ammo.len >= box.max_ammo)
+				break
+			if(box.caliber == bullet.caliber && bullet.BB)
+				if (boolets < 1)
+					to_chat(user, "<span class='notice'>You start collecting shells.</span>") // Say it here so it doesn't get said if we don't find anything useful.
+				if(do_after(usr,5,box))
```
Atermonera, line 62: `user` should be valid here, it's best practice to avoid using usr where possible

ForFoxSake: Whoops, that one slipped past my attention.

ForFoxSake:
Thrown in a changelog, fingers crossed it's descriptive enough.

PR #7392, opened 2020-07-30
Add a 'Restarting world!' notice.

lbnesquik:
Basically works like this, which is the restart vote code:

		sleep(50)
		log_game("Rebooting due to restart vote")
		world.Reboot()


Cerebulon:
You wanna put a bangin' donk on it!

PR #7398, opened 2020-07-31
Stickynote tweaks

Cerebulon:
Apparently stickynote crates were An Issue. I think it may have just been the latter but I changed 2 things for niceness:
Defined a specific crate for them to arrive in.
Changed picking up stickynotes from "be on grab intent" (unique afaik?) to "click drag to self" (same as paper bins, which are similar)

### DIFF
--- a/code/modules/paperwork/paper_sticky.dm
+++ b/code/modules/paperwork/paper_sticky.dm
@@ -1,6 +1,7 @@
 /obj/item/sticky_pad
 	name = "sticky note pad"
 	desc = "A pad of densely packed sticky notes."
+	description_info = "Click to remove a sticky note from the pile. Click-drag to yourself to pick up the stack. Sticky notes stuck to surfaces/objects will persist for 50 rounds."
 	color = COLOR_YELLOW
 	icon = 'icons/obj/stickynotes.dmi'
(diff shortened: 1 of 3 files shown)


lbnesquik:
Cheers.

code/modules/paperwork/paper_sticky.dm
```diff
@@ -50,24 +50,38 @@
...
-		to_chat(user, SPAN_NOTICE("You pull \the [paper] off \the [src]."))
-		papers--
-		if(papers <= 0)
-			qdel(src)
-		else
-			update_icon()
+		update_icon()
+
+/obj/item/sticky_pad/MouseDrop(mob/user as mob)
+	if((user == usr && (!( usr.restrained() ) && (!( usr.stat ) && (usr.contents.Find(src) || in_range(src, usr))))))
```
Atermonera, line 69: `if(user == usr && !(usr.restrained() || usr.stat) && (usr.contents.Find(src) || in_range(src, usr)))` doesn't use quite so many parentheses

Atermonera:
Oh, and please add the new usage to the desc (extended or not) so it's apparent at a glance how players can use them

PR #7403, opened 2020-07-31
Mechtoy and arcade

dwinters99:
Adds a 'Random Mech Toy' object and replaces arcade machine prize table bloat with it.


Heroman3003:
While a more 'clean' solution, I feel like it should be two or three weight worth, considering how there are single-type non-random toys in there with more weight. Overall less likelyhood of mech toys is good, but this just makes actually getting whole collection from just the arcade a looot harder.

dwinters99:
I suppose it's harder, yeah, but you can beat Orion in less than 60 seconds so any loot being 'rare' is a bit of a non-issue in my opinion.

PR #7404, opened 2020-07-31
Fuel spills no longer persistent

Cerebulon:
Title.
Due to terrible, flaming circumstances. Will now appear as generic filth cross-round.


Greenjoe12345:
could it be set up in a way so there's one or two puddles of fuel left, but most is replaced with filth?
that way it won't be as deadly as halls full of fuel

zekesulastin:
I'd rather just not have random fuel persist per round with both antag rounds and random events waiting to turn every damn room into a fire hazard.

PR #7406, opened 2020-08-01
Adds phase projectile as a firemode for tasers.

Celestius357:
Does as it says. Adds an alternative firemode to tasers so security can deal with the pesky spiders and the like.


Cyantime:
Tasers can already kill spiders though?? If security is wanting to kill angry mobs busting in with more effectiveness, shouldn't they get actual weapons from the armory?

Atermonera:
Taser base shot should be more effective at killing spiders, both in damage and range.

Atermonera:
Staff input: Tasers are already perfectly functional for the stated task, and this change wouldn't add anything meaningful. Nae

PR #7408, opened 2020-08-01
adds cold discomfort messages for tesh + one new for heat

Greenjoe12345:
messages such as "You fluff up your feathers against the cold."
though something with cold_discomfort_levels is broken, these messages can be here for when it is fixed, when that is I might see about making teshari cold resistance more realistic, and not them being able to wander around at like -180c


Greenjoe12345:
okay uh cause of github stuff being weird my New PR pulled this into it too so the features of this will be there...?

Greenjoe12345:
or not ty for helping fix it cerb

PR #7409, opened 2020-08-01
Adjusts Skrellian hair names to fit descriptions on wiki

listerla:
Very Short -> Short (Rasktina)
Short -> Average (Malish, Kanin)
Average -> Long (Talum)
Long -> Very Long (Qerr)


DasIrrlicht:
We need to add 'missing headtail left/right' for every length.

listerla:
@DasIrrlicht make the sprites and i'll dunk it into another PR

PR #7423, opened 2020-08-04
Relocate handle_ambience() to human/life() to reduce CPU usage

Rykka-Stormheart:
Turns out the /mob/living/Life() being called every second MAY be causing movement/lag issues. If this resolves it, then we know it was this, if not, then there's a further yet-to-be-found cause.


Rykka-Stormheart:
Doing an alternate fix someone suggested instead

PR #7426, opened 2020-08-04
Disable breaking fragile floor tiles with the crowbar

lbnesquik:
It's not fun, it adds nothing to the gameplay and it isn't clear what breaks and what does. I see seriously zero reasons to keep this.
This include wood and some of the floor in robotics for example.

### DIFF
--- a/code/game/turfs/simulated/floor_attackby.dm
+++ b/code/game/turfs/simulated/floor_attackby.dm
@@ -121,9 +121,9 @@
 		if(broken || burnt)
 			to_chat(user, "<span class='notice'>You remove the broken [flooring.descriptor].</span>")
 			make_plating()
-		else if(flooring.flags & TURF_IS_FRAGILE)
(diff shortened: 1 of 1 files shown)


Greenjoe12345:
maybe set it to only do it on harm intent, so it's still possible, but you can't do it quite so easily on accident

lbnesquik:
Why would you willingly want to do that? And you can just stuff it into an autolathe, disposal or chuck it into a corner if you don't want to see it.

Cyantime:
Not all flooring types have tiles, without the ability to do this there'd be no way to get to the plating under certain floors.

lbnesquik:
This doesn't disable crowbaring them. It disables BREAKING them. You can still use a crowbar to get everything out.

lbnesquik:
Updated my PR after testing. This one does work to pry off wooden tiles.

Atermonera:
Staff input: Yae

PR #7428, opened 2020-08-05
Small tweaks to cyborg verbs and icon

lbnesquik:
- Add two little sentences of feedback to the 'Emit sparks' and 'namepick' verbs.
- Change the default HUD to 'Pick me', so new players might be a tad less confused about this clunky mess.

I actually made this commit like a week ago but forgot.


lbnesquik:
Labels are outdated

code/modules/mob/living/silicon/robot/robot.dm
```diff
@@ -325,6 +325,7 @@
 /mob/living/silicon/robot/verb/Namepick()
 	set category = "Robot Commands"
 	if(custom_name)
+		to_chat(usr, "You can't pick another custom name. Go ask for a name change.")
```
Atermonera, line 328: Missing filter tags, should be of the format `<span class='[label]'> [message]</span>`, where label is either a valid styling label (notice, warning, etc.,), or a filter label (filter_notice will send the messages to the same VChat filter as notice), so the messages get properly sorted by VChat.
The other change is also missing a label.

PR #7430, opened 2020-08-06
Mulebot and vehicle updates - they now run over mobs!

Rykka-Stormheart:
Mulebots and vehicles will actually run over mobs properly now. 

Pathing fixes will be in a second PR.


Rykka-Stormheart:
Fixed issue with runtime and mulebots killing themselves, please remove DNM hold!

PR #7444, opened 2020-08-08
Removes the ability to forge plasteel from steel.

Mechoid:


polaris.dme
```diff
@@ -3017,7 +3017,7 @@
 #include "code\ZAS\Zone.dm"
 #include "interface\interface.dm"
 #include "interface\skin.dmf"
-#include "maps\southern_cross\southern_cross.dm"
+#include "maps\example\example.dm"
```
listerla, line 3020: example.dm

Mechoid: ..feck

PR #7452, opened 2020-08-09
Gives FBP's a proc to change their monitor display without using a mask item.

SplinterGP:
Makes a proc and gives them to FBP's so they can change their monitor display (if they have one), it uses same system as the monitor mask item to get the sprites and all, except it does not need an item and applies the monitor icon as the eye sprite.

had to add three lines on update_icons.dm so it updates the cache for the monitor display as otherwise it will not update the sprite correctly when you change it. the monitor display also accepts different colors by changing eye color and just applying it over the monitor sprite (depends of the image being displayed on the monitor, and dark colors work better), default for the normal sprite is black.

ezgif-2-319c37aac72a


SplinterGP:
also gave a quick fix on a monitor sprite that had a extra line out of bounds.

code/modules/mob/living/carbon/human/human_powers.dm
```diff
@@ -347,3 +347,28 @@
...
+/mob/living/carbon/human/proc/setmonitor_state()
+	set name = "Set monitor display"
+	set desc = "Set your monitor display"
+	set category = "IC"
+	if(stat)
+		to_chat(src,"<span class='warning'>You must be awake and standing to perform this action!</span>")
+		return
+	var/obj/item/organ/external/head/E = organs_by_name[BP_HEAD]
+	var/datum/robolimb/robohead = all_robolimbs[E.model]
+	if(!E)
```
Atermonera, line 360: You should enforce E (and E.model, don't know off the top of my head what happens when you index a list at null) _before_ you try and use it.

SplinterGP: oh right, i'll check that.

SplinterGP: i believe when you index a list at null or something it doens't has it just returns null, i'll move E.model below if(!E) but i think E = organs_by_name[BP_HEAD] if no head just gives back null. i'll check and update it.

Atermonera: If it's an associative list, then yeah I think that's the behaviour. Numerical lists like to runtime when you try similar stuff though, because woooo consistency!

PR #7454, opened 2020-08-10
Adds AlliCo Grab-A-Gift

Schnayy:
A claw machine just as hellish to play as the real ones. Uses TGUI & was good practice for it. Should be excluded from the mapped random arcade spawns, only gives stuffed animals, and is unique in that it demands money. Not mapping it in just yet. Tested it every way I could think of.

image

Basically the directional buttons are pointless clickspam, each win adds to the probability of winning a prize........ which starts at 0.

Once a prize is won, it's reset back to 0.

Also if you emag it the machine will steal all of your account funds and reward you with a primed frag grenade to the face.


code/game/machinery/computer/arcade.dm
```diff
@@ -1075,3 +1075,272 @@
...
+	var/winscreen = ""
+
+/// Payment
+/obj/machinery/computer/arcade/clawmachine/attackby(obj/item/I as obj, mob/user as mob)
+	if(..())
+		return
+
+	var/obj/item/weapon/card/id/W = I.GetID()
+
+	if(!emagged)
```
Atermonera, line 1106: The if and else blocks are almost entirely duplicate code. The check for emag should just go with the actual special behaviour: the cases of a wallet and cash.
You've also included some extra code in the else block that isn't in the if block.

code/game/machinery/computer/arcade.dm
```diff
@@ -1075,3 +1075,272 @@
...
+			else if(istype(I, /obj/item/weapon/spacecash))
+				playsound(src, 'sound/arcade/steal.ogg', 50, 1, extrarange = -3, falloff = 0.1, ignore_walls = FALSE)
+				to_chat(user, "<span class='info'>It doesn't seem to accept that! Seem you'll need to swipe a valid ID.</span>")
+				handled = 1
+			if(paid)
+				gamepaid = 1
+				instructions = "Hit start to play!"
+				return
+			else if(handled)
+				return // don't smack that machine with your 2 thalers
```
Atermonera, line 1149: Handled doesn't do anything here, because the code would naturally fall out the end of this block and hit the end of the proc anyways.

code/game/machinery/computer/arcade.dm
```diff
@@ -1075,3 +1075,272 @@
...
+				var/obj/item/weapon/spacecash/ewallet/C = I
+				paid = pay_with_ewallet(C)
+				handled = 1
+			else if(istype(I, /obj/item/weapon/spacecash))
+				var/obj/item/weapon/spacecash/C = I
+				paid = pay_with_cash(C, user)
+				handled = 1
+
+			if(paid)
+				gamepaid = 1
```
Atermonera, line 1124: You can just assign `gamepaid = [pay_woth_x()]` and check the value of that in this if, the paid car is redundant

code/game/machinery/computer/arcade.dm
```diff
@@ -1075,3 +1075,272 @@
...
+	if(action == "pointless" && wintick >= 10)
+		instructions = "Insert 1 thaler or swipe a card to play!"
+		clawvend()
+
+/obj/machinery/computer/arcade/clawmachine/proc/clawvend() /// True to a real claw machine, it's NEARLY impossible to win.
+	winprob += 1 /// Yeah.
+
+	if(prob(winprob)) /// YEAH.
+		if(!emagged)
+			playsound(src, 'sound/arcade/Ori_win.ogg', 50, 1, extrarange = -3, falloff = 0.1, ignore_walls = FALSE)
```
Atermonera, line 1313: Playsound() and Winston are independent of the if-else clause

Atermonera:
@Schnayy this still has merge conflicts, any plans on fixing those?

Schnayy:
Ack, yes, will get to it soon.

code/game/machinery/computer/arcade.dm
Schnayy: Not sure what you mean here.

Schnayy:
Still need to do a few more changes, so don't merge until I've had a chance to test.

PR #7467, opened 2020-08-11
Adds New Prosthetics Set

SubberTheFabulous:
Clothing and mismatching friendly, supposed to represent an outdated design of an already existing Cyber Solutions set, or whatever.

Deletes/updates Cyber Solutions - Array. I don't think anyone ever used it, anyway.


Cerebulon:
I'd prefer this didn't remove the old cybersolutions alt-head. It's entirely possible somebody's character IS using it, and it's pretty visually different. Could you readd that as a new alternate maybe?

SubberTheFabulous:
I feel like it's not that visually different. Outdated head is just a slightly chonkier version of Array, but, sure, I'll readd it.

PR #7471, opened 2020-08-12
Reverts the plasteel cost for the durand and their fabrication time.

lbnesquik:
- I have no idea what the doubled fabrication time added to the experience, and the plasteel cost is just a pain in the butt.

### DIFF
--- a/code/modules/research/mechfab_designs.dm
+++ b/code/modules/research/mechfab_designs.dm
@@ -177,57 +177,57 @@
 	name = "Durand Chassis"
 	id = "durand_chassis"
 	build_path = /obj/item/mecha_parts/chassis/durand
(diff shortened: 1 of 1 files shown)


Mechoid:
https://github.com/PolarisSS13/Polaris/pull/7473

Plasteel was kept specifically due to the durability of the Durand.

lbnesquik:
I would had rather not buffed the durand and kept the price the same.

Atermonera:
Staff input: going with mechoid's tweaks, then evaluating the results.
And I don't know precisely when the durand was buffed, but I don't believe it was strictly accompanied by the cost changes, they were rather applied after the fact. @Mechoid I think might be able to confirm/deny this

PR #7475, opened 2020-08-13
Hydro upkeep

Mechoid:
Renames Space Spice to Wurmwoad, the name of the new "plant" that produces it.

Adds Wurmwoad, a suspiciously worm-like plant that produces pods of spice.
Adds Wurmwoad to the service borg synthesizer.
Allows borgs to work with the New Kitchenware.


Atermonera:
Borg gripper stuff has been tested

PR #7479, opened 2020-08-13
Replaces water + fuel tanks and bed

listerla:
Sprites ported from Aurora + CEV-Eris.

Also changes objects.dmi bed sprite.

Don't merge unless staff have agreed on these ones during the next meeting because I recall them being controversial in the past.


listerla:
Removed dispenser change, will have as a separate PR.

PR #7486, opened 2020-08-13
Adds metal detectors

listerla:
From https://github.com/GeneriedJenelle/The-World-Server-Redux/pull/932

Other enhancements also taken from that PR.

Metal detectors check if you have a gun, ammo, or similar and if so blare an alarm and inform security. If you have security access, this does not occur.


code/_helpers/unsorted.dm
```diff
@@ -311,26 +311,20 @@ Turf and target are seperate in case you want to teleport some distance from a t
 					break
 
 		//update our pda and id if we have them on our person
-		var/list/searching = GetAllContents(searchDepth = 3)
-		var/search_id = 1
-		var/search_pda = 1
+		var/list/searching = src.GetAllContents()
```
ShadowLarkens, line 314: no src.

Why did you get rid of the searchDepth option in GetAllContents? It's deliberately there to prevent way too much recursion from happening

code/_helpers/unsorted.dm
```diff
@@ -311,26 +311,20 @@ Turf and target are seperate in case you want to teleport some distance from a t
...
 
 		for(var/A in searching)
-			if( search_id && istype(A,/obj/item/weapon/card/id) )
+			if(istype(A,/obj/item/weapon/card/id))
```
ShadowLarkens, line 317: Similarly, why did you get rid of the search_id thing here? It's there to keep this loop from arbitrarily looping over the whole list once it's found it's actual target. Is there some need to make sure that every single ID and PDA on someone with their registered name is updated?

code/_helpers/unsorted.dm
```diff
@@ -608,15 +602,25 @@ proc/GaussRandRound(var/sigma,var/roundto)
...
-/atom/proc/GetAllContents(searchDepth = 5)
-	var/list/toReturn = list()
-
-	for(var/atom/part in contents)
-		toReturn += part
-		if(part.contents.len && searchDepth)
-			toReturn += part.GetAllContents(searchDepth - 1)
-
-	return toReturn
+/atom/proc/GetAllContents(var/T)
```
ShadowLarkens, line 605: If you absolutely must change this proc, at least change it to /tg/'s version [here](https://github.com/tgstation/tgstation/blob/master/code/__HELPERS/unsorted.dm) that avoids unnecessary list creation and calls to Cut().

code/game/machinery/metaldetector.dm
```diff
@@ -0,0 +1,87 @@
+GLOBAL_LIST_INIT(metal_detector_items, typecacheof(list(
+	/obj/item/weapon/gun,
+	/obj/item/weapon/material,
+	/obj/item/weapon/melee,
+	/obj/item/device/transfer_valve,
+	/obj/item/weapon/grenade/,
```
ShadowLarkens, line 6: Paths shouldn't be ended with /

code/game/machinery/metaldetector.dm
```diff
@@ -0,0 +1,87 @@
...
+	/obj/item/ammo_casing/,
```
ShadowLarkens, line 7: Paths shouldn't be ended with /

code/game/machinery/metaldetector.dm
```diff
@@ -0,0 +1,87 @@
...
+
+	if((stat & (NOPOWER|BROKEN)))
+		return
+
+	if(!istype(M))
+		return
+	if(allowed(M))
+		return
+
+	var/list/items_to_check = M.GetAllContents()
```
ShadowLarkens, line 68: Again with the lack of recursion limits, I'm a little concerned that someone with a large enough inventory could cause actual lag spikes by just walking through a metal detector

code/game/machinery/metaldetector.dm
```diff
@@ -0,0 +1,87 @@
...
+	idle_power_usage = 100
+
+	plane = MOB_PLANE
+	layer = ABOVE_MOB_LAYER
+
+	var/next_announcement_time	// so this doesn't get spammy.
+
+	var/alarm_delay = 2000
+
+/obj/machinery/metal_detector/New()
```
Rykka-Stormheart, line 27: This should be Initialize()

code/game/machinery/metaldetector.dm
```diff
@@ -0,0 +1,87 @@
...
+	..()
```
Rykka-Stormheart, line 28: This should be . = ..()

code/game/machinery/metaldetector.dm
```diff
@@ -0,0 +1,87 @@
...
+	next_announcement_time = get_game_time() - alarm_delay
+
+	component_parts = list()
```
Rykka-Stormheart, line 31: This entire section should use `default_apply_parts()`, instead of defining components individually. The proc is found in machinery.dm line 308.

Make a circuitboard:
`/obj/item/weapon/circuitboard/metaldetector`, and place it in `code/game/objects/items/weapons/circuitboards/machinery/metaldetectors.dm`

Define the name, built_path as the metaldetector, the board_type as a frame (you make the frame in `code/game/machinery/frame.dm` and path it `/datum/frame/frame_types/metaldetector`), you define the matter as whatever you prefer for making boards, default is wallmaterial = 50, and req_components is that components list you had originally.

For reference, said components list would look like:
`req_components = list(
							/obj/item/weapon/stock_parts/console_screen = 1,
							/obj/item/weapon/stock_parts/scanning_module = 3)`
Circuitboard not included because we define it earlier on the object.

Then, to enable it being printable, go to:
`code/modules/research/designs/circuits/circuits.dm` and add a build path:
`/datum/design/circuit/metaldetector` and set the build_path to the circuitboard we defined earlier.

Then, in metaldetector.dm, set `circuit = /obj/item/weapon/circuitboard/metaldetector`

And that's it, your items are now createable in-game, using the correct method for generating components!
You'll also want to remove `RefreshParts()` on the init because default_apply_parts() does that for you.

listerla:
@ShadowLarkens @Rykka-Stormheart Thanks for the reviews! This was pretty much a straight port, I'm not super familiar w/ BYOND, but I think I'm starting to get a few things from this.

code/game/machinery/metaldetector.dm
listerla: I actually knew this one, I just forgot to push the commit for it. I'll batch all the fixes into one commit now.

code/game/machinery/metaldetector.dm
```diff
@@ -0,0 +1,82 @@
...
+	var/alarm_delay = 2000
+
+/obj/machinery/metal_detector/Initialize()
+	. = ..()
+	next_announcement_time = get_game_time() - alarm_delay
+	default_apply_parts()
+
+/obj/machinery/metal_detector/examine(mob/user)
+	if((stat & NOPOWER))
+		to_chat(user, "An advanced metal detector used to detect weapons. It is currently unpowered.")
```
Atermonera, line 36: Noncompliant with current examine text usage. Syntax should be:
> . = ..()
> . += [message]

code/game/machinery/metaldetector.dm
Atermonera: Seconded.

listerla: Would the recursion limits be the thing I had prior removed out of stupidity and now reinstated, or is there something that must be changed here?

Atermonera: With `helpers/unstorted.dm` untouched, this will use the default depth of 5.
The reason that a lack of limitation is concerning is because, if someone has boxes stuffed inside backpacks stuffed inside bags of holding ad nauseam, etc., then getAll Contents will iterate over every single item, which for inventories with a large number of items, could potentially cause the server to hang.

Atermonera:
@listerla do you have any plans to address the changes requested by both myself and Nyria?

Atermonera:
No response for 30+ days, closing. If the author wishes for this to be reopened, then they'll have to get in contact with one of the devs.

PR #7487, opened 2020-08-13
Adds shift-middle click pointing

listerla:
The title seems self-explanatory. Taken from Citadel Station 13 RP.


lbnesquik:
Love you for this.

code/modules/mob/living/living.dm
```diff
@@ -46,7 +46,7 @@
 
 //mob verbs are faster than object verbs. See above.
 /mob/living/pointed(atom/A as mob|obj|turf in view())
-	if(src.stat || !src.canmove || src.restrained())
```
listerla, line 49: This is so you can point while buckled, by the way.

PR #7496, opened 2020-08-13
Add a plastic knife to the pizzzaaaaa crate.

lbnesquik:
- That one friend who always has a knife will be disappointed now.

### DIFF
--- a/code/datums/supplypacks/hospitality.dm
+++ b/code/datums/supplypacks/hospitality.dm
@@ -57,13 +57,12 @@
 	group = "Hospitality"
 
 /datum/supply_pack/randomised/hospitality/pizza
-	num_contained = 5
 	contains = list(
-			/obj/item/pizzabox/margherita,
-			/obj/item/pizzabox/mushroom,
-			/obj/item/pizzabox/meat,
-			/obj/item/pizzabox/vegetable,
-			/obj/item/pizzabox/pineapple,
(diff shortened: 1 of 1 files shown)


code/datums/supplypacks/hospitality.dm
```diff
@@ -63,7 +63,8 @@
 			/obj/item/pizzabox/mushroom,
 			/obj/item/pizzabox/meat,
 			/obj/item/pizzabox/vegetable,
-			/obj/item/pizzabox/pineapple
+			/obj/item/pizzabox/pineapple,
+			/obj/item/weapon/material/knife/plastic
```
Atermonera, line 67: Unfortunately, because this crate has `	num_contained = 5` defined, it picks [5] elements from this list at random. Which, with this change, _could_ include 1 knife. Or 5 knives and no pizza.

Atermonera:
DNM on resolving the noted issue re: 5 knives and no pizza

code/datums/supplypacks/hospitality.dm
lbnesquik: Hnngh.

lbnesquik:
@Atermonera I just made all of the pizza random.

Atermonera:
If that mysteriously doesn't work then I guess someone's gonna have a bad pizza party. Looks like it should be fine to me tho

PR #7459, opened 2020-08-11
Massive Ambience and Sound Overhaul + Addition

Rykka-Stormheart:
Includes changes from PR's:
https://github.com/Aurorastation/Aurora.3/pull/8861
https://github.com/tgstation/tgstation/pull/46677
https://github.com/tgstation/tgstation/pull/46387
https://github.com/CHOMPStation2/CHOMPStation2/pull/463
https://github.com/CHOMPStation2/CHOMPStation2/pull/108
https://github.com/tgstation/tgstation/pull/45462

Other changes include:
- New Security Alert for Red + Delta
- New Ambience per-area
- New SM sounds
- New Destructive Analyzer Sounds
- New Scanner/Cloner/Resleever Sounds
- Knuckle-cracking *crack emote
- Hyena *cackle emote (Virgo only)
- New Sounds for Oven/Grill/Candy/Cereal/Mixer Machines
- New Sound for Pre-Game Lobby (ping)
- Adds Preference for Pickup Sounds (Same as Drop Sounds)

PR #7504, opened 2020-08-15
Beginnings of a global volume control system

ShadowLarkens:
This adds the backend for allowing clients to control different volume "channels" (read: not byond sound channel, just a conceptual channel) almost like this game wasn't made in the 2000s.  You can set a given volume channel (conceptual channel) between 0 and 200%, which acts as a multiplier for all sounds using that (conceptual) channel.

as a note, I added a TGUI panel, but it's disabled in this PR pending #7397 

Note that this system is intentionally super barebones. @Rykka-Stormheart was interested in deciding what new "channels" to make and going through the rather extensive process of going through different playsound() calls and deciding what channel they should use.

Also note that this doesn't remove or touch anything to do with the existing sound on/off preferences. They should probably be moved over to just being sliders where 0 is off eventually, but I don't wanna cause too many breaking changes with this PR.


Rykka-Stormheart:
DNM this before #7397, this has dependency on the menu.

@ me or @ Shadow once 4.0 is in and we will uncomment the menu. Merge this BEFORE #7459 if possible.

Thanks!

PR #7511, opened 2020-08-16
Raises Xion and Grayson monitor animation delay from 1 to 6

listerla:
Xion and Grayson monitors are no longer strobe lights, they're actually tolerable.


Atermonera:
@listerla this still has merge conflicts, any plans on fixing those?

PR #7512, opened 2020-08-16
Changes is_type(x, /path) to applicable helpers

Shadow-Quill:
# For the love of all that is holy, please testmerge this before merging.

ie: `is_type(M, /mob)` would be changed to `ismob(M)`

This will almost certainly conflict with some other PRs.
This also adds `#define isdrone(A)		istype(A, /mob/living/silicon/robot/drone)` and
`#define issimulated(A)  istype(A, /turf/simulated)`

This also probably doesn't need a changelog, as all code changed should have no effect on gameplay.

TODO:
- [x] Check datum/weakref helpers
- [x] isatom & isclient
- [x] /list/ helper
- [x] /area/ helper
- [X] /mob/ helpers
- [x] Writhe in pain
- [x] /turf/ helpers

###### PS: Maintainers, I'm sorry.


Shadow-Quill:
Please let me know if I overlooked typing something (like using isanimal for a simple animal), or some other change snuck in.

Atermonera:
`[2020-08-21T06:02:58] Runtime in atoms.dm,507: undefined proc or verb /atom/movable/lighting_overlay/show message().`

Shadow-Quill:
Fixed.
I do not promise copious amounts of runtimes won't happen if on live.

Atermonera (changes requested):
Finally done, this has taken a lot longer than it need have because I had to scrutinize each change so much to spot errors.
>10% /file failure rate at file 532/556, reviewing to completion because there's so little left.
Final per-file failure rate is approximately 56/556 files, or about 10% of all files have some sort of error. When I discovered the first major errors at file 128 or so, it was not so burdensome to re-review those files to find other instances. When I discovered a whole new class of error on file _40_8, I _cannot_ re-review all previous files.

While I appreciate the desire to get all of these done at once, and am sorry that this has taken so long, the rate of errors and effort required to satisfactorily determine that all errors are resolved means that this PR _must_ be split up into multiple PRs. These are too many slightly-similar changes stuck together, that the risk of errors like those found causing subtle (or extremely flagrant) shifts in behaviour is unacceptably high. I am sorry that more work is required of this, but having reviewed all 556 files here, that is my conclusion as a developer and code maintainer.

PR #7522, opened 2020-08-19
Improve cleanbot ai slightly

Meghan-Rossi:
*Multiple cleanbots will no longer attempt to clean the same tile at the same time.
*Cleanbots will now clean tiles closer to them first.


Atermonera:
Performing a test:

PR #7530, opened 2020-08-20
Disables Swarmer Drones

Neerti:
I actually forgot to do this before the event system went permanently live so that's on me.
Anyways this turns off swarmer drones from being able to be picked by the GM system, since right now they're, to put it charitably, poorly balanced and implemented (being super-maint drones is very clunky to me but thats not the main thing).
When the GM system spawns one in, it tends to result in a ghost becoming the drone, then that player essentially never being able to lose due to infinite re-tries and the crew has to fight off drones for the rest of the round or until they or the drone gets bored.
There's been quite a few number of rounds that have devolved due to them so for now it's probably best to turn them off for the time being.


Atermonera:
Doing a quick test:

PR #7554, opened 2020-08-23
Fixes turf slowdown not applying on non-water/snow turfs

Heroman3003:
why.


Heroman3003:
Okay, I genuinely have no idea why travis decided to commit die on this.

PR #7579, opened 2020-08-27
Makes borg cell replacement take priority over petting

Heroman3003:
Because its dumb that fluff interactions override important ones in such obtuse ways. Should be other way around. Now if their cover is open any hand interaction takes out the cell properly.


listerla:
This is way better, thanks. Sorry I didn't even think to implement it this way in the first place.

PR #7583, opened 2020-08-27
FBP Internals Additions

KillianKirilenko:
Having a synthetic bloodstream isn't very useful if you can't inject things into it directly. Something advanced like a Vey-Med should be able to handle this, and most other models probably have ports/taps for maintenance purposes.

I also briefly tried to make it so you could draw out oil/coolant but it wouldn't play nice, so I left that untouched for now. It's a one-way trip.

This also fixes the borg hypodermic being under borg**hydro**.dm, for some reason.


KillianKirilenko:
I could've sworn I made a changelog for this one. Give me a minute.

PR #7585, opened 2020-08-28
Adds two effect types for PoIs and theoretical future bossmobs.

Mechoid:
On the tin, yes the icons are Sad and lame for now.


code/game/objects/effects/prop/snake.dm
```diff
@@ -0,0 +1,125 @@
...
+	if(Creator)
+		creator = Creator
+
+	..()
+
+/obj/effect/temporary_effect/pulse/snake/pulse_loop()	// Override needed unfortunately to handle the possibility of not finding a target turf.
+	set waitfor = FALSE
+
+	while(pulses_remaining)
+		sleep(pulse_delay)
```
Atermonera, line 38: Is this inside of a spawn, or will it hang everything?

Mechoid: waitfor = FALSE sets it so that it's functionally in a spawn

PR #7595, opened 2020-08-29
Adjusts a few of the new airlock sounds

Woodratt:
- Departmental airlocks now all sound the same (same sound as the default airlock, one with the green stripes)
- Glass Airlocks now sound like the command airlocks sounded previously
- Maintenance airlocks now sound like the glass airlock sounded previously


Atermonera:
DNM pending results of #7597

PR #7597, opened 2020-08-29
Adds Toggles to Revert to Legacy Door Sounds and Enable Department Doors

Rykka-Stormheart:
Self-explanatory.

Adds toggle to go back to the old-school SS13 legacy airlock sounds, including the glass windoor airlock sounds, and a toggle to disable the per-department door sounds, which some found jarring/unrealistic.

If Per-Department sounds are disabled, the department doors will use the default sounds except for open/close except for glass airlocks, which use cmd3o/c, and maintenance, which uses hall1o/c, following maintainer-requested behavior for these doors.
PR with changes from staff: https://github.com/PolarisSS13/Polaris/pull/7595

This also relocates the following sounds to be inside machines/door for more consistency:
- airlock.ogg
- airlockclose.ogg
- airlockforced.ogg
- airlock_creaking.ogg
- boltsup.ogg
- boltsdown.ogg
- blastdoorclose.ogg
- blastdooropen.ogg
- windowdoor.ogg 

Demonstration Here (recorded before the change to make department doors use maintainer-chosen audio if department was disabled): https://giant.gfycat.com/WhichUniformBrahmancow.webm


Rykka-Stormheart:
Additionally: This supersedes https://github.com/PolarisSS13/Polaris/pull/7595 because it allows the new department sounds to be toggle-able - and if requested, the toggle can default off for department-specific doors.

code/game/machinery/doors/airlock.dm
```diff
@@ -1083,10 +1088,41 @@ About the new airlock wires panel:
 	use_power(360)	//360 W seems much more appropriate for an actuator moving an industrial door capable of crushing people
 
 	//if the door is unpowered then it doesn't make sense to hear the woosh of a pneumatic actuator
-	if(arePowerSystemsOn())
-		playsound(src, open_sound_powered, 50, 1, volume_channel = VOLUME_CHANNEL_DOORS)
-	else
-		playsound(src, open_sound_unpowered, 75, 1, volume_channel = VOLUME_CHANNEL_DOORS)
+	for(var/P in player_list)
```
Atermonera, line 1091: I feel like it may be more efficient to just move all of this into a `playsound_if_pref_enabled()` proc that just takes ~~the pref-on and pref-off sounds and the path of the pref to check~~ matching lists of sounds and preferences to check, in order of priority, in addition to all of playsound()'s arguments. I think a list of toggles is more extensible than trying to put the two preferences into a single numeric pref that acts as a finite-state-machine.

Certainly better than just copypasting it into doorcode and making it kinda snowflakey

code/game/machinery/doors/airlock.dm
```diff
@@ -1083,10 +1088,41 @@ About the new airlock wires panel:
...
+		var/department_door_sounds = M.client.is_preference_enabled(/datum/client_preference/department_door_sounds)
+		var/sound
+		var/volume
+		if(old_sounds) // Do we have old sounds enabled? Play these even if we have department door sounds enabled.
+			if(arePowerSystemsOn())
+				sound = legacy_open_powered
+				volume = 50
+			else
+				sound = open_sound_unpowered
+				volume = 75
```
Atermonera, line 1105: Failing the first suggestion, this repeated else should probably instead go in an `if(!arePoweredSystemsOn())` check, the else of which instead contains the if-else that selects among the other sounds.

code/game/machinery/doors/airlock.dm
```diff
@@ -1181,10 +1217,41 @@ About the new airlock wires panel:
...
+				sound = close_sound_powered
+				volume = 50
+			else
+				sound = open_sound_unpowered
+				volume = 75
+
+		var/turf/T = get_turf(M)
+		var/distance = get_dist(T, get_turf(src))
+		if(distance <= world.view * 2)
+			if(T && T.z == get_z(src))
```
Atermonera, line 1253: Out of scope, I think (P. sure playsound doesn't do any multiz stuff) but should doors seen from above play their opening sound? Or more broadly should playsound be given a multiz-considering counterpart?

PR #7599, opened 2020-08-30
Replaces modular components with sprites from Baystation

listerla:
Source: https://github.com/Baystation12/Baystation12/pull/27786


Cyantime:
Those battery sprites are worse, they got rid of actual different sized batteries for the same sprite with a different color stripe.

Atermonera:
That isn't so different from the precedent of power cells, which just have the red spots on top to indicate capacity rather than any sort of size difference (Though I won't argue with any claims that those sprites are themselves old and decrepit).

In either case this was discussed at the staff meeting and nobody was particularly against it, but most were surprised to learn that these things actually had sprites and weren't just nebulous things that lived inside the computers.

Cerebulon:
Why are you closed? Why are you closed to the public?

listerla:
thinki

listerla:
I agree that some are at best sidegrades after looking over it (not CPUs tho), so I'd like this looked over for what's to be kept and what's to be tossed.

Schnayy:
Okay, genuine apology that this has taken so long. Are you ready for this to be merged as-is?

PR #7601, opened 2020-08-30
Remove Step_delay from all mech equipment.

lbnesquik:
My issue is that slowdown is one of the least fun things to have to deal with in this game. To my understanding, the goal of most mech PR's lately wa make mechs more fun, so this is pretty much in direct conflict.
The info on slowdown is also very much hidden with essentially zero way for someone to know that this particular thing causes slowdown.
And if the weapon is not balanced, i'd much rather we ask ourselves if that weapon really is unbalanced and take other actions to make it so.
Not to mention with how many novelty weapons we have, there is a LOT of them which are just underpowered and situational as hell. Like the incendiary weapons which don't really affect creatures, the main target of mechs in our servers.

### DIFF
--- a/code/game/mecha/equipment/tools/armor_melee.dm
+++ b/code/game/mecha/equipment/tools/armor_melee.dm
@@ -9,8 +9,6 @@
 	var/deflect_coeff = 1.15
 	var/damage_coeff = 0.8
 
-	step_delay = 0.5
-
 	equip_type = EQUIP_HULL
 
 /obj/item/mecha_parts/mecha_equipment/anticcw_armor_booster/get_equip_info()
--- a/code/game/mecha/equipment/tools/armor_ranged.dm
+++ b/code/game/mecha/equipment/tools/armor_ranged.dm
@@ -9,8 +9,6 @@
 	var/deflect_coeff = 1.15
 	var/damage_coeff = 0.8
 
-	step_delay = 1
-
 	equip_type = EQUIP_HULL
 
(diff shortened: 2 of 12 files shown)


Mechoid:
Hokay. Do you have a recording of your testing done to show what you're classifying as too extreme of slowdown? As the PRs have been intended to be both fun and maintain a measure of balance. The problem is not specifically any one weapon or armor, it is combinations of such that can cause spikes of power that aren't intended. A Durand with a ranged armor booster, scattershot, AC 2, and energy relay, is generally strong enough to hold its own against a crew, or swarm of spiders.

lbnesquik:
Any crew member equipped with gear that takes as much effort as a Durand to get can hold its own against spiders and crew, provided they don't blindly rush forward. 

If you can get a Durand, you can get a laser canon, self recharging laser, SABER, EMP shells from R&D. 
And from cargo, you can order combat shotguns, extra armor, ion rifle, phoron bores, extra lasers and so on. 
And they can also make EMP grenades from Chemistry/R&D with the uranium. 
And there is always the whole ion rifle at the armory. 

Given it is an HRP server, we can assume that people don't just randomly start shit for the heck of it every single shift. So i can't think of many scenarios where a Durand or Gygax ends up in a firefight with crew where no one has any counter.

lbnesquik:
Ion weapons could be made easier to receive as a compromise.

lbnesquik:
#7620

DerBlitzkrieg:
Hell yea, make mechs not painful to play.

Mechoid:
Exosuits have been given an encumbrance system, which has, seemingly, solved the problem. If there's further concerns after using it, feel free to re-open this or another PR.

PR #7608, opened 2020-08-31
Kaleidoscope, Toolbox

Mechoid:
Adds:
  - rscadd: "Cosmetic genemods added."
  - rscadd: "Virgo sprite accessories ported, most in background for event toolbox."


code/modules/mob/new_player/sprite_accessories.dm
```diff
@@ -1124,25 +1168,25 @@
 
 //Skrell 'hairstyles'
 /datum/sprite_accessory/hair/skr_tentacle_veryshort
-	name = "Skrell Short Tentacles"
-	icon_state = "skrell_hair_short"
-	species_allowed = list(SPECIES_SKRELL)
+	name = "Skrell Very Short Tentacles"
```
listerla, line 1171: You've reverted the change to Skrell hair here, which is a step back regarding accuracy to the wiki and will also break save files made since then.

Novacat:
Hm, if we have to make an override of those files, we would probably have to do something wierd like _vr_vr

code/_helpers/global_lists.dm
```diff
@@ -231,3 +231,68 @@ var/global/list/string_slot_flags = list(
...
 //Hexidecimal numbers
 var/global/list/hexNums = list("0", "1", "2", "3", "4", "5", "6", "7", "8", "9", "A", "B", "C", "D", "E", "F")
+
+/**
+ * VOREStation global lists
+*/
+
+var/global/list/ear_styles_list = list()	// Stores /datum/sprite_accessory/ears indexed by type
+var/global/list/tail_styles_list = list()	// Stores /datum/sprite_accessory/tail indexed by type
+var/global/list/wing_styles_list = list()	// Stores /datum/sprite_accessory/wing indexed by type
```
Atermonera, line 241: Should these not live on glob? And also you appear to have misplaced some vore edit tags

code/game/objects/items.dm
```diff
@@ -787,7 +787,7 @@ modules/mob/living/carbon/human/life.dm if you die, you will be zoomed out.
 	icon = 'icons/obj/device.dmi'
 
 //Worn icon generation for on-mob sprites
-/obj/item/proc/make_worn_icon(var/body_type,var/slot_name,var/inhands,var/default_icon,var/default_layer)
+/obj/item/proc/make_worn_icon(var/body_type,var/slot_name,var/inhands,var/default_icon,var/default_layer,var/icon/clip_mask = null) //VOREStation edit - add
```
Atermonera, line 790: couple more edit tags in here

code/modules/client/preference_setup/virgo/01_spriteaccessories.dm
```diff
@@ -0,0 +1,298 @@
+// Global stuff that will put us on the map
```
Atermonera, line 1: What parts of this are used?

code/modules/clothing/clothing.dm
```diff
@@ -987,6 +987,45 @@
...
 	update_clothing_icon()
 
+//Switch to taur sprites if a taur equips
+/obj/item/clothing/suit
+	var/taurized = FALSE
+
+/obj/item/clothing/suit/equipped(var/mob/user, var/slot)
+	var/normalize = TRUE
+
+	//Pyramid of doom-y. Improve somehow?
```
Atermonera, line 997: > if(ishuman(user))
> > if((taurized && !isTaurTail(H.tail_style)) || (!taurized && isTaurTail(H.tail_style))
> > > taurize()

And then extract the actual sprite swapping stuff to a separate proc. I combined the if statements such that a single proc can be used to toggle it on and off, switching off of `taurized`

code/modules/clothing/clothing.dm
```diff
@@ -987,6 +987,45 @@
...
+	if(!taurized && slot == slot_wear_suit && ishuman(user))
+		var/mob/living/carbon/human/H = user
+		if(isTaurTail(H.tail_style))
+			var/datum/sprite_accessory/tail/taur/taurtail = H.tail_style
+			if(taurtail.suit_sprites && (get_worn_icon_state(slot_wear_suit_str) in cached_icon_states(taurtail.suit_sprites)))
+				icon_override = taurtail.suit_sprites
+				normalize = FALSE
+				taurized = TRUE
+
+	if(normalize && taurized)
```
Atermonera, line 1007: if a taur equips this, and then another taur equips it (consecutively), it will return to the non-taur sprite for the second taur. `normalize` becomes useless with the above suggested changes to the nested ifs

code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -86,11 +87,13 @@ var/global/list/damage_icon_parts = list() //see UpdateDamageIcon()
 #define LEGCUFF_LAYER			25		//Same as handcuffs, for legcuffs
 #define L_HAND_LAYER			26		//Left-hand item
 #define R_HAND_LAYER			27		//Right-hand item
-#define MODIFIER_EFFECTS_LAYER	28		//Effects drawn by modifiers
-#define FIRE_LAYER				29		//'Mob on fire' overlay layer
-#define WATER_LAYER				30		//'Mob submerged' overlay layer
-#define TARGETED_LAYER			31		//'Aimed at' overlay layer
-#define TOTAL_LAYERS			31//<---- KEEP THIS UPDATED, should always equal the highest number here, used to initialize a list.
+#define WING_LAYER				28		//VOREStation edit. Simply move this up a number if things are added.
```
Atermonera, line 90: vore edit tags

code/modules/mob/new_player/sprite_accessories_extra.dm
```diff
@@ -0,0 +1,2361 @@
+/datum/sprite_accessory/marking/vr
```
Atermonera, line 1: This file should _prooobably_ be broken up by subtype

Atermonera (changes requested):

code/_helpers/global_lists.dm
Mechoid: These and the other preference setup lists are not GLOB'd assumedly because they need to exist from before GLOB is actually initialized.

code/modules/client/preference_setup/virgo/01_spriteaccessories.dm
Mechoid: All parts, but not all are 'used' for non-event species.

PR #7617, opened 2020-09-02
MECH TRAMPLING

lbnesquik:
It is a port from Aurora.
Any resting mob will get 15 DMG if they get walked over by a mech because stompy. Small simple mobs get gibbed if they are dead.
Damage based on actuators atm, given we don't have legs modifiers.
https://github.com/Aurorastation/Aurora.3/pull/7791

### DIFF
--- a/code/game/mecha/mecha.dm
+++ b/code/game/mecha/mecha.dm
@@ -869,7 +869,7 @@
 			if(istype(L, /mob/living/simple_mob))
 				if(issmall(L) && (L.stat == DEAD))
 					L.gib()
-					playsound(src, 'sound/effects/splat.ogg' 50, 1)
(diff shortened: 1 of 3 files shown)


DerBlitzkrieg:
This is an amazing and pretty amusing addition.

code/game/mecha/mecha.dm
```diff
@@ -843,6 +843,40 @@
 		. = ..(obstacle)
 	return
 
+
+/obj/mecha/proc/trample(var/mob/living/H)
+	if(!isliving(H))
+		return
+
+	var/obj/item/mecha_parts/component/actuator/actuator = internal_components[MECH_ACTUATOR]
```
Mechoid, line 851: This should probably check if the exosuit is phasing, since they're not entirely corporeal.

PR #7618, opened 2020-09-02
Revert strafing speed nerf.

lbnesquik:
A 1.5x modifier was added to strafing speed without much of any info. This is not something that should had been added without at least clearly saying it in the PR that introduced it.

Maybe it could comeback for shitty actuators, but seeing as we only have a single actuator variant at the moment, this has to go.

For an example, strafing with a Durand drops your speed down to un-upgraded upgraded Ripley levels. That's SLOW.


Mechoid:
Going to mark this as a results-to-be-reviewed, as I am still skeptical of the balance involved.

PR #7621, opened 2020-09-03
Bring shotgun slugs AP potential to 20%

lbnesquik:
This is to help them combat certain armored threats.


Mechoid:
May see issues arise as they are already the primary shotgun shell used, though numbers can be re-tweaked if so.

PR #7622, opened 2020-09-03
Changes default skin tone from 35 to 110

listerla:
Why? It's the center ground between 1 and 220.

### DIFF
--- a/code/modules/client/preference_setup/general/03_body.dm
+++ b/code/modules/client/preference_setup/general/03_body.dm
@@ -449,7 +449,7 @@ var/global/list/valid_bloodtypes = list("A+", "A-", "B+", "B-", "AB+", "AB-", "O
 			pref.r_hair = 0//hex2num(copytext(new_hair, 2, 4))
 			pref.g_hair = 0//hex2num(copytext(new_hair, 4, 6))
 			pref.b_hair = 0//hex2num(copytext(new_hair, 6, 8))
-			pref.s_tone = 0
+			pref.s_tone = -75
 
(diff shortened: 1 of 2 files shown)


Mechoid:
Changes aren't actually contentious, but changing the default skin-tone doesn't touch any existing characters, and is a tweak most people set anyway.

code/modules/client/preference_setup/general/03_body.dm
```diff
@@ -449,7 +449,7 @@ var/global/list/valid_bloodtypes = list("A+", "A-", "B+", "B-", "AB+", "AB-", "O
 			pref.r_hair = 0//hex2num(copytext(new_hair, 2, 4))
 			pref.g_hair = 0//hex2num(copytext(new_hair, 4, 6))
 			pref.b_hair = 0//hex2num(copytext(new_hair, 6, 8))
-			pref.s_tone = 0
+			pref.s_tone = -75
```
Mechoid, line 452: Have you tested that this doesn't mess with non-human colors?

listerla: Played fine with my Skrell.

Mechoid: Even on a new slot changing from human to skrell etc?

listerla: Correct. I'll test it with other species some time today.

Mechoid: Other species use the same system, tis fine if it worked on the Squid

PR #7623, opened 2020-09-04
Added missing spellbook sprites

Kazzc:
As the title says.


Mechoid:
To be tweaked later for all spellbooks to match the normal book appearance. Since they don't appear anywhere aside Bus, not critical to prevent the PR.

PR #7629, opened 2020-09-05
Fresh Meat, Leatherworking

Mechoid:
Butchering more modular.

Animals can be butchered for organs and hide. Requires scraping (sharp), washing (water or washing machine), and then drying (bonfire or drying rack).
Spiders can be butchered for chitin, instead of hide. Requires no prep, but can't be used to make clothing like leather.

Leather and Chitin can be exported.

Organs can be butchered for meat, named "[organ] meat". Heart meat, liver meat, etc.
Brains from player mobs cannot be butchered.


lbnesquik:
Consider adding a bit of value to it as an export product.

Meghan-Rossi:
Simplemobs spawning with organs sounds like it could be bad for performance.  It might be better to have the organs only be created while the mob is being butchered.

Mechoid:
If it ends up causing a huge spike in performance problems, I'll move it over to that, but the idea and hope, is that the organs would eventually provide ways for the mobs to have different passive effects modularly, which could then be implanted into a humanmob, for the same effects.

Atermonera:
Do we have the "nobody's on my Z level, I should just sleep" check for simplemobs? I remember that being discussed as a way to cut down on spider chitter spam but cannot conclusively remember if we implemented it or not.
That would help limit the processing overhead, potentially even more if you try and further isolate their processing by player proximity (Though that seems excessive to me, it may be necessary for other servers)

code/modules/mob/living/organs.dm
```diff
@@ -0,0 +1,28 @@
+/mob/living
+	var/list/internal_organs = list()
+	var/list/organs = list()
+	var/list/organs_by_name = list() // map organ names to organs
+	var/list/internal_organs_by_name = list() // so internal organs have less ickiness too
```
Atermonera, line 5: This will also give these vars to silicons

code/modules/organs/organ.dm
```diff
@@ -134,7 +172,7 @@ var/list/organ_cache = list()
 	handle_organ_proc_special()
 
 	//Process infections
-	if(robotic >= ORGAN_ROBOT || (owner && owner.species && (owner.species.flags & IS_PLANT || (owner.species.flags & NO_INFECT))))
+	if(robotic >= ORGAN_ROBOT || (istype(owner) && (owner && owner.species && (owner.species.flags & IS_PLANT || (owner.species.flags & NO_INFECT)))))
```
Atermonera, line 175: `istype(owner)` enforces `owner`, so the former can replace the latter, rather than being needed in addition to

code/modules/mob/living/organs.dm
Mechoid: Correct. They don't have any use on Silicons, but they are needed on simple_mobs, and it's the latest common branch.

code/modules/organs/organ.dm
```diff
@@ -172,7 +172,7 @@ var/list/organ_cache = list()
 	handle_organ_proc_special()
 
 	//Process infections
-	if(robotic >= ORGAN_ROBOT || (istype(owner) && (owner && owner.species && (owner.species.flags & IS_PLANT || (owner.species.flags & NO_INFECT)))))
+	if(robotic >= ORGAN_ROBOT || (istype(owner) && (owner.species && (owner.species.flags & IS_PLANT || (owner.species.flags & NO_INFECT)))))
```
Atermonera, line 175: These are also bitflags being checked, such that `(owner.species.flags & IS_PLANT) || (owner.species.flags & NO_INFECT)` is equivalent to `(owner.species.flags & (IS_PLANT | NO_INFECT))` (Note the single | for logical OR)

PR #7630, opened 2020-09-05
Exo Encumberence

Mechoid:
Adds an Exosuit Encumberance system for equipment.

Exosuits can carry extra weight from equipment up to a certain value, at which weight will begin causing slowdown.

An exosuit defaulting at 4 weight 1 gap, when carrying 1.5 weight-units in equipment, will only register as 4.5 total, rather than 5.5.

Gap values:

Default: 1
Combat: 1.5
Alien: 2

Fix: Overload actually makes you faster, instead of slower.


lbnesquik:
This COULD work as a better system than having slowdown no matter what. Still not super confident about the whole thing.

code/game/mecha/combat/combat.dm
```diff
@@ -17,6 +17,8 @@
 	max_special_equip = 1
 	cargo_capacity = 1
 
+	encumberence_gap = 1.5
```
listerla, line 20: I hate to tell you, but the word is 'encumbrance'

Mechoid: ..Fuck.

Screemonster:
I can't help but feel that the ripley, which is slow enough as it is, should have a pretty high gap value. Like, it's a loader. Its entire point is to lift a crapload of heavy stuff.

TheDavestDave:
To add to this the ripley kinda just sucks. It's okay in combat but at mining it's often just easier to go in and mine yourself. A decent set up or weapon + speed boost+ drill+ clamp makes it useable. And that tends to use most of the slots.

DerBlitzkrieg:
Why is this being added to the ripley, a purpose built hauler of a mech? Honestly I'm unsure why its being added at all. If you're gonna use the extra time and effort to make more equipment for mechs, not sure why that should give you a direct handicap when mech equipment is already fairly specialized.

Mechoid:
Ripleys are already affected by weighted equipments, though I have increased their limit to 2 given they _are_ Automated **Power Loading** Units. Most tools do not have weight values as of now.

Atermonera:
Staff input: Until people start throwing numbers around about what "is" and "is not" ""slow,"" the term will remain nebulously defined in peoples' headcannons as various _different_ numbers and thus we won't be able to have an intelligent conversation on what is good and what is bad.

And this system looks good and makes sense.

FenodyreeAv:
Here's a custom build Durand. In the current version, before this change (meaning this Durand is 1.5 weight higher than it will be.)
https://streamable.com/7f1trr
Double shielded, ranged weapon armour, xray laser.
Stripped out life support, overclocked actuator, efficience harness, blackops plating, reinforced hull.

Marginally slower than the old Durand, immune to anything with less than 15AP (immune to normal lasers & bullets), 0.1 less armour across the board, higher EMP resistance.

If you know what you're doing, Mechs got MORE powerful, not less. Variance is good. Rewarding good players is good.

Now for fun, here's a Durand crossing the station in 15 seconds flat. Imagine this chasing and punching you.
https://streamable.com/e6wahp

PR #7638, opened 2020-09-08
Remove broken, unused commcard system from communicators

ShadowLarkens:
This is being done to help facilitate the transfer to TGUI. These
commcards are part of a project that never really got finished, and the
majority of them do not even function in their current implementation,
with only barebone templates and half-made integration.

While it would be neat to have these function one day, I believe it
would be for the best to remove their current implementation wholesale
as this PR does, and allow them to be reintegrated at a later time
when communicators have been converted to TGUI and NanoUI has been
eliminated.

It's important to realize that these are already completely unobtainable outside of admin spawn and by and large *do not function whatsoever*. You can look at templates such as emp_records.tmpl to see that the code to support all of the features it wants to simply doesn't currently exist.


ShadowLarkens:
To be specific, the reason I'd like these to be removed is because it would be very difficult to convert communicators from NanoUI to TGUI without taking on the side project of "Completely remake commcards into something functional", which I'm not inclined to do.

It wouldn't really be possible to convert currently completely broken things to use TGUI, as it's really difficult to make an *equivalently **broken*** feature out of an abandoned project.

PR #7640, opened 2020-09-08
Cycler made more rugged and options for no cosmetic changes added to all cyclers.

cadyn:
Made cyclers more rugged by adding a check to make sure that the sprite for the suit and species combo exists before allowing it to be cycled. Also added an option so that users can retain the cosmetics of the suit while cycling it.


Cerebulon:
This appears to be causing problems/not working. One human person was allowed to cycle their sec suit to riot control, but it didn't actually change the sprite. Another promethean was given the "No combination available" when trying to make the same selection (security->riot control) despite prometheans using the human spritesheet.

cadyn:
> This appears to be causing problems/not working. One human person was allowed to cycle their sec suit to riot control, but it didn't actually change the sprite. Another promethean was given the "No combination available" when trying to make the same selection (security->riot control) despite prometheans using the human spritesheet.

Beh, I have an idea about the issue, I'll pump out fixes tomorrow if no-one beats me to it. Sorry for the oversight.

PR #7657, opened 2020-09-16
Should fix unset PDA name glitches

### LINKED ISSUE #7644: Wearing un-set PDA glitches the game out in many ways
Wearing a unset PDA while also being in 'unknown' status glitches the game out in multible ways.
To be just 'Unknown'.
#### What actually happened
-Items used in-hand do not realise that there is a user, since user blank, and do not work (For example, Stunprods won't toggle, Syndiradio won't open, probably many more that are worse)
-You are even more invisible then the AI-Eye and can not be spoted by turf-view
-You are more invisible then ghosts and can not be spoted by turf-view of the death (speculation)
-Outside of shortcut-clicking, you can not be examined.
-You display no name at all if examined, talking or gesturing.
-Probably other ways, too!
#### Steps to reproduce
Get a full body cover, like a spacesuit and wear a PDA that has never had contact with a ID or Name in the ID spot.
#### Additional info:
This has the potential to break so many things...

Atermonera:
Haven't tested because the documentation of what was specifically broken wasn't terribly clear, but this should reliably cause the user's name to not be set to `null`, and if the glitches persist, then I will just find where PDAs set `owner` to null and switching those to something intelligible.
fixes #7644 and if it doesn't then I expect people will come screaming at me in fairly short order.

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -259,7 +259,7 @@
 		return
 
 	spread_fire(AM)
-	
+
 	..() // call parent because we moved behavior to parent
 
 // Get rank from ID, ID inside PDA, PDA, ID in wallet, etc.
@@ -269,7 +269,7 @@
 		if (pda.id)
 			return pda.id.rank ? pda.id.rank : if_no_job
 		else
-			return pda.ownrank
+			return pda.ownrank ? pda.ownrank : if_no_job
 	else
 		var/obj/item/weapon/card/id/id = get_idcard()
 		if(id)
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/human.dm
```diff
@@ -269,7 +269,7 @@
 		if (pda.id)
 			return pda.id.rank ? pda.id.rank : if_no_job
 		else
-			return pda.ownrank
+			return pda.ownrank ? pda.own_rank : if_no_job
```
Meghan-Rossi, line 272: own_rank should be ownrank.

Atermonera: Bah, that snuck out of the commit. Fixed now

PR #7667, opened 2020-09-16
Adds lrifle_open sprite for lrifle set

### LINKED ISSUE #7520: LR1 Shishi - missing sprite when cell is removed
title
for the weapon to not go invisible when the power cell is removed
#### What actually happened
the gun vanished when I removed the cell and I had to put it on the ground and use altclick on the tile to reload it
#### Steps to reproduce
get the LR1 Shishi (gun/energy/laser/sleek ) and remove the power cell
#### Additional info:

Atermonera:
Fixes #7520


Cerebulon:
Oh I was just about to fix this for the PDW sprite, which needs a PDWkill_open and PDWstun_open state... But since you're also touchy that file, could you touchy that?

Issue #7661, opened 2020-09-16
'Dib into Beaker'-Mechanic has unintended side effects


DasIrrlicht:
#### Brief description of the issue
Trieing to create glycerol, I had to combine Cornoil & Sulphoric Acid. The 'Dip into Beaker' mecahnic had different plans.
#### What you expected to happen
Clicking with the Cornoil bottle at the beaker with the acid would pour it into the beaker, like it used to do (and still does with non-acid beakers).
#### What actually happened
Cornoil-bottle gets dipped into beaker, swallowing it whole and deleting bottle & reagents, making it impossible to pour anything into a acid-beaker.
#### Steps to reproduce
Get cornoil, or probably any other bottle with reagents. Click with the bottle at a beacer holding some kind of acid.
#### Additional info:
 - **Anything else you may wish to add**
Suggested fix: make the 'dip' mechanic only work when on a non-help intend.

Issue #7309, opened 2020-06-26
"You dip (object) into (container)" spam


Wickedtemp:
#### Brief description of the issue
So, this has been a thing for a while, but only recently have i actually been inconvenienced enough to make a git report on it. 

Whenever you use an item on a container with a reagent, whether you're transferring from one container to another or using a syringe or dropper, or writing labels.. you "dip the item into the container". Every time. And while usually it's not a big deal, you can just ignore the spam... 

Acid. 

Turns out it melts stuff.

Stuff you.."dip into the container"

Reagent scanners, droppers.. anything that isn't acid-proof I'm assuming. So, while I can see this as being a fun tool to use and abuse.. I think its time we've fixed everything being dipped into containers for now and save that stuff for a harm intent interaction or something.

PR #7669, opened 2020-09-17
Fix Beaker Dipping

Mechoid:
Dipping things into beakers now does not occur on help, since you're probably being careful.

Fix https://github.com/PolarisSS13/Polaris/issues/7661
Fix https://github.com/PolarisSS13/Polaris/issues/7309


Wickedtemp:
Also fixes #7309

PR #7671, opened 2020-09-17
Self-Surgery

Mechoid:
Allows organics to self-surgery after toggling the surgical safety in the object tab. It follows the same rules as normal surgery.

FBPs cannot repair their own internal damage, but a Human with a prosthetic limb can repair its internal damage.


Schnayy:
Probably going to be flagged with staff discussion, but throwing my vote in for it in case I'm not around for the staff meeting. Sounds fun, and can be adjusted to still be balanced.

Wickedtemp:
Is there a way to prevent misclicks on yourself from starting surgery?

TheFurryFeline:
> Is there a way to prevent misclicks on yourself from starting surgery?

Probably easily enough with a popup prompt about warning the self-surgeon about the attempt being ill-advised. "Are you sure you want to make an incision on yourself?"

Wickedtemp:
So, I'm kinda split on this. Mostly, I'm kinda just curious as to what happened to result in this PR and what the intended-usage would be for self-surgery. My best guess is that it'd be a sort of dead-shift insurance in case the only doctor gets rekt, so now they'd at least be able to fix themselves (assuming they live and escaped whatever danger was present) and now that they're all good, they can fix any other wounded. Hard to continue a round if there's only one or two medical and both need surgery and are effectively incapacitated and everyone else is kinda just left to slowly die of whatever injury they'd sustained.

The primary concerns I'd have is that.. this seems like something that would best be tied to a fully-fleshed-out skills system in order to prevent anyone and everyone with the tools from operating on themselves. Outside of purely RP reasons, there's little to no reason for me to go to Medical if I happen to acquire the surgical tools (or even ghetto tools and some chems). If we're honest, even the rng-chance to fail due to improper tools or a lack of OR table really isn't too dangerous. If you've got bicaridine, you can just heal the damage like nothing's happened (I personally think failing a surgical step should be much more damaging, but that's something for another PR). 

On the flip-side, there are many, many instances of self-operation occurring IRL, so this is something that is well within the realms of possibility. But, most of the time, it's done by someone who at least has some manner of training, or they're left scarred or disfigured or permanently injured due to a screwup, or just straight up die. We can't really effectively emulate long-term consequences, so it kinda goes back to the whole.. "why shouldn't people just rely on this feature". 

Now, this is stuff that can be added in later. Things can be adjusted and modified as we see how a feature fits in. I think, for the most part, I'd kinda like to see where this goes. Worst case, it's heavily abused and a few powergamers get slapped and a couple catches get implemented later. 

(though I would like something to make sure that if I'm about to operate on someone else, that I won't start cutting myself open due to a misclick)

code/modules/mob/living/living_powers.dm
```diff
@@ -15,3 +15,12 @@
...
 		to_chat(src,"<span class='notice'>You are now hiding.</span>")
+
+/mob/living/proc/toggle_selfsurgery()
+	set name = "Allow Self Surgery"
+	set desc = "Toggles the 'safeties' on self-surgery, allowing you to do so."
+	set category = "Object"
+
+	allow_self_surgery = !allow_self_surgery
+
+	to_chat(usr, "<span class='notice'>You will [allow_self_surgery ? "now" : "no longer"] attempt to operate upon yourself.</span>")
```
TheFurryFeline, line 26: Imo, I think this should have an admin-notify thing just in case someone powergames it or something. Which is of course at admins' discretion for actions taken against the said odd case

Atermonera: I'd be more worried about someone spamming that for the sake of spamming adminlogs, in that case. To be clear, the change was made as a result of staff concerns over it being too easy to accidentally start doing surgery on yourself, where you happen to be stood still for 5 seconds and then misclick yourself with a surgery tool (Which notably includes _ghetto_ surgery tools, which are actually pretty common). Suddenly, what was originally a bruise and maybe a cut has turned into an open surgery incision that could become septic. Especially concerning for new players that like to click on things with other things just to see what will happen, who may then be surprised that they're doing surgery and try to move to cancel it, causing the surgery _fail_ step, and confusing them greatly.

Mostly, it's an issue of this not really being a precedented behaviour, where you have to stand still _before_ doing something for it to work (Lots of things where you have to stand still _while_ you do something though). Making it so you toggle it on forces players to actively choose to start doing it, instead of being able to accidentally start doing it without realizing that they could.

TheFurryFeline: I meant to answer sooner after I got notification of this response this morning but here I am. The concern about someone being a troll enough to blatantly toggle this off/on to spam adminlogs is in itself something for an admin to handle, too, imo. But I digress and have a mindset forged from playing on an HRP server where it's more respectful to let the doctors to the work they sign up for and not "I know the mechanics, therefore I am a doctor/engineer/scientist Mary Sue."

PR #7679, opened 2020-09-21
Cycler sprite check fixes

cadyn:
Neglected human sprites last time, since apparently they don't have a sprite sheet defined and instead them not having a sprite sheet is used to separate their sprites for some ungodly reason.


Atermonera:
_~quality code!~_

PR #7680, opened 2020-09-22
Phoron bore utility improvements

Cyantime:
The phoron bore: 

- Uses *significantly* less cell charge per shot. 750 gives you 20 shots with a regular 15k cell (ignoring material ammo cost), each of which will dig about six tiles. Compared to a drill, or a KA, or basically anything and this is woefully expensive for the actual mineral output. It now uses 100 charge per shot.

- Can be reloaded with raw plasma ore in addition to plasma sheets. Each ore gives 800u of plasma, compared to 2000 per sheet. Two ores compress into one sheet, minus some inefficiency for using unrefined material. Carrying a stack of sheets is still more effective, but it can now be reloaded in the field with extracted material.

- Has a tiny generator built in that runs off the stored plasma. It charges the cell, or capacitor directly if there's no installed cell, at a rate of 20kW. This is the same as a basic pacman set to output level one, but uses about 25% more fuel and produces exhaust that can potentially be a hazard in very enclosed spaces with no scrubbers.

- Comes pre-installed with a basic power cell and capacitor when bought from the mining vendor.

- Does some extra damage to objects.

Also did some minor cleanups in the magnetic/bore files. Screwdrivers/crowbars use their toolsound, removed some unneeded forceMoves, changed a modulo that gave infinite ammo to a division that doesn't, etc.


code/modules/projectiles/guns/magnetic/bore.dm
```diff
@@ -77,91 +74,169 @@
...
+			if(!do_after(user, 2 SECONDS, src))
+				generator_state = GEN_OFF
+				return
+			pull--
+		soundloop.start()
+		cell?.use(100)
+		audible_message(SPAN_NOTICE("\The [src] starts chugging."),SPAN_NOTICE("A motor noise starts up."))
+		generator_state = GEN_IDLE
+
+	else if(generator_state > GEN_OFF && time_started + 3 SECONDS < world.time)
```
Atermonera, line 235: You don't appear to set `time_started` anywhere. And this appears to just stop the generator 3 seconds after it's started?

code/modules/projectiles/guns/magnetic/bore.dm
```diff
@@ -77,91 +74,169 @@
...
+/obj/item/weapon/gun/magnetic/matfed/phoronbore/ui_action_click()
+	toggle_generator(usr)
+
+/obj/item/weapon/gun/magnetic/matfed/phoronbore/process()
+	if(generator_state && !mat_storage)
+		audible_message(SPAN_NOTICE("\The [src] goes quiet."),SPAN_NOTICE("A motor noise cuts out."))
+		soundloop.stop()
+		generator_state = GEN_OFF
+
+	else if(generator_state)
```
Atermonera, line 187: This can try to set the soundloop speed and generate power while the generator is being started.

Cyantime: Whoops, the starting state was a late change and I missed this

code/modules/projectiles/guns/magnetic/bore.dm
Cyantime: This prevents you from mashing the toggle hud icon to turn the generator off too quickly. If a soundloop is stop()ed while it's still playing the start sound, it doesn't stop and will keep going and I don't want to try and get into the soundloop code to try and change that right now.

code/modules/projectiles/guns/magnetic/bore.dm
```diff
@@ -77,91 +74,170 @@
...
+#define GEN_OFF 0
+#define GEN_IDLE 1
+#define GEN_ACTIVE 2
+
+/obj/item/weapon/gun/magnetic/matfed/phoronbore
+	name = "portable phoron bore"
+	desc = "A large man-portable tunnel bore, using phorogenic plasma blasts. Point away from user."
+	description_fluff = "An aging Grayson Manufactories mining tool used for rapidly digging through rock. Mass production was discontinued when many of the devices were stolen and used to break into a high security facility by Boiling Point drones."
+	description_antag = "This device is exceptional at breaking down walls, though it is incredibly loud when doing so."
+	description_info = "The projectile of this tool will travel six tiles before dissipating, excavating mineral walls as it does so. It can be reloaded with phoron sheets."
```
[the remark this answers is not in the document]
Cyantime: Done

PR #7695, opened 2020-09-26
Changes 'Me' Verb to use a proper textbox instead of single-line input.

Andromeda-K22:
This is a one line quality of life change.


Atermonera:
>Not just typing me into the command bar the same as typing say :V

Andromeda-K22:
> typing say or me into the command bar

PR #7703, opened 2020-09-29
Make the tesla ball eventually run out of energy

Meghan-Rossi:
Currently the tesla ball only loses energy until it runs out of orbiting mini-balls, then sticks around forever.  This makes it keep losing energy until it disappears.


TheFurryFeline:
THANK YOU, this was a problem for me downstream where I wanted to switch to Lord Singuloth long after Lady Tesla was started one shift.

Meghan-Rossi:
Yeah, it looks like originally there was no way for it to disappear at all, then code was added to make it disappear at zero energy, but there still wasn't a way for it to reach zero energy.

PR #7731, opened 2020-10-12
Adds the ability to choose your default unarmed attack

Greenjoe12345:
there is a new button in the IC tab: Check Attacks, it will let you choose which unarmed attack to use.
If you cannot use the attack you chose, you will use the default unarmed attack for your species.
also changes the teshari default attack to claws so they're not being such savages biting people by default.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_attackhand.dm
+++ b/code/modules/mob/living/carbon/human/human_attackhand.dm
@@ -1,3 +1,6 @@
+/mob/living/carbon/human
+	var/datum/unarmed_attack/default_attack
+
 /mob/living/carbon/human/proc/get_unarmed_attack(var/mob/living/carbon/human/target, var/hit_zone)
 	if(src.gloves)
 		var/obj/item/clothing/gloves/G = src.gloves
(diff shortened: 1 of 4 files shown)


code/modules/mob/living/carbon/human/human_attackhand.dm
```diff
@@ -418,3 +427,63 @@
...
+	bite/strong/attack_name = "strong bite"
+	punch/attack_name = "punch"
+	kick/attack_name = "kick"
+	stomp/attack_name = "stomp"
+	stomp/weak/attack_name = "weak stomp"
+	light_strike/attack_name = "light hit"
+	diona attack_name = "tendrils"
+	claws/attack_name = "claws"
+	claws/strong/attack_name = "strong claws"
+	slime_glomp/attack_name = "glomp"
```
Mechoid, line 489: These should probably be assigned in the attack datums themselves, instead of here, as it will cause issues if any of these typepaths are changed elsewhere.

Mechoid (changes requested):

code/modules/mob/living/carbon/human/human_attackhand.dm
Greenjoe12345: Right.. Hopefully I can figure that out otherwise this PR is dead

code/modules/mob/living/carbon/human/human_attackhand.dm
```diff
@@ -418,3 +427,46 @@
...
+
+	var/dat = "<b><font size = 5>Known Attacks</font></b><br/><br/>"
+
+	for(var/datum/unarmed_attack/u_attack in species.unarmed_attacks)
+		dat += "<b>Primarily [u_attack.attack_name] </b><br/><br/><br/>"
+
+	src << browse(dat, "window=checkattack")
+	return
+
+/mob/living/carbon/human/check_attacks()
```
Mechoid, line 444: Are both of these meant to be left here?

PR #7738, opened 2020-10-18
Stops slime securitrons from going ballistic when they get scratched

greggbot:
…so they can actually discipline properly.

Right now, the retaliation component of securitrons, which is fine and makes sense, also applies to slime securitrons.
This prevents them from administering proper discipline, as they just go crazy and repeatedly baton the slimes as soon as they recieve the inevitable glomp, as they can't be "arrested". Eventually the slimes enrage and you have a pen full of very angry slimes - the exact opposite of what it's meant to do.

So, I added a simple var to the securitron to determine if it should care about retaliating or not, and set it to false for the slime securitron. Since they already detected and disciplined disobedient slimes, this will allow them to perform their function normally.


Shadow-Quill:
Would a better way of handling this just be overriding the `react_to_attack` proc on slime secbots?

Mechoid:
Having it var-controlled does mean it could be added as an interface toggle later. Though, I believe this would also need to be added to the slime ED

PR #7531, opened 2020-08-20
teshari spase suit update

KBraid:
replaces placeholder teshari space suit and adds matching helmet

PR #7740, opened 2020-10-18
Kbraid.space+radsuitsprites

KBraid:
image


KBraid:
guess not

Cerebulon:
Hi, you did this PR before ( #7531 ) where I explained that this is not wanted in this form. I'll copy paste the comment.

"Goofy ill-fitting sprites were kind of intentional to indicate that there isn't purpose-made stuff for Tesh. I'm planning on reverting the radiation suit change and implementing the 'tailored fit' suit as a specific custom-order alternative in the same way as the adaptive voidsuits are exclusive of tesh due to tesh's non-humanoid shape, and would prefer this were a separate item also.
Generic space suits especially so since they don't even usually have Taj/Unathi sprites afaik? They're on-size-fits-all in the uncomfortable way only one-size-fits-all products can ever achieve."

If you want to add these sprites, add them as a separate, teshari-only version of the items.

KBraid:
in terms of the space suit, it's back to the original however i noticed the helmet was still missing in your current version so i made one based off of the original.
and as for 'making it fitted' i'm not familiar enough with the code to actually set that up.

Cerebulon:
Alright, so these _fitted sprites do not actually replace the current rad/space suit sprites? That's alright then. If you can't code them into actual 'fitted' versions of the item I can probably do that later.

PR #7745, opened 2020-10-21
Adds one (1) PoI

SubberTheFabulous:
Located in plains, the VR pods are a core (and only) feature. Currently the only way to access VR.

Can be removed or adjusted, if VR ever gets added elsewhere.

### DIFF
--- a/maps/submaps/surface_submaps/plains/VRDen.dmm
+++ b/maps/submaps/surface_submaps/plains/VRDen.dmm
@@ -1,0 +1,72 @@
+"a" = (/turf/template_noop,/area/template_noop)
+"b" = (/turf/template_noop,/area/submap/VRDen)
+"c" = (/turf/simulated/wall,/area/submap/VRDen)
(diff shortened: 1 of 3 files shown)


Cerebulon:
This seems like it might be weird once there's actually VR on the station? As in, why would some random pods in the middle of nowhere be hooked up to the same 'network' as the station pods?

SubberTheFabulous:
Hence "Can be removed or adjusted, if VR ever gets added elsewhere."

Atermonera:
Without the VR spawn landmarks, the pods are useless. `/obj/effect/landmark/virtual_reality` are required _somewhere_ in the world for the pods to do anything when you climb into them.

SubberTheFabulous:
@Atermonera 
Already present, at a separate location on CentComm map.

Atermonera:
Ah, one of the PRs _other people_ merged while I was buried in the tgui pr.

PR #7748, opened 2020-10-23
Adds new balloons

Schnayy:
Adds a recolorable balloon and two giant halloween balloons.

You can pick up the former, the latter two are structures.

image
image

Please get this in before the gala & in time for an update.


PacifistDalek:
I'm trying to port the basic balloons to Paradise, but I'm running into the issue of the in-hand sprites not changing colours when you change the item's color var (the item sprite does change) :/

PR #7749, opened 2020-10-28
Stowaway Cryopod is now invisible, and incorporeal.

Mechoid:
Makes the stowaway cryopod more mysterious by making it only visible to ghosts. The manual cryopod remains normal.

### DIFF
--- a/code/game/objects/structures/ghost_pods/human.dm
+++ b/code/game/objects/structures/ghost_pods/human.dm
@@ -8,9 +8,10 @@
 	icon = 'icons/obj/Cryogenic2.dmi'
 	icon_state = "sleeper_1"
 	icon_state_opened = "sleeper_0"
-	density = TRUE
+	density = FALSE
 	ghost_query_type = /datum/ghost_query/stowaway
-	anchored = FALSE
+	anchored = TRUE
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/ghost_pods/human.dm
```diff
@@ -123,7 +124,9 @@
 	if(allow_appearance_change)
 		H.change_appearance(APPEARANCE_ALL, H, check_species_whitelist = 1)
 
-	visible_message("<span class='aliem'>\The [src] [pick("gurgles", "seizes", "clangs")] before releasing \the [H]!</span>")
+//	visible_message("<span class='aliem'>\The [src] [pick("gurgles", "seizes", "clangs")] before releasing \the [H]!</span>")
```
Shadow-Quill, line 127: If you don't plan on doing something with this commented out bit, why not just remove it.

Mechoid: You ever just type something out, forget about it, then restart your browser? Yeah I did that like 8 times on this aaaaaaa sorry.

I left it in comment, mostly as reference, but if anyone wanted to go back and restore them to being visible, it's still there.

PR #7753, opened 2020-11-01
No more unnearsightedness!

Shadow-Quill:
As per title.

All this does is make it if you have nearsightedness, you won't lose it by overusing a welder.


Shadow-Quill:
I fixed the travis error on oops last night, but forgot to push it x3

PR #7756, opened 2020-11-13
Matter Matters

Mechoid:
Begins work on maintaining equivalent exchange.

Autolathes, protolathes, and sheet crafting will respect equivalent exchange.

Autolathes and sheet-crafting can be very wasteful of materials, when their products are recycled. Stacks of items are treated, in manual sheet crafting, as their material counterparts are, wherein the stack has the matter of a single 'sheet', rather than the sum total. 2000 is the standard per sheet, as per the lathes.

### DIFF
--- a/polaris.dme
+++ b/polaris.dme
@@ -3096,7 +3096,7 @@
 #include "code\ZAS\Zone.dm"
 #include "interface\interface.dm"
 #include "interface\skin.dmf"
-#include "maps\example\example.dm"
+#include "maps\southern_cross\southern_cross.dm"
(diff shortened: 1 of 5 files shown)


polaris.dme
```diff
@@ -3096,7 +3096,7 @@
 #include "code\ZAS\Zone.dm"
 #include "interface\interface.dm"
 #include "interface\skin.dmf"
-#include "maps\southern_cross\southern_cross.dm"
```
Atermonera, line 3099: Oops

PR #7768, opened 2020-11-26
Move away from Travis CI to GitHub Actions

Atermonera:
Port of [VORESTATION/VORESTATION#9337](https://github.com/VOREStation/VOREStation/pull/9337)
Travis is dead, long live CI! (We need to find a good name for CI because psi is already taken by a prominent coder)

@Arokha This is a cherry pick of the merge commit from the linked PR and while I _think_ I correctly excised all the stuff specific to virgo, this is babby's first port so do tell if I missed anything that's important or will mess with your code.


KasparoVy:
CIpher, c'mon! Low hanging fruit

.github/workflows/ci.yml
```diff
@@ -0,0 +1,68 @@
+name: CI
+on:
+  pull_request:
+    branches:
+      - master
+
+env:
+  BASENAME: "vorestation"
```
zekesulastin, line 8: This probably needs to be changed to "polaris" to make the checker work properly, but I don't quite have the experience (or setup) with Github CI to prove it.

PR #7771, opened 2020-12-01
Creates framework for species to experience traumatic shock at different rates

Runa-Dacino:
I am making a trait downstream for custom species to be able to take twice the traumatic_shock from being injured in any way, and to do it I need to change base polaris files unfortunately. Same deal as spice_mod and other VORESstation custom species traits.

This does not affect polaris gameplay at all as it is implemented, but may be used to differentiate some species further if so desired by the community just like brute_mod and other traits allow differentiation.

Testing shows it works as intended, and will NOT affect Polaris.

Before the change:
100 oxloss led to 100 traumatic shock.
After the change, without trait that changes trauma_mod to be equal to 2 (essentially, Polaris experience):
100 oxloss led to 100 traumatic shock.
After the change, with trait:
100 oxloss led to 200 traumatic shock.

Also works for other sources of trauma, but they produce less pretty numbers.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/species.dm
+++ b/code/modules/mob/living/carbon/human/species/species.dm
@@ -102,6 +102,7 @@
 	var/alcohol_mod =		1						// Multiplier to alcohol strength; 0.5 = half, 0 = no effect at all, 2 = double, etc.
 	var/pain_mod =			1						// Multiplier to pain effects; 0.5 = half, 0 = no effect (equal to NO_PAIN, really), 2 = double, etc.
 	var/spice_mod =			1						// Multiplier to spice/capsaicin/frostoil effects; 0.5 = half, 0 = no effect (immunity), 2 = double, etc.
+	var/trauma_mod = 		1						// Affects traumatic shock (how fast pain crit happens). 0 = no effect (immunity to pain crit), 2 = double etc.Overriden by "can_feel_pain" var	
 	// set below is EMP interactivity for nonsynth carbons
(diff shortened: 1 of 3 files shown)


Atermonera:
Shock here being halloss, yes? Or is there some other mechanic? (Also closing/reopening the PR to update the workflow file for CI)

Runa-Dacino:
> Shock here being halloss, yes? Or is there some other mechanic? (Also closing/reopening the PR to update the workflow file for CI)

Originally I was going to only use Hallos, but it turned out to not affect how a character experiences pain when taking lethal damage. You only suffer hallos when taking non-lethal sources of damage like holodeck weapons and jointlock.

Shock.dm is the file that governs getting into pain-crit. Each source of lethal damage AND hallos (brute,  burn, oxy, toxin, hallos, clone) is multiplied by a number (oxy by 1) and added together. Your organ damage is added on top of this. 

This PR takes said sum and multiplies it with a trauma_mod to make it that you go into pain crit way only 50 traumatic shock rather than the mandatory 100 if a species has it set to 2.

Originally the calculator subtracted painkillers before adding organ damage, i moved that addition at the end of the file, before it returns traumatic_shock to src to make it that sensitivity to pain does NOT make painkillers more or less effective.

The default trauma_mod is 1, meaning unless someone gives a species a different value in the station defines, it wont affect Polaris at all.

On Virgo downstream there will be a custom species negative trait that will cost -1 points and set this value to 2.

PR #7774, opened 2020-12-01
Prevents infinite money

Shadow-Quill:
Not tested, bc mobile webedit, but theoretically should function as intended.


Shadow-Quill:
scree, EOF changes.

Issue #6600, opened 2020-01-04
Dismantling an Employee Record computer with an ID in it deletes the ID


SpadesNeil:
#### Brief description of the issue
See title.

#### What you expected to happen
The ID to be ejected onto the floor.

#### What actually happened
Yeetus deletus.

#### Steps to reproduce
1. Put an ID into an employee record console.
2. Use a screwdriver on the console.

#### Additional info:
 - **Server Revision**: https://github.com/VOREStation/VOREStation/commit/0c7edc74cb02628d7ca76ad47b7f9c66d52b34a9
 - **Anything else you may wish to add** Might occur with other computers. Don't know.

lbnesquik:
Has been fixed

PR #7776, opened 2020-12-02
Fixes dismantled machines deleting inserted IDs

Verkister:
Machines no longer delete people's ID cards whenever they attempt the most obvious route of getting their card back from an unpowered machine.


zekesulastin:
For reference, fixes #6600 https://github.com/PolarisSS13/Polaris/issues/6600

PR #7781, opened 2020-12-10
Blob Genesis

Mechoid:
Additions:
Allows the regeneration of blobs from the core chunks they leave behind on death, in two methods. (If you kill it once to get the chunk, it's within the realm of possibility for killing it again.)

The first, 60 units pure phoron, produces a hostile blob on the chunk's turf.
The second, using 40 units hydrophoron, 20 peridaxon, 20 mutagen, produces a neutral blob. Crew can walk through and on these blob tiles with no harm done.

Tweaks:
Player blobs can move as long as they can see a piece of blob.

Blobs can fight eachother if they have differing factions. Grey Tide is by default the only non-standard faction blob, under Nanomachines.

Standard blob walls will gesticulate on pulse. This is surprisingly disturbing in large blobs.

Cryoblob chunks can produce cryoslurry.

Blazing chunks can produce thermite venom, the same injected by firespiders.

### DIFF
--- a/code/__defines/species_languages.dm
+++ b/code/__defines/species_languages.dm
@@ -66,7 +66,7 @@
 #define LANGUAGE_ALAI "Alai"
 #define LANGUAGE_ZADDAT "Vedahq"
 #define LANGUAGE_PROMETHEAN "Promethean Biolinguistics"
-#define LANGUAGE_BLOB "Blob"
+#define LANGUAGE_BLOB "Chemosense Transmission"
 #define LANGUAGE_GIBBERISH "Babel"
 
 // Language flags.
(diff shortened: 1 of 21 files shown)


Atermonera:
Will friendly blobs still break walls, structures, etc., that are blocking their expansion?
If one blob defeats another, will it still subsume all of the defeated blob's remaining mass to make one really big blob?

Mechoid:
Yes and yes.

Mechoid:
Though on the second, it will take time, as it's based entirely on node pulses.

Atermonera:
Please provide evidence of your testing.

Mechoid:
image

PR #7803, opened 2020-12-23
Fixes dark click catcher

### LINKED ISSUE #1997: Dark Click Catcher no longer catches clicks

When clicking upon a black square, the character does not interact with that area.

Character in the direction of the black square.
#### What actually happened

Nothing.
#### Steps to reproduce

Join.
Find spot where you cannot see a section of the screen.
Face away.
Click on dark spot.
#### Additional info:
  6374bc2c4c342b53bbf2950b9e792ec7e2084495 

Atermonera:
Fixes #1997
:cl:
fix - Dark click catcher has been repaired, and you can now click on unseen turfs (The other side of walls, for instance) to turn your character
/:cl:
[Tested](https://puu.sh/H0dCP/d3454a23af.mp4)


code/_onclick/click.dm
```diff
@@ -338,27 +338,22 @@
...
-	screen_loc = "CENTER-7,CENTER-7"
-
-/obj/screen/click_catcher/proc/MakeGreed()
-	. = list()
-	for(var/i = 0, i<15, i++)
-		for(var/j = 0, j<15, j++)
-			var/obj/screen/click_catcher/CC = new()
-			CC.screen_loc = "NORTH-[i],EAST-[j]"
-			. += CC
+	screen_loc = "1,1 to 15,15"
```
Neerti, line 347: It might be better to use `screen_loc = "SOUTHWEST to NORTHEAST"`, since then it will scale automatically if the client can see more than 15x15 tiles. iirc that comes into play with scopes/binoculars and an admin button.

The `"SOUTHWEST to NORTHEAST"` thing was used to make goggle vision scale automatically in #6809 if you want to see an example.

Atermonera: I originally changed that when I found that screenloc2turf couldn't parse that syntax, but seeing as I later refactored it to be a single catcher rather than one for each tile of the viewport, I suppose that concern is nullified.

PR #7815, opened 2021-01-02
New ID Card Design

KillianKirilenko:
A surprisingly quick and easy port of [CitRP's fancier, shinier ID cards](https://github.com/Citadel-Station-13/Citadel-Station-13-RP/pull/823). If I'd known it was going to be this easy I would've done it sooner.

tl;dr newer, shinier, updated IDs that are composed of several elements overlaid/blended. super easy to tweak and update and redesign if you don't like them, either by just changing things or adding entirely new elements as-needed.

I've taken the base system and cleaned up some redundancies/streamlined a few things, as well as adjusted the overall 'themes' to hopefully be a bit more coherent.

preview of a bunch of different designs;
image
Sec-Detective-Warden-HOS
Clown-genericx2-Cargo-Miner-QM
Mime-NT-VIP-ERT-Journalist
Assistant-Bartender-Botanist-Chaplain-Chef-Internal Affairs
Guest-Engi-Atmos-CE-Janitor
Secretary-HOP-Cap-Spare
Med-Chem-EMT-Psych-CMO
Sci-Robotics-RD
Civ-Chaplain-Clown-Mime-IAA
Pilot-Explorer-Pathfinder-SAR

:cl:
rscadd - Updated ID designs using new sprites, finally replacing those dusty old ones that are at least ten years old.
rscadd - Added a bunch of new ID designs for jobs
/:cl:


MrSign:
I prefer the current designs better, nothing wrong with them. Sorry but I don't believe this should be ported in. The current sprites serve just fine.

JoanRisu:
I want new ID cards but I have to wonder if there is too much going on in these cards that isn't already being conveyed by a simpler design. 

For example, why do the ID cards need visible text on them when the information is already conveyed by color (or by the little image box in the case of the chaplain)? More does not necessarily mean fancier either, so I don't think the text is actually adding anything of visual value here. I would like a version of these cards that does not have that text (and isn't as long as they appear here because they look like they are teetering on the edge of being bookmarks. )

To give whoever designed these cards some credit, they resolved an issue that I had with the older sprites, and that is the color bar that is going down the middle of the card. When people try to recreate those cards in an illustration, it becomes a graphics logistical nightmare to resolve because of the visual contrasts and separation that bar of color creates. People put black boxes behind text or overlay the bar over the text, etc etc. It ends up not looking very good. The last time I spoke about designing a new card, I wanted the color bar to be pushed to the side with the photobox or as they are here, so credit where it is due, this design does solve that particular issue.

Very-Soft:
I'm not **opposed** to this, I'd not whine if they were on the server, but I do prefer the way our existing cards look to these. It essentially takes the existing design and adds more complicated elements. But where the originals look a bit flat, these look very flat. Like, I get that it's an ID card and it is literally flat, but speaking more strictly from an art standpoint, I prefer the old ones, as they convey the department and rank easily by the colors on the card while the angle and art style hides a bit of the flatness. 

I certainly am not against the idea of new ID cards, but I'd honestly like ID cards that look more like actual ID cards as opposed to the long credit card.

JoanRisu:
I jusst realized this was the polaris repo, woopsie. 

My point kinda still stands though, I think the cards can do without the text imo but that's up to you folks.

KillianKirilenko:
> I certainly am not against the idea of new ID cards, but I'd honestly like ID cards that look more like actual ID cards as opposed to the long credit card.

The general aesthetic and proportions are something I mulled over as well for sure, so you are not alone there in the slightest. On the plus side the stacks system massively simplifies the effort of converting IDs over to whichever new style is desired, as you would only need to create a relatively small selection of base assets rather than needing to sprite a couple dozen+ new ID designs one by one.

The tricky part I think is nailing it so that designs can be consistent as a whole and within departments (and beyond, if you're adding unique designs for offmap roles).

I'm not very good at doing things like shading or I might actually take a crack at it myself.

> I jusst realized this was the polaris repo, woopsie.
> 
> My point kinda still stands though, I think the cards can do without the text imo but that's up to you folks.

Hey, it's still potentially useful feedback, I will say this though; if Polaris really want to keep the lettering but Virgo absolutely wants it gone, it's very simple to just make a Virgo-specific stacks override that removes all the lettering elements but keeps everything else.

Both sides get to have their way, and no poor spriter has to spend hours adjusting every single ID by hand. Everyone wins.

Atermonera:
Finally got around to looking at it, and I _think_ the thing that's putting me off the most is that the `N`s are all different colors, especially within departments (Basically, the white text-background and blue N for heads versus the gray text-background and department-specific N color for everyone else).
Initially, I feel like the color-coded bars (/notches?) for jobs w/in department are too hard to recognize easily, but I also feel like that's the sort of thing that you just get used to over time. I do like the department-coded color swatches.
I think the `N`s themselves are forcing the sprite as a whole to be too wide in order to make it feel like there's more information on there besides just a big fat NT logo, but that's definitely also in part that the text-lines next to them are trying to be long enough to reasonably contain the wealth of information that goes on an ID when that's not strictly necessary.

One idea that I had while mulling this over was to move the job title text down to where the main text is, and have the department-coded color swatch at the top be clean except for job-specific bars, and possibly even to try and abbreviate job titles in place of the department name that most of those represent, but that might be troublesome to reach a consensus upon for everyone.  Make the cards more like corporate ID cards today where they're something that's big enough to show at a glance who you are and what you're doing and less like drivers licenses (Though I will admit some balance between the two would be more practical in-setting).

I will try to track down other folks to get their thoughts on it over the next couple days.

KillianKirilenko:
Would've replied sooner but I ended up leaving my email client closed for an unusually long time, oops.

Anyway, I think this is also a good time to mention also that @Very-Soft has a redesign of their own in the works that might be of interest, though I don't have an up-to-date preview to hand. That might well be of interest to people too.

KillianKirilenko:
Drafted pending resolution of tech issues. Please hold.

Atermonera:
Right, sorry for the delay. Input from the staff meeting was mostly Abstains, a few other people shared similar concerns that it's too noisy, as well that the gold stripe for head IDs isn't as substantial when _all_ IDs have some colored stripe there.

Atermonera:
@KillianKirilenko interest check, any plans/updates on this?

KillianKirilenko:
Still delayed by issues. Will get around to it when I can.

KillianKirilenko:
Able to resume work on this. Slight curveball in that I thought I was missing a file that may never have existed in the first place, but no matter.

Currently my plan is to strip about 2px off each short end (to make the whole thing 4px thinner), probably 1px off the long sides (2px thinner total), and 1-3px off the bottom of the top bar. Other elements will be adjusted as needed;

KillianKirilenko:
@Atermonera Here's my revised version so far;

image

Pips indicate *subdepartment*, whilst a silver or gold stripe and silver base indicates *head-of-staff*. Alternate options include a cross stamp for chaplain, a headshot/bust silhouette, a black S, a lanyard clip, and some other odds and ends.

Verkister:
At glance they kinda look like wall signs without the angle.

Atermonera:
Ok, sorry for the further delay. Talked with staff, only real concern is that the dither on the top color bar in some cases makes it really hard to distinguish the color of the pips. Might try removing it // removing it specifically around pips? Unless you've got any other ideas for making that a bit more readable?

KillianKirilenko:
Switched from dither to a stripe for all stack-assembled cards;
image

Didn't edit the premade cards but I can if these look satisfactory.

Atermonera:
:+1: Yeah, that's definitely a lot cleaner.

Atermonera:
@KillianKirilenko just verifying, is this ready for merge? (If not just let me know when)

KillianKirilenko:
Yup, it's good to go.

PR #7817, opened 2021-01-03
Adds StrongDMM to the main repository

Atermonera:
I've been using it for a while and if we're to avoid a mapwar I should probably include it so other people start using it too. 

TODO: 

- [x] Github wiki guide to using it. 
- [x] Readme in the file, with appropriate licensing.
- [x] Also an actual wiki entry? Will pester cerebul about that.


Atermonera:
Wrote the github wiki entry. As it turns out, this is licensed with GPL-V3, so I'm waiting to hear back from ZekeS on what all I need to do to cover our asses by including the executables in the repo. And also including both windows and linux executables while I'm at it because Neerti and Zekes aren't the only ones that struggle with Wine.

PR #7818, opened 2021-01-03
Mech Appearance Update

Mechoid:
:cl:
rscadd - Mechs are scaled to be more in-line with proper size.
bugfix - Taur appearances now properly cut the leg sprites away.
/:cl:


Atermonera:
Pictures of the new ripley, as well as size comparisons of those that were resized?

Mechoid:
image

PR #7820, opened 2021-01-04
Adds the Restraining Bolt.

Mechoid:
:cl:
rscadd - Adds the Restraining Bolt to Cargo.
/:cl:

Restraining Bolts:
When implanted into human/likes, mechanical augments are unable to function.

When installed into a Cyborg, movement is slowed, wireless and access are disabled, and outbound binary communication is lost.
A Cyborg can take 90 seconds to break the restraining bolt, with similar messages as a human breaking handcuffs. It's important to note that **a cyborg with a restraining bolt still has full module functions, and remains dangerous**, but significantly less so due to the hinderances applied by the bolt.


Greenjoe12345:
I feel this should be two separate items
one restraining bolt for borgs
one augment disruptor implant for humans
putting a restraining bolt on a human seems.. odd

code/modules/mob/living/silicon/robot/robot.dm
```diff
@@ -475,6 +477,20 @@
 
 				return
 
+		if(istype(W, /obj/item/weapon/implant/restrainingbolt) && !cell)
+			if(bolt)
+				to_chat(user, "<span class='notice'>There is already a restraining bolt installed in this cyborg.</span>")
```
Atermonera, line 482: So if a borg breaks their bolt, they can't be re-bolted until someone with access actually takes out the defective bolt?

code/modules/mob/living/carbon/human/human_helpers.dm
```diff
@@ -202,6 +202,20 @@
...
+	var/obj/item/weapon/implant/restrainingbolt/RB
+
+	for(var/obj/item/organ/external/EX in organs)
+		RB = locate() in EX
+		if(istype(RB))
+			break
+
+	if(RB)
+		if(!RB.malfunction)
+			return TRUE
```
Atermonera, line 215: If a human has two bolts implanted, only one broken, and this happens to find the broken one first, the working one won't function at all until the broken one is removed.

code/modules/mob/living/silicon/robot/robot.dm
Mechoid: It takes access to place the bolt _in_ the borg, as the panel needs to be open at the moment.

code/modules/mob/living/carbon/human/human_helpers.dm
Mechoid: Fixt.

PR #7834, opened 2021-01-14
fixes some sprites, adds a thin blindfold

Greenjoe12345:
This PR fixes some clothing item sprites that were in the wrong .dmi file. It also adds a thin white blindfold, this gives you the same screen overlay as if you wore a pair of welding goggles, but doesn't give you welding protection. The blindfold onmob sprites for Teshari are also adjusted to fit better.


Schnayy:
Flash protection is subject to change if we find it to be a concern, but will go ahead and merge.

PR #7839, opened 2021-01-24
Fixed global_announce autospeaker

Cerebulon:
#7687 didn't fully port the global_announce changes from https://github.com/VOREStation/VOREStation/pull/7917 so it borked a whole lot!
Maybe this will incidentally fix arrivals announcements, but probably not lol.


Cerebulon:
Ok why the fuck are old commits being appended because I didn't put them in Master or anything dumb like that this is literally a branch

PR #7844, opened 2021-01-26
Makes wrenches only anchor lockers if they're open

Shadow-Quill:
It doesn't make sense to me that lockers can only be anchored if they're welded shut.
So I've made it that a locker has to be open in order to be (un)bolted down.

:cl:
tweak: Lockers can now only be (un)wrenched if the door is open.
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/__closets.dm
+++ b/code/game/objects/structures/crates_lockers/__closets.dm
@@ -272,7 +272,7 @@
 				anchored = !anchored
 				return
 		else
-			to_chat(user, "<span class='notice'>You can't reach the anchoring bolts when \the door is closed!</span>")
+			to_chat(user, "<span class='notice'>You can't reach the anchoring bolts when the door is closed!</span>")
 	else if(opened)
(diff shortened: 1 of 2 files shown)


Atermonera:
If applying a wrench to a locker when it's open toggles anchored, how do you put a wrench into a locker?

Shadow-Quill:
> If applying a wrench to a locker when it's open toggles anchored, how do you put a wrench into a locker?

Same way you put a wrench on a table, by throwing it.

KillianKirilenko:
Click-dragging tools from the hand slot to the locker might also work, I think it works for wrenches and tables at least? Would probably be easy enough to add the same logic if it doesn't already exist.

You could also just move into the open locker tile and drop the wrench since open lockers aren't dense/don't block movement.

Schnayy:
Will click-dragging from the hand work on the locker, the same as it does for tables?

KillianKirilenko:
Just tested locally, and no, click-dragging wrenches onto lockers will not place the wrench in the locker. Probably easy enough to add though.

PR #7853, opened 2021-02-06
QoL Door Closing

Shadow-Quill:
Clicking on the tile (with an empty hand) of an airlock or firelock, will cause you to click on the air/firelock instead.

This way, you can close airlocks and/or firelocks easier, just by clicking the tile that the open door is on.

Airlocks take priority over firelocks.

:cl:
add: QoL feature that lets you close doors/firelocks while they're open, simply by clicking on the tile. (Airlocks take priority over firelocks, so if you need to close a firelock that's on the same tile as an airlock, you'll have to click on the pixels.)
/:cl:

### DIFF
--- a/code/game/turfs/turf.dm
+++ b/code/game/turfs/turf.dm
@@ -66,6 +66,16 @@
 	return 1
 
 /turf/attack_hand(mob/user)
+	//QOL feature, clicking on turf can toogle doors
+	var/obj/machinery/door/airlock/AL = locate(/obj/machinery/door/airlock) in src.contents
+	if(AL)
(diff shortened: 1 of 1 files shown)


Atermonera:
What if it's a light tile  (Are there any other floortiles with special attack_hand logic?)

Shadow-Quill:
> What if it's a light tile (Are there any other floortiles with special attack_hand logic?)

Toggles the light, and I don't think other turfs have special attack_hand stuff?
The toggling thing isn't an issue, imo. ~~Because who even makes light floors anymore? /s~~

PR #7863, opened 2021-02-09
Makes event managers admins as far as TGUI is concerned

Cerebulon:
TGUI finally defeated?????


Cerebulon:
Not sure why this is runtiming in unit tests, it says the error is in the trait menu which... Works fine?

Atermonera:
It says the error is in specialized department goals

PR #7866, opened 2021-02-10
New Shutters and Sprite replacements

Woodratt:
- Ported shutters from World
- Replaced request console, lightswitch, and newscaster sprites, (from Virgo)


code/game/machinery/doors/blast_door.dm
```diff
@@ -323,5 +328,64 @@ obj/machinery/door/blast/regular/open
...
+	maxhealth = 400
+	block_air_zones = 0
+	opacity = 0
+	istransparent = 1
+
+obj/machinery/door/blast/gate/open
+	icon_state = "tshutter0"
+	density = 0
+
+obj/machinery/door/blast/gate/proc/force_close_transparent()
```
Schnayy, line 350: To my understanding, force_close already checks to see if the source should be transparent, making this proc unneeded.

Schnayy (changes requested):
I believe you have an unneeded proc in here.

Schnayy (approved):

PR #7881, opened 2021-02-13
Adds area soundproofing.

Verkister:
Areas now have a var for soundproofing in order to block sounds from other areas from entering or sounds from within leaving the area.


KillianKirilenko:
So heads up, I think this throws runtimes if the turf loc is null. Mostly an issue for new-player 'mobs' on the login screen by the looks of things.

```
Runtime in sound.dm,7: Cannot read null.loc
proc name: playsound (/proc/playsound)
usr: (MYCKEY) (/mob/new_player) @ (NULL) (0,0,0) (NULL) (ckey=MYCKEY)
usr.loc: *null*
```

PR #7888, opened 2021-02-15
Rebases onto VOREStation (You may need to reload the page to view it)

Atermonera:
### Does this mean we're adding vore?
# No. We are not adding vore.
# The goal of this is to have precisely the same set of features as before, with any extra features that Vore (The server) hasn't already backported to us, that we decide we want. The only mechanical changes that should result from this (After appropriate live tests) are those that we actually want.

Making this so I can parse the diff a bit more intelligently while sorting through all the changes to find what we do/don't want (Especially .dmi changes). If anyone else wants to look through this stuff, please make review comments because there's literally 3020 changed files and I'm only gonna scroll through them so many times. This will be tested live, because that's just the sane thing to do, and I'm pretty sure no other dev is crazy enough to review this with me. When this actually approaches a state of completion, I will instate a Code Freeze.

Why are we doing this (now)?
Most of the code we've been getting lately has been backports from Vore, and with our own much more limited supply of devpower (Whether that will grow in the near future is entirely speculative), our de facto relationship is really one where we're the downstream, where the de jure relationship is one where they're the downstream. This leaves Vore in a position of relative power over us if we want to keep getting shiny things from them (And thus keep getting shiny things in _general_), because if they choose to stop backporting to us, there's not much we can do besides start doing so ourselves (Which is work, which requires people we frankly don't have). By formally reversing the relationship, we can just set up a mirror for all of their PRs, and decide ourselves whether we want any particular changes or tweaks, much as they have done for the past few years.
This is something that I personally have waffled on for the past few months, and input from bringing it up at staff discussions has been "If you're willing to put in the work, and you think it'll make your job easier, then go for it."


zekesulastin:
merge posthaste

KillianKirilenko:
...well this is one move *I* definitely didn't see coming.

LunarFleet:
https://www.youtube.com/watch?v=jBOMlWl7fFk

Razgriz1032:
Hey just a heads up, I kinda tried this a year ago. It was BAD. What I did instead was take Virgo's code, turn on Southern Cross instead, and port individual features instead of entire systems.

Razgriz1032:
Oh yeah, and if you do try porting it, be prepared to deal with this comms bug. It has plagued us for months and Virgo hasn't even looked. It'll fuck up any robusting.
https://github.com/VOREStation/VOREStation/issues/8454

Screemonster:
We've been wanting to make the vore stuff a config option for a while so downstreams can just turn it off without having to rip it out of the code.

TheFurryFeline:
> We've been wanting to make the vore stuff a config option for a while so downstreams can just turn it off without having to rip it out of the code.

Lot of stuff to refactor to be a configurable, suffice to say. Doable on paper, but in practice may well be a headache to make work eventually.

Novacat:
Do note that we will likely have to do a massive reorganization of our own code if this is approved.

TheFurryFeline:
> Do note that we will likely have to do a massive reorganization of our own code if this is approved.

I wonder what'll have to be done with our mirrorbot in turn.... 🤔

MistakeNot4892:
Not to pick a fight or anything, I don't go here really, but this is literally unreviewable.

TheFurryFeline:
Ater knows that and said it in the op

MistakeNot4892:
I was responding to:
`Making this so I can parse the diff a bit more intelligently while sorting through all the changes`
`If anyone else wants to look through this stuff, please make review comments`

Because: 
image

TheFurryFeline:
> I was responding to:
> `Making this so I can parse the diff a bit more intelligently while sorting through all the changes`
> `If anyone else wants to look through this stuff, please make review comments`
> 
> Because:
> image

Ah, fair enough, forgot that can happen.

Wickedtemp:
So... in a way, Vorestation just ate Polaris.

Atermonera:
Haven't had internet for the past two days which has sorta interrupted this when I wanted to use GitHub to make a more readable diff. 

@MistakeNot4892 to be fair, it's 3000/3020 files, but I was moreso asking if anyone wanted to review _pieces_ of it at their leisure because I'm gonna be working on this for a while. 

If I think of a more manageable way of doing this, then I may well try and switch over to that (just flat out swapping the code over to vore master is appealing but for the fact that I/we'd still have to review all the vorestation edits and appraise all the other missing/different features, where this method at least compiles them into a diff. The commits I make working on this will probably be in a more or less linear order, but a part of the goal here is so that if/when this lands on the mirror (ideally as the _last_ thing to land on it, at least without adjustments to it), it will at least more easily document what changes vore will want/need to address to become the official upstream

drexample:
based

Atermonera:
Update: 1266/3012 files reviewed. Most of which are map files, sounds, icon files, etc, that I don't really need to care about.

PR #7889, opened 2021-02-15
Makes simplemob braingibs not create brainmobs

Verkister:
Every single simplemob that was gibbed or butchered would spawn a new carbonmob inside their braingib.


Cyantime:
Maybe make it work for simplemobs that have a client too, for dumb brain transplant science

PR #7892, opened 2021-02-15
Changed ore New() to Initialize(), calls parent.

MistakeNot4892:
Might fix hard dels, either way this is an invalid New().


MistakeNot4892:
Ah I didn't see #7891.
Would suggest replacing New() with Initialize() @Novacat

PR #7893, opened 2021-02-15
Sideports a couple of init unit tests from Neb.

MistakeNot4892:
- Ports two atom subsystem unit tests from Neb/Bay. I couldn't get them all working by the time I ran out of motivation, so they are currently unticked. I would recommend trying to sort out the turf and decal issues discussed in the comments of this PR and enabling them at some point.
- Cleans up a bunch of Initialize() procs that didn't call parent.
- Adds some explicit bypassing of init for unsimulated turfs rather than relying on `initialized = TRUE`
- Replaces several `del` calls with `qdel` for non-client non-image atoms.
- Moves sheet logic onto Initialize() from New().
- Makes the sheet stacker use initial() instead of creating and discarding instances.
- Probably other small stuff to get as many checks passing as possible.

**Some commentary on the issues and advice for future PRs:**
- Don't use `del()` for anything that isn't a client or an image. Use `qdel()` and return early if you need to. The garbage collection system exists to avoid hard deletions, not using it is just deciding you want more lag for no benefit.
- Always call `..()` in `Initialize()`. Don't rely on `initialized = TRUE`, you are breaking a critical call chain if you don't call all the way to `/datum/proc/Initialize()`. If you DO refrain from calling `..()`, you MUST ensure you set `initialized` appropriately and return an appropriate hint, as well as making sure you manually handle any other state setting defined in `Initialize()` lower in the chain.
- Always return a value to `Initialize()`, either via `. = ..()` or manually (ie. when something should qdel). Values should be initialization hints `(INITIALIZE_HINT_BLAH`). The result of Initialize() is used to detect bad Initialize() calls that could be introducing unwanted behavior, like the recent ore hard del fix.
- Never `qdel()` an atom in `New()` or `Initialize()`, return `INITIALIZE_HINT_QDEL` to `Initialize()` instead. Initialization can often rely on neighboring or holding atoms to exist before an individual atom is initialized, so returning the hint means the item will be kept around until it is no longer potentially needed, then it will be qdeleted.
- Likewise never `qdel()` other atoms in `New()` or `Initialize()`, as they may not have initialized themselves. In cases like merging decals, update the existing atom and qdel yourself instead.
- Always return a value to `/hook` procs, it uses the return value to check if anything runtimed or broke during the hook. If you don't return a value the hook will erroneously report as being broken.
- Never `sleep()` in `New()` or `Initialize()`, it messes with `SSatoms`. Use `addtimer(CALLBACK(src, /some/proc, some_args), some_time_in_deciseconds)` and put your logic in `/some/proc` instead. ~~Honestly this goes for almost all usage of `sleep()` and `spawn()`, use a timer instead~~.


MistakeNot4892:
Struggling to chase down the remaining init failures:
````
Path : /obj/item/weapon/storage/bag/circuits/basic 
- Slept during Initialize()
Path : /obj/effect/floor_decal/borderfloorblack 
- Qdel'd in New()
Path : /obj/effect/floor_decal/industrial/danger 
- Qdel'd in New()
````
And the remaining turf failures:
````
(the floor) (131,70,2) (/turf/simulated/floor/tiled/dark)
(the silver blue carpet) (181,112,2) (/turf/simulated/floor/carpet/sblucarpet)
(the silver blue carpet) (181,112,2) (/turf/simulated/floor/carpet/sblucarpet)
(the silver blue carpet) (181,112,2) (/turf/simulated/floor/carpet/sblucarpet)
(the silver blue carpet) (181,112,2) (/turf/simulated/floor/carpet/sblucarpet)
(the silver blue carpet) (181,112,2) (/turf/simulated/floor/carpet/sblucarpet)
(the linoleum) (121,159,2) (/turf/simulated/floor/lino)
(the linoleum) (121,159,2) (/turf/simulated/floor/lino)
(the linoleum) (121,159,2) (/turf/simulated/floor/lino)
(the linoleum) (121,159,2) (/turf/simulated/floor/lino)
(the floor) (140,161,2) (/turf/simulated/floor/tiled/dark)
(the floor) (140,161,2) (/turf/simulated/floor/tiled/dark)
(the linoleum) (117,164,2) (/turf/simulated/floor/lino)
(the linoleum) (117,164,2) (/turf/simulated/floor/lino)
(the linoleum) (117,164,2) (/turf/simulated/floor/lino)
(the linoleum) (117,164,2) (/turf/simulated/floor/lino)
(the blue carpet) (142,124,3) (/turf/simulated/floor/carpet/blue)
(the blue carpet) (142,124,3) (/turf/simulated/floor/carpet/blue)
(the floor) (139,201,5) (/turf/simulated/floor/tiled/steel_grid)
(the floor) (99,202,5) (/turf/simulated/floor/tiled)
(the floor) (221,145,8) (/turf/unsimulated/floor)
(TTS Circuit) (0,0,0) (/mob/living/voice)
(B-9 (AI Eye)) (0,0,0) (/mob/observer/eye/aiEye)
(B-9) (0,0,0) (/mob/living/silicon/ai/announcer)
````

If anyone is aware of mechanics that would be reversing a successful initialization, or messing with decals on a turf, let me know, because I am stumped on these ones. Their Initialize() chains are complete, and the only qdel is via hint, but something is qdelling them during their init anyway.

MistakeNot4892:
Rolling back some of the initialization changes because of unexplained behavoior, specifically floor decals qdeling themselves.

MistakeNot4892:
I've tried all day to isolate the last ones and failed, so, rip.

Atermonera:
Re:
```Path : /obj/item/weapon/storage/bag/circuits/basic 
- Slept during Initialize()```

It looks like `/obj/item/integrated_circuit/input/signaler` can call `sleep()` on Initialize when it tries to `set_frequency()` if the global `radio_controller` hasn't initialized yet? Might be the cause, sorta hard to parse sleep calls when they're spread across 621 calls in 266 files.

MistakeNot4892:
Still getting the sleep fail unfortunately. Was a good guess though. Will try sorting through the other components in case they can be singled out as causes.

PR #7903, opened 2021-02-18
Ports TGUI Inventory Screen

Cerebulon:
Port of https://github.com/VOREStation/VOREStation/pull/8835

### DIFF
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -193,10 +193,6 @@
 				client.eye = loc
 		return TRUE
 
-
-/mob/proc/show_inv(mob/user as mob)
-	return
-
 //mob verbs are faster than object verbs. See http://www.byond.com/forum/?post=1326139&page=2#comment8198716 for why this isn't atom/verb/examine()
 /mob/verb/examinate(atom/A as mob|obj|turf in view())
 	set name = "Examine"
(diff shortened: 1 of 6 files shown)


code/modules/mob/mob.dm
```diff
@@ -478,16 +474,6 @@
 /mob/proc/pull_damage()
 	return 0
 
-/mob/MouseDrop(mob/M as mob)
-	..()
-	if(M != usr) return
-	if(usr == src) return
-	if(!Adjacent(usr)) return
-	if(usr.incapacitated(INCAPACITATION_STUNNED | INCAPACITATION_FORCELYING | INCAPACITATION_KNOCKOUT | INCAPACITATION_RESTRAINED)) return //Incapacitated.
```
Atermonera, line 486: It doesn't look like this particular check was transferred over to the new UI? Or is that handled by the tgui checks?

Cerebulon: If I'm not mistaken this is the code which opens the inventory when you click drag from a person (unless its you, out of range, incapped)? If so it seems to be handled elsewhere because the inventory does open correctly when you do that. I can double check the conditions tonight but it seems like if this check was entirely absent now then it wouldn't function at all.

Cerebulon: Ok yeah it's covered elsewhere. Now instead of not opening the inventory menu at all (except adjacency stuff), these checks are in TGUI and just don't let you interact with the menu (it's greyed out if you're, say, handcuffed)

PR #7914, opened 2021-02-21
Universal Anomalies

Mechoid:
:cl:
rscadd - Adds the ability for any atom to be Anomalous through the power of DCS.
rscadd - Adds visual effects for anomalous objects, denoting their activity.
rscadd - Anything that counts as 'anomalous' through the is_anomalous() proc can be stored in an anomaly container.
/:cl:


code/modules/xenoarcheaology/effect_master.dm
```diff
@@ -0,0 +1,379 @@
...
+/*
+ * Here there be dragons.
+ */
+
+/atom/var/datum/artifact_master/artifact_master
+
+/atom/Initialize()
+	. = ..()
+
+	if(ispath(artifact_master))
```
MistakeNot4892, line 11: I would very strongly recommend against this. Many atoms don't call the entire Initialize() chain, some deliberately, and putting it here means _every other non-datum non-constant in the game_ is going to be doing this check during init, for the sake of what looks like is primarily for adminbuse?

This is the exact use-case that components and extensions were intended for, and Polaris definitely seems to have components - it might be worth looking into them rather than tighly coupling it to atoms.

code/modules/xenoarcheaology/effect_master.dm
```diff
@@ -0,0 +1,379 @@
...
+		artifact_master = new artifact_master(src)
+
+/atom/Destroy()
+	if(artifact_master)
+		QDEL_NULL(artifact_master)
```
MistakeNot4892, line 16: QDEL_NULL handles null vars fine, you don't need to check it beforehand.

code/modules/xenoarcheaology/effect_master.dm
```diff
@@ -0,0 +1,379 @@
...
+	if(artifact_master)
+		artifact_master.on_attack_hand(user)
+
+/atom/attackby(obj/item/weapon/W as obj, mob/living/user as mob)
+	. = ..()
+
+	if(artifact_master)
+		artifact_master.on_attackby(W, user)
+
+/atom/Bumped(M as mob|obj)
```
MistakeNot4892, line 32: I don't believe `as mob|obj` does anything in this context. The useful places for `as blah` are inputs, verb args, and for loops.

MistakeNot4892:
There seems to be a huge amount of copypaste in this PR - definitely feel like it would benefit from properly datumizing the xenoarch effects and using a component instead of just jamming all of this logic into the root `/atom` procs.

Mechoid:
I have learned the power of DCS. My power is infinite and my day is good.

code/modules/xenoarcheaology/effect_master.dm
```diff
@@ -0,0 +1,383 @@
...
+					my_effect.ToggleActivate()
+					triggered = TRUE
+
+	if(triggered)
+		return COMPONENT_IGNORE_EXPLOSION
+
+	return
+
+/datum/component/artifact_master/proc/on_bullet()
+	var/obj/item/projectile/P = args[2]
```
Atermonera, line 163: does the callback usage prevent you from outright naming the arguments as parameters?

code/modules/xenoarcheaology/effect_master.dm
```diff
@@ -0,0 +1,383 @@
...
+
+		my_effect.process()
+
+		//COLD ACTIVATION
+		if(trigger_cold)
+			if(my_effect.trigger == TRIGGER_COLD && !my_effect.activated)
+				my_effect.ToggleActivate()
+		else
+			if(my_effect.trigger == TRIGGER_COLD && my_effect.activated)
+				my_effect.ToggleActivate()
```
Atermonera, line 343: All of these checks specifically want to perform the activation when one condition is true and the other is false, so you could just use XOR
```suggestion
if(my_effect.trigger == TRIGGER_COLD && (trigger_cold ^ my_effect.activated))
	my_effect.ToggleActivate()
```
That is, it'll run on `(A & !B) | (!A & B)` ~= `!((A & B) | (!A & !B))`

code/modules/xenoarcheaology/effect.dm
```diff
@@ -3,16 +3,24 @@
...
 	var/chargelevelmax = 10
 	var/artifact_id = ""
 	var/effect_type = 0
 
-/datum/artifact_effect/New(var/atom/location)
+	var/req_type = /atom/movable
+
+/datum/artifact_effect/Destroy()
+	if(master)
+		master = null
```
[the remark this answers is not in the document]
Mechoid: The only time the effect is destroyed is via verb or master destruction, and it handles the deletion of the effects it controls.

code/modules/xenoarcheaology/effect_master.dm
Mechoid: Due to the system, I'd had trouble getting it to work right without manually specifying from the args list.

code/__defines/dcs/signals.dm
```diff
@@ -55,8 +55,8 @@
 #define COMSIG_ATOM_AFTER_SUCCESSFUL_INITIALIZE "atom_init_success"
 ///from base of atom/attackby(): (/obj/item, /mob/living, params)
 #define COMSIG_PARENT_ATTACKBY "atom_attackby"
-///Return this in response if you don't want afterattack to be called
-	#define COMPONENT_NO_AFTERATTACK (1<<0)
+///Return this in response if you don't want later item attack procs to be called.
+	#define COMPONENT_CANCEL_ATTACK_CHAIN (1<<0)
```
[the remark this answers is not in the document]
Mechoid: Both component returns did the exact same thing, but NO_AFTERATTACK was named more specifically than CANCEL_ATTACK_CHAIN, so I generalized what was present.

Cerebulon:
Personal feedback from the test today:
- Make adding anomalous properties (anomaly master + tech stuff, currently an if statement) into a proccall for wider use + ingame calls might be good.
- Using scanner outline system or similar to provide visual feedback to an object being actively weird. Maybe not always idk.
- Object with anom master VV'd could not go in anom container. Seems to apply to any small-nomaly.

code/_onclick/item_attack.dm
```diff
@@ -25,27 +25,46 @@ avoid code duplication. This includes items that may sometimes act as a standard
...
+	if(SEND_SIGNAL(src, COMSIG_ITEM_PRE_ATTACK, A, user, params) & COMPONENT_CANCEL_ATTACK_CHAIN)
+		return TRUE
+	return FALSE //return TRUE to avoid calling attackby after this proc does stuff
 
 //I would prefer to rename this to attack(), but that would involve touching hundreds of files.
 /obj/item/proc/resolve_attackby(atom/A, mob/user, var/attack_modifier = 1, var/click_parameters)
-	pre_attack(A, user)
 	add_fingerprint(user)
+	. = pre_attack(A, user, click_parameters)
+	if(.)	// We're returning the value of pre_attack, in the event it has a special return rather than a boolean.
```
MistakeNot4892, line 48: Comment is misleading - this will return for any truthy value, not simply a special (assuming string/instance/numeric?) value.

code/game/atoms.dm
```diff
@@ -157,6 +159,9 @@
 	return
 
 /atom/proc/bullet_act(obj/item/projectile/P, def_zone)
+	if(SEND_SIGNAL(src, COMSIG_ATOM_BULLET_ACT, P, def_zone) & COMPONENT_CANCEL_ATTACK_CHAIN)
+		return
```
MistakeNot4892, line 163: This seems like a strange place to use that component given that bullet_act isn't really an attack chain.
Also unclear why you want to avoid calling `P.on_hit()` considering this will mean shooting an artifact with an ion rifle or such will just ignore the projectile.

code/modules/xenoarcheaology/anomaly_container.dm
```diff
@@ -37,7 +46,16 @@
...
 
+/atom/MouseDrop(var/obj/structure/anomaly_container/over_object)
+	. = ..()
+
+	if(src && istype(loc, /turf) && is_anomalous() && Adjacent(over_object) && CanMouseDrop(over_object, usr))
+		Bumped(usr)
+		over_object.contain(src)
+
+/*
 /obj/machinery/artifact/MouseDrop(var/obj/structure/anomaly_container/over_object)
```
MistakeNot4892, line 57: Seems a bit unnecessary to keep this commented code here given that the above mousedrop makes it completely redundant.

code/modules/xenoarcheaology/anomaly_container.dm
```diff
@@ -37,7 +46,16 @@
 	underlays.Cut()
 	desc = initial(desc)
 
+/atom/MouseDrop(var/obj/structure/anomaly_container/over_object)
+	. = ..()
+
+	if(src && istype(loc, /turf) && is_anomalous() && Adjacent(over_object) && CanMouseDrop(over_object, usr))
```
MistakeNot4892, line 52: Need to typecheck `over_object`.

MistakeNot4892: Also probably `QDELETED(src)` rather than just `src`.

code/modules/xenoarcheaology/artifacts/artifact.dm
```diff
@@ -5,25 +5,39 @@
...
 	density = 1
+
+	var/predefined_icon_num
+
+	var/datum/component/artifact_master/artifact_master = /datum/component/artifact_master
+
+	var/being_used = 0
+
+/*
 	var/datum/artifact_effect/my_effect
```
MistakeNot4892, line 16: More commented code being kept around - generally not advisable for housekeeping/searching purposes. The logic is in the git history if someone wants to go digging for the old implementation.

code/modules/xenoarcheaology/artifacts/artifact.dm
```diff
@@ -5,25 +5,39 @@
...
+*/
 
 /obj/machinery/artifact/New()
 	..()
 
+	if(ispath(artifact_master))
+		AddComponent(artifact_master)
+
+		artifact_master = GetComponent(artifact_master)
+	else
```
MistakeNot4892, line 37: I feel like this `else` should be `if(!istype(artifact_master)` since the following lines assume it is an instance of the correct type.

code/modules/xenoarcheaology/artifacts/artifact.dm
```diff
@@ -77,24 +93,15 @@
...
+/obj/machinery/artifact/update_icon()
+	..()
 
+	if(LAZYLEN(artifact_master.get_active_effects()))
+		icon_state = "ano[icon_num]1"
+	else
+		icon_state = "ano[icon_num]0"
 
+/*
 /obj/machinery/artifact/process()
```
MistakeNot4892, line 105: More commented code >:

code/modules/xenoarcheaology/effect.dm
```diff
@@ -3,19 +3,45 @@
...
+
 	..()
-	holder = location
+
+/datum/artifact_effect/New(var/datum/component/artifact_master/newmaster)
+	..()
+	if(!newmaster)
+		qdel(src)
```
MistakeNot4892, line 32: Hmmmm not sure about this. Letting it runtime would probably be more informative than (for example) something trying to create an instance, getting a qdeleted instance, and keeping a reference to it anyway.

code/modules/xenoarcheaology/effect.dm
```diff
@@ -3,19 +3,45 @@
...
+		return
+
+	master = newmaster
 	effect = rand(0, MAX_EFFECT)
 	trigger = rand(0, MAX_TRIGGER)
 
+	if(effect_icon && effect_state)
+		if(effect_state == "sparkles")
+			effect_state = "sparkles_[rand(1,4)]"
+		active_effect = image(effect_icon, null, effect_state)
```
MistakeNot4892, line 42: Second arg for image() is state isn't it?

code/modules/xenoarcheaology/effect.dm
```diff
@@ -37,23 +63,31 @@
 /datum/artifact_effect/proc/ToggleActivate(var/reveal_toggle = 1)
 	//so that other stuff happens first
 	spawn(0)
```
MistakeNot4892, line 65: Would be better to change this to a `set waitfor = FALSE` while you're in the area.

code/_onclick/item_attack.dm
```diff
@@ -25,19 +25,35 @@ avoid code duplication. This includes items that may sometimes act as a standard
...
+		. = A.attackby(src, user, attack_modifier, click_parameters)
+
+	return .
 
 // No comment
 /atom/proc/attackby(obj/item/W, mob/user, var/attack_modifier, var/click_parameters)
-	if(SEND_SIGNAL(src, COMSIG_PARENT_ATTACKBY, W, user, click_parameters) & COMPONENT_NO_AFTERATTACK)
+	if(SEND_SIGNAL(src, COMSIG_PARENT_ATTACKBY, W, user, click_parameters) & COMPONENT_CANCEL_ATTACK_CHAIN)
 		return TRUE
 	return FALSE
```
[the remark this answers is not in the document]
MistakeNot4892: Not strictly speaking 1:1, `&` is bitwise and will return a truthy value but not explicitly TRUE necessarily.
`return !!(SEND_SIGNAL(src) & FART)` would do it though.

code/game/atoms.dm
Mechoid: This allows something like, say, a signal-based shield to negate impacts.

code/modules/xenoarcheaology/effect.dm
Mechoid: If it is, that's strange. VSCode lookup said it's loc

MistakeNot4892: Might be magic BYOND horseshit, other usage of image() in polaris goes `image(icon, icon_state)`

Atermonera: Yup, it's BYOND esoterica ([docs](https://www.byond.com/docs/ref/#/proc/image):
The official invocation is
`    image(icon,loc,icon_state,layer,dir) `
but there's a special case where the second argument is a string:

> Another common use of images is in making an overlay:
> ```
> overlays += image('pants.dmi',icon_state = "red")
> ```
> Since the loc argument could never be a text string, the above statement can be further shortened:
> ```
> overlays += image('pants.dmi',"red")
> ```

MistakeNot4892 (approved):
Some of this is over my head (signal stuff particularly) but nothing leaps out as horrible in the logic. If this has been tested thoroughly then I have no issue with it being merged.

MistakeNot4892:
Needs a squash though.

PR #7925, opened 2021-02-23
Makes custom cigarettes even more custom-ier

BillyBangles:
Adds an intermediary step between rolling paper and joint wherein you can add more dried plants to the rolling paper before you roll it. Once you're satisfied, you can click on the rolling paper in an active hand to finish it. Also adds blunt wraps and blunts. Blunts can hold more reagents than joints, burn for longer, and contain a bit of nicotine as well. Now you can finally add dried banana and berries to your ambrosia for the fruity smoking experience all smokers crave [citation needed]. Alternatively you can dry some cabbage, tomatoes, and carrots and smoke a salad if that's what you really want to do.

Thing that this doesn't do that I would like it to do in the future: secondhand smoke. The ideal culmination of all this is to be able to weld the vents shut, turn off the scrubbers, and hotbox yourself in one of the dorm rooms or something. But that's atmos code and that scares me, so we'll get to that in a different PR. I also might consider the percentage of reagents you actually inhale from cigarettes. Currently you only inhale about 20% of the reagents in the cigarette, and I'm wondering if that's worth changing. Again, probably for another PR.

🆑 Billy Bangles
rsctweak - Adds the ability to add more than one item to a rolling paper before rolling it.
rscadd - Adds blunt wraps and blunts, which have more reagent capacity than joints.
/🆑


code/game/objects/items/weapons/cigs_lighters.dm
```diff
@@ -490,33 +490,92 @@ CIGARETTE PACKETS ARE IN FANCY.DM
...
+
+/obj/item/weapon/reagent_containers/rollingpaper/full
+	icon_state = "paper_full"
+
+/obj/item/weapon/reagent_containers/rollingpaper/attackby(obj/item/weapon/W as obj, mob/user as mob)
 	if (istype(W, /obj/item/weapon/reagent_containers/food/snacks))
 		var/obj/item/weapon/reagent_containers/food/snacks/grown/G = W
 		if (!G.dry)
-			to_chat(user, "<span class='notice'>[G] must be dried before you roll it into [src].</span>")
+			to_chat(usr, "<span class='notice'>[G.name] must be dried before you add it to [src].</span>")
```
Atermonera, line 526: That should still use `user`, using `usr` outside of verbs is generally risky behaviour.

code/game/objects/items/weapons/cigs_lighters.dm
```diff
@@ -490,33 +490,92 @@ CIGARETTE PACKETS ARE IN FANCY.DM
...
 			return
-		var/obj/item/clothing/mask/smokable/cigarette/joint/J = new /obj/item/clothing/mask/smokable/cigarette/joint(user.loc)
-		to_chat(usr, "<span class='notice'>You roll the [G.name] into a joint!</span>")
-		J.add_fingerprint(user)
+		var/obj/item/weapon/reagent_containers/rollingpaper/full/P = new /obj/item/weapon/reagent_containers/rollingpaper/full
```
Atermonera, line 528: What purpose does P serve? If your goal is simply to update the icon_state, you should just set icon_state directly, or put code to handle that in `update_icon()`

new() also doesn't require a typepath when that typepath is explicitly the type of the variable it's assigning something to. So in this case, `var/obj/item/weapon/reagent_containers/rollingpaper/full/P = new()` is sufficient

code/game/objects/items/weapons/cigs_lighters.dm
```diff
@@ -490,33 +490,92 @@ CIGARETTE PACKETS ARE IN FANCY.DM
...
+		P.reagents.maximum_volume = 25
+		if (istype(src, /obj/item/weapon/reagent_containers/rollingpaper/blunt))
+			P.name = "blunt wrap"
+			P.desc = "A small piece of easily flammable paper similar to that which encases cigars. It's made out of tobacco, bigger than a standard rolling paper, and will last longer."
+			P.icon_state = "blunt_full"
+			P.reagents.maximum_volume = 45
+		to_chat(usr, "<span class='notice'>You add the [G.name] to the [P.name].</span>")
```
Atermonera, line 535: user

code/game/objects/items/weapons/cigs_lighters.dm
```diff
@@ -490,33 +490,92 @@ CIGARETTE PACKETS ARE IN FANCY.DM
...
 		var/obj/item/weapon/reagent_containers/food/snacks/grown/G = W
 		if (!G.dry)
-			to_chat(user, "<span class='notice'>[G] must be dried before you roll it into [src].</span>")
+			to_chat(usr, "<span class='notice'>[G.name] must be dried before you add it to [src].</span>")
 			return
-		var/obj/item/clothing/mask/smokable/cigarette/joint/J = new /obj/item/clothing/mask/smokable/cigarette/joint(user.loc)
-		to_chat(usr, "<span class='notice'>You roll the [G.name] into a joint!</span>")
-		J.add_fingerprint(user)
+		var/obj/item/weapon/reagent_containers/rollingpaper/full/P = new /obj/item/weapon/reagent_containers/rollingpaper/full
+		P.reagents.maximum_volume = 25
```
Atermonera, line 529: I feel like this should be handled by `new()`. While doing this operation here is probably fine, elsewhere it might lead to risky behaviour (Things getting more reagents than their max volume). I'd rather avoid setting any (further) precedents for this operation, and afaik the max volume of a reagent container should be a constant that can be set at initialization

After some code diving, setting the `volume` var on each type will cause their reagents datums to be initialized to that max volume.

code/game/objects/items/weapons/cigs_lighters.dm
```diff
@@ -490,33 +490,92 @@ CIGARETTE PACKETS ARE IN FANCY.DM
...
+		if (istype(src, /obj/item/weapon/reagent_containers/rollingpaper/blunt))
+			P.name = "blunt wrap"
+			P.desc = "A small piece of easily flammable paper similar to that which encases cigars. It's made out of tobacco, bigger than a standard rolling paper, and will last longer."
+			P.icon_state = "blunt_full"
```
Atermonera, line 533: Again, P doesn't really do anything here, because you could just use `src` for exactly the same function

code/game/objects/items/weapons/cigs_lighters.dm
```diff
@@ -490,33 +490,92 @@ CIGARETTE PACKETS ARE IN FANCY.DM
...
-			G.reagents.trans_to_obj(J, G.reagents.total_volume)
-		J.name = "[G.name] joint"
-		J.desc = "A joint lovingly rolled and filled with [G.name]. Blaze it."
+			G.reagents.trans_to_obj(P, G.reagents.total_volume)
 		qdel(G)
+		user.put_in_hands(P)
+		user.drop_from_inventory(src)
+		qdel(src)
+
+/obj/item/weapon/reagent_containers/rollingpaper/full/attackby(obj/item/weapon/W as obj, mob/user as mob)
```
Atermonera, line 544: How much of this proc is handled by the parent `/obj/item/weapon/reagent_containers/rollingpaper`'s `attackby()`? Because from what I can see this doesn't have any unique operations

code/game/objects/items/weapons/cigs_lighters.dm
```diff
@@ -490,33 +490,92 @@ CIGARETTE PACKETS ARE IN FANCY.DM
...
+			to_chat(usr, "<span class='warning'>The [src] is too full to add [G.name].</span>")
+			return
+		to_chat(usr, "<span class='notice'>You add the [G.name] to the [src].</span>")
+		if(G.reagents)
+			G.reagents.trans_to_obj(src, G.reagents.total_volume)
+		qdel(G)
+
+/obj/item/weapon/reagent_containers/rollingpaper/full/attack_self(mob/living/user)
+	if(!src.reagents)
+		to_chat(usr, "<span class='warning'>There is nothing in [src]. Add something to it first.</span>")
```
Atermonera, line 560: user

code/game/objects/items/weapons/cigs_lighters.dm
```diff
@@ -490,33 +490,92 @@ CIGARETTE PACKETS ARE IN FANCY.DM
...
+		to_chat(usr, "<span class='notice'>You add the [G.name] to the [P.name].</span>")
+		P.add_fingerprint(user)
 		if(G.reagents)
-			G.reagents.trans_to_obj(J, G.reagents.total_volume)
-		J.name = "[G.name] joint"
-		J.desc = "A joint lovingly rolled and filled with [G.name]. Blaze it."
+			G.reagents.trans_to_obj(P, G.reagents.total_volume)
 		qdel(G)
+		user.put_in_hands(P)
+		user.drop_from_inventory(src)
```
Atermonera, line 541: put_in_hands() should come *after* the `drop_from_inventory(), or you might not have a free hand to use. G should also probably get dropped before its qdel call

Atermonera (changes requested):
In summary, the changes proposed are to:
* Repath `[...]/rollingpaper/full/attackby()` onto just `[...]/rollingpaper`,
* Set the `volume` var on each of these subtypes to correctly limit the maximum reagent volume,
* Toss out the `[...]/rollingpaper/full` subtype entirely, I'm genuinely not sure what it's trying to do that should be any different from the other subtypes.
* On attackby(), just set the icon_state directly to the `_full` state

code/game/objects/items/weapons/cigs_lighters.dm
```diff
@@ -490,33 +490,77 @@ CIGARETTE PACKETS ARE IN FANCY.DM
...
-		J.add_fingerprint(user)
+		if (G.reagents.total_volume + src.reagents.total_volume > src.reagents.maximum_volume)               //Check that we don't have too much already in the paper before adding things
+			to_chat(user, "<span class='warning'>The [src] is too full to add [G.name].</span>")
+			return
+		if (src.reagents.total_volume == 0)
+			if (istype(src, /obj/item/weapon/reagent_containers/rollingpaper/blunt))                         //update the icon if this is the first thing we're adding to the paper
+				src.icon_state = "blunt_full"
+			else
+				src.icon_state = "paper_full"
+		to_chat(user, "<span class='notice'>You add the [G.name] to the [src.name].</span>")
```
MistakeNot4892, line 536: A good pattern to use for stuff like this is `if(!user.unEquip(G, src)) return` - it will move the item into `src` and return `TRUE` if the mob can unequip the item and it successfully moves. Saves having to manually call `drop_from_inventory` and prevents stuff like borg modules ending up in toilets.

code/game/objects/items/weapons/cigs_lighters.dm
```diff
@@ -490,33 +490,77 @@ CIGARETTE PACKETS ARE IN FANCY.DM
...
-		J.desc = "A joint lovingly rolled and filled with [G.name]. Blaze it."
+			G.reagents.trans_to_obj(src, G.reagents.total_volume)                                            //adds the reagents from the plant into the paper
+		user.drop_from_inventory(G)
 		qdel(G)
+
+/obj/item/weapon/reagent_containers/rollingpaper/attack_self(mob/living/user)
+	if(!src.reagents)                                                                                        //don't roll an empty joint
+		to_chat(user, "<span class='warning'>There is nothing in [src]. Add something to it first.</span>")
+		return
+	if (istype(src, /obj/item/weapon/reagent_containers/rollingpaper/blunt))
```
MistakeNot4892, line 547: This would be better as an override on `/obj/item/weapon/reagent_containers/rollingpaper/blunt`. Generally speaking if you're ever doing `istype(src)` your design needs reworking.

MistakeNot4892: You could also set a var on `/obj/item/weapon/reagent_containers/rollingpaper` like `var/joint_type = /obj/item/clothing/mask/smokable/cigarette/joint` and set it to the blunt type on `/rollingpaper/blunt`.

MistakeNot4892:
You seem to have committed an empty file called `cigarettes.dmi`.

BillyBangles:
I'm not entirely sure why it's saying that when viewing it still shows all the proper contents.

Atermonera (approved):

PR #7928, opened 2021-02-24
Added a GPS compass.

MistakeNot4892:
image

:cl:
rscadd: Holding an active GPS in your hand will overlay a compass, which can track individual signals on your Z-level.
/:cl:


zekesulastin:
Is there a way to change or hide specifically the text flag in the compass window?  The Unidentified Signal is pretty big :p

code/game/objects/items/devices/gps.dm
```diff
@@ -17,15 +18,98 @@ var/list/GPS_list = list()
...
+	update_holder()
+
+/obj/item/device/gps/dropped()
+	. = ..()
+	update_holder()
+
+/obj/item/device/gps/process()
+	if(!tracking)
+		return PROCESS_KILL
+	update_holder()
```
Atermonera, line 74: What circumstances does this try to handle that the others don't? I'm a little wary of the `if(!(src in SSobj.processing))` on account of not knowing how big that list is, on average.

code/game/objects/items/devices/gps.dm
```diff
@@ -17,15 +18,98 @@ var/list/GPS_list = list()
...
+		holder = null
+
+	if(istype(loc, /mob))
+		holder = loc
+		GLOB.moved_event.register(holder, src, .proc/update_compass)
+		GLOB.dir_set_event.register(holder, src, .proc/update_compass)
+
+	if(holder && tracking)
+		if(!(src in SSobj.processing))
+			START_PROCESSING(SSobj, src)
```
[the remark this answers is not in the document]
MistakeNot4892: It's purely a decision by whoever implemented it, presumably because `|=` is slower than `+=` (particularly on enormous lists like the processing systems end up with). I could copy START_PROCESSING to add one with `|=` if it would be clearer.

code/game/objects/items/devices/gps.dm
```diff
@@ -17,15 +18,98 @@ var/list/GPS_list = list()
...
+		if(holder?.client)
```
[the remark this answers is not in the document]
MistakeNot4892: Good catch.

code/game/objects/items/devices/gps.dm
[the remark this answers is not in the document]
MistakeNot4892: Decided to track it on the object with a bool since that seems to be the least potentially disruptive solution.

code/game/objects/items/devices/gps.dm
MistakeNot4892: Seems to be unnecessary, I assume it was to handle an edge case that I then found a proc to hook for.

MistakeNot4892: Ah found it. equipped() doesn't fire on putting something in your pocket.

MistakeNot4892: Hm odd, it should be firing but isn't. If I proc call update_holder when it's in my pocket, it properly removes the overlay.

MistakeNot4892: Given that the processing list isn't getting checked endlessly now I think it should be fine to leave update_holder in process to catch edge cases like pockets. Can't work out why they're not updating properly in any case.

MistakeNot4892:
Ready for review again.

PR #7935, opened 2021-02-28
Added think-tanks.

MistakeNot4892:
image

Platforms/think-tanks are a robot subtype that do not select a custom icon, but assemble one from a set of greyscale icons by Pawn (thanks Pawn). They are based on tachikoma/fuchikoma from GITS both in appearance and roughly in conception. Lorewise they are flexible adaptive drones produced by NT to assist explorers on the surface in a context where crewbots would be unable to stay charged or receive repairs. I am not mapping them into the outpost garage in this PR as I'd like to see how they go onserver with supervision before making them permanently available.

Main differences from basic robots:
- They have limited access, largely revolving around Cargo, Mining and Science.
- They require access_explorer (for Recon) or access_cargo_bot (for Logistics) to open their panel, instead of access_robotics.
- They have twice the armour of basic robots and cannot be pushed (though they can be dragged).
- They spawn with a mech-equivalent cell.
- They recharge their cell when on an exterior turf (planet or space).
- They can have a cell inserted into them to recharge it from their internal battery.
- They can carry two crates or humans inside themselves (they do not provide an air supply or climate control, but it's faster than dragging them, protects from bullets, and won't exacerbate injuries.
- They do not have an associated job and cannot select a module; they are mapped in as inactive chasses and ghosts click on them to control them, so long as the ghost isn't banned from robot roles.
- They can change the colour of their eyes, and the eyes will glow over darkness.

They can also be customized with a floor painter (renamed to paint sprayer in this PR):
image

Their module consists of:
- Standard robot gear (crowbar, gps, flash)
- A phase carbine.
- A wrench.
- A screwdriver.
- A plasma cutter.
- A chainsaw.

Their emag item is an 'overvolted phase carbine' which is just a laser carbine. 

:cl:
add: Adds new Thinktank robot type specifically outfitted for exploration, cargo, and mining.
add: Thinktanks have an internal storage compartment that can carry two crates or one human mob
add: Thinktanks recharge their batteries when standing on outdoor turfs, or in space
add: Thinktanks are mapped into cargo and the planetside outpost, and can be entered as a ghost by clicking on it after the round has started.
/:cl:


code/modules/mob/living/silicon/robot/subtypes/thinktank/thinktank_interactions.dm
```diff
@@ -0,0 +1,57 @@
...
+			recharge_complete = FALSE
+			user.visible_message(SPAN_NOTICE("\The [user] slots \the [recharging] into \the [src]'s recharging port."))
+		return TRUE
+
+	if(istype(W, /obj/item/device/floor_painter))
+		return FALSE // Paint sprayer wil call try_paint() in afterattack()
+
+	. = ..()
+
+/mob/living/silicon/robot/platform/attack_ghost(mob/observer/ghost/user)
```
Atermonera, line 39: If these are mapped in, could ghosts activate them and start tooling around before round-start (Possibly even before initializations are done)?

MistakeNot4892: Not sure. It hasn't been an issue with attack_ghost stuff on Bay to my knowledge but I can put in an initialization check if you like.

code/modules/mob/living/silicon/robot/subtypes/thinktank/thinktank_interactions.dm
```diff
@@ -0,0 +1,57 @@
...
+			return TRUE
+		if(length(stored_atoms))
+			var/weakref/remove_ref = stored_atoms[length(stored_atoms)]
+			var/atom/movable/removing = remove_ref?.resolve()
+			if(!istype(removing) || QDELETED(removing) || removing.loc != src)
+				LAZYREMOVE(stored_atoms, remove_ref)
+			else
+				user.visible_message(SPAN_NOTICE("\The [user] begins unloading \the [removing] from \the [src]'s cargo compartment."))
+				if(do_after(user, 3 SECONDS, src) && !QDELETED(removing) && removing.loc == src)
+					drop_stored_atom(removing, user)
```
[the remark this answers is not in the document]
MistakeNot4892: Nope, if the user or target changes loc the do_after fails.

code/modules/mob/living/silicon/robot/subtypes/thinktank/thinktank_interactions.dm
Atermonera: I can give it a try on a live test

MistakeNot4892: Added a ticker check to attack_ghost so should be sorted now either way.

MistakeNot4892:
We had a couple of rounds of live testing, I'm confident these are fine for merge after I get them added to the map.

PR #7936, opened 2021-02-28
Ports Tertiary Colors

Novacat:
May not be complete.


code/modules/client/preference_setup/general/03_body.dm
```diff
@@ -13,20 +13,29 @@ var/global/list/valid_bloodtypes = list("A+", "A-", "B+", "B-", "AB+", "AB-", "O
 	var/r_ears2 = 30	// Ear extra color.
 	var/g_ears2 = 30	// Ear extra color
 	var/b_ears2 = 30	// Ear extra color
+	var/r_ears3 = 30	// Ear tertiary color.
+	var/g_ears3 = 30	// Ear tertiary color
+	var/b_ears3 = 30	// Ear tertiary color
```
MistakeNot4892, line 18: This is getting to a point where you should probably just be serializing/deserializing a colour matrix :P (this is a joke, not a change request)

code/modules/client/preference_setup/general/03_body.dm
```diff
@@ -1063,6 +1114,15 @@ var/global/list/valid_bloodtypes = list("A+", "A-", "B+", "B-", "AB+", "AB-", "O
 			pref.b_ears2 = hex2num(copytext(new_earc2, 6, 8))
 			return TOPIC_REFRESH_UPDATE_PREVIEW
 
+	else if(href_list["ear_color3"])
+		var/new_earc3 = input(user, "Choose your character's tertiary ear colour:", "Character Preference",
+			rgb(pref.r_ears3, pref.g_ears3, pref.b_ears3)) as color|null
+		if(new_earc3)
+			pref.r_ears3 = hex2num(copytext(new_earc3, 2, 4))
+			pref.g_ears3 = hex2num(copytext(new_earc3, 4, 6))
+			pref.b_ears3 = hex2num(copytext(new_earc3, 6, 8))
```
MistakeNot4892, line 1123: Similar to above, the various colour combination procs could probably stand to be genericized at some point. They're pretty boilerplate-y, especially since unless Polaris/Virgo are doing something very different to Bay with the component colours you can get away with turning all the `(r|g|b)_(skin|ears|eyes|etc)` vars into hex strings. Not in scope for this PR, but worth considering I think.

code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -1112,6 +1112,17 @@ var/global/list/damage_icon_parts = list() //see UpdateDamageIcon()
 			overlay.Blend(rgb(src.r_wing2, src.g_wing2, src.b_wing2), wing_style.color_blend_mode)
 			wing_s.Blend(overlay, ICON_OVERLAY)
 			qdel(overlay)
+		if(wing_style.extra_overlay2)
+			var/icon/overlay = new/icon("icon" = wing_style.icon, "icon_state" = wing_style.extra_overlay2)
+			if(wing_style.ani_state)
```
MistakeNot4892, line 1117: Seems to be wasting a new icon operation here, which is bad news. Haven't pulled it down and tested, but I think this block could be simplified to:
````
var/icon/overlay = new/icon("icon" = wing_style.icon, "icon_state" = (wing_style.ani_state ? wing_style.extra_overlay2_w : wing_style.extra_overlay2))
overlay.Blend(rgb(src.r_wing3, src.g_wing3, src.b_wing3), wing_style.color_blend_mode)
wing_s.Blend(overlay, ICON_OVERLAY)
qdel(overlay)
````
Also not sure if you need to qdel `overlay` as it's an icon and should naturally fall out of scope and get GC'd.

File: code/modules/mob/living/carbon/human/update_icons.dm [hunk 4]
```diff
@@ -1153,6 +1169,18 @@ var/global/list/damage_icon_parts = list() //see UpdateDamageIcon()
 				tail_s.Blend(overlay, ICON_OVERLAY)
 				qdel(overlay)
 
+		if(tail_style.extra_overlay2)
+			var/icon/overlay = new/icon("icon" = tail_style.icon, "icon_state" = tail_style.extra_overlay2)
+			if(wagging && tail_style.ani_state)
```
MistakeNot4892, line 1174: Similar to above, this is doing an icon op that it immediately replaces if a condition is met - could do something similar to above with a ternary condition.

code/modules/mob/new_player/sprite_accessories_tail.dm
```diff
@@ -12,11 +12,13 @@
 
 	color_blend_mode = ICON_ADD // Only appliciable if do_coloration = 1
 	var/extra_overlay // Icon state of an additional overlay to blend in.
+	var/extra_overlay2 //Tertiary.
 	var/show_species_tail = 0 // If false, do not render species' tail.
 	var/clothing_can_hide = 1 // If true, clothing with HIDETAIL hides it
 	var/desc = "You should not see this..."
 	var/ani_state // State when wagging/animated
 	var/extra_overlay_w // Wagging state for extra overlay
+	var/extra_overlay2_w // Tertiary wagging.
```
MistakeNot4892, line 21: These var names aren't super informative but that's definitely out of scope for your PR.

Atermonera (commented):
Agree with zuh that the colors should be refactored (A list of hex strings?), but also that it's out of scope for this.

PR #7940, opened 2021-03-01
Diet/Allergy Rework & Refactor Pt2

KillianKirilenko:
Refactors and builds on #7864 to allow for;
1) Purely lethal or purely nonlethal reactions
2) Differing severities for lethal/nonlethal reactions

I've moved the entire thing up to the /datum/reagent/ level where it probably should've been done in the first place to reduce code duplication. Thanks, past me!

__**IMPORTANT NOTICE:**__ this PR also adjusts the base amount of damage taken from allergens *quite significantly*. I suspect nobody ever actually *tested* the allergen reactions before my last PR, because the damage amount was barely a slap on the wrist. Now a few (~5) units of a reagent you're allergic to won't kill you outright, but it will be very, very bad for you and necessitate medical attention. Anything more than that will prove fatal within the span of a few minutes, but people can still save you if they're fairly quick in administering dylo/dex/carbon.

To put things in perspective, 5u of coffee will cause about 40-50 tox damage for a tajara, and if dylovene isn't administered immediately this will cause nasty liver damage over the long term. If they slam an entire cup of coffee they will probably explode*.

*Not literally. But they might if brute damage was a possible allergy reaction.

As always the values and reaction type can be tweaked on a per-species basis.

:cl:
tweak - allergies expanded again! Food allergies are also now actually dangerous.
/:cl:

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents.dm
+++ b/code/modules/reagents/Chemistry-Reagents.dm
@@ -162,24 +162,26 @@
 				affect_touch(M, alien, removed)
 	if(overdose && (volume > overdose * M?.species.chemOD_threshold) && (active_metab.metabolism_class != CHEM_TOUCH && !can_overdose_touch))
 		overdose(M, alien, removed)
-	if(M.species.allergens & allergen_type)	//uhoh, we can't handle this!		
+	if(M.species.allergens & allergen_type)	//uhoh, we can't handle this!	
+		var/damage_severity = M.species.allergen_damage_severity*allergen_factor
+		var/disable_severity = M.species.allergen_disable_severity*allergen_factor	
 		if(M.species.allergen_reaction & AG_TOX_DMG)
-			M.adjustToxLoss(M.species.allergen_damage_severity*allergen_factor)
+			M.adjustToxLoss(damage_severity)
 		if(M.species.allergen_reaction & AG_OXY_DMG)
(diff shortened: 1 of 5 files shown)


MistakeNot4892 (commented):
Code seems fine to me, only quibbles I have are cosmetic (using tabs to space your defines, and one extraneous trailing `return`).

code/modules/reagents/Chemistry-Reagents.dm
```diff
@@ -159,6 +162,24 @@
 				affect_touch(M, alien, removed)
 	if(overdose && (volume > overdose * M?.species.chemOD_threshold) && (active_metab.metabolism_class != CHEM_TOUCH && !can_overdose_touch))
 		overdose(M, alien, removed)
+	if(M.species.allergens & allergen_type)	//uhoh, we can't handle this!
```
cadyn, line 165: I feel like multiplying each value just once is absolutely the move here, mostly for cleanliness.
```suggestion
	if(M.species.allergens & allergen_type)	//uhoh, we can't handle this!		
		var/damage_severity = M.species.allergen_damage_severity*allergen_factor
		var/disable_severity = M.species.allergen_disable_severity*allergen_factor
```

KillianKirilenko: Good idea.

PR #7943, opened 2021-03-02
Stair tweaks

### LINKED ISSUE #7934: New stairs break pulls and cause runtimes
Pulls get broken if a user continues holding the movement key while going up stairs.
The pull to still work, especially when I was in range still.
#### What actually happened
The pull was broken, and a runtime was produced:
#### Steps to reproduce
1. Start pulling somebody (If you're on a test server, a carbon/human/dummy works great.)
2. Line yourself up with the stairs so that you can just hold down a movement key, and keep moving in the direction of the stairs.
3. Hold the movement key down as you go up the stairs
4. Your pull is broken. GG, especially if they were cuffed.
#### Additional info:
Runtime:
image

Atermonera:
Fixes #7934 , at least as far as I can understand/reproduce the issues listed. Lighting issues are a mapping issue, mostly because I don't want to go rooting through lighting code right now, and that solves it (Just replace the open_space under the top stair piece with an actual tile).

Player-controlled mobs will now transition up/down stairs by moving in the direction of the stairs (i.e., to go up, you step from the bottom-most stair into the middle stair). Works similar to old stairs, but the transition is in the visual middle.

Buckled things are brought with the mob, works similar to pulled objects.
Throwing things up/down stairs won't hurl them to the max range once they're up there, but instead the distance the user actually tried to throw it.

AI-piloted mobs will use the ""new"" interaction where they simply have to step on the stair turfs to transition, in keeping with the original goal of reworking stairs such that they are a destination that a mob can path to in order to move between z-levels, rather than a boundary which a mob must move from to perform the transition.

### DIFF
--- a/code/game/atoms_movable.dm
+++ b/code/game/atoms_movable.dm
@@ -450,8 +450,10 @@
 		minor_dir = dx
 		minor_dist = dist_x
 
+	range = min(dist_x + dist_y, range)
+
 	while(src && target && src.throwing && istype(src.loc, /turf) \
-		  && ((abs(target.x - src.x)+abs(target.y - src.y) > 0 && dist_travelled < (dist_x + dist_y)) \
+		  && ((abs(target.x - src.x)+abs(target.y - src.y) > 0 && dist_travelled < range) \
 		  	   || (a && a.has_gravity == 0) \
 			   || istype(src.loc, /turf/space)))
 		// only stop when we've gone the whole distance (or max throw range) and are on a non-space tile, or hit something, or hit the end of the map, or someone picks it up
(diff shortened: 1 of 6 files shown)


MistakeNot4892 (commented):
Going over some PRs idly due to work being quiet, apologies if this is unwanted, just let me know and I'll refrain.

PR seems good to me, nothing springs out as dangerous/busted. Only quibble I have is that `instant_stairs` isn't a super informative proc name.

It also looks like nothing ever overrides `has_buckled_mobs`, so it might be worth replacing it with the LAZYLEN check in the future to avoid the proc overhead. Doesn't seem like a hot proc though so likely isn't worth the effort.

PR #7948, opened 2021-03-04
Fix stacking machine not stacking most materials

Meghan-Rossi:
🆑
bugfix - Fixed the stacking machine not stacking most materials
/🆑
Tested

### DIFF
--- a/code/modules/mining/machine_stacking.dm
+++ b/code/modules/mining/machine_stacking.dm
@@ -86,18 +86,11 @@
 /obj/machinery/mineral/stacking_machine/New()
 	..()
 
-	for(var/stacktype in subtypesof(/obj/item/stack/material))
-		var/obj/item/stack/S = stacktype
-		var/s_name = initial(S.name)
-		stack_storage[s_name] = 0
-		stack_paths[s_name] = stacktype
-
-	stack_storage["glass"] = 0
-	stack_paths["glass"] = /obj/item/stack/material/glass
-	stack_storage[DEFAULT_WALL_MATERIAL] = 0
-	stack_paths[DEFAULT_WALL_MATERIAL] = /obj/item/stack/material/steel
-	stack_storage["plasteel"] = 0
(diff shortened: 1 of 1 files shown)


code/modules/mining/machine_stacking.dm
```diff
@@ -86,18 +86,11 @@
...
-	stack_paths["glass"] = /obj/item/stack/material/glass
-	stack_storage[DEFAULT_WALL_MATERIAL] = 0
-	stack_paths[DEFAULT_WALL_MATERIAL] = /obj/item/stack/material/steel
-	stack_storage["plasteel"] = 0
-	stack_paths["plasteel"] = /obj/item/stack/material/plasteel
+	for(var/stacktype in (subtypesof(/obj/item/stack/material) - typesof(/obj/item/stack/material/cyborg)))
+		var/obj/item/stack/material/S = stacktype
+		var/s_matname = initial(S.default_type)
+		stack_storage[s_matname] = 0
+		stack_paths[s_matname] = stacktype
```
MistakeNot4892, line 93: Not sure how feasible this is, but it might be worth trying to retrieve the material datum and check if you can use `material.stack_type` rather than using `stacktype` itself - there are overrides that set `amount` for example (ten/fifty) which will be included in the subtypesof loop and may override the path setting and get spawned by the stacker.

Alternatively you could iterate subtypesof() to build a list of materials, and then build the storage/path lists from the material names in a single pass.

Meghan-Rossi: It looks like we don't actually have subtypes for stacks with a particular amount, we use an separate object that spawns a stack and sets its amount - see https://github.com/PolarisSS13/Polaris/blob/b22a0568a3169027fe4884138033a50e92df782d/code/game/objects/items/stacks/fifty_spawner.dm

MistakeNot4892: Ah cool, my mistake then.

PR #7950, opened 2021-03-05
Allergens v2

cadyn:
So, what changed? Mostly nuances. First of all, food should now be spawning proper allergen reagents in themselves, so that Allergens will actually apply to the foods which are supposed to contain them. This is done using a little bit of code that turns the nutriment_desc variable and tries to find actual IDs from it and then adds those reagents in proportion. This code only runs for food/snack types.

Second of all, I did not like the way the allergens were handled in Chemistry-Reagents-Food-Drinks.dm, so I reviewed the whole file, and made changes to allergens where they were missing or didn't make sense, and added comments to every single allergen definition, so it's absolutely clear why those allergens are there. It's not that much for the first bit of the file, but once you get to the cocktails, if you're making modifications, you'll be glad those comments are there.


Cerebulon:
Eggs are not dairy, but I think that was placed under that umbrella because e.g. Skrell should also be 'allergic' to eggs. So you might want to either just call them dairy for that purpose, or add a specific Egg Allergen (Which also happens to be a real allergy).

cadyn:
> Eggs are not dairy, but I think that was placed under that umbrella because e.g. Skrell should also be 'allergic' to eggs. So you might want to either just call them dairy for that purpose, or add a specific Egg Allergen (Which also happens to be a real allergy).

Added egg allergy.

code/__defines/species_languages.dm
```diff
@@ -35,7 +35,8 @@
 #define DAIRY		0x80	// Lactose intolerance, ho! Also bad for skrell.
 #define FUNGI		0x100	// Delicious shrooms.
 #define COFFEE		0x200	// Mostly here for tajara.
-#define GENERIC		0x400	// Catchall for stuff that doesn't fall into the groups above. You shouldn't be allergic to this type, ever.
+#define EGGS		0x400	// For skrell. Just eggs.
+#define GENERIC		0x800	// Catchall for stuff that doesn't fall into the groups above. You shouldn't be allergic to this type, ever.
```
KillianKirilenko, line 39: #7940 is adding sugars as 0x800 and leaving generic as 0x400. I recommend just adding eggs at 0x1000 or else there's probably going to be conflicts.

cadyn:
Conflicts resolved

PR #7952, opened 2021-03-05
Allows removing omni filters/mixers at any pressure (same as non-omni)

Meghan-Rossi:
🆑
tweak - Omni filters and mixers can now be removed regardless of pressure, just like their non-omni versions.
/🆑
Tested.


Meghan-Rossi:
Restarting CI

PR #7956, opened 2021-03-07
Ported Poppy the possum.

MistakeNot4892:
Ported from Bay with some rewrites for Polaris compat.
:cl:
rscadd Adds Poppy, the Engineering Possum.
/:cl:


code/modules/tgui/states/notcontained.dm
```diff
@@ -21,6 +21,3 @@ GLOBAL_DATUM_INIT(tgui_notcontained_state, /datum/tgui_state/notcontained_state,
 
 /mob/living/silicon/notcontained_can_use_tgui_topic(src_object)
 	return default_can_use_tgui_topic(src_object) // Silicons use default bevhavior.
-
-/mob/living/simple_animal/drone/notcontained_can_use_tgui_topic(src_object)
```
MistakeNot4892, line 25: This is a path that no longer exists, just noticed it when I mistyped while spawning Poppy for testing.

Atermonera (approved):

PR #7959, opened 2021-03-08
Examining dark click catcher doesn't show turfs

### LINKED ISSUE #7890: Click Catcher returns Turfs on Shift-Click
So, #7803 fixed the darkness. Sort of. But I've found something weird with it: if you shift-click to examine, what you get instead of the Darkness is the name and description of the turf underneath the darkness. This gives you a very weird and limited ability to sense through walls, as you can gauge what's on the opposite side of something (plating? space? more wall?).

Either nothing, a "you can't see what's there" type remark, or some snark about casting magic missile at the darkness.

#### What actually happened
Limited x-ray vision.

#### Steps to reproduce
Shift-click on a darkness-obscured turf, such as one behind a wall.

#### Additional info:
None.

Atermonera:
Fixes #7890 
Tested, not really much to show when it doesn't output anything. I could make it call `face_atom()` so it still turns you, if people want that.

### DIFF
--- a/code/_onclick/click.dm
+++ b/code/_onclick/click.dm
@@ -355,5 +355,8 @@
 		var/list/P = params2list(params)
 		var/turf/T = screen_loc2turf(P["screen-loc"], get_turf(usr))
 		if(T)
+			if(modifiers["shift"])
+				usr.face_atom(T)
+				return 1
 			T.Click(location, control, params)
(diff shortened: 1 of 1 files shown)


KillianKirilenko:
I'd keep the `face_atom` thing, seems like it'd be useful.

PR #7961, opened 2021-03-08
Re-adds flashing on heavy-duty cell chargers

Shadow-Quill:
Scree.

For whatever reason, #7006 undid my changes, but not for the 0-25% charge level. Why? I 'unno.
(These changes were already accepted in #7030, if that makes any difference)


Shadow-Quill:
I did a whoops.

PR #7969, opened 2021-03-09
Adds the bluespace worm

Schnayy:
Adds the bluespace worm. For events!
image

Should work exactly the same as the previous space worm. Drops void opals rather than phoron, meat contains different chems.

Dead sprite is not finished, names may be changed, putting the DNM tag on for the meantime.


code/modules/mob/living/simple_mob/subtypes/animal/space/bluespaceworm.dm
```diff
@@ -0,0 +1,432 @@
...
+		if(previous.z != z)
+			previous.z_transitioning = TRUE
+		else
+			previous.z_transitioning = FALSE
+		previous.forceMove(old_loc)	// None of this 'ripped in half by an airlock' business.
+	update_icon()
+
+/mob/living/simple_mob/animal/space/bluespace_worm/head/Bump(atom/obstacle)
+	if(open_maw && !stat && obstacle != previous)
+		spawn(1)
```
SplinterGP, line 223: what's this spawn() for even ? spawn() should rarely be used because mostly you don't need to in 99% of cases

SplinterGP (changes requested):
helping ater a bit

code/modules/mob/living/simple_mob/subtypes/animal/space/bluespaceworm.dm
Atermonera: In this case, using `spawn()` creates a delay between the worm Bumping something/one and it actually consuming them. This could lead to circumstances where the bumped thing moves away from the worm, breaking any adjacency requirements. That is to say, the spawn here is bad.

Atermonera:
If there aren't any significant mechanical differences between this and the original space worm, you're just changing what it drops, then this should inherit all identical behaviour _from_ the original worm (Allowing most of this code to not need to be copypasted), and then modifying/adding variables to the base type definition that describe what its original drops were, so you can override that definition on the subtype with what the new drops are.

Atermonera:
@Schnayy reports this'll need to more or less start over. Closing because stale

PR #7973, opened 2021-03-10
News Update  + Communicator Lag Fix

### LINKED ISSUE #7942: New news channel is too chonk for JS-based UIs
Opening my communicator UI makes my client fall over and die. Testing showed that other NanoUI windows that remained, like guest pass terminals and web shuttle controls, did not fall over, thus hinting that the issue probably wasn't NanoUI.

Two `git bisect`ions later and it said that `d0b71c8c527409261a511a9f6c4d97e1530fcf21` was the first bad commit.

#7856 introduced a fourth newscaster channel which contains iirc every single lore news thing that's been written, and based on the paths for each article, seems to be over a hundred articles. The NanoUI and the TGUI news viewers seem to choke on having to do string operations on over a hundred articles (every update?).

To be able to open my communicator.

#### What actually happened
My client froze and I had to kill the process. Other people reported really bad FPS instead.

#### Steps to reproduce
Join as any role.
Look at an ordinary newscaster on the wall, which uses plain HTML (not even with css!)
Observe the `Vir News Network` channel and note how huge it is. The scrollbar is so tiny.
Tiny scrollbar
Close that window and now go find a communicator. Click it inhand and watch your client slow or stop entirely instantly when used.
Recover from that, and then this time open your TGUI-based PDA. The PDA is better behaved since it doesn't seem to actually look at news articles until you ask it to, unlike communicators, which seem to load everything first. Open the news app and your PDA window will likely begin chugging, as it only shows the latest article. Open Vir News Network specifically and it will die alongside the communicator.
TGUI
Recover again, and if you're feeling up to it, open the admin newscaster window with the verb, or just comment out the articles and recompile.
Redact everything
Try the communicator and PDA again and note how it doesn't die.
NanoUI lives
If you _really_ wanna be sure, checkout a commit prior to d0b71c8c527409261a511a9f6c4d97e1530fcf21.

#### Additional info:

Cerebulon:
The communicator no longer attempts to load the entire 'lore news' channel contents which lags it to shit and back because it's actually got formatting n shit. Just outright excluded it since the communicator ALREADY has a 'most recent' news section and it seemed weird to double yknow. Should fix #7942

### DIFF
--- a/code/game/objects/items/devices/communicator/helper.dm
+++ b/code/game/objects/items/devices/communicator/helper.dm
@@ -41,22 +41,23 @@
 	for(var/datum/feed_channel/channel in news_network.network_channels)
 		var/list/messages = list()
 		if(!channel.censored)
-			var/index = 0
-			for(var/datum/feed_message/FM in channel.messages)
-				index++
-				if(FM.img)
-					usr << browse_rsc(FM.img, "pda_news_tmp_photo_[feeds["channel"]]_[index].png")
-				// News stories are HTML-stripped but require newline replacement to be properly displayed in NanoUI
-				var/body = replacetext(FM.body, "\n", "<br>")
-				messages[++messages.len] = list(
-						"author" = FM.author,
-						"body" = body,
-						"message_type" = FM.message_type,
-						"time_stamp" = FM.time_stamp,
-						"has_image" = (FM.img != null),
-						"caption" = FM.caption,
-						"index" = index
-					)
+			if(channel.channel_name != "Vir News Network") //Do not load the 'IC news' channel as it is simply too long.
+				var/index = 0
+				for(var/datum/feed_message/FM in channel.messages)
+					index++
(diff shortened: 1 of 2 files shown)


code/game/objects/items/devices/communicator/helper.dm
```diff
@@ -41,22 +41,23 @@
...
-				messages[++messages.len] = list(
-						"author" = FM.author,
-						"body" = body,
-						"message_type" = FM.message_type,
-						"time_stamp" = FM.time_stamp,
-						"has_image" = (FM.img != null),
-						"caption" = FM.caption,
-						"index" = index
-					)
+			if(channel.channel_name != "Vir News Network") //Do not load the 'IC news' channel as it is simply too long.
```
Atermonera, line 44: This if can be safely combined with the previous if to require only the existing level of indenation

PR #7976, opened 2021-03-12
Fixes Global and Fixes Eyes

### LINKED ISSUE #7850: Heterochomia body marking is broken
The heterochromia marking no longer shows up on characters, possibly eyes now render over body markings?
for the marking to show
#### What actually happened
it's likely hidden behind the eyes now, so does nothing visually
#### Steps to reproduce
select heterochomia body marking in character setup
#### Additional info:

Novacat:
Basically changing it to a GLOB just had it add all the species into one global variable instead of a proper list.

Also fixes eyes.

Also removes a bunch of //VOREStation edits that should not be there. Fixes #7850

### DIFF
--- a/code/__defines/sound.dm
+++ b/code/__defines/sound.dm
@@ -111,7 +111,7 @@
 	'sound/ambience/generic/generic2.ogg',\
 	'sound/ambience/generic/generic3.ogg'\
 	)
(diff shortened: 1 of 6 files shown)


Atermonera:
Aha, a compile warning. @Novacat 
`code/modules/organs/subtypes/standard.dm:336:warning: eyecon: variable defined but not used`
Sadly wasn't caught until we updated this onto our box

PR #7982, opened 2021-03-13
Adds random wilderness encounter event.

FenodyreeAv:
Spawns 1-3 mobs, native to the Sif wilderness, and sends them to a player.

Does not trigger in the plains, caves or on someone who is inside.

https://streamable.com/ieex5c

Doesn't spawn in the view range of any players.
https://streamable.com/plda9m
Tested!

🆑 Fenodyree
rscadd - Adds adds random encounters with Sif mobs in the wilderness.
/🆑


maps/southern_cross/events/wildlife_encounter.dm
```diff
@@ -0,0 +1,94 @@
...
+		if(L.stat)
+			continue // Don't want dead people.
+		if(istype(get_turf(L), /turf/simulated/floor/outdoors) && istype(get_area(L),/area/surface/outside/wilderness))
+			potential_victims += L
+			log_debug("Found victim")
+
+	if(potential_victims.len)
+		victim = pick(potential_victims)
+
+/datum/event2/event/wildlife_encounter/start()
```
MistakeNot4892, line 35: You need to null `victim` at the end of this to avoid GC failures.

maps/southern_cross/events/wildlife_encounter.dm
```diff
@@ -0,0 +1,94 @@
...
+	if(!explorers)
+		return 0
+	return 20 + explorers * 50
+
+/datum/event2/event/wildlife_encounter
+	var/mob/living/victim = null
+
+/datum/event2/event/wildlife_encounter/set_up()
+	var/list/potential_victims = list()
+	for(var/mob/living/L in player_list)
```
MistakeNot4892, line 21: I would suggest using `for(var/client/C)` and checking `C.mob` instead of `player_list`, otherwise you could be spawning critters and sending them to kill someone who is SSD.

MistakeNot4892: Just noticed the `stat` check, still not sure if `player_list` is the best approach.

maps/southern_cross/events/wildlife_encounter.dm
```diff
@@ -0,0 +1,94 @@
...
+		//if(!(L.z in get_location_z_levels()))
+		//	log_debug("Not on the right z-level")
+		//	continue // Not on the right z-level.
+
+		if(L.stat)
+			continue // Don't want dead people.
+		if(istype(get_turf(L), /turf/simulated/floor/outdoors) && istype(get_area(L),/area/surface/outside/wilderness))
+			potential_victims += L
+			log_debug("Found victim")
```
Atermonera, line 30: Not sure I, as a dev who often has debug logs turned on, need to see 7 lines of this output to chat when this event fires because there happens to be as many people on the planet. Also not sure how useful this piece of information would be in server logs after the fact.
Other debug logs in here are probably fine in that regard, but you should probably prefix them with something to indicate what is causing the log

FenodyreeAv: The GM outputs what event just fired in the line above.
Should show this now.

DEBUG: Event /datum/event2/event/wildlife_encounter is now being ran. 
DEBUG: Sending 3 /mob/living/simple_mob/animal/sif/shantaks after Michaela Karkee. 
DEBUG: Event /datum/event2/event/wildlife_encounter has finished.

maps/southern_cross/events/wildlife_encounter.dm
FenodyreeAv: This is how it's done in the carp event that this one is based off of. I tried looking for an example of how C.mob works but github search had other ideas.

maps/southern_cross/events/wildlife_encounter.dm
FenodyreeAv: Nulled it, I don't know how to use a weak reference in DM.

Is it similar to a weak pointer in C++?
/weakref/proc/resolve() returns the result locate(ref), so I'd call resolve() on my weak reference when I need to access it?

If there's documentation on how to use these, I can't find it. ):

maps/southern_cross/events/wildlife_encounter.dm
FenodyreeAv: Something like this? 
```
	for(var/client/C in GLOB.clients)
		if(!C.mob)
			if(istype(get_turf(C.mob), /turf/simulated/floor/outdoors) && istype(get_area(C.mob),/area/surface/outside/wilderness))
				potential_victims += C.mob
```

Why is this more desirable than getting the mob from the player list? Because the stat check is particularly expensive?

maps/southern_cross/events/wildlife_encounter.dm
MistakeNot4892: Yep, basically all you need to know is:
````
var/weakref/foo = weakref(someatom)
````
````
var/atom/someatom = foo.resolve()
````
The main caveats are that you check victim is the expected type (and hasn't been qdeleted) since references can be reused.

Comment (MistakeNot4892) [reply maps/southern_cross/events/wildlife_encounter.dm:36]: Also, you never delete weakrefs, you just null them and let them GC.

File: maps/southern_cross/events/wildlife_encounter.dm (hunk 2 above)
Comment (MistakeNot4892) [reply maps/southern_cross/events/wildlife_encounter.dm:22]: Having checked Polariscode it looks like player_list is fine, main reason I was suggesting clients was avoiding logged-out mobs resulting from client getting unexpectedly deleted. If player_list is consistent then yeah you can ignore my comment.

Just for reference `for(var/client/C)` (no `in foo`) is some BYOND sorcery that the compiler optimizes.

File: maps/southern_cross/events/wildlife_encounter.dm [hunk 4]
```diff
@@ -0,0 +1,95 @@
...
+		abort()
+		return
+
+	var/number_of_packs = rand(1, 3)
+	// Getting off screen tiles is kind of tricky due to potential edge cases that could arise.
+	// The method we're gonna do is make a big square around the victim, then
+	// subtract a smaller square in the middle for the default vision range.
+
+	var/list/outer_square = get_safe_square(victim, world.view + 5)
+	var/list/inner_square = get_safe_square(victim, world.view + 1)
```
Atermonera, line 48: Given that this doesn't consider the positions of any other players, it's _possible_ that this will spawn mobs within the view of another player. However, I'm not sure that trying to consider the views of other players is necessarily feasible, since you could have a large number of players to consider (Even if you track those players that could interfere from the victim selection), which could altogether disqualify a particular victim because nowhere in the predefined donut is not visible by _any_ player.

So I'm mostly just noting that this behaviour may be observed by players, and (especially without any announcement), it could be considered buggy by such players, particularly if/when a hostile mob like a phoron spider happens, through random chance, to spawn right next to someone.

maps/southern_cross/events/wildlife_encounter.dm
```diff
@@ -0,0 +1,95 @@
...
+			donut -= T
+
+	var/build_path = item_to_spawn()
+	var/turf/spawning_turf = pick(donut)
+	log_debug("Sending [number_of_packs] [build_path]\s after \the [victim].")
+	for(var/i = 1 to number_of_packs)
+
+		if(spawning_turf)
+			var/mob/living/simple_mob/M = new build_path(spawning_turf)
+			M.ai_holder?.give_destination(get_turf(victim))
```
Atermonera, line 62: What happens if there's a suitable target closer to the mob than the selected victim (general AI pathing question)? Will a hungry spider walk _away_ from a potential meal in search of a more distant one? What if the closer meal starts shooting at the spider? (Last question is probably answered by "it'll turn around and bite back", but I would like you to verify that)

Atermonera (commented):
Nothing that strictly requires coding changes, but some further testing for specific edge cases would be appreciated.

maps/southern_cross/events/wildlife_encounter.dm
FenodyreeAv: give.destination is the same thing that the move command with build mode does. So the spider will stop moving and attack as soon as it sees a target.

Someone shooting it from off-screen won't cause it to change movement though, but if you're on screen it'll try and eat you even if you don't shoot it.

maps/southern_cross/events/wildlife_encounter.dm
FenodyreeAv: It now checks if anyone else on the victim's list is within 7 tiles of the spawn turf, so it shouldn't be possible to see unless you're using binoculars, and looking in exactly the direction it decides to spawn.

Atermonera (approved):

PR #7993, opened 2021-03-21
Removed inactive platforms from the end of round laws output, gives them pilot access.

MistakeNot4892:


MistakeNot4892:
Need to make sure AI can mess with doors even if borgs are access-checked properly.

MistakeNot4892:
Will do that in another PR.

PR #7994, opened 2021-03-21
Fixes manifest datacore failing to find job datums for alt titles

Meghan-Rossi:
I don't think this actually affects anything on Polaris, but it's causing bugs downstream


Novacat:
It mostly has to do with off-station job roles which Polaris does not have yet

PR #7997, opened 2021-03-22
Fixes naked ghosts

Verkister:
Fixes ghosts going all naked as soon as they summon a typing bubble or anything.


MistakeNot4892:
Whoops. Did not observe this happening during testing but that's definitely an oversight on my part, apologies.

PR #8006, opened 2021-03-23
Prevented Teshari from selecting positronic brains.

MistakeNot4892:


Cerebulon:
Fine with relaxing rule on rules, but this is still useful for enforcing the positronic brain thing, thanks.

PR #8013, opened 2021-03-26
Oops, I refactored Chemistry!

Atermonera:
Contains commits from #8007 because I started doing this while working on that project and realized I should make a separate PR for this independent of the rest of the stuff I'm doing for that project.
This signficantly reorganizes the directory structure of the reagents module folder, it's now much more in line with modules like projectiles in structure. Different machines are in their own files, but reactions and reagents are still largely grouped into categories.
SSChemistry was switched from a Processing subsystem (Like SSObj) to a NO_FIRE subsystem (Like jobs). `/datum/reagents` would STOP_PROCESSING() on it when destroyed, but they never START_PROCESSING(), so it never actually processed anything. 
Chemical reactions were refactored to /decl, they are only ever process due to some distinct action occurring to trigger the proc, and they are only ever referenced from SSchemistry, which acts as a cache for all of the singleton instances. I almost did the same for reagent datums, but that would have quickly ballooned into a very large undertaking.

To update existing chemical reactions, you'll want to do the following find::replace operations, in order
`datum/chemical_reaction/distilling`::`decl/chemical_reaction/distilling` // Convert distillations first
`datum/chemical_reaction`::`decl/chemical_reaction/instant` // Then convert all others to subtypes of /instant, which will maintain existing operation.

SSchemistry will sort unrecognized subtypes as though they are /instant, but relying on that is bad practice.

Distillery and regular reactions were tested, and still work appropriately.


code/controllers/subsystems/chemistry.dm
```diff
@@ -0,0 +1,59 @@
...
+// For example:
+// chemical_reactions_by_reagent["phoron"] is a list of all reactions relating to phoron
+// Note that entries in the list are NOT duplicated. So if a reaction pertains to
+// more than one chemical it will still only appear in only one of the sublists.
+/datum/controller/subsystem/chemistry/proc/initialize_chemical_reactions()
+	var/list/paths = decls_repository.decls_of_subtype(/decl/chemical_reaction)
+	chemical_reactions.len = paths.len // Preallocate space
+
+	for(var/path in paths)
+		var/decl/chemical_reaction/D = new path
```
MistakeNot4892, line 37: `paths[path]` instead of `new path`.

code/controllers/subsystems/chemistry.dm
```diff
@@ -0,0 +1,59 @@
...
+
+//Chemical Reactions - Initialises all /decl/chemical_reaction into a list
+// It is filtered into multiple lists within a list.
+// For example:
+// chemical_reactions_by_reagent["phoron"] is a list of all reactions relating to phoron
+// Note that entries in the list are NOT duplicated. So if a reaction pertains to
+// more than one chemical it will still only appear in only one of the sublists.
+/datum/controller/subsystem/chemistry/proc/initialize_chemical_reactions()
+	var/list/paths = decls_repository.decls_of_subtype(/decl/chemical_reaction)
+	chemical_reactions.len = paths.len // Preallocate space
```
MistakeNot4892, line 34: As discussed on Discord, preallocating list size inserts a bunch of `null` entries, since you're doing `chemical_reactions += D` later you're just ending up with a list with a billion `null`s before the reaction instances.

code/controllers/subsystems/chemistry.dm
```diff
@@ -0,0 +1,59 @@
...
+			LAZYINITLIST(add_to[reagent_id])
+			add_to[reagent_id] += D
+
+//Chemical Reagents - Initialises all /datum/reagent into a list indexed by reagent id
+/datum/controller/subsystem/chemistry/proc/initialize_chemical_reagents()
+	var/paths = subtypesof(/datum/reagent)
+	chemical_reagents = list()
+	for(var/path in paths)
+		var/datum/reagent/D = new path()
+		if(!D.name)
```
MistakeNot4892, line 57: Could potentially do:
````
```
	for(var/path in paths)
		var/datum/reagent/D = new path()
		if(!D.name)
			continue
		chemical_reagents[D.id] = D
```
to
```
	for(var/path in paths)
		var/datum/reagent/D = path
		if(!initial(D.name))
			continue
		chemical_reagents[initial(D.id)] = new D
```
to save some creation overhead, but possibly not enough null-name reagents to bother.

File: code/modules/power/fusion/core/core_control.dm [hunk 4]
```diff
@@ -37,182 +26,83 @@
...
+	var/list/cores = list()
+
+	for(var/obj/machinery/power/fusion_core/C in GLOB.fusion_cores)
+		if(C.id_tag == id_tag&& (get_dist(get_turf(C), get_turf(src)) <= scan_range))
+
+			var/list/reactants = list()
+
+			if(C.owned_field)
+				for(var/reagent in C.owned_field.dormant_reactant_quantities)
+					reactants.Add(list(list(
```
MistakeNot4892, line 87: You don't seem to actually do anything with `reactants`.

code/modules/power/fusion/fuel_assembly/fuel_compressor.dm
```diff
@@ -54,7 +54,7 @@
 			to_chat(user, "<span class='warning'>It would be pointless to make a fuel rod out of [mat.use_name].</span>")
 			return
 		if(M.get_amount() < 15)
```
MistakeNot4892, line 56: Would probably be better to make this a constant or define somewhere while you're tweaking it.

code/modules/power/fusion/fusion_reagents.dm
```diff
@@ -1,18 +1,10 @@
...
 	desc = "A fueltank."
 	icon = 'icons/obj/objects.dmi'
 	icon_state = "weldtank"
 	amount_per_transfer_from_this = 10
-	New()
-		..()
-		reagents.add_reagent("helium-3",1000)
\ No newline at end of file
+
+/obj/structure/reagent_dispensers/he3/New()
```
MistakeNot4892, line 8: I know you're just reindenting it, but while you're here it might be worth switching New() out for Initialize(). Can't recall specifically what required it for the Neb reagent refactors but I vaguely feel like it'll save you tons of time down the track if you ever do anything with reagents that depends on atom init order.

code/modules/reagents/machinery/distillery.dm
```diff
@@ -53,31 +53,17 @@
...
-/obj/item/weapon/reagent_containers/glass/distilling/Destroy()
-	Master = null
-	..()
-
 /obj/machinery/portable_atmospherics/powered/reagent_distillery/Initialize()
 	. = ..()
 
-	Reservoir = new (src)
-	Reservoir.Master = src
+	reagents = new /datum/reagents/distilling(600, src)
```
MistakeNot4892, line 65: Personally I would suggest allowing `create_reagents` to take a type argument for the new reagent instance rather than doing a naked `new`.

code/modules/reagents/reactions/distilling/distilling.dm
```diff
@@ -26,32 +26,19 @@
 	var/list/temp_range = list(T0C, T20C)
 	var/temp_shift = 0 // How much the temperature changes when the reaction occurs.
 
-/datum/chemical_reaction/distilling/can_happen(var/datum/reagents/holder)
-	//check that all the required reagents are present
-	if(!holder.has_all_reagents(required_reagents))
-		return 0
+/decl/chemical_reaction/distilling/can_happen(var/datum/reagents/holder)
+	if(!(istype(holder, /datum/reagents/distilling) && istype(holder.my_atom, /obj/machinery/portable_atmospherics/powered/reagent_distillery)))
```
MistakeNot4892, line 30: Purely a nitpick, but `(!(foo && bar))` always seems less readable than `(!foo || !bar)` to me.

MistakeNot4892:
Nothing else springs out as problematic with this PR other than a few TGUI files from the RUST stuff you still have committed.

PR #8014, opened 2021-03-26
rutile is an alloying material

Hatterhat:
if it's not alloy = 1 then it makes slag when you try to alloy with it. this is kind of dumb but here have a webedit one line fix for plastitanium or w/e

### DIFF
--- a/code/modules/mining/ore_datum.dm
+++ b/code/modules/mining/ore_datum.dm
@@ -224,6 +224,7 @@ var/global/list/ore_data = list()
 	smelts_to = "titanium"
 	result_amount = 5
 	spread_chance = 12
+        alloy = 1
 	ore = /obj/item/weapon/ore/rutile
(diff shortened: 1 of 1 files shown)


code/modules/mining/ore_datum.dm
```diff
@@ -224,6 +224,7 @@ var/global/list/ore_data = list()
 	smelts_to = "titanium"
 	result_amount = 5
 	spread_chance = 12
+        alloy = 1
```
Novacat, line 227: Indentation error.

Novacat (changes requested):

PrismaticGynoid:
Titanium... isn't something you can do alloys with over here?

Hatterhat:
well hey pal i saw a bug downstream someone told me "throw this upstream" and i'm just goin with the flow

PR #8018, opened 2021-03-29
Makes manifest preserve original position of Custom jobs

Heroman3003:
Title. Screw all modifications out of the general alt-titles throwing the position down to Miscellaneous.

### DIFF
--- a/code/datums/datacore.dm
+++ b/code/datums/datacore.dm
@@ -310,7 +310,6 @@ var/global/list/PDA_Manifest = list()
 			break
 		else
 			var/list/alttitles = get_alternate_titles(J.title)
-			if(!J)	continue
 			if(assignment in alttitles)
 				real_title = J.title
 				break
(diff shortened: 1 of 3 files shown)


code/datums/datacore.dm
```diff
@@ -302,11 +302,18 @@ var/global/list/PDA_Manifest = list()
...
-			real_title = J.title
+		if(J.title == rank)					//If we have a rank, just default to using that.
+			real_title = rank
 			break
+		else if(J.title == assignment)
+			real_title = rank
+			break
+		else
+			var/list/alttitles = get_alternate_titles(J.title)
+			if(!J)	continue
```
MistakeNot4892, line 313: Why do you check `J` after the point where a falsey value in `J` would have runtimed?

Heroman3003: Thats the part I did not modify. I have not a single clue why it was like that in the first place, but it wasn't my place to judge.

MistakeNot4892: You can safely delete it. `J` is implicitly typechecked by the for loop.

Heroman3003: Sure. I was wandering why it was like that when its basically a check for an always false thing, but I decided that I'm a bit too sleepy to try and delete extremely ambigous-looking things that were there before me.

PR #8031, opened 2021-04-10
Adds whitelist system for ear/tail genemods/augments

Atermonera:
Species-specific mods will continue to be available to anyone that can select that species (tesh ears, variant taj ears, etc).
The whitelist exists to allow characters access to a more expanded set of ears/tails, and restrict people just memeing the system. Specific choices subject to people saying "Hey that looks dumb".
Will discuss this further with staff, but I may restrict the extra options per-species, such that say, unathi, cannot have teddybear ears.
Details of the whitelist application to be discussed at a later date.

### DIFF
--- a/code/modules/mob/new_player/sprite_accessories_ear.dm
+++ b/code/modules/mob/new_player/sprite_accessories_ear.dm
@@ -30,7 +30,7 @@
 	icon_state = "ears_plain"
 	do_colouration = 1
 	color_blend_mode = ICON_MULTIPLY
-	species_allowed = list(SPECIES_TAJ, SPECIES_EVENT1, SPECIES_EVENT2, SPECIES_EVENT3)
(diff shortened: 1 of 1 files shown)


Atermonera:
Following discussion at staff meeting: The system will be more general (which I expected, I just saw this as the thing that specifically needed to be discussed), and will be locked behind a whitelist (Details to be discussed, possibly similar to a blend between species and language apps?). This system will not be merged/enabled _until_ the timeskip and setting changes take place.

Atermonera:
Update 2: in order to have a discussion that _isn't_ in outright bad faith about the concerns with adding stuff like this with the very concerned parties that are against it, I will be closing this, and will post results of such discussions on discord, and in any future PRs that I may make to add anything in the vein of this PR.

PR #8034, opened 2021-04-12
Random spawner refactor.

MistakeNot4892:
Allows mapped random object spawners to set `dir` on the result.


MistakeNot4892:
Unlikely to get back to this.

MistakeNot4892:
Had a brainfart and forgot I never pushed the horrible part of this PR.

PR #8052, opened 2021-04-26
Simplemobs seek lost targets

Atermonera:
[Tested (.mp4)](https://puu.sh/HBzqX/d5df557b84.mp4)

When simplemobs lose sight of their target, they'll drop the target (So if they see another, they'll switch). If the mob cannot shoot at the target's last position, it will advance towards it. If it is in melee range, it'll try to whack something in the tile the target was last seen, and if that fails, it'll consider possible escape routes. These routes include stairs, open non-glass doors, and if the mob is intelligent and has hands or is floating, ladders. It'll ignore objects that are behind it, which can make it stuck, but should prevent them from doubling back into a door they just walked out of. It'll pick randomly from the set of closest such objects to investigate.
They do get caught on railings, not really an easy fix for that for exactly the same reason I reworked stairs to avoid leaving-edge transitions.

I also tested carp, to make sure that they aren't any more eager to break windows into maintenance, and I didn't observe any new issues there.

There's a pretty good chance that this'll actually make spiders and suchlike _too_ lethal, in which case further adjustments will have to be considered, or even outright removal/limiting this to AI_SMART mobs (Sapphire slimes, humanoid mobs).

### DIFF
--- a/code/modules/ai/ai_holder_combat_unseen.dm
+++ b/code/modules/ai/ai_holder_combat_unseen.dm
@@ -7,27 +7,30 @@
 	if(conserve_ammo || !holder.ICheckRangedAttack(target_last_seen_turf))
 		if(get_dist(holder, target_last_seen_turf) > 1) // We last saw them over there.
 			// Go to where you last saw the enemy.
-			give_destination(target_last_seen_turf, 1, TRUE) // Sets stance as well
-		else if(lose_target_time == world.time) // We last saw them next to us, so do a blind attack on that tile.
-			melee_on_tile(target_last_seen_turf)
+			return give_destination(target_last_seen_turf, 1, TRUE) // Sets stance as well
+		else if(lose_target_time < world.time) // We last saw them next to us, so do a blind attack on that tile.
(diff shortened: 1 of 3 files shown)


code/modules/ai/ai_holder_combat_unseen.dm
```diff
@@ -36,9 +39,53 @@
...
+	var/list/closest_escape = list()
+	var/closest_dist = world.view // We can't see any further than this
+	var/list/possible_escape_types = list(
+		/obj/machinery/door,
+		/obj/structure/stairs/top,
+		/obj/structure/stairs/bottom
+	)
+	
+	if(intelligence_level >= AI_SMART)
+		possible_escape_types += /obj/structure/ladder
```
TheFurryFeline, line 63: 😨 In before they learn to use ladders....

Very-Soft:
THIS LOOKS AWESOME I CAN'T WAIT

Atermonera:
Aaaand it's been 3 weeks

PR #8056, opened 2021-04-27
weather sounds adjustments

Greenjoe12345:
Makes all the weather sounds much quieter, so blizzards aren't deafening anymore
also enables sounds for rain on sif that were previously commented out, along with a louder rain sound for storms.
as well as this, it adds a very quiet wind sound for the 'calm' weathers (clear, overcast, light snow), to give some ambiance to Sif during those times, a light breeze, you could call it.


Novacat:
Merge this immediately.

PR #8057, opened 2021-04-30
Body style selection and sanitizing refactor.

MistakeNot4892:
Cleans up sprite accessory selection and standardizes some generic proc helpers for it. Should fix current issues with overzealous sanitizing or incorrect list usage - style selection as admin and non-admin works on my machine:tm: at time of writing.


MistakeNot4892:
CI is failing to find BYOND, not sure what that has to do with my changes.

MistakeNot4892:
DNM please, need to verify if this is going to break older saves or not.

PR #8061, opened 2021-05-04
Commits Genocide against the /decl and /repository typepaths

ShadowLarkens:
There is no reason for these to exist.
   - They provide no advantages over defining /datum
   - They mask the fact that what you're actually defining is a /datum/decl/x
   - Almost every usage of them did not even take advantage of the one fucking thing that actually was implemented on the /decl typepath, the decl repository
   - It adds overhead by requiring you to access stuff through procs on the decl repository instead of accessing a GLOB list
   - The semantic meaning doesn't even make sense when 99% of our singletons use datums because they were written by people who aren't insane


MistakeNot4892:
I'm going to respond to this as I seem to be the designated spokesperson for the decl system on Polaris and I feel like there's quite a lot of vitriol being mixed in with objective reasoning here. I'm not saying you should like decls or that your viewpoint here is invalid, but I do think your criticism is somewhat skewed and that there's quite a lot of emotion involved.

`They provide no advantages over defining /datum`
They provide a centralized repository that allows for control over ordering and initialization, and in theory allows unit checking of singleton validity, all that business. The Polaris/Virgo decl repositories and decl implementations are woefully inconsistent/underdeveloped and make it hard to actually capitalize on the benefits, but they do exist.

`They mask the fact that what you're actually defining is a /datum/decl/x`
By the same argument, `/obj` is masking `/datum/atom/movable/obj`. I do think /decl is an uninformative name but /singleton bloats paths out significantly, so heigh ho.

`Almost every usage of them did not even take advantage of the one fucking thing that actually was implemented on the /decl typepath, the decl repository`
I don't follow this logic. The primary benefit of the repository is that you're getting a managed instance and don't need to worry about explicitly checking that it was only created once. If you're talking about Initialize(), refer to my point below about the repository being undeveloped and other new/init inconsistencies being more immediate problems.

`It adds overhead by requiring you to access stuff through procs on the decl repository instead of accessing a GLOB list`
This can be mitigated pretty easily (for reference, the `GET_DECL` macro that you deleted reduced overhead from extremely hot code in Nebula's gas/materials processing logic to damn near 1:1 with pre-decl code), and the overhead from a single proc call is rarely significant. Accessing `GLOB` is itself not free, as you're referencing an instance variable and then doing a list lookup regardless, so it comes out more or less the same in the end.

`The semantic meaning doesn't even make sense when 99% of our singletons use datums because they were written by people who aren't insane`
There's really no need to be rude. You don't like it, I get that, but not liking something isn't carte blanche to call people who disagree with you names.

Ultimately killing off the system is a bit w/e because it's so poorly integrated on these codebases that it hardly matters, and there are other far more severe init and object creation problems already. However removing the repository in favour of runtime-built global lists is adding more potentially brittle, order-dependent code spread across multiple points of instantiation, and additionally is a huge pain to unit test or otherwise validate in the future, as you will need to keep track of multiple lists of multiple different object types. 

If you wanted to fix/improve the system rather than shift-delete it for a moral victory I would suggest killing the /decl subtype but letting the decl_repository remain as a singleton_repository or such that will accept arbitrary datum types and return a managed instance to the proc.

If you're at all interested in what can be done to properly integrate the decl repository and implement validation, you could reference:
https://github.com/NebulaSS13/Nebula/pull/64
https://github.com/NebulaSS13/Nebula/pull/1576
https://github.com/NebulaSS13/Nebula/pull/1575
https://github.com/NebulaSS13/Nebula/pull/1577
https://github.com/NebulaSS13/Nebula/pull/1578

There's a bunch more Initialize/New changes in the Neb history but I can't find specific PRs. The point comes across regardless.

Poojawa:
centralizing anything in Byond code under the guise of improvement is just a resurgence of the PoolorNew craze that everyone thought was the secret to fixing effects lag. GLOB already has round start lists *if* it's populated. That's your central repo. GLOB is also documented and has had proven metric keeping done showing how, in conjunction to Initialize() and async stuff from the Lazy lists, it resulted in tremendous performance gains. 

TG's signals are more sophisticated, and have better documentation as to what they do, why you should use them, and provable evidence they do actually help overhead. 

While it's tempting to reinvent the wheel, there's honestly no reason to when better tried and tested systems exist.

MistakeNot4892:
`centralizing anything in Byond code under the guise of improvement is just a resurgence of the PoolorNew craze that everyone thought was the secret to fixing effects lag.`
At no point have I argued decls are relevant to reducing lag. All the repository does is provide a managed singleton, it has no bearing on pooling or lag.

`GLOB already has round start lists if it's populated. That's your central repo.`
`GLOB` does not provide any kind of management other than VVability and some var protection, and is completely redundant on 512 or above due to having access to `global`. My comment regarding order-dependant/manual instantiation refers to the fact you have to explicitly populate global lists with `init_subtypes` or such, while the decl repository will create, store and return the singletons you ask for regardless of when or where you call the proc (assuming it's post-`world/New`. `GLOB` is _objectively less performant_ than a naked globally scoped list and _nearly exactly as performant_ as `GET_DECL`, ignoring the miniscule overhead from the `ispath()` check.

`GLOB is also documented and has had proven metric keeping done showing how, in conjunction to Initialize() and async stuff from the Lazy lists, it resulted in tremendous performance gains.`
I'm not sure I follow how this relates, particularly async stuff and lazylists, which are completely unrelated to the decls repository or type. Are you referring to the performance gains from deferring instance creation into a managed system via SSatoms and Initialize? Correct usage of the decl repository (ie. not using bare-assed `new` to create them like /decl/observ does on Polariscode) falls into line with the same paradigm.

`TG's signals are more sophisticated, and have better documentation as to what they do, why you should use them, and provable evidence they do actually help overhead.`
I really don't get this. How do signals relate to a singleton repository?

`While it's tempting to reinvent the wheel, there's honestly no reason to when better tried and tested systems exist.`
If you were using a better tried and tested system I wouldn't have felt obligated to comment. Lists on `GLOB` populated in subsystem/MC procs are not 1:1 with the decl repository.

`also for real why is it called decl`
According to @mloc it's short for declaration.

Poojawa:
if you have some testing examples of how your system is as good/superior to datums and Globals, I would be interested in seeing it.

silicons:
@ShadowLarkens you are a hero

MistakeNot4892:
`if you have some testing examples of how your system is as good/superior to datums and Globals, I would be interested in seeing it.`
I'm not sure how else I can express that the decl repository provides specific functionality that `GLOB` and global lists don't. You're asking me to provide a testing metric for two completely unrelated things.

I've also been informed that you personally don't want to listen to anything I have to say, so I think I will call it here as I've expressed my main thoughts on the PR.

I won't respond to @silicons here since we've been chatting on Discord anyway and it's 99% unrelated.

Andromeda-K22:
I'll throw my hat into the ring here for a little bit.

I'm going to be honest it seems like both people in this comment thread who support the removal are completely out of their depth. The global singleton repository is stupid useful for not having to keep track of a thousand different datum instances of the same thing, the overhead from get_decl is pretty minimal as well. This just strikes me as a bizarre, bad-faith PR because someone can't grasp the idea behind an underutilized system.

The response is even more bewildering when MistakeNot offered actual, useful feedback to make the system better. 
A global singleton repo has nothing to do with signals/events. Where did you even *get* that notion from?
GLOB's clunky and as stated, *redundant* now that you can utilize actual globals in the same manner GLOB was intended for. afaik GLOB was a work-around for globals not working that way ... so why embrace the use of a system that was ostensibly a work-around???

0/10 all around confusing and bizzare PR.

MistakeNot4892:
Regarding `GLOB`:
image

The decl repository and other global instances have identical overhead, but there seemed to be some confusion about what I was saying about GLOB's performance, so I benchmarked it.

mloc:
/decl was notation for types that would never be used as datums, which were only used for their compile-time structure - no continuous runtime logic or mutation, they're only instantiated to query their contents.

This existed before I started calling things /decl, I just wanted to have some consistent name for this type-tree-as-DSL stuff. The decl repository is a natural progression to ease querying these.


I haven't even read the code here, I'm just giving some context. I also don't know if all of the above applies anymore, I've been OOTL for too long.

Atermonera:
Alright. After discussing this with both sides to get as many takes as possible, I will not be merging this. I see value in having a shared infrastructure to identify and manage something as a singleton, and GLOB is now redundant with native Byond globals as of 512 adding the `global.var` accessor. The only other function that GLOB specifically provides is GLOBAL_PROTECTED, which can be implemented quite easily by adding the particular var to the list that's hidden by the `VV_hidden()` list (or modifying the #defines of GLOB to handle such changes automatically.

Use cases that handle /decl's incorrectly should be corrected. The underlying mechanism isn't faulty, just underutilized because nobody knew what /decl's were supposed to be used for.
The name can and probably should be changed, but that's a very simple fix, and the only hard part is finding some particular term that more intuitively encapsulates the behaviour.

As for overhead, as other people have described, the computational overhead is comparable to GLOB. VV may become more tedious, but I think that's something that can be accommodated through other means. The coding overhead to convert something from a /decl to a /datum is trivial, and converting something from a global /datum to a /decl shouldn't be too much harder (Just change the type, undeclare the var and see what breaks). I do think that using /decl's more proactively could actually improve global initialization times, by allowing certain things (Like the plant controller) to be initialized only when it is actually needed, instead of with everything else.

As for claims that this is in bad faith, I think it's more likely a case of separate factors occurring in conjunction: A) the author's opinion on zuhayr, who as the person that's been explaining to everyone what /decl's are supposed to be used for has in a sense become the spokesperson for them, B) the author's opinions on the usefulness (or lack thereof) of this infrastructure, and C) the author's personal habits with regards to phrasing for PRs in this vein, that remove unwanted features. Talking with the author, I suspect the three factors to be separate in nature. However, in light of these three factors occurring in conjunction, I can see how people would interpret the various pieces of hostility together as ultimately a bad-faith change targeted at particular people.

If I were to merge this PR, then it would potentially be more difficult to port features from servers like Bay and Nebula, where /decl's are much more commonplace. I am actually interested in porting features from those servers, even if only in overall concepts rather than specific application (Due to missing other infrastructure) The author has voiced to me that they don't see any value in the sort of infrastructure provided by the /decl and /repository structures, so getting a replacement for them that is actually properly used and documented is unlikely.

If I were to reject this PR, then it would increase the discrepancy between our code and Virgo's in a time when I'm actively rebasing our code onto theirs, but my current approach to reviewing the rebase shouldn't actually add any overhead to the process. It would also add overhead to the process of receiving updates from them in the future. However, I anticipate such updates already being a significantly more involved process than has been the case for them receiving updates from us because of their dev/release production cycle. That cycle permits them to commit incomplete/nonfunctional code to their dev branch, and that it only needs to be properly tested and fixed when a release is published, which means we can only safely pull from them when such releases are made. Those releases will inevitably be a group of unrelated changes, which will make reviewing them more difficult.

I think it also worth mentioning that there are likely to be use cases where we might benefit from storing something on a /decl, when Virgo necessarily cannot. One such example is species datums: we currently maintain a single global list of species datums, and all mobs have a reference to a datum in that list. Virgo has custom species and traits that necessitate each mob have a unique species datum, with extra code to produce useful copies on mob instantiation. Since we don't have plans to use that system, we can safely ignore that code and instead treat them as singletons.

Overall, I think we can continue to appreciate the benefits of using /decl's, and the downsides are generally things that can be corrected rather than issues intrinsic to the system. Particularly with the looming rebase to Virgo, this decision really need only affect us, and it shouldn't really introduce much more overhead to the process of receiving updates from Virgo, even if they elect to remove /decl's on their end (Which I highly suspect they will after this is closed).

ike709:
Polaris moment

PR #8065, opened 2021-05-05
Standardizes Boxes

Novacat:
This standardizes all boxes to 7 slots, instead of the current system where we have a ton of 4-slot boxes, and a ton of 7-slot boxes.


code/__defines/inventory_sizes.dm
```diff
@@ -22,4 +22,4 @@
 // Container sizes.  Note that different containers can hold a maximum ITEMSIZE.
 #define INVENTORY_STANDARD_SPACE	ITEMSIZE_COST_NORMAL * 7 // 28
 #define INVENTORY_DUFFLEBAG_SPACE	ITEMSIZE_COST_NORMAL * 9 // 36
-#define INVENTORY_BOX_SPACE			ITEMSIZE_COST_SMALL * 4 // 8
\ No newline at end of file
+#define INVENTORY_BOX_SPACE			ITEMSIZE_COST_SMALL * 7 // 8
```
Atermonera, line 25: Comment should be updated to reflect the new value as well

PR #8071, opened 2021-05-10
Ports Nebula's Discord Webhook Integration

Neerti:
Ports the ability for the game to utilize webhooks for various purposes, from Nebula. Currently, this PR adds them for the server starting up, round starting, round ending, adminhelp actions (opening, closing, handling, resolving, etc), faxes (both sending and replying), and custom events. Other purposes can easily be added.

I intend to write a github wiki page with setup instructions for the webhook integration, since it depends on an external library called [byhttp](https://github.com/Lohikar/byhttp) and will need to be built from source for servers wanting to use it. For this PR, have a [copypasted readme](https://github.com/NebulaSS13/Nebula/blob/dev/docs/installation.md#webhooks) from Nebula.
```
WEBHOOKS

If you wish to use Discord webhooks, which are a way of passing information from the server to a Discord channel, you will need to copy webhooks.json into config/ from config/example/ and add definitions pointing the desired event at the desired Discord webhook URL. Valid webhook IDs as of time of writing are as follows:

    webhook_roundend: The round has ended. Will include the mode name and summarize survivors and ghosts.
    webhook_roundstart: The master controller has finished initializing and the round will begin soon.
    webhook_submap_loaded: A submap has been loaded and placed, and is available for people to join. Includes the name of the submap.
    webhook_custom_event: The custom event text for the round has been set or changed.

Each definition can optionally include an array of roles to mention when the webhook is called. Roles must be provided using the role ID (ex. <@&555231866735689749>), which can be obtained by writing \@somerole into the chat, in order for pinging to work correctly. Including the mentions as the value keyed to a URL will send the mentions only with that URL, while using the "mentions" field will attach the mentions to all URLs.

Webhooks additionally require a HTTP POST library called byhttp. The compiled lib, byhttp.dll on Windows or libbyhttp.so on Linux, must be placed in the lib directory by default in order for webhooks to function. The DLL location can be customized by supplying WINDOWS_HTTP_POST_DLL_LOCATION UNIX_HTTP_POST_DLL_LOCATION, or HTTP_POST_DLL_LOCATION as preprocessor macros containing the desired path.
```
N.B. `webhook_submap_loaded` doesn't exist in this port because it's not really something we need right now (I assume it's more like away missions). It also doesn't mention `webhook_[ahelp|fax]_sent`, but the wiki page should.


MistakeNot4892:
Submaps are basically PoI with joinable jobs that show up on the lobby screen, Polaris has no equivalent.

PR #8078, opened 2021-05-11
Refactors IsSynthetic

Novacat:
Port of https://github.com/VOREStation/VOREStation/pull/10298 by Aronai


code/modules/mob/living/carbon/human/human_powers.dm
```diff
@@ -218,7 +218,6 @@
 		if(isSynthetic())
 			output += "Current Battery Charge: [nutrition]\n"
 
-		if(isSynthetic())
```
TheFurryFeline, line 221: Pain. Why did this part need that if(isSynthetic()) twice? Spaghetti bruh moment.

PR #8080, opened 2021-05-12
Adds admin verb to hopefully fix database connection failures

Atermonera:
Will be tested live on the server if/when the issue happens again and someone remembers to test this.
Has lots of output to try and determine what actually goes wrong with it.

> `view-global-variables`
> > Select `dbcon`
> > > Call Proc `Disconnect`, no parameters
> 
> `Advanced-ProcCall`, proc is NOT owned by something, `/proc/establish_database_connection`, no parameters
> `SDQL2-query` "CALL log_client_to_db() ON /mob WHERE client MAP client"

### DIFF
--- a/code/modules/admin/verbs/dbcon_fix.dm
+++ b/code/modules/admin/verbs/dbcon_fix.dm
@@ -1,0 +1,39 @@
+// Will hoepfully fix the database when it breaks
+// Use case: if the server is left in the lobby for long enough,  players that join will see player_age = 0, restricting them from all age-locked jobs.
+/client/proc/dbcon_fix()
+	set name = "Fix Database Connection"
+	set category = "Server"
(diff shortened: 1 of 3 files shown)


code/modules/admin/verbs/dbcon_fix.dm
```diff
@@ -0,0 +1,39 @@
...
+	var/errno = dbcon.ErrorMsg()
+	if(errno)
+		admin_log("Database connection returned error message `[errno]`. Aborting.")
+		return FALSE
+	if(!dbcon.IsConnected())
+		admin_log("Database could not be reconnected! Aborting.")
+		return FALSE
+	admin_log("Database reconnected. Fixing player ages...")	
+
+	int num = 0
```
Neerti, line 28: Ater is a silly goose.

PR #8085, opened 2021-05-13
Linter Introduction + Cleanup

Neerti:
This adds the required defines to use the SpacemanDMM DreamChecker linter, which can be used inside of certain IDEs such as Visual Studio Code. It can also be integrated into CI, so that someday in the future we can have our CI point out spotty code.
Documentation for that can be seen here: https://github.com/SpaceManiac/SpacemanDMM/tree/master/src/dreamchecker

Unfortunately, Polaris has over 500 problems raised by the linter, and I assume Virgo has even more, so I figure I would do my part and try to reduce the number down so that someday we can become A Prim And Proper Codebase:tm: like Nebula. I've shoved linter error fixes into seperate commits so that it might be easier to review, or to be cherrypicked into a seperate PR if @Atermonera wants that. 

Incidental fixes as a result:
* This also fixes a bug with fax machines, where you cannot actually fax the `Supply` admin department due to the check if it was SifGov always evaluating to true.


code/game/objects/structures/crates_lockers/__closets.dm
```diff
@@ -434,10 +434,6 @@
 			breakout = 0
 			return
 
-			playsound(src, breakout_sound, 100, 1)
-			animate_shake()
-			add_fingerprint(escapee)
```
Novacat, line 439: Huh, I always wondered what broke the animation to break out of lockers. Could this code be fixed by just having the proper number of tabs?

code/modules/mob/living/living_defense.dm
```diff
@@ -481,7 +481,6 @@
 
 /mob/living/proc/reagent_permeability()
 	return 1
-	return round(FIRESUIT_MAX_HEAT_PROTECTION_TEMPERATURE*(fire_stacks/FIRE_MAX_FIRESUIT_STACKS)**2)
```
Novacat, line 484: This might be responsible for the bug where it is possible to apply reagents in certain ways through any amount of protection, including voidsuits?

Neerti: A quick look at the code suggests that human mobs use an overrided version of that proc, so if there is an issue, I don't think it's due to the `/mob/living` level proc.
https://github.com/PolarisSS13/Polaris/blob/9942d9cb25f0f9128880cf540646d8e833411e46/code/modules/mob/living/carbon/human/human_defense.dm#L561

Neerti:
Now Polaris should not have anymore linter issues, assuming another PR didn't add them in before this gets in.
Also I didn't make it look for stuff like relative pathing since that seems better for a seperate PR and technically is less likely to find actual bugs.

code/game/objects/structures/crates_lockers/__closets.dm
Neerti: Locker animation has been fixed now.

code/game/machinery/computer/syndicate_specops_shuttle.dm
```diff
@@ -1,257 +1,6 @@
...
-#define SYNDICATE_ELITE_DOCK_AREATYPE "/area/shuttle/syndicate_elite/mothership"	//Type of the spec ops shuttle area for dock
-
-var/syndicate_elite_shuttle_moving_to_station = 0
-var/syndicate_elite_shuttle_moving_to_mothership = 0
-var/syndicate_elite_shuttle_at_station = 0
-var/syndicate_elite_shuttle_can_send = 1
-var/syndicate_elite_shuttle_time = 0
-var/syndicate_elite_shuttle_timeleft = 0
-
 /obj/machinery/computer/syndicate_elite_shuttle
```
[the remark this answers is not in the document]
Neerti: The console is still mapped in on the Southern Cross, and I didn't want to edit the map and risk merge conflicts.

PR #8089, opened 2021-05-15
[Bugfix] Maybe Fix Cleanbots

TheFurryFeline:
Removes blood stuff, since it's already a cleanable, which is what a cleanbot is to clean, anything cleanable. Also effectively removes anything in the target_types list, replace with the base cleanable obj path so it cleans any children of cleanable. Tested, spawned stuff, it cleaned pie splatter, cobwebs, etc.

The way this was before, if there was still a cleanable that wasn't listed in target_types, the bot would get stuck forever without the un-included cleanable showing up in their target.

🆑 
bugfix - Cleanbots no longer get stuck trying to clean something not in its target list
/🆑 

Edit for clarification: only removed the targeting stuff for blood. The gib splattering effect from emagged cleanbot has not been removed with this PR.

### DIFF
--- a/code/modules/mob/living/bot/cleanbot.dm
+++ b/code/modules/mob/living/bot/cleanbot.dm
@@ -14,7 +14,6 @@
 	var/cleaning = 0
 	var/wet_floors = 0
 	var/spray_blood = 0
-	var/blood = 1
 	var/list/target_types = list()
 
 /mob/living/bot/cleanbot/New()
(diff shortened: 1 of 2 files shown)


TheFurryFeline:
I've been notified it's intentional to have the portion of being able to make a cleanbot ignore cleaning blood spills. I can restore this but I'd prefer to ensure that anything else is cleaned to negate any further incidences of being trapped in a loop of sitting underneath a cobweb in a corner and never being able to get it assigned to their target because target_type did not have it in the list.

PR #8094, opened 2021-05-18
phoron bore stock parts now affect range, along with other bits

### LINKED ISSUE #7779: Portable phoron bores can be reloaded forever with single bars of phoron
As in VOREStation/VOREStation#9407, a stack with one bar of phoron in it will never be consumed when reloading a portable phoron bore. 

For the phoron to be used up when loading in the last bar in the stack. 

#### What actually happened
The reload cycle repeats until the bore is filled up and you're still in possession of the solitary piece of phoron, allowing for infinite reloads. Since a phoron bore can also recharge its cell by burning its fuel, you can use even a stock bore to clear entire mining zones or kill anything you face with enough patience. 

edit: raw phoron crystals load correctly.

#### Steps to reproduce
1. Acquire a portable phoron bore and a stack of more than one phoron.
2. Load the bore normally, and observe how a single bar is used per 2000 loaded units. 
3. Split off a single piece of phoron from the stack and attempt to load that.
4. Witness how you can keep reloading a bore as many times as you need with that one bar of phoron.  

#### Additional info:
Server revision: Tested it on a private server, with the most recent commit: 
5d041b9dcc3b957bb74b8ecca52c0c775041eed4

Hatterhat:
phoron bores:
- examining them now shows if there's a manipulator loaded, and how much material it will consume per shot
(if there isn't, a "manipulator missing" message is shown instead)
- fired phorogenic blasts can now gain a **!!RANGE BUFF!!** depending on the rating of the stock parts (capacitor, manipulator) loaded in
- bores no longer display their ammo count twice
- closes #7779. by fixing it

phoron bore range buff mechanic:
- requires both a manipulator and a capacitor to be loaded in
- adds the rating of the capacitor AND manipulator to the range of the blast
(this can get rather Quirky, even with conventional T3 parts)

i'm open to changing the range buff mechanic to average the capacitor and manipulator ratings, rounding up, but until i'm told otherwise it's staying as is

### DIFF
--- a/code/modules/projectiles/guns/magnetic/bore.dm
+++ b/code/modules/projectiles/guns/magnetic/bore.dm
@@ -6,7 +6,14 @@
 	var/mat_cost = 500			// How much material is used per-shot?
 	var/ammo_material
 	var/obj/item/weapon/stock_parts/manipulator/manipulator    // Installed manipulator. Mostly for Phoron Bore, higher rating == less mats consumed upon firing. Set to a path to spawn with one of that type.
+	var/rating_modifier = 0 // rating of installed capacitor + manipulator
 	var/loading = FALSE
+
+/obj/item/weapon/gun/magnetic/matfed/proc/update_rating_mod()
+	if(capacitor && manipulator)
+		rating_modifier = capacitor.get_rating() + manipulator.get_rating()
+	else
+		rating_modifier = FALSE
 
 /obj/item/weapon/gun/magnetic/matfed/Initialize()
 	. = ..()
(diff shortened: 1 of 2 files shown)


Wrackbang:
Phoron bores have an infinite reload bug as described in PolarisSS13/Polaris#7779. May I ask for a fix for that in this PR?

Hatterhat:
k

Hatterhat:
any%

Wrackbang:
Thank you.

PR #8099, opened 2021-05-21
Tweaks to emote system to avoid memory overhead and make runtime validation easier.

MistakeNot4892:
- Mobs don't maintain lists of usable emotes on themselves, they look them up ad hoc and validate them at time of use.
- Deathgasp doesn't make you puke anymore.

### DIFF
--- a/code/modules/emotes/definitions/human.dm
+++ b/code/modules/emotes/definitions/human.dm
@@ -1,11 +1,5 @@
-/decl/emote/human
-	key = "vomit"
-
-/decl/emote/human/check_user(var/mob/living/carbon/human/user)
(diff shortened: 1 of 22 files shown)


MistakeNot4892:
This has been mergable for almost a week. Pls.

code/modules/emotes/definitions/human.dm
```diff
@@ -1,11 +1,5 @@
-/decl/emote/human
-	key = "vomit"
```
TheFurryFeline, line 2: So THAT'S why vomit was happening for deathgasp... Inheritance gay, man. lol

PR #8100, opened 2021-05-21
Implements A General Serializer

Neerti:
Putting this up early so people can :eyes:s it and say if something looks wrong.

## Rationale
This is intended to be a system that's up for the task to 'scale up' in serializing things ranging from simple `/datum` subtypes to more complex objects such as items and structures. Maybe someday it could scale up to include turfs, areas, and complex mobs like humans.

I imagine it could also be used to serialize non-objects as well, just without the bells and whistles that the object part has.

I actually wrote this for World Server, before World decided to call it quits, so now Polaris might get to have it instead. On World, there were many different implementations of persistence systems, generally one for each feature, with the most complex one being the map loading/saving system. I was writing a thing that would've needed another new system, so instead I decided to try to write a better generic system, use my new feature for it, and fold the other persistence features into my system one-by-one, to avoid https://xkcd.com/927/.

## Features
Oftentimes I see persistence systems utilize a list of variable names to save, like `list("name", "desc", "color")`, then when it's getting saved, the game just iterates that list and retrieves the var from `vars[]` directly. This has several issues.
- Using strings to address variables has many disadvantages. It's kinda ugly, but it's also prone to errors. For builtin vars like `name` it's probably safe, but subtypes will also define new vars that might need to get saved, and there's a chance that they could get renamed, and the person who does the rename might forget to update the list, causing it to silently fail.
- By reading from the vars list directly/writing to the var directly, this excludes the ability to selectively save a specific variable by using a getter, or load one using a setter.
- Certain variables, especially involving references, will require complex behavior that is difficult to accommodate with a basic string list system, for example, storing which material something is made out of by storing the material's typepath, then loading it by using it to index a global list to get the material reference back.

Instead, my system ~~ab~~uses macros, which has the following benefits.
- Addressing variables does not use strings, but the variable name itself.
  - This means if the variable stops existing, the code won't compile, instead of silently failing.
  - Protects from typos, e.g. `desc` becoming `dsec`.
  - Generally you only need to type the variable name once with the macro, verses twice if you were to do it manually. Technically you'd write it twice, one for serializing and one for deserializing, but it'd be four times otherwise!
  - Advanced automated tooling such as VSC that replaces specifically variable names and not just all mentions of `foobar` are likely to be able to replace the variable name `foobar` there as well, where it would most likely ignore the same name `"foobar" if it's a string`
  - It looks nicer, imo. `SERIALIZE_VAR(foobar)` > `var/list/vars_to_save = list("foobar")`
- Variables can have individualized serialization behavior if needed.
  - 98% of vars will probably not need that, but it's nice to have it be an option for more complex objects.
- The list of variables leverages inheritance, so that subtypes won't need to add too much on top of what's already being saved for base types.
- Will probably add more stuff here later to shill my system when I can think of more benefits.

## Demo
As a demo, the code currently has the capacity to save a bunch of default datum/atom vars, and a number of vars for material stacks, as a proof of concept since Ater already has persisting material stacks, and I feel that my system would be better suited for doing the data encoding and decoding.

Might also do reagent container saving as a demo since that seems like a complex object to save. Might be worth asking former World devs on serialization woes they had and seeing if my system can overcome similar scenarios.

Demo's also to show the relative ease in including subtype vars and leveraging inheritance in order to not need to maintain weird lists of strings.

## Obstacles
So far, there is no support for positional serialization, that is, objects that are saved currently don't remember where they were when saved. That will need to be fixed if we want to be able to serialize turf tiles and areas, ideally in a way that won't commit us to a specific thing (maybe using the options list fed into the serializing proc, similar to whether or not to recursively save a specific object?) Would also like the option to save some "imaginary" variables like relative coords, so that deserializing could use that as an option or use absolute coordinates.

Turf and area loading are not supported right now. Would need to forbid weird things like lighting objects on turfs from getting saved.

Will need to find a method to attempt to preserve 'internal' references, which will be needed for complex objects like humans and their organs. That's sort of the ideas for the `persistent_address` var but it might be a bad idea, since objects getting loaded twice will have duplicate persistent addresses...

Would be nice to have some kind of automatic variable renaming migration system, so that if a var gets renamed, then an object is loaded with the old var name, the system would know to treat it like the new var name instead.

### SSpersistence
One thing to note is that this kinda competes with `SSpersistence` since that uses json to save and load stuff as well, compared to my `SSserialization`, so we might need to decide what to do with that.

### Chonky JSON
Basic `/obj` objects have over 200 variables. Humans have over 600 vars + everything inside them like organs. While a lot of those vars hold transient data or otherwise have things that shouldn't be saved, like airflow stuff, other stuff still should be saved. This means each individual object might inflate the resulting json file.

A lot of that bloat is because most variables don't change, or at least aren't different from their compile-time values (which can be checked with `initial()`. One method to 'compress' the json file output is to not save variables that haven't changed. That saves a lot of space, but it also introduces additional complexity and would also require baking in assumptions into the system that might be dodgy.

Not saving vars if they don't differ implies that if an object is being deserialized, and if it looks for a variable and doesn't find one, the system needs to assume that the value got optimized out, and to instead overwrite the variable to `initial()` value instead. This means that if the compile-time value gets changed inbetween saving and loading, the loaded object will receive the new default instead of what it actually was when it was saved, which could be a problem.

This optimization system also introduces headaches involving `null`. There isn't a nice and easy way to distinguish the system seeing `null` because a variable isn't in the json file, verses if it actually is in the file but the value is actually null. That CAN be done but it involves code that I assume is O(N) to see if the list contains a value, verses O(1) for index access.

For example, say there's an object with `color = null`, and the default is actually `color = #ff0000`. It doesn't get optimized out since it differed, but when the json is read again, it will look for `color`, find `null`, and possibly think 'okay it was default so it's actually `#ff0000`, but that's incorrect.

One method to fix that is to use a null marker for legitimate nulls, instead of storing actual nulls. That could solve the above problem, but it introduces even more complexity.

Is all of that worth the space savings?

## JSON Structure
Currently, the whole object is held inside of a JSON object (which gets turned into a byond assoc list, then an object instance). iirc, TG does it a bit differently, with the object being held inside a JSON object, but that object is held inside a metadata JSON object. Would we benefit from that? It might be a good place to shove 'imaginary' variables like relative position coords, but it also adds complexity...


MistakeNot4892:
SSpersistence could easily be rolled into SSserialization though conceptually there might be some oddness there. At the very least I don't think they need to compete, SSpersistece can handle the idea of how and why things persist between rounds and then call out to SSserialization to load/save without there being too much overlap or redundancy.

With regard to `null`s, I think it would make more sense to write-out some substituted value (`NULL` or such) and then read them back in to a proper `null`. I think this is what you mean by complexity? Having something to handle situational values like that wouldn't be too complex I think, maybe just an assoc list of values to substitutions in case it needs to be expanded later (or in case there's a way to compress serialized data by using constants to substitute for large/complex values).

Changes in compile-time values are pretty unavoidable, similar to changes in type path. You would need to write some kind of metastructure to track and report on changes to the object structure via Github and ultimately I don't think it's worth the headache.

code/controllers/subsystems/serialization.dm
```diff
@@ -0,0 +1,316 @@
...
+		CRASH("Tried to open file '[file_path]', but it does not exist or cannot be accessed.")
+	
+	var/json_file = file(file_path)
+	return file2text(json_file)
+
+// Adds some whitespace to make the json formatting easier for humans to read.
+// `json_decode()` ignores it so it's purely for the benefit of humans.
+// However it will add a small amount of overhead when something is being saved.
+// The file size will also be slightly bigger due to the added whitespace.
+/datum/controller/subsystem/serialization/proc/pretty_json(json_string)
```
MistakeNot4892, line 139: Might be worth gating this behind an #ifdef or config or something, this is significantly more than `a little` overhead due to all the string operations.

code/controllers/subsystems/serialization.dm
```diff
@@ -0,0 +1,316 @@
...
+					depth--
+					string_to_add = "\n[indent(depth)][char]"
+			if(",")
+				if(!inside_string)
+					string_to_add = "[char]\n[indent(depth)]"
+		new_json_string_list += string_to_add
+	return new_json_string_list.Join()
+
+// Used for above proc.
+/datum/controller/subsystem/serialization/proc/indent(amount)
```
MistakeNot4892, line 173: `pad_right()` or `pad_left()` on Neb might be worth porting as they're useful procs in general.

Spookerton:
@Neerti is this still a technical goal to implement/use?

Atermonera:
@Neerti isn't capable of working on byond atm due to computer issues

PR #8104, opened 2021-05-24
Adds new rocks + resprites mining outcrops

Cerebulon:
Rocks and water rocks:
https://i.gyazo.com/b33773fb23a658349188c8bec22d2b45.mp4
Rocks existed in the code but weren't used. Water effects are new, for mapping into water areas. All purely cosmetic and visually distinct from...

Extremely out of place rocky outcrop sprites replaced with rocks that blend in with the underground environment more.
rocks


Atermonera:
icondiff looks like it did an oopsie?
mp4

Cerebulon:
If you download the dmi there's no fucked up frame in the actual file so... Shrugs.

PR #8109, opened 2021-05-25
Misc Bits

Woodratt:
Misc bits for the new map.

- Another statue sprite.
- Shuttle handrails that can be buckled into.


Novacat:
Wait, did we not port our handrails upstream?

Woodratt:
Correct

PR #8115, opened 2021-05-28
Random mob spawners can accept faction input now

Very-Soft:
The variable was there, just not used for anything, so I fix

### DIFF
--- a/code/game/objects/random/mob.dm
+++ b/code/game/objects/random/mob.dm
@@ -61,7 +61,7 @@
 		M.pixel_y = pixel_y
 
 	if(mob_faction)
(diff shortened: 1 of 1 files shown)


code/game/objects/random/mob.dm
```diff
@@ -60,6 +60,10 @@
 		M.pixel_x = pixel_x
 		M.pixel_y = pixel_y
 
+	if(mob_faction)
+	M.faction = mob_faction
```
Heroman3003, line 64:
```suggestion
	if(mob_faction)
		M.faction = mob_faction
```

PR #8116, opened 2021-05-28
Fixes some unseen combat ai jank

Verkister:
haha lack of exit condition go brrrr

kuva


Neerti:
Does this conflict with #8112?

Verkister:
Should be fine on the same spot with just adjusting the tabbing. Aro might probably still slap a small fix in that PR considering there's a > facing wrong way that they spotted. Basically it works with the current code, but Aro might have their own approach planned to replace this part for their PR.

Atermonera:
Was this addressed by aro's PR?

Verkister:
Wasn't addressed, just mitigated by the timeout part there. The spam would still happen for the timeout wait period. This PR makes the spam not happen in the first place.

PR #8122, opened 2021-05-31
Make req_[one_]access lazy

Arokha:
Remove these two lists from being instantiated on every /obj

### DIFF
--- a/code/modules/customitems/item_spawning.dm
+++ b/code/modules/customitems/item_spawning.dm
@@ -190,7 +190,7 @@
 
 		// Check for required access.
 		var/obj/item/weapon/card/id/current_id = M.wear_id
-		if(citem.req_access && citem.req_access > 0)
+		if(citem.req_access && citem.req_access > 0) // These are numbers, not lists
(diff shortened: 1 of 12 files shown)


MarinaGryphon (changes requested):

MarinaGryphon (approved):
Also fixes `LAZYCLEARLIST(L)` to null lists out after cutting, because that's how lazylists work.
Yes, the cut is necessary for GC reasons. I asked.

code/modules/customitems/item_spawning.dm
```diff
@@ -190,7 +190,7 @@
 
 		// Check for required access.
 		var/obj/item/weapon/card/id/current_id = M.wear_id
-		if(citem.req_access && citem.req_access > 0)
+		if(citem.req_access && citem.req_access > 0) // These are numbers, not lists
```
Atermonera, line 193: is the first clause really necessary? 0 and null are both false-y values, and so evaluate to 0, which should fail the second clause.

Arokha: True, though not code I added or changed

PR #8138, opened 2021-06-14
fixes coughs screaming in the chat

Verkister:


MistakeNot4892:
Copypaste strikes again rip

PR #8139, opened 2021-06-15
Configure some VV bitfields and fix editor

Leshana:
- Configure vv bitfields for appearance_flags, vis_flags, and sight.
flags with bitfields
- Fix unwanted line wraps in vv bitfield editor.
flags bitfields editor


Atermonera:
Would something like this be appropriate for `dir`?

Leshana:
> 
> 
> Would something like this be appropriate for `dir`?

I think `dir` ishandled better by the specific dir editor we already have in the upper left.
image

PR #8144, opened 2021-06-15
Indescribable! - A Food Flavor Fixing Pass

MarinaGryphon:
*You keep using that word. I do not think it means what you think it means.*
  - Food will no longer erroneously have double the nutriment it's supposed to.
  - Fixed a lot of invalid food nutriment tastes and spruced up some old ones.
  - Made all foods use `nutriment_amt` where applicable
  - Made those not applicable to the former use `nutriment_desc` when manually adding nutriment
  - Moved setting bitesize out of Initialize, because... why

### DIFF
--- a/code/modules/food/food/snacks.dm
+++ b/code/modules/food/food/snacks.dm
@@ -4183,9 +4183,6 @@
 	nutriment_amt = 4
 	nutriment_desc = list("soggy fries" = 4)
 	center_of_mass = list("x"=16, "y"=11)
-
-/obj/item/weapon/reagent_containers/food/snacks/microchips/Initialize()
-	. = ..()
 	bitesize = 2
 
 /obj/item/weapon/reagent_containers/food/snacks/ovenchips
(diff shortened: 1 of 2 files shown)


Cerebulon:
> Moved setting bitesize out of Initialize, because... why

These were all SPECIFICALLY moved to be defined post-reagent-adding in initialize quite recently because otherwise absolutely inexplicably, the reagents.add_reagent code in init just would not work if bitesize was defined in the object itself. Any food with it defined how you're changing it back to would be missing its special reagents.
If you've managed to fix that somehow, then great. Try verifying that a food with unusual reagents actually contains it in-game, e.g. Chaos Bagels should contain Defective Nanites as an obvious one.

MarinaGryphon:
> Edit: Actually, that isn't a good test because you didn't move the bitesize on that particular food.

That's... not correct. I just moved it locally and tested, and it worked fine. I also tested chaos donuts and a number of other foods.

Chances are, someone simply made a mistake and blamed it on bitesize; I also checked for any kind of causal mechanism, and there is none.

image

MistakeNot4892:
Can confirm bitesize has been on the item definitions on Neb (and I am pretty sure Bay) for years, whoever did the original tests likely conflated two or more changes with the bug and didn't properly bisect their code to resolve it.

Cerebulon:
Yeah nevermind I was misremembering and was thinking of the exact opposite. Moving bitesize out of init FIXES the thing I was talking about, and I guess I just never did a thorough job of it previously and this rectifies that.

PR #8145, opened 2021-06-15
Make item/attackby call ..(), multitool tool quality fix

ShadowLarkens:
Also includes https://github.com/VOREStation/VOREStation/pull/10658
because the fix otherwise makes items randomly whine about being hit and it's just a mess
there's no reason to print this message unless it's actually doing damage

### DIFF
--- a/code/_onclick/item_attack.dm
+++ b/code/_onclick/item_attack.dm
@@ -41,11 +41,6 @@ avoid code duplication. This includes items that may sometimes act as a standard
 		return TRUE
 	return FALSE
 
-/atom/movable/attackby(obj/item/W, mob/user, var/attack_modifier, var/click_parameters)
-	. = ..()
-	if(!. && !(W.flags & NOBLUDGEON))
-		visible_message("<span class='danger'>[src] has been hit by [user] with [W].</span>")
-
 /mob/living/attackby(obj/item/I, mob/user, var/attack_modifier, var/click_parameters)
 	if(!ismob(user))
 		return 0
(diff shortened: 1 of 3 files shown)


MistakeNot4892:
Generally IMO you want to call ..() at the end of procs like attackby() that assume only one interaction will result. Otherwise if someone defines new behavior on /obj or /atom etc. you'll potentially end up with extra interactions and messages.

ShadowLarkens:
> Generally IMO you want to call ..() at the end of procs like attackby() that assume only one interaction will result. Otherwise if someone defines new behavior on /obj or /atom etc. you'll potentially end up with extra interactions and messages.

you know, exactly as you would expect to happen rather than "oh, items don't follow the parent?

as far as I can tell it's not applicable here anyways because I don't think storage stuff has return values

MistakeNot4892:
I would not personally expect to get several instantaneous interactions and outputs from one attackby, I would expect it to evaluate each one once then proceed to the parent if it isn't dealt with by that proc. Otherwise you might end up with a low level proc moving the atom around and then a high level proc getting confused about an inconsistent location, etc.

If that's not the expectation on this code then my comment doesn't apply.

ShadowLarkens:
> 
> 
> I would not personally expect to get several instantaneous interactions and outputs from one attackby, I would expect it to evaluate each one once then proceed to the parent if it isn't dealt with by that proc. Otherwise you might end up with a low level proc moving the atom around and then a high level proc getting confused about an inconsistent location, etc.
> 
> If that's not the expectation on this code then my comment doesn't apply.

it's very much not because components should always take priority so they can cancel the rest of the attackby and they're on /atom/attackby

PR #8154, opened 2021-06-21
Closet appearance decls correctly use the repository

Atermonera:
Tested, doesn't break everything:
dreamseeker_2021-06-20_23-53-01

Moves decl repo init back out of global init

### DIFF
--- a/code/datums/repositories/decls.dm
+++ b/code/datums/repositories/decls.dm
@@ -14,8 +14,7 @@
 //     III. Thou shalt not write a decl that relies on arguments supplied to New().
 //     IV.  Thou shalt not call Initialize() on a /decl.
 
-var/repository/decls/decls_repository // Initialiozed in /datum/global_init/New()
-
+var/repository/decls/decls_repository = new()
 /repository/decls
 	var/list/fetched_decls
 	var/list/fetched_decl_types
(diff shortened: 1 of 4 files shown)


code/datums/repositories/decls.dm
```diff
@@ -14,8 +14,7 @@
 //     III. Thou shalt not write a decl that relies on arguments supplied to New().
 //     IV.  Thou shalt not call Initialize() on a /decl.
 
-var/repository/decls/decls_repository // Initialiozed in /datum/global_init/New()
-
+var/repository/decls/decls_repository = new()
```
MistakeNot4892, line 17: What is the purpose of moving it to world new?

code/game/objects/structures/crates_lockers/__closets.dm
```diff
@@ -59,7 +59,7 @@
 			storage_capacity = content_size + 5
 
 	if(ispath(closet_appearance))
-		var/decl/closet_appearance/app = GLOB.closet_appearances[closet_appearance]
+		var/decl/closet_appearance/app = decls_repository.get_decl(closet_appearance)
```
Neerti, line 62: I think you can also use the `GET_DECL()` macro.

code/datums/repositories/decls.dm
Atermonera: Lazy fix for it needing to exist in global_init because I didn't bother removing it from GLOB or checking to update literally the only usage of the list

PR #8155, opened 2021-06-21
Adds optional area-based definition for outdoors-ness

Neerti:
This is mostly for the new map, but it might be helpful to have now and outside of the map PR.

This essentially converts the `outdoors` var on turfs from a binary value to a ternary, with init values being able to be 'yes', 'no', and a new option, 'area', which makes it outdoors or indoors based on what the area wants it to be, with a newly added value that can hold a 'yes' or 'no'.

The advantage is that this helps avoid ten thousand `/outdoors` subtypes for turfs (tho that ship has kinda sailed, but we can now get rid of those types and replace with the base one instead). It also helps avoid bugs with single turfs that are indoors when they're intended to be outdoors (or the other way around). It also helps cut down on map overrided turfs to set `outdoors` to 0 or 1.

It should be noted the turfs themselves have the final say on whether or not they're outdoors. If the value is 'yes' or 'no', the area's value is irrelevant. Also, the 'yes' value is 1, and the 'no' value is 0, which mirrors the binary system and thus this should be backwards compatible with current maps.

By default, turfs defer to the area. Also by default, areas default to being indoors. This is done since the majority of areas are indoors. Most outdoor areas are under a general outdoor area anyways, so it's easy to apply it.

### DIFF
--- a/code/controllers/subsystems/planets.dm
+++ b/code/controllers/subsystems/planets.dm
@@ -42,7 +42,7 @@ SUBSYSTEM_DEF(planets)
 			return
 		if(istype(T, /turf/unsimulated/wall/planetary))	
 			P.planet_walls += T
-		else if(istype(T, /turf/simulated) && T.outdoors)
+		else if(istype(T, /turf/simulated) && T.outdoors == OUTDOORS_YES)
 			P.planet_floors += T
 			T.vis_contents |= P.weather_holder.visuals
 			T.vis_contents |= P.weather_holder.special_visuals		
--- a/code/__defines/turfs.dm
+++ b/code/__defines/turfs.dm
@@ -20,3 +20,9 @@
 #define isDiagonal(x)			(x == NORTHEAST || x == SOUTHEAST || x == NORTHWEST || x == SOUTHWEST)
 
 #define FOOTSTEP_SPRITE_AMT 2
+
+// Used to designate if a turf (or its area) should initialize as outdoors or not.
+#define OUTDOORS_YES		1	// This being 1 helps with backwards compatibility.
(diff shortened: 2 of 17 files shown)


code/controllers/subsystems/planets.dm
```diff
@@ -42,7 +42,7 @@ SUBSYSTEM_DEF(planets)
 			return
 		if(istype(T, /turf/unsimulated/wall/planetary))	
 			P.planet_walls += T
-		else if(istype(T, /turf/simulated) && T.outdoors)
+		else if(istype(T, /turf/simulated) && T.outdoors == OUTDOORS_YES)
```
Atermonera, line 45: Should all of these checks use a ternary statement (probably make it a function it for simplicity, #defines might be messy) to check both turf and area outdoors where necessary?
i.e.,
```
/proc/is_outdoors(atom/A)
    var/turf/T = get_turf(A)
    if(istype(T))
         switch(T.outdoors)
            if(OUTDOORS_YES):
                return TRUE
            if(OUTDOORS_AREA):
                 var/area/A = get_area(T)
                 return A.outdoors
    return FALSE
```

Neerti: When a turf gets initialized and turned into an outdoor turf, even if it's by area, it's `outdoors` var is set to `OUTDOORS_YES`.
I now realize that I don't believe that occurs for default indoor turfs so that might need to get tweaked, or I can remove init turning `outdoors` into a binary value and instead keep `OUTDOORS_AREA` and use a proc, tho I'm not sure how many situations where keeping it a ternary after init would be useful.

PR #8184, opened 2021-07-25
Refactors crafting to support requirement alternatives

Atermonera:
Adds recipe for spears
Fixes crafting menu not opening
Adds mechanism for providing multiple options to fulfill a particular requirement for crafting. **Note that existing recipes will need to be updated to wrap their current requirements in a containing list (so the reqs list is a list of lists in the reqs[requirement][option to fulfill requirement] format)**
Tested: Unfortunately, I don't yet have a good way to select one requirement or the other without duplicating the recipe (ick), so it's random if there's multiple things that could fulfill it. I think the best solution would be to add some selection to the crafting menu itself, but that's out of scope here.
Also SIGNIFICANTLY cleans up the code involved, that stuff is a mess.
Current UI status:

Following shows the actual usage of the recipes, from a pile of ingredients at the user's feet. One shows construction of a glass spear, given both a glass shard and a steel blade to randomly choose from. The other shows construction of a steel spear from a steel blade.

https://user-images.githubusercontent.com/8475252/126916893-3c233172-343b-43ea-b0e8-1c026de30483.mp4

https://user-images.githubusercontent.com/8475252/126916898-fac2ec2d-b67d-48fc-b514-cff5a4864816.mp4

### DIFF
--- a/code/datums/components/crafting/crafting.dm
+++ b/code/datums/components/crafting/crafting.dm
@@ -75,28 +75,34 @@
 	var/list/requirements_list = list()
 
 	// Process all requirements
-	for(var/requirement_path in R.reqs)
-		// Check we have the appropriate amount available in the contents list
-		var/needed_amount = R.reqs[requirement_path]
-		for(var/content_item_path in contents)
-			// Right path and not blacklisted
-			if(!ispath(content_item_path, requirement_path) || R.blacklist.Find(content_item_path))
+	for(var/list/requirement in R.reqs)
+		var/satisfied = FALSE
+		for(var/requirement_path in requirement)
+			// Check we have the appropriate amount available in the contents list
+			var/needed_amount = requirement[requirement_path]
+			for(var/content_item_path in contents)
+				// Right path and not blacklisted
+				if(!ispath(content_item_path, requirement_path) || R.blacklist.Find(content_item_path))
+					continue
+
+				needed_amount -= contents[content_item_path]
+				if(needed_amount <= 0)
+					break
+
+			if(needed_amount > 0)
 				continue
 
-			needed_amount -= contents[content_item_path]
-			if(needed_amount <= 0)
-				break
+			// Store the instances of what we will use for R.check_requirements() for requirement_path
+			var/list/instances_list = list()
+			for(var/instance_path in item_instances)
+				if(ispath(instance_path, requirement_path))
+					instances_list += item_instances[instance_path]
 
-		if(needed_amount > 0)
+			requirements_list[requirement_path] = instances_list
+			satisfied = TRUE
+			break
+		if(!satisfied)
 			return FALSE
-
-		// Store the instances of what we will use for R.check_requirements() for requirement_path
-		var/list/instances_list = list()
-		for(var/instance_path in item_instances)
-			if(ispath(instance_path, requirement_path))
-				instances_list += item_instances[instance_path]
-
-		requirements_list[requirement_path] = instances_list
 
 	for(var/requirement_path in R.chem_catalysts)
 		if(contents[requirement_path] < R.chem_catalysts[requirement_path])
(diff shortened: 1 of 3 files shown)


code/datums/components/crafting/crafting.dm
```diff
@@ -119,13 +126,13 @@
...
-
+// Returns an associative list containing the types of tools available, and the paths of objects available
 /datum/component/personal_crafting/proc/get_surroundings(atom/a, list/blacklist=null)
 	. = list()
-	.["tool_qualities"] = list()
-	.["other"] = list()
-	.["instances"] = list()
-	.["machinery"] = list()
+	.["tool_qualities"] = list() // List of tool types available
+	.["other"] = list()		// List of reagents/material stacks available
```
MistakeNot4892, line 133: Nit: This would render a bit nicer if indented with spaces rather than tabs.

code/datums/components/crafting/crafting.dm
```diff
@@ -75,29 +75,35 @@
...
+		var/satisfied = FALSE
+		for(var/requirement_path in requirement)
+			// Check we have the appropriate amount available in the contents list
+			var/needed_amount = requirement[requirement_path]
+			for(var/content_item_path in contents)
+				// Right path and not blacklisted
+				if(!ispath(content_item_path, requirement_path) || R.blacklist.Find(content_item_path))
+					continue
+
+				needed_amount -= contents[content_item_path]
```
MistakeNot4892, line 88: Nit: seeing the var named `contents` in a datum context was a bit confusing, might be worth renaming to be distinct from the inbuilt atom var at some point.

code/datums/components/crafting/recipes.dm
```diff
@@ -48,9 +48,40 @@
 /datum/crafting_recipe/stunprod
 	name = "Stunprod"
 	result = /obj/item/weapon/melee/baton/cattleprod
-	reqs = list(/obj/item/weapon/handcuffs/cable = 1,
-				/obj/item/stack/rods = 1,
-				/obj/item/weapon/tool/wirecutters = 1)
+	reqs = list(list(/obj/item/weapon/handcuffs/cable = 1),
```
MistakeNot4892, line 51: Nit: This is uncomfortable. Not sure how else you'd do alternatives without switching this to some kind of datum system but nested lists are a bit urgh generally.

code/datums/components/crafting/crafting_external.dm
```diff
@@ -12,21 +12,27 @@
...
-				if(isliving(M.loc))
-					var/mob/living/L = M.loc
-					L.unEquip(M, target = src)
-				else
-					M.forceMove(src)
-				SEND_SIGNAL(M, COMSIG_ATOM_USED_IN_CRAFT, src)
+	if(LAZYLEN(parts_list))
+		for(var/datum/reagent/A as anything in parts_list["reagents"])
+			if(!reagents)
+				reagents = new()
```
MistakeNot4892, line 18: `create_reagents(someval)`

code/datums/components/crafting/crafting_external.dm
```diff
@@ -12,21 +12,27 @@
...
-			else if(ismovable(A))
-				var/atom/movable/M = A
-				if(isliving(M.loc))
-					var/mob/living/L = M.loc
-					L.unEquip(M, target = src)
-				else
-					M.forceMove(src)
-				SEND_SIGNAL(M, COMSIG_ATOM_USED_IN_CRAFT, src)
+	if(LAZYLEN(parts_list))
+		for(var/datum/reagent/A as anything in parts_list["reagents"])
```
MistakeNot4892, line 16: Would it be feasible to pass the reagents datum instead of a list of individual reagents? You could then either `trans_to_holder` or just set myatom on the reagents to the new atom.

Atermonera:
Retested with new code, still works:
https://puu.sh/I5067/4a2fce56b8.png

MistakeNot4892 (approved):
I keep coming back to try and review this but a lot of the code makes me go a bit cross-eyed so I'm not confident in my ability to say that the flow is sound and there aren't possible issues in the logic. Since you've tested it, it works, and I can't see any serious issues with the code itself I'm going to approve it and we can deal with issues if/as they arise.

PR #8199, opened 2021-08-08
Ports Diagonal Movement

Cerebulon:
https://i.gyazo.com/d74fba16abce0589d02408333528407c.mp4

for years, man has gone sideways but now
man go diagonal

https://github.com/VOREStation/VOREStation/pull/10928
https://github.com/VOREStation/VOREStation/pull/10975

:cl:
rscadd - Implemented diagonal movement when two directional keys are pressed simultaneously.
/:cl:


code/__defines/_compile_options.dm
```diff
@@ -11,6 +11,9 @@
 //#define ZASDBG	 	// Uncomment to turn on super detailed ZAS debugging that probably won't even compile.
 #define MULTIZAS		// Uncomment to turn on Multi-Z ZAS Support!
 
+// Movement Compile Options
+//#define CARDINAL_INPUT_ONLY // Uncomment to disable diagonal player movement (restore previous cardinal-moves-only behavior)
```
MistakeNot4892, line 15: I feel like this should be a config var rather than a compile time var.

code/__defines/subsystems.dm
```diff
@@ -55,6 +55,7 @@ var/global/list/runlevel_flags = list(RUNLEVEL_LOBBY, RUNLEVEL_SETUP, RUNLEVEL_G
 #define INIT_ORDER_WEBHOOKS		50
 #define INIT_ORDER_SQLITE		40
 #define INIT_ORDER_CHEMISTRY	35
+#define INIT_ORDER_INPUT		37
```
MistakeNot4892, line 58: Values are out of order. :(

code/modules/mob/mob.dm
```diff
@@ -618,6 +619,9 @@
 				stat("CPU:","[world.cpu]")
 				stat("Instances:","[world.contents.len]")
 				stat(null, "Time Dilation: [round(SStime_track.time_dilation_current,1)]% AVG:([round(SStime_track.time_dilation_avg_fast,1)]%, [round(SStime_track.time_dilation_avg,1)]%, [round(SStime_track.time_dilation_avg_slow,1)]%)")
+				stat("Keys Held", keys2text(client.move_keys_held | client.mod_keys_held))
```
MistakeNot4892, line 622: Why is this being thrown into the status panel? Won't that be a fair bit of extra overhead for no useful purpose?

code/modules/mob/mob_movement.dm
```diff
@@ -137,25 +137,36 @@
...
+		DEBUG_INPUT("--------")
+	next_move_dir_add = 0	// This one I *think* exists so you can tap move and it will move even if delay isn't quite up.
+	next_move_dir_sub = 0 	// I'm not really sure why next_move_dir_sub even exists.
+
+	if(!n || !direct)
+		return
 
 	// If dead and we try to move in our mob, it leaves our body
 	if(my_mob.stat == DEAD && isliving(my_mob) && !my_mob.forbid_seeing_deadchat)
+		my_mob.setMoveCooldown(my_mob.movement_delay(n, direct))
```
MistakeNot4892, line 169: What's the purpose of setting move delay on a dead mob?

MistakeNot4892:
Generally looks good, but would suggest a live test given that we've had a few odd bugs pop up on Neb that we missed in review for our port.

code/__defines/_compile_options.dm
MistakeNot4892: On reflection performance would take a hit. Not sure it's worth including a debug flag for it though.

code/modules/mob/mob_movement.dm
MistakeNot4892: Discussed on Discord, may solve issues with ghost gliding being wonky.

MistakeNot4892 (dismissed):
Looks fine to me though a test would still be smart imo.

MistakeNot4892:
@Schnayy is going to raise this at the next staff meeting to discuss scheduling a test.

code/_helpers/type2type.dm
```diff
@@ -96,6 +97,19 @@
...
+	if(!direction)
+		return ""
+	var/list/dirs = list()
+	if(direction & NORTH) dirs += "NORTH"
+	if(direction & SOUTH) dirs += "SOUTH"
+	if(direction & EAST) dirs += "EAST"
+	if(direction & WEST) dirs += "WEST"
+	if(direction & UP) dirs += "UP"
+	if(direction & DOWN) dirs += "DOWN"
+	return dirs.Join(" ")
```
Atermonera, line 111:
```
// Turns a direction into text showing all bits set
/proc/dirs2text(direction)
    . = ""
    if(direction & [DIR]) . += "DIR "
    [...]
    . = copytext(., 1, (.).len - 1)
```

code/modules/keybindings/bindings_atom.dm
```diff
@@ -0,0 +1,45 @@
...
+	if(user.mod_keys_held & CTRL_KEY)
+		return
+
+	var/must_call_move = FALSE
+	var/movement_dir = MOVEMENT_KEYS_TO_DIR(user.move_keys_held)
+	if(user.next_move_dir_add)
+		must_call_move = TRUE // So that next_move_dir_add gets cleared if its time.
+		movement_dir |= user.next_move_dir_add
+	if(user.next_move_dir_sub)
+		DEBUG_INPUT("[(user.next_move_dir_sub & movement_dir) ? "Actually" : "Want to"] subtract [dirs2text(user.next_move_dir_sub)] from [dirs2text(movement_dir)]")
```
Atermonera, line 18: Where does this get output to? How often will this generate logs?

code/__defines/flags.dm
```diff
@@ -13,6 +13,10 @@
 
 GLOBAL_LIST_INIT(bitflags, list(1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192, 16384, 32768))
 
+/* Directions */
+///All the cardinal direction bitflags.
+#define ALL_CARDINALS (NORTH|SOUTH|EAST|WEST)
+
```
Atermonera, line 19: This should go with the other defines that actually reference it in `input.dm`

code/modules/keybindings/bindings_movekeys.dm
```diff
@@ -0,0 +1,77 @@
...
+	set instant = TRUE
+	set hidden = TRUE
+	// set name = ".movekeyup"
+	set name = "KeyUp"
+
+	// Map text sent by skin.dmf to our numeric codes. (This can be optimized away once we update skin.dmf)
+	var/movekey = MOVE_KEY_MAPPINGS[movekeyName]
+
+	// Validate input.  Must be one (and only one) of the key codes)
+	if(isnull(movekey) || (movekey & ~0xFFF) || (movekey & (movekey - 1)))
```
Atermonera, line 59: `isnull()` is more expressively caught by the second check ("" == null).
`0xFFF` feels fragile. It should be a define that is documented as capturing all the valid movement keys.

code/modules/keybindings/bindings_movekeys.dm
```diff
@@ -0,0 +1,77 @@
...
+		log_debug("Client [ckey] sent an illegal movement key up: [movekeyName] ([movekey])")
+		return
+
+	// Clear bit indicating we were holding the key
+	move_keys_held &= ~movekey
+	mod_keys_held &= ~movekey
+
+	// If we were holding at the start of this move cycle and now relased it, remember that.
+	var/movement = MOVEMENT_KEYS_TO_DIR(movekey)
```
Atermonera, line 68: It feels like this key-text-bitfield mapping is invoked in about 3 different naming schemes, can we unify this?
i.e., MOVE_KEY_MAPPINGS, MOVEMENT_KEYS_TO_DIR, and dirs_to_text should all use similar format (capitalization, proc/define, etc.,)

code/modules/keybindings/setup.dm
```diff
@@ -0,0 +1,34 @@
...
+/mob/proc/set_focus(datum/new_focus)
+	if(focus == new_focus)
+		return
+	focus = new_focus
+
+/// Turns a keys bitfield into text showing all bits set
+/proc/keys2text(keys)
+	if(!keys)
+		return ""
+	var/list/out = list()
```
Atermonera, line 22: See also above. Don't actually know if list operations are cheaper than string operations but for a completely self-contained set of strings, generating a list and then concatenating that list feels a bit excessive.

Atermonera (commented):
I'm not... really a fan of switching the current interrupt-based system for a polling system, but looking at how our skin currently handles repeated input, I don't think it really lends itself to diagonal movement. Also not really a fan of how this is split between movekeys, bitfields, and textstrings. It feels kinda spaghetti.

code/__defines/input.dm
```diff
@@ -0,0 +1,21 @@
+// Bitflags for the move_keys_held bitfield.
+#define NORTH_KEY   (1<<0)
+#define SOUTH_KEY   (1<<1)
+#define EAST_KEY    (1<<2)
+#define WEST_KEY    (1<<3)
+#define W_KEY       (1<<4)
+#define S_KEY       (1<<5)
+#define D_KEY       (1<<6)
+#define A_KEY       (1<<7)
```
Atermonera, line 9: Challenge: Why do we need to track WASD separately from NESW? It adds a lot of logical complexity that makes this stuff harder to read. If we drop it, the only change I see is that if you're holding both UP and W, the first one you release will stop movement, and I don't necessarily see that as a bad thing.

Atermonera (changes requested):

code/_helpers/type2type.dm
MistakeNot4892: This code makes no sense. DIR is an integer and copytext and repeated string operations aren't an improvement on Join().

code/__defines/input.dm
MistakeNot4892: SSinput allows for keybinding in the future and it is generally more robust for minimal additional complexity. I don't think it's significantly harder to read personally.

Atermonera (approved):

Atermonera (changes requested):
Ctrl+wasd does not work when not in hotkey mode. Arrow keys work both in and out of hotkey mode, and wasd works in hotkey mode. This must be fixed for this PR to be merged.

FenodyreeAv:
This should have mathamatically correct slowdown. Without it it feels strange.
There are also edge cases, moving in a 1x1 corridoor, pressing the diagonal movement keys causes slowdown even when not moving diagonally.

FenodyreeAv:
Also, wierd interaction with pushing objects.
https://streamable.com/q9h1dg

code/modules/keybindings/setup.dm
Spookerton: Byond strings are cstrings, which are immutable. Concatenating many together in sequence means having the initial string, then for each chunk added creating the string to add, creating the concatenated result, and then discarding the original and the add string.

List and Join can be dramatically cheaper in this style of use, which is effectively a budget StringBuilder. Most of the intermediate allocation and freeing is skipped, and the final join is done as a single act from the combined set length, the length of the members, and the joining particle.

code/modules/keybindings/setup.dm
```diff
@@ -0,0 +1,34 @@
...
+/datum/proc/keyLoop(client/user) // Called once every server tick
+	set waitfor = FALSE
+	return
+
+/// Set mob's focus
+/// TODO - Do we even need this concept?
+/mob/proc/set_focus(datum/new_focus)
+	if(focus == new_focus)
+		return
+	focus = new_focus
```
Spookerton, line 16:
```suggestion
// Set a client's focus to an object and override these procs on that object to let it handle keypresses

/// Called when a key is pressed down initially
/datum/proc/key_down(key, client/user)
	return

/// Called when a key is released
/datum/proc/key_up(key, client/user)
	return

/// Called once every server tick
/datum/proc/keyLoop(client/user)
	set waitfor = FALSE
	return

/// Set mob's focus. TODO: Decide if required.
/mob/proc/set_focus(datum/new_focus)
	if(focus == new_focus)
		return
	focus = new_focus
```
[code standards - readability](https://github.com/PolarisSS13/Polaris/wiki/Code-Standards#code-readability)

code/_helpers/type2type.dm
```diff
@@ -96,6 +97,19 @@
...
+/proc/dirs2text(direction)
+	if(!direction)
+		return ""
+	var/list/dirs = list()
+	if(direction & NORTH) dirs += "NORTH"
+	if(direction & SOUTH) dirs += "SOUTH"
+	if(direction & EAST) dirs += "EAST"
+	if(direction & WEST) dirs += "WEST"
+	if(direction & UP) dirs += "UP"
+	if(direction & DOWN) dirs += "DOWN"
```
Spookerton, line 110:
```suggestion
	if(direction & NORTH)
		dirs += "NORTH"
	if(direction & SOUTH)
		dirs += "SOUTH"
	if(direction & EAST)
		dirs += "EAST"
	if(direction & WEST)
		dirs += "WEST"
	if(direction & UP)
		dirs += "UP"
	if(direction & DOWN)
		dirs += "DOWN"
```
[code standards - readability](https://github.com/PolarisSS13/Polaris/wiki/Code-Standards#code-readability)

code/modules/keybindings/setup.dm
```diff
@@ -0,0 +1,34 @@
...
+	if(keys & SOUTH_KEY) out += "SOUTH"
+	if(keys & EAST_KEY) out += "EAST"
+	if(keys & WEST_KEY) out += "WEST"
+	if(keys & W_KEY) out += "W"
+	if(keys & A_KEY) out += "A"
+	if(keys & S_KEY) out += "S"
+	if(keys & D_KEY) out += "D"
+	if(keys & CTRL_KEY) out += "CTRL"
+	if(keys & SHIFT_KEY) out += "SHIFT"
+	if(keys & ALT_KEY) out += "ALT"
```
Spookerton, line 33:
```suggestion
	if(keys & NORTH_KEY)
		out += "NORTH"
	if(keys & SOUTH_KEY)
		out += "SOUTH"
	if(keys & EAST_KEY)
		out += "EAST"
	if(keys & WEST_KEY)
		out += "WEST"
	if(keys & W_KEY)
		out += "W"
	if(keys & A_KEY)
		out += "A"
	if(keys & S_KEY)
		out += "S"
	if(keys & D_KEY)
		out += "D"
	if(keys & CTRL_KEY)
		out += "CTRL"
	if(keys & SHIFT_KEY)
		out += "SHIFT"
	if(keys & ALT_KEY)
		out += "ALT"
```
[code standards - readability](https://github.com/PolarisSS13/Polaris/wiki/Code-Standards#code-readability)

Spookerton (changes requested):

code/modules/keybindings/bindings_atom.dm
Spookerton:
```
// Uncomment to get a lot of debug logging for movement keys.
// #define DEBUG_INPUT(A) to_world_log(A)
#define DEBUG_INPUT(A) 
```
From elsewhere in the PR.

I'm not convinced this is a good pattern to include, and would be better handled at least with an ifdef rather telling the programmer to switch line commenting, but:

When in the uncommented form, is modified to an empty line by the compiler (and thus creates no work and has no outputs). When in the commented form, outputs directly to the world log.

code/modules/keybindings/bindings_movekeys.dm
Spookerton: isnull is generally preferable for little micro purposes - <https://pastebin.com/raw/k5pnnXhf>
@Cerebulon defines for magic bitfields still needs done

code/modules/mob/living/living_movement.dm
```diff
@@ -147,7 +147,15 @@ default behaviour is:
 				now_pushing = 0
 				return
 
-		step(AM, t)
+		var/turf/T = AM.loc
+		var/turf/T2 = get_step(AM,t)
+		if(!T2) // Map edge
+			now_pushing = 0
+			return
+		var/move_time = movement_delay(loc, t)
```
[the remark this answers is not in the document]
Spookerton: movement_delay is a tick-time between movements where higher is slower. `* sqrt(2) / 2` comes out at a fraction of 1, making diagonal movements faster compared to cardinal. Removing the / 2 is then correct for euclidean distance - but it is also much preferable to implement this with a constant:
```
var/global/const/SQRT_TWO = 1.41421356237
...
var/move_time = movement_delay(loc, t) * SQRT_TWO
```


As to whether euclidean distance is a reasonable multiplier, I disagree. People *not* moving diagonally are then actively disadvantaged, because they take the full time of two turf changes to move. Most reasonable is simply to consider a diagonal movement the same as its component x/y, and multiply move time by two:
```
var/move_time = movement_delay(loc, t) * 2
```

code/modules/keybindings/bindings_atom.dm
```diff
@@ -0,0 +1,45 @@
+// You might be wondering why this isn't client level. If focus is null, we don't want you to move.
+// Only way to do that is to tie the behavior into the focus's keyLoop().
+
+// THE TRADITIONAL STYLE FROM /TG (modified)
+
+/atom/movable/keyLoop(client/user)
+	// Bail out if the user is holding the "face direction" key (Maybe?)
+	// TODO - I think this breaks non-hotkeys WASD movement, so maybe adopt the /tg solution)
+	if(user.mod_keys_held & CTRL_KEY)
+		return
```
Atermonera, line 10: This one here, officer. These are the lines that hurt me.
[mp3 of me commenting these out and testing with ctrl+WASD locally.](https://puu.sh/J3bD9/ee54c74a7c.mp4)

What is this even supposed to do? `ALT_KEY` is the face-direction.

Atermonera (changes requested):

Greenjoe12345:
A few issues I've found from today's test merge
Movement speed in general seems faster then normal?
moving while looking through binoculars (and I'd guess scoped weapons too) doesn't reset looking through them, so you move while doing so.

Greenjoe12345:
Teshari cannot swap places with some other mobs, the mob acting like a solid wall to them, even with some simplemobs like diaaybs

Greenjoe12345:
Teshari gliding seems to sometimes fail when jumping off a z level diagonally

Cerebulon:
Wasn't able to find any problem with binoculars (they don't cancel with cardinal **or** diagonal movement), unsure if that code was changed at some point in the last year. Teshari losing in size-battles is, as far as I know, consistent with current behaviour and not related to this change.

Atermonera (approved):

PR #8162, opened 2021-06-28
Removes popups from logging in.

Neerti:
Removes VChat failing to start popup if it was due to player pref setting.
News popup is now opt-in instead of always happening.

PR #8209, opened 2021-08-11
Prevents Pop-up for Already Seen Lore News

Cerebulon:
The "Player Already Saw News" variable is now actually saved between connections, preventing the news from popping up for people who have already seen it.

Also incorporates the preference option from #8162 for those who do not want to have the popup under any circumstances, and the planned bolding of the main menu option for those with the popup off.
The preference for the popup is on by default due to the far less intrusive nature of it after this change.

### DIFF
--- a/code/modules/lore_codex/news_data/main.dm
+++ b/code/modules/lore_codex/news_data/main.dm
@@ -112,6 +112,8 @@
 		/datum/lore/codex/page/about_news,
 		)
 
+	var/newsindex = "103" //Update with number of latest article
+
 /datum/lore/codex/page/about_news
 	name = "About the Publisher"
 	data = "The <i>Daedalus Pocket Newscaster</i> is produced and maintained by Occulum Broadcast, the foremost authority on media distribution \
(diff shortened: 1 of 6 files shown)


code/modules/lore_codex/news_data/main.dm
```diff
@@ -112,6 +112,8 @@
 		/datum/lore/codex/page/about_news,
 		)
 
+	var/newsindex = "103" //Update with number of latest article
```
MistakeNot4892, line 115: Surely this should be set on the fly, not hardcoded.

Cerebulon: Ideally yes, but considering news articles ALSO have to be hard-coded currently I didn't see any benefit to complex code like the Game Update News system uses.
If I ever work out/want to put in the excessive effort to add an easier way to upload news articles than going through the PR process, then that would absolutely be something that would change to match.

MistakeNot4892: That was my thought. Either grab some key from the news datum or just use the length of the list.

MistakeNot4892 (approved):
I have a vague suspicion that `LAZYLEN(children)` is going to break on categories but we can burn that bridge when we get to it.

PR #8214, opened 2021-08-12
Ports Aurora-style written languages.

Cerebulon:
Based on: https://github.com/Aurorastation/Aurora.3/pull/11624 and subsequent bugfixes.
Originally ported the bay version but this one seems more flexible in its applications even if hunting down fonts and their licenses was a huge pain!
Fonts are licensed under the Open Font License which is included in their folder along with required credits.


code/modules/paperwork/paper.dm
```diff
@@ -153,13 +153,22 @@
...
+	paper_win.set_content(get_content(user, can_read(user, forceshow)))
+	paper_win.add_stylesheet("paper_languages", 'html/browser/paper_languages.css')
+	paper_win.open()
+
+/obj/item/weapon/paper/proc/can_read(var/mob/user, var/forceshow = FALSE)
+	var/can_read = ((istype(user, /mob/living/carbon/human) || istype(user, /mob/observer/dead) || istype(user, /mob/living/silicon)) || forceshow)
+	if(!forceshow && istype(user,/mob/living/silicon/ai))
+		var/mob/living/silicon/ai/AI
+		can_read = get_dist(src, AI.camera) < 2
+	return can_read
```
Atermonera, line 167:
```suggestion
	if(!forceshow && isAI(user))
		var/mob/living/silicon/ai/ai = user
		return get_dist(src, AI.camera) < 2
	return forceshow || \
		ishuman(user) || \
		isobserver(user) || \
		issilicon(user)
```

code/modules/paperwork/paper.dm
```diff
@@ -153,13 +153,22 @@
...
+
+/obj/item/weapon/paper/proc/get_content(var/mob/user, var/can_read = TRUE)
+	return "<head><title>[capitalize(name)]</title><style>body {background-color: [color];}</style></head><body>[can_read ? parse_languages(user, info) : stars(info)][stamps]</body>"
```
Atermonera, line 170: In the case where the user can't read it at all, do we really have to process the whole string? Can we not just prefabricate some generic "You cannot read this" response?

code/modules/paperwork/paper.dm
```diff
@@ -391,19 +402,65 @@
...
+ * with scrambled and properly fonted content depending on what languages the user knows.
+ */
+/obj/item/weapon/paper/proc/parse_languages(mob/user, input, language_check = FALSE)
+	// Just a safety fallback.
+	if (!user)
+		return input
+
+	var/static/regex/written_lang_regex
+	if (!written_lang_regex)
+		written_lang_regex = new(@"(\[lang=([#_a-zA-Z0-9\^]{1})])(.*?)(\[\/lang])", "g")
```
Atermonera, line 441: Surprised to see that close-brackets don't need an escape. A bit harder to read this without.

code/modules/paperwork/paper.dm
```diff
@@ -391,19 +402,65 @@
...
+		var/content = written_lang_regex.group[3]
+		var/reader_understands = user.say_understands(null, L)
+
+		// Replace the content with <p>content here</p>
+		if(!reader_understands)
+			if(language_check)
+				. = replacetext(., written_lang_regex.match, "")
 			else
-				to_chat(user, "<font color='red'>You must hold \the [P] steady to burn \the [src].</font>")
+				content = L.scramble(content)
```
Atermonera, line 459: Are either of these deterministic? That is, if someone reads a paper twice, will they get the same result? There doesn't need to be meaning, so much as repeatable results, because it doesn't make sense for someone to read the same text twice and come up with different characters.

code/modules/paperwork/paper.dm
Atermonera: Actually, how does this behave if someone forgets to terminate a tag? What if someone else later writes more, and _does_ terminate the tag?

Atermonera (commented):
I'm dubious about storing the whole text as a single string and having to parse it every time someone reads it.
Probably low-impact memory/runtime optimization, but would it be possible to apply this regex on paper-write, and store on the paper a list of text fragments on per-language basis? Would allow a far simpler deterministic encryption by only requiring a single processing step to generate what everyone sees if they can't understand it.
Under that system, if someone doesn't terminate a tag at the end, we don't need to expect a termination on further writes: better to just implicitly terminate the tag, so the next person that writes something doesn't get a surprise when they write something in Siik.

Possibly out of scope of porting this, but at least something to consider. As is I can foresee a few bits of weird behaviour, noted below.

Atermonera:
No plans for further work

PR #8215, opened 2021-08-12
Tweak some nutrition values

lbnesquik:
## About this pull request 
- Gives cream & Milk higher nutrition values 
- Make ethanol give at least 2 nutrition per unit as a failsafe for any drinks. It could be used to remove some values bloat as well.
- Indents some comments to look prettier.

Milk and cream are acceptably thick in nutrition as it should be.
Code is slightly prettier to read
This makes drinks behave a tad more realistically without much cost. 

It could still use a more thorough pass but as you can see in this example, it isn't clear why some drinks give you some, and others give you zero.
image

### DIFF
--- a/code/modules/reagents/reagents/_reagents.dm
+++ b/code/modules/reagents/reagents/_reagents.dm
@@ -6,23 +6,23 @@
 	var/id = "reagent"
 	var/description = "A non-descript chemical."
 	var/taste_description = "bitterness"
-	var/taste_mult = 1 //how this taste compares to others. Higher values means it is more noticable
-	var/nutriment_factor = 2
+	var/taste_mult = 1 		//how this taste compares to others. Higher values means it is more noticable
 	var/datum/reagents/holder = null
 	var/reagent_state = SOLID
 	var/list/data = null
 	var/volume = 0
-	var/metabolism = REM // This would be 0.2 normally
+	var/metabolism = REM 	// This would be 0.2 normally
 	var/list/filtered_organs = list()	// Organs that will slow the processing of this chemical.
 	var/mrate_static = FALSE	//If the reagent should always process at the same speed, regardless of species, make this TRUE
 	var/ingest_met = 0
 	var/touch_met = 0
 	var/dose = 0
 	var/max_dose = 0
-	var/overdose = 0		//Amount at which overdose starts
-	var/overdose_mod = 1	//Modifier to overdose damage
+	var/overdose = 0		// Amount at which overdose starts
+	var/overdose_mod = 1	// Modifier to overdose damage
 	var/can_overdose_touch = FALSE	// Can the chemical OD when processing on touch?
-	var/scannable = 0 // Shows up on health analyzers.
+	var/injectable = 0		// Is this allowed as an injection
+	var/scannable = 0 		// Shows up on health analyzers.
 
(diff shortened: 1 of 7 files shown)


lbnesquik:
I moved nutrition straight to reagents, so we don't have that SAME var defined in two places.

MistakeNot4892 (changes requested):
You'll need to move/redefine/unify some logic from `/datum/reagent/nutriment` and `/datum/reagent/ethanol` to `/datum/reagent`, specifically the adjust_nutrition calls:
````
/datum/reagent/ethanol/affect_ingest(var/mob/living/carbon/M, var/alien, var/removed)
	if(issmall(M)) removed *= 2
	if(!(M.species.allergens & allergen_type))	//assuming it doesn't cause a horrible reaction, we get the nutrition effects
		M.adjust_nutrition(nutriment_factor * removed)
		
/datum/reagent/nutriment/affect_ingest(var/mob/living/carbon/M, var/alien, var/removed)
	switch(alien)
		if(IS_DIONA) return
		if(IS_UNATHI) removed *= 0.5
	if(issmall(M)) removed *= 2 // Small bodymass, more effect from lower volume.
	if(!(M.species.allergens & allergen_type))	//assuming it doesn't cause a horrible reaction, we'll be ok!
		M.heal_organ_damage(0.5 * removed, 0)
		M.adjust_nutrition(nutriment_factor * removed)
		M.add_chemical_effect(CE_BLOODRESTORE, 4 * removed)

**Comment by lbnesquik** [comment]
I moved some things around and it should be good to go this time around.

**Comment by lbnesquik** [comment]
Was clearing branches, soz

File: code/modules/reagents/reagents/_reagents.dm [hunk 1]
```diff
@@ -186,13 +203,136 @@
...
+
+		if(halluci)
+			M.hallucination = max(M.hallucination, halluci*3)
+
+
+
 	return
 
 /datum/reagent/proc/affect_ingest(var/mob/living/carbon/M, var/alien, var/removed)
+	if(nutriment_factor)
```
MistakeNot4892, line 274: Huge amount of duplicate code here. Should really move it into a proc or just have affect_blood call affect_ingest.

lbnesquik: Well i am just moving things around. A proper brush through to see what the deal is with those variables is a bit outside the scope of this PR

MistakeNot4892: I'm not talking about the variables - you have added identical code to affect_ingest and affect_blood that could be moved into a single proc and called, reducing the amount of code you're adding/that has to be maintained.

lbnesquik: I don't know how to say it politely but i really do not feel like putting 3/4 hours into making a proper, unified proc to handle this. 
And yes, while there is duplicate code, this is not an addition, given it was all pretty much here before, if only scattered across multiple files.

MistakeNot4892: It's nearly identical code. It should take maybe ten minutes to split it out.
Someone else will need tonreview/approve if you don't want to do that as this level of copypaste is something that contributes to poor long term code quality and maintainability.

PR #8216, opened 2021-08-13
Randomized Vendor Spawners for Mapping

Cerebulon:
Generic (for PoIs and crap), Food (Snacks!) and Soda vendors so there isn't a need to map in all the specific vendors and they can randomized per round.


MistakeNot4892 (approved):

MistakeNot4892:
Only nitpick is that personally I'd format the lists like:
````
foobar = list(
    foo,
    bar
)
````
rather than the multitab.

PR #8220, opened 2021-08-16
What was supposed to be another straightforward major system overhaul that once again spiraled out of control

Atermonera:
At this point I should just get this code compiling and working, which I think mostly involves fixing _all_ the uses of toolspeed and makes surgery work on tool qualities.

- [X] Updates surgery tools to actually use tool_qualities
- [X] Makes tool_qualities numerical instead of just true/false (with #define'd presets)
- [x] Makes toolspeed determined by tool quality
- [ ] ~Makes surgery use tool_qualities instead of hardcoded types.~
- [X] Continues my eternal crusade against bad/old code.

And the best part is _most_ of these changes are trivial, but several are less so.

Followup for welding tool usage, because some things which should act like welding tools aren't welding tools: #8284 
Followup for sharp and edge tool quality types: #8285


Atermonera (commented):
Oh apparently github doesn't like it when you don't have comments and don't explicitly approve.

I made a full pass over this code (resulting in some changes to ~40 files). But would like other eyes on it.

Atermonera:
Fallout from live test:
```
You can't unwrench pipes
undefined proc or verb /turf/simulated/floor/plating/dismantle wall().
     proc name: attackby (/turf/simulated/wall/attackby)
Runtime in twohanded.dm,108,114: bad index
Runtime in frame.dm,346: Cannot read "teleporter_hub".name
Surgery is instant.
```

PR #8221, opened 2021-08-17
Decl Music and Ported Music + Licenses

Cerebulon:
- Ported a ton of music tracks from Bay/Nebula.
- Adjusted how ear protection works (Also from Nebula)
- Added headphones being able to actually play music (and providing more ear protection when doing so).
- Added ability to define per-map lobby music lists (Not currently defined so picks one at random - this is unfinished)

Putting up as a draft so I can stare and also run the unit tests.
~~Current issues:~~

~~1. On every compile, only a seemingly random selection of the music tracks will play - the rest will be totally silent. The decls appear to be initialized and the lobby cycler is able to identify their names/artists/licenses, it just isn't playing certain files. But always DIFFERENT files.~~
~~2. Still fucking with headphone UI stuff.~~
~~3. Seemingly random specific files SOMETIMES crash the client entirely. I deleted one set that caused this, and all it did was transfer that property onto different seemingly random files. This time it includes Beyond, a track which is already in the game.~~

~~As of latest push:~~
~~Added debug crap all over the place. Things appear to be firing, the decls reference the files correctly, they're just not ACTUALLY playing the .ogg files~~
~~If the 'Play different lobby track' OOC verb is used BEFORE geometry processing I am yet to have a single track fail to play which is random-picked at that stage. After geometry it tends to fail to play any tracks 'first loaded' after that point. Problem with where init happens or caching or something???~~
~~If 'play different lobby track' OOC verb is used too many times... The client hangs and the NanoUI lobby windows unload. There is also some threshold I haven't found where you won't crash immediately, but will when trying to load into the game on actual roundstart.~~
~~Removed the tracks that were just straight up crashing the client 20% of the time. This did nothing. The client still crashes on DIFFERENT songs now. Totally insane.~~

It works now.


Cerebulon:
I'm pretty stumped on this. Could somebody make a local copy and confirm that the Juke only playing a random handful of tracks isn't just an issue on my end for whatever reason?

Atermonera (approved):

PR #8225, opened 2021-08-21
Seed Contraband Changes

Schnayy:
Changes the contraband seeds to be more interesting and have probability. 6 list selections.

Happy to hear any balance concerns about the plant choices, but I thought it silly we had so many plants we never ever see. Of about 17 spawned vendors in testing, most had produce/mushrooms, a few of the cold plants and survivalist, about 3 poison, and none of the extra poison.


MistakeNot4892 (approved):
`pick()` so ugly :(

PR #8226, opened 2021-08-21
Fixes infinite naming loop

Schnayy:
Ater's requested changes to the surgery code, should fix the bug where entering a name infinitely loops and requires a relog to get rid of the prompt.

I could not test this personally because it requires two people fyi


MistakeNot4892 (approved):
Seems fine to me. Odd that it was sanitizing twice.

PR #8227, opened 2021-08-21
RIGAI Update

Mechoid:
:cl:
tweak - RIG AIs can now actually use the component toggling again.
rscadd - Adds a yet-more-cyberpunkian RIG module to allow an AI in the suit to take control of a body.
/:cl:

RIG outline:
- Allow Suit-AI to take control of an individual wearing their RIG, if AI control is enabled.
- AI cannot speak if the pilot is not at least partially mechanical. Neural implant, MMI, or robotic brains are required. If this check fails, the control override will also cause brain-damage.
- RIG AI is force-reverted to their module if a borer makes its home in the pilot. No brain worm mind swaps, sorry.
- RIG AI may attempt to control a corpse pilot. If the pilot body dies while controlled, or is controlled after death, the suit will attempt to defibrillate them, adding a unique modifier denoting corpsehood. If the AI leaves control, even after the body received medical aid, the body will return to being a corpse. This defibrillation can cause the suit to lose power rather quickly, if attempted on an invalid corpse.
- RIG AI control may be negated before enaction by disabling AI control on the suit.
- RIG AIs negate oxyloss, aside from the extremes that would immediately cause death, such as incredible bloodloss or heart removal. This just outright kills them before the cycle.


code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
```diff
@@ -15,7 +15,9 @@
 		to_chat(usr, "Your module is not installed in a hardsuit.")
 		return
 
-	module.holder.tgui_interact(usr, custom_state = GLOB.tgui_contained_state)
+	var/obj/item/weapon/rig/RIG = module.holder
+	var/mob/living/L = usr
+	RIG.tgui_interact(L, custom_state = GLOB.tgui_contained_state)
```
MistakeNot4892, line 20: What's the purpose of casting the mob and holder?

code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
```diff
@@ -54,14 +57,16 @@
 			SetupStat(rig)
 
 /obj/item/rig_module/ai_container/proc/update_verb_holder()
-	if(!verb_holder)
-		verb_holder = new(src)
+	if(!istype(verb_holder))
```
MistakeNot4892, line 60: `if(ispath(verb_holder))` would likely make more sense here since you're newing it immediately afterwards.

code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
```diff
@@ -129,12 +134,16 @@
 	if(!..())
 		return 0
 
+	if(ai_locked)
+		to_chat(usr, "\The [src] gives an error code response.")
```
MistakeNot4892, line 138: SPAN_WARNING

MistakeNot4892: Several other naked warning messages below as well.

code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
```diff
@@ -213,3 +228,237 @@
 		to_chat(user, "<span class='warning'>There is no active AI within \the [ai].</span>")
 	update_verb_holder()
 	return
+
+/*
+ * The realm of cyberpunk, a Neural Jack-enabled IIS.
+ * Components unashamedly utilizing the groundwork of borers.
```
MistakeNot4892, line 234: Oh god. This got an actual out loud despairing 'Noooo.' out of me.
Borers are a fractal nightmare of barely functional code and problematic hacks. Ominous.

code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
```diff
@@ -213,3 +228,237 @@
...
+
+	var/obj/item/rig_module/ai_container/advanced/module = usr.loc.loc
+	if(!module.holder)
+		to_chat(usr, "Your module is not installed in a hardsuit.")
+		return
+
+	module.jack_brain()
+
+/obj/item/rig_module/ai_container/advanced
+	name = "AINI module"
```
MistakeNot4892, line 253: `\improper`

code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
```diff
@@ -213,3 +228,237 @@
...
+ * The realm of cyberpunk, a Neural Jack-enabled IIS.
+ * Components unashamedly utilizing the groundwork of borers.
+ */
+/obj/item/ai_verbs/advanced/verb/assume_control()
+	set category = "Hardsuit"
+	set name = "Enable Direct Control"
+	set src in usr
+
+	if(!usr.loc || !usr.loc.loc || !istype(usr.loc.loc, /obj/item/rig_module))
+		to_chat(usr, "You are not loaded into a hardsuit.")
```
MistakeNot4892, line 242: SPAN_WARNING on these too.

code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
```diff
@@ -213,3 +228,237 @@
...
+			to_chat(integrated_ai, "<span class='danger'>The neural jack signals a warning abruptly, before rapidly retracting.</span>")
+			to_chat(H, "<span class='warning'>You are shoved into consciousness as though you have been dropped into freezing water.</span>")
+			return
+
+		var/obj/item/organ/internal/brain/Brain = H.internal_organs_by_name[O_BRAIN]
+		if(Brain)
+			if(Brain.robotic < ORGAN_ASSISTED)
+				H.add_modifier(/datum/modifier/mute, 30 SECONDS, suppress_failure = TRUE)
+
+		if(H.stat == DEAD && world.time > last_defib + 1 MINUTE && (H.getBruteLoss() + H.getFireLoss()) < (H.getMaxHealth() * 2 - H.getCloneLoss() * 2))
```
MistakeNot4892, line 296: This is a lot of boilerplate. Can't you call out to the defib to apply all this somehow rather than reimplementing?

code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
```diff
@@ -213,3 +228,237 @@
...
+				imperfect_connection = TRUE
+
+			else
+				to_chat(H, "<span class='warning'>You feel a strange shifting sensation behind your eyes as a consciousness displaces yours.</span>")
+
+			if(H.stat == DEAD)
+				corpse_pilot = TRUE
+
+			// pilot -> brain
+			var/h2b_id = H.computer_id
```
MistakeNot4892, line 376: This code and the brain transfer code in general should be generalized into a proc, maybe on the captive_brain, rather than copypasted.

code/modules/clothing/spacesuits/rig/rig_tgui.dm
```diff
@@ -138,7 +140,7 @@
 			locked = !locked
 			. = TRUE
 		if("toggle_piece")
-			if(ishuman(usr) && (usr.stat || usr.stunned || usr.lying))
+			if(/*ishuman(usr) && */(usr.stat || usr.stunned || usr.lying))
```
MistakeNot4892, line 143: Why leave this inline?

MistakeNot4892 (changes requested):
Some of the individual components of this PR could be split out and merged easily enough (tweaks to the rig AI and the topic stuff for example) but as a whole this is a bit spaghetti. Lots of copypasted boilerplate that needs tidying.

Also not completely sure it's good to be adding another borer-like control mechanic given how unpopular borers were/are, but that's more subjective.

code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
Mechoid: Readability was the primary purpose.

code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
Mechoid: My eyes have been opened to a whole new feature I didn't know we have. The tedium of manual SPANs is at an end.

code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
Mechoid: This is the least-boiler of the methods I found. It needed to circumnavigate the normal checks the DEFIB does, and this method allows the proper associated power-draw.

code/modules/clothing/spacesuits/rig/rig_tgui.dm
Mechoid: A consideration of downstream.

code/game/objects/items/devices/defib.dm
```diff
@@ -574,6 +574,31 @@
...
+	icon_state = "defibpaddles0"
+	item_state = "defibpaddles0"
+	cooldowntime = (3 SECONDS)
+
+/obj/item/weapon/shockpaddles/rig/check_charge(var/charge_amt)
+	if(istype(loc, /obj/item/rig_module))
+		var/obj/item/rig_module/RM = src.loc
+		var/obj/item/weapon/rig/Rig = RM.holder
+		if(!Rig)
+			return FALSE
```
MistakeNot4892, line 591: Would suggest moving the various rig casting and checking into a get_cell() override then calling that in both check_charge and checked_use, rather than duplicating code.

code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
Atermonera: What specific checks by the defib need to be avoided? Can you refactor these checks to a separate proc on the defib?

Mechoid: The standard attackby chain.

code/game/objects/items/devices/defib.dm
```diff
@@ -574,6 +577,33 @@
...
+		var/obj/item/rig_module/RM = src.loc
+		if(istype(RM))
+			return RM.get_cell()
+
+	return FALSE
+
+/obj/item/weapon/shockpaddles/rig/check_charge(var/charge_amt)
+	var/obj/item/weapon/cell/C = get_cell()
+
+	return (C && C.check_charge(charge_amt))
```
MistakeNot4892, line 600: Minor, but you could do `C?.check_charge(charge_amt)` here and below.

code/modules/clothing/spacesuits/rig/modules/modules.dm
```diff
@@ -108,6 +108,11 @@
 		return
 	..()
 
+/obj/item/rig_module/get_cell()
+	if(holder)
```
MistakeNot4892, line 112: Similar to above, could do `return holder?.get_cell()`

code/modules/mob/living/say.dm
```diff
@@ -94,19 +94,58 @@ var/list/channel_to_radio_key = new
...
 
-		if((HULK in mutations) && health >= 25 && length(S.message))
+		var/force_slurring = 0
+		var/force_shouting = 0
+		var/force_stuttering = 0
+		var/force_mute = 0
+
+		for(var/datum/modifier/M in modifiers)
+			if(M.speech_slurring)
+				force_slurring += M.speech_slurring
```
Atermonera, line 104: So if I have two modifiers that want to make me _not_ slur, but only one that _does_, I won't slur? Is that the pattern we want to go for here?

code/modules/mob/living/say.dm
```diff
@@ -94,19 +94,58 @@ var/list/channel_to_radio_key = new
...
+				force_stuttering += M.speech_stuttering
+			if(M.speech_shouting)
+				force_shouting += M.speech_shouting
+			if(M.speech_mute)
+				force_mute += M.speech_mute
+
+		if(force_slurring && force_slurring > 0)
+			force_slurring = TRUE
+		else
+			force_slurring = FALSE
```
Atermonera, line 115:
```
if(force_slurring < 1)
    force_slurring = FALSE
```
(And so on and so forth. If it's greater than 1 it's true, and that's the paradigm this uses)

code/modules/mob/mob_helpers.dm
```diff
@@ -685,3 +685,25 @@ var/global/image/backplane
...
+ * 
+ * Returns FALSE if it is unable to transfer the player.
+ * allow_overwrite determines if it will fail when targeting a mob with a ckey.
+ */
+/mob/proc/transfer_player(var/mob/target = null, var/allow_overwrite = FALSE)
+	if(!target)
+		return FALSE
+
+	if(target.ckey && !allow_overwrite)
+		return FALSE
```
Atermonera, line 700: So we can't do it even if they logged out?

code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
```diff
@@ -199,17 +212,200 @@
...
+	desc = "An artificial intelligence neural integration module, suitable for most hardsuits."
+	description_fluff = "The unassuming device has no exterior markings, save for a network usage monitor and tiny warning label. On closer inspection, it reads, \"Warning: Device may cause damage if used without proper neural interface firmware.\" Something in your core urges you to heed that warning."
+	description_info = "Unless used by a robotic RIG pilot, or an organic one with a neural implant, the module will cause brain-damage on activation of its override mode."
+	icon_state = "AIIS"
+	var/controlling = FALSE
+
+	verb_holder = /obj/item/ai_verbs/advanced
+
+	var/last_defib = 1
+	var/mob/living/captive_brain/pilot_brain
```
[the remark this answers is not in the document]
Mechoid: I assume yes, as while it provides the same features as the baseline, it _is_ different enough to warrant, I believe.

code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
```diff
@@ -199,17 +212,200 @@
...
+			if(corpse_pilot)
+				H.setOxyLoss(0)
+				H.add_modifier(/datum/modifier/risen_corpse, 30 SECONDS, suppress_failure = TRUE)
+				last_defib = world.time
+			if(H.sleeping)
+				H.AdjustSleeping(-10)
+
+		if(H.has_modifier_of_type(/datum/modifier/risen_corpse))
+			H.setToxLoss(min(H.getToxLoss(), 30))
+			H.setOxyLoss(min(H.getOxyLoss(), 20))
```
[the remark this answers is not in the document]
Mechoid: Toxloss, yes, oxyloss, no. Fixing.

code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
```diff
@@ -199,17 +212,200 @@
...
+
+				paddles.busy = TRUE
+				paddles.do_revive(H, integrated_ai)
+				paddles.busy = FALSE
+				if(H.stat != DEAD)
+					corpse_pilot = TRUE
+
+		if(H.stat != DEAD)
+			if(corpse_pilot)
+				H.setOxyLoss(0)
```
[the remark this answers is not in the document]
Mechoid: Oxyloss is cleared as muscles do not require oxygen to operate, however they generate damaging waste. Accounting for it now.
FBPs gain toxloss in accordance with a damaged system.

code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
```diff
@@ -199,17 +212,200 @@
...
+			H.setToxLoss(0)
+
+			if(paddles && !paddles.busy)
+				if(H.isSynthetic())
+					paddles.use_on_synthetic = TRUE
+				else
+					paddles.use_on_synthetic = FALSE
+
+				paddles.busy = TRUE
+				paddles.do_revive(H, integrated_ai)
```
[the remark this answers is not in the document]
Mechoid: Messages are intentional, yes.

code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
```diff
@@ -199,17 +212,200 @@
...
+				H.add_modifier(/datum/modifier/mute, 30 SECONDS, suppress_failure = TRUE)
+
+		if(H.stat == DEAD && world.time > last_defib + 1 MINUTE && (H.getBruteLoss() + H.getFireLoss()) < (H.getMaxHealth() * 2 - H.getCloneLoss() * 2))
+			H.setToxLoss(0)
+
+			if(paddles && !paddles.busy)
+				if(H.isSynthetic())
+					paddles.use_on_synthetic = TRUE
+				else
+					paddles.use_on_synthetic = FALSE
```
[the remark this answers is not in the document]
Mechoid: Done.

code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
```diff
@@ -199,17 +212,200 @@
...
+			to_chat(H, SPAN_WARNING("You are shoved into consciousness as though you have been dropped into freezing water."))
+			return
+
+		var/obj/item/organ/internal/brain/Brain = H.internal_organs_by_name[O_BRAIN]
+		if(Brain)
+			if(Brain.robotic < ORGAN_ASSISTED)
+				H.add_modifier(/datum/modifier/mute, 30 SECONDS, suppress_failure = TRUE)
+
+		if(H.stat == DEAD && world.time > last_defib + 1 MINUTE && (H.getBruteLoss() + H.getFireLoss()) < (H.getMaxHealth() * 2 - H.getCloneLoss() * 2))
+			H.setToxLoss(0)
```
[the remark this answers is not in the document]
Mechoid: Resetting toxloss as a way to provide a means for a critical body to be raised, even if only temporarily, as tweaks are making it possible to build up again.

code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
Mechoid: Math oofs and, maxhealth specifically should be doubled as humanoids go into the negative for their health calculation.

code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
```diff
@@ -199,17 +212,200 @@
...
+/obj/item/rig_module/ai_container/advanced/removed()
+	revert()
+	..()
+
+/obj/item/rig_module/ai_container/advanced/process()
+	if((holder.offline || !holder.ai_override_enabled) && controlling)
+		revert()
+
+	if(controlling)
+		var/mob/living/carbon/human/H = holder.wearer
```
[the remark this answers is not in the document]
Mechoid: Accounting now

code/modules/mob/mob_helpers.dm
Mechoid: I don't know if there's a way of _determining_ that.

code/modules/mob/living/say.dm
Mechoid: I can't quite imagine another way to do it that wouldn't be infuriating (influencing RNG)

code/modules/mob/mob_helpers.dm
Atermonera: `target.client`, I think?

MistakeNot4892: `target.client` is null and `target.ckey` is not if they have logged out.

code/game/objects/items/devices/defib.dm
```diff
@@ -574,6 +577,45 @@
...
+		return "buzzes, \"Incompatible physiology. Operation aborted.\""
+
+	if(H.stat != DEAD)
+		return "buzzes, \"Patient is not in a valid state. Operation aborted.\""
+
+	if(!check_contact(H))
+		return "buzzes, \"Patient's chest is obstructed. Operation aborted.\""
+
+	return null
+
```
Atermonera, line 618:
```suggestion

/obj/item/weapon/shockpaddles/rig/universal/can_use(mob/user, mob/M)
	if(busy)
		return FALSE
	if(!check_charge(chargecost))
		to_chat(user, "<span class='warning'>\The [src] doesn't have enough charge left to do that.</span>")
		return FALSE
	if(cooldown)
		to_chat(user, "<span class='warning'>\The [src] are re-energizing!</span>")
		return FALSE
	return TRUE
```
Not actually sure how user might be populated here, so the to_chat()s may end up being meaningless

code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
```diff
@@ -175,41 +184,214 @@
...
+				to_chat(integrated_ai, SPAN_DANGER("<span class='danger'>The neural jack signals a warning abruptly, before rapidly retracting.</span>"))
+				to_chat(H, SPAN_WARNING("You are shoved into consciousness as though you have been dropped into freezing water."))
+				return
+
+			var/obj/item/organ/internal/brain/Brain = H.internal_organs_by_name[O_BRAIN]
+			if(Brain)
+				if(Brain.robotic < ORGAN_ASSISTED)
+					H.add_modifier(/datum/modifier/mute, 30 SECONDS, suppress_failure = TRUE)
+
+			if(H.stat == DEAD && world.time > last_defib + 1 MINUTE && (H.getBruteLoss() + H.getFireLoss()) < (H.getMaxHealth() * 2 - H.getCloneLoss()))
```
Atermonera, line 290:
```suggestion
			if(H.stat == DEAD && !paddles.cooldown && !paddles.can_revive(H))
```
That... should be sufficient to handle this check?
`paddles.can_revive()` returns an error string on failure, null on success, so we check that it's falsey.
You might need a little bit more logic to handle the setToxLoss() bit below, but that feels... extremely strong in the edge case where the user died entirely to toxloss and it just blows it all away.

code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
```diff
@@ -175,41 +184,214 @@
...
+				if(H.stat != DEAD)
+					corpse_pilot = TRUE
+
+			if(H.stat != DEAD)
+				if(corpse_pilot)
+					H.setOxyLoss(0)
+					H.add_modifier(/datum/modifier/risen_corpse, 30 SECONDS, suppress_failure = TRUE)
+					last_defib = world.time
+				if(H.sleeping)
+					H.AdjustSleeping(-10)
```
Atermonera, line 302:
```suggestion
				if(H.stat != DEAD)
					H.setOxyLoss(0)
					H.add_modifier(/datum/modifier/risen_corpse, 30 SECONDS, suppress_failure = TRUE)
				else // They're still dead, defib failed: no point trying this again
					controlling = FALSE
			if(H.stat != DEAD && H.sleeping)
				H.adjustSleeping(-10)
```

code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
```diff
@@ -175,41 +184,214 @@
...
+
+			if(H.has_modifier_of_type(/datum/modifier/risen_corpse))
+				H.adjustToxLoss(0.1)
```
Atermonera, line 305: Shouldn't this just be handled by the modifier itself?

code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
```diff
@@ -175,41 +184,214 @@
...
+				to_chat(H, SPAN_WARNING("You feel a strange shifting sensation behind your eyes as a consciousness displaces yours."))
+
+			if(H.stat == DEAD)
+				corpse_pilot = TRUE
+
+			// pilot -> brain
+			qdel(pilot_brain)
+			pilot_brain = new(src)
+
+			H.transfer_player(pilot_brain)
```
Atermonera, line 363: What happens if the wearer is dead and ghosted? Do they get stuffed into the holder anyways?

code/game/objects/items/devices/defib.dm
Atermonera: If the user being an... AI? pAI? would actually pass the parent type, then feel free to ignore this

code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
Mechoid: This was left to the RIG, in the event we acquire other means of creating "risen corpses", as zombies are, universally, probably good at taking hits, but they are not all going to need to auto-degrade.

code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
Mechoid: The defib does not occur until after the puppeteer is in the body, and corpse_pilot is a necessary var to ensure that a risen corpse properly _returns_ to corpsehood after the puppeteer lets go of the strings.

code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
Mechoid: Correct. Though I assume most would stay in their body if they knew they had the module installed.

code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
Mechoid: Toxloss in robots is purely from rads and tick damage from organs, which would still be present, in addition to the presence of the damage caused by the risen_corpse check.

Toxloss in meats is theoretically associated with organ failure and blood toxicity, which are both things generally disregarded by forcing the muscles to function.

In either case, the cause of damage will still probably be present in the body to cause expiry again, such as chems, rads, or damaged/rotting organs. At the point this is _strong_ is the case where, feasibly, the short moments the puppeteer gets to repair the body (if they can) or move them to safety are enough to maintain function until help arrives, or get somewhere others _can_ help, respectively.

code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
Atermonera: So... a subtype of the modifier that does handle degeneration? This feels really weird being on the rig.

code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
Atermonera: So the risen corpse modifier only lasts for 30s, but the AI can control the mob for longer than that? This feels like a _really_ clunky way of handling whether the mob should fall back over or not, using both a modifier _and_ a variable on this module.

What would happen if the module or the rig were removed from the wearer (by any mechanism) while this is active?

code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
```diff
@@ -175,41 +184,210 @@
...
+
+		if(istype(H))
+			controlling = FALSE
+			H.verbs -= /mob/living/carbon/proc/release_control
+
+			H.transfer_player(integrated_ai)
+
+			to_chat(pilot_brain, "<span class='notice'>Your mind returns from a strange limbo, the sensations of the body returning.[imperfect_connection ? " Your voice returns shakily." : ""][corpse_pilot ? " <span class='warning'>And then everything comes crashing down.</span>" : ""]</span>")
+
+			pilot_brain.transfer_player(H)
```
Atermonera, line 382: What happens if the player ghosts out of the pilot brain?

code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
Atermonera: Please anwer the previous question: What checks in the standard attackby chain are you specifically trying to avoid here?

code/modules/clothing/spacesuits/rig/modules/specific/ai_container.dm
Mechoid: The pilot brain is the same one the borer uses, so it would be the same fate. They can resist to return to their body, and given this is unlikely to be used on unwilling individuals due to how easy it is , they probably won't do so out of frustration.

Atermonera:
If you still intend to work on this, please make a new PR so that I can start tracking discussion with it again.

PR #8233, opened 2021-08-22
Ports sliding things along tables

Cerebulon:
Click dragging an object which is already on a table to another section of the table that you can reach visibly slides the object along the table.
https://github.com/Aurorastation/Aurora.3/pull/11778


MistakeNot4892 (approved):
Bit iffy on some of the logical structure here but code looks largely fine.

PR #8234, opened 2021-08-23
Adds Qipaos

Cerebulon:
Technically the same thing as Cheongsam but these are different because uh... Free colour selection I guess.


MistakeNot4892 (approved):
Code looks fine.

PR #8244, opened 2021-08-23
Ports Pixel Shifting

Cerebulon:
CTRL + SHIFT + Movement Keys to scoot around juuuuuust a little.
https://github.com/VOREStation/VOREStation/pull/9265 + various fixes

:cl:
rscadd - Added pixel shift movement. Ctrl+shift+movement key to move around within a single tile.
/:cl:

### DIFF
--- a/code/modules/mob/living/living_movement.dm
+++ b/code/modules/mob/living/living_movement.dm
@@ -210,6 +210,10 @@ default behaviour is:
 /mob/living/Moved(var/atom/oldloc, direct, forced, movetime)
 	. = ..()
 	handle_footstep(loc)
+	if(is_shifted)
+		is_shifted = FALSE
+		pixel_x = default_pixel_x
+		pixel_y = default_pixel_y
 
 	if(pulling) // we were pulling a thing and didn't lose it during our move.
 		var/pull_dir = get_dir(src, pulling)
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -1032,7 +1032,7 @@
 		pixel_x--
 		is_shifted = TRUE
 
-mob/verb/shifteast()
+/mob/verb/shifteast()
 	set hidden = TRUE
 	if(!canface())
 		return FALSE
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -671,10 +671,9 @@
 
 // facing verbs
 /mob/proc/canface()
-	if(!canmove)						return 0
 	if(stat)							return 0
 	if(anchored)						return 0
-	if(transforming)						return 0
+	if(transforming)					return 0
 	return 1
 
 // Not sure what to call this. Used to check if humans are wearing an AI-controlled exosuit and hence don't need to fall over yet.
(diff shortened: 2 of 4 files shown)


zekesulastin:
Aside from the CI failure, it also looks like you didn’t add CTRL+SHIFT+W/A/S/D for hotkey mode :(

Cerebulon:
> 
> 
> Aside from the CI failure, it also looks like you didn’t add CTRL+SHIFT+W/A/S/D for hotkey mode :(

NORTH, SOUTH, EAST and WEST should reference the actual movement keys afaik?

zekesulastin:
> 
> 
> > Aside from the CI failure, it also looks like you didn’t add CTRL+SHIFT+W/A/S/D for hotkey mode :(
> 
> NORTH, SOUTH, EAST and WEST should reference the actual movement keys afaik?

Nope - they are defined further down in the alphabetical section of the file

MistakeNot4892:
Personally I really don't like this feature. It allows you to mess with offsets and interrupt animations for the sake of the minor RP benefit of being to pose yourself. Positioning on turfs can already be a bit confusing and I think this has the potential to make it worse.

Not blocking or anything, but won't be approving this personally. Sort of an abstaining review I guess?

Atermonera:
This feels like it will introduce a lot of unintuitive behaviour where animations are interrupted or offsets reset unexpectedly. It might introduce weird behaviour in combat from projectile weapon targeting?

code/modules/mob/mob.dm
```diff
@@ -1009,6 +1008,38 @@
 	set hidden = 1
 	set_face_dir(client.client_dir(WEST))
 
+/mob/verb/shiftnorth()
+	set hidden = TRUE
+	if(!canface())
+		return FALSE
+	if(pixel_y <= (default_pixel_y + 16))
+		pixel_y++
+		is_shifted = TRUE
```
Atermonera, line 1017:
```suggestion
	if(custom_pixel_y < 16)
		pixel_y++
		custom_pixel_y++
```
And so on and so forth.

code/modules/mob/living/living_movement.dm
```diff
@@ -210,6 +210,10 @@ default behaviour is:
 /mob/living/Moved(var/atom/oldloc, direct, forced, movetime)
 	. = ..()
 	handle_footstep(loc)
+	if(is_shifted)
+		is_shifted = FALSE
+		pixel_x = default_pixel_x
+		pixel_y = default_pixel_y
```
Atermonera, line 216: By tracking custom offsets independently, we just subtract pixel_x/y by those values and set the tracking vars to 0
```suggestion
if(custom_pixel_x)
	pixel_x -= custom_pixel_x
	custom_pixel_x = 0

if(custom_pixel_y)
	pixel_y -= custom_pixel_y
	custom_pixel_y = 0
```

code/modules/mob/mob.dm
Atermonera: Also, I'm guessing hotkey macros are why there's 4 separate procs here, and not one that just takes in a DIR?

Atermonera (changes requested):

Atermonera:
Stale, no plans to work on it due to opposition

Spookerton:
@Atermonera reopen intended?

Spookerton:
I will assume not, given a month.

As far as I'm concerned the opposition to this is entirely correct. Being able to do some little pixel shifting to the detriment of numerous other real mechanics is not a good trade.

PR #8245, opened 2021-08-23
New dirt and paved sidewalk sprites

Cerebulon:
Much nicer dirt (and sidewalk) sprites from https://github.com/VOREStation/VOREStation/pull/10476 other elements of which may be ported separately. Colour corrected for sif-blue plants.


code/game/turfs/simulated/outdoors/dirt.dm
```diff
@@ -6,3 +6,107 @@
...
+		'sound/effects/footstep/asteroid3.ogg',
+		'sound/effects/footstep/asteroid4.ogg',
+		'sound/effects/footstep/asteroid5.ogg',
+		'sound/effects/footstep/MedDirt1.ogg',
+		'sound/effects/footstep/MedDirt2.ogg',
+		'sound/effects/footstep/MedDirt3.ogg',
+		'sound/effects/footstep/MedDirt4.ogg'))
+
+/turf/simulated/floor/outdoors/newdirt/Initialize(mapload)
+	var/possibledirts = list(
```
MistakeNot4892, line 36: Please make this a global var outside of Initialize(), or a static var on the turf Declaring it and discarding it every time one of these turfs init is expensive. Same for the other instances below.

code/game/turfs/simulated/outdoors/sidewalk.dm
```diff
@@ -0,0 +1,33 @@
...
+	icon = 'icons/turf/outdoors.dmi'
+	icon_base = "sidewalk"
+	footstep_sounds = list("human" = list(
+		'sound/effects/footstep/LightStone1.ogg',
+		'sound/effects/footstep/LightStone2.ogg',
+		'sound/effects/footstep/LightStone3.ogg',
+		'sound/effects/footstep/LightStone4.ogg',))
+
+/turf/simulated/floor/outdoors/sidewalk/Initialize(mapload)
+	var/possibledirts = list(
```
MistakeNot4892, line 13: You missed one!

Cerebulon: turns out the entire contents of sidewalk.dm was duplicated in dirt.dm for no reason so I didn't change that version.

code/game/turfs/simulated/outdoors/dirt.dm
```diff
@@ -6,3 +6,64 @@
...
+		'sound/effects/footstep/asteroid3.ogg',
+		'sound/effects/footstep/asteroid4.ogg',
+		'sound/effects/footstep/asteroid5.ogg',
+		'sound/effects/footstep/MedDirt1.ogg',
+		'sound/effects/footstep/MedDirt2.ogg',
+		'sound/effects/footstep/MedDirt3.ogg',
+		'sound/effects/footstep/MedDirt4.ogg'))
+
+/turf/simulated/floor/outdoors/newdirt/Initialize(mapload)
+	flooring_override = pickweight(possibledirts)
```
MistakeNot4892, line 48: Is this missing an `initial(icon_state)`?

code/game/turfs/simulated/outdoors/dirt.dm
```diff
@@ -6,3 +6,64 @@
...
+	var/static/possibledirts = list(
+		"dirt0" = 200,
+		"dirt3" = 20,
+		"dirt4" = 3,
+		"dirt5" = 3,
+		"dirt6" = 1
+	)
+
+/turf/simulated/floor/outdoors/newdirt_nograss/Initialize(mapload)
+	flooring_override = pickweight(possibledirts)
```
MistakeNot4892, line 68: As above.

code/game/turfs/simulated/outdoors/dirt.dm
Cerebulon: The dirts don't need it because they don't have subtypes that use anything other than the base set of icons. Sidewalks needed it because they use completely different sets for the sideways subtype.

PR #8247, opened 2021-08-24
Throw Mode QoL Tweaks

Cerebulon:
Requested combo of https://github.com/Aurorastation/Aurora.3/pull/12094 and https://github.com/Aurorastation/Aurora.3/pull/12161

:cl:
rscadd - Clicking on an adjacent person while on help intent with throw mode on will 'Give' them the item. If both parties have help intent throw mode on the item will automatically be accepted.
tweak - Situations where clicking in throw mode would not actually throw an item will no longer cancel throw mode.
/:cl:

Also give code is now on /living instead of /living/carbon/human for maximum giving.

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -839,9 +839,9 @@
 
 	if(lying)
 		density = 0
-		if(l_hand) 
+		if(l_hand)
 			unEquip(l_hand)
-		if(r_hand) 
+		if(r_hand)
 			unEquip(r_hand)
 		for(var/obj/item/weapon/holder/holder in get_mob_riding_slots())
 			unEquip(holder)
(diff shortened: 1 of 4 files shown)


code/modules/mob/living/living.dm
```diff
@@ -1011,10 +1010,29 @@
...
+				do_give(H)
+			return TRUE
+		make_item_drop_sound(I)
+		I.forceMove(get_turf(target))
+		return TRUE
 
-	src.drop_from_inventory(item)
-	if(!item || !isturf(item.loc))
-		return
+	drop_from_inventory(item)
```
MistakeNot4892, line 1035: Need to check `!item || QDELETED(item)` after this for stuff like ninja sword that deletes itself on drop.

MistakeNot4892 (approved):
Think this will be fine to approve after the qdel check is added.

MistakeNot4892 (changes requested):
Hit the wrong button yesterday apparently.

MistakeNot4892 (approved):

PR #8250, opened 2021-08-26
Adds unique projectile hit messages to simplemobs

Cerebulon:
No more shooting the giant spider in the left hand.
Based on https://github.com/Aurorastation/Aurora.3/pull/10125

:cl:
rscadd - Added unique hit messages for shooting mobs with non-human anatomy.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_mob/simple_mob.dm
+++ b/code/modules/mob/living/simple_mob/simple_mob.dm
@@ -173,6 +173,9 @@
 	if(has_eye_glow)
 		add_eyes()
 
+	if(organ_names)
+		organ_names = GET_DECL(organ_names)
+
 	return ..()
 
 /mob/living/simple_mob/Destroy()
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -670,7 +670,7 @@
 	var/impacted_organ = parse_zone(def_zone)
 	if(istype(target_mob, /mob/living/simple_mob))
 		var/mob/living/simple_mob/SM = target_mob
-		var/decl/mob_organ_names/organ_plan = GET_DECL(SM.organ_names)
+		var/decl/mob_organ_names/organ_plan = SM.organ_names
 		impacted_organ = pick(organ_plan.hit_zones)
 
 	//hit messages


MistakeNot4892:
Not sure I'm a fan of having tons of non-static compile time lists on mobs from a memory and init time standpoint, especially given the comments from downstream regarding map performance with large numbers of mobs. You could possibly use JSON strings instead of lists but that is going to need you to either cache or deserialize on the fly. Alternatively make it a proc override that returns the lists on demand, but that has other performance/memory impacts.

Will have a think.

Atermonera:
Could use a decl (or extension-like object with a per-type static list) to centralize the lists?

MistakeNot4892:
That might be a good way to do it.
````
/decl/bodyplan
  var/list/hit_zones = list(foo)
  
/mob/living/simple_mob
  var/organ_names = /decl/bodyplan
  
...

  var/decl/bodyplan/organ_plan = GET_DECL(organ_names)
  pick(organ_plan.hit_zones)
````

**Review by Atermonera — CHANGES_REQUESTED** [review]
Please centralize these lists per-type. You probably can't make this static, because it might be shared by literally everything, but as discussed you most like can use decls or similar to create a single instance of the list on a roughly per-mob-type basis. There's not much reason two different spiders should have two separate lists of what limbs can be shot at.

**Comment by Cerebulon** [comment]
Lists unlisted. Also meant I could just point some shit at existing decls instead of having two identical crab-part lists under separate subtypes for example.

File: code/modules/mob/living/simple_mob/simple_mob.dm [hunk 1]
```diff
@@ -59,8 +60,7 @@
 	var/harm_intent_damage = 3		// How much an unarmed harm click does to this mob.
 	var/list/loot_list = list()		// The list of lootable objects to drop, with "/path = prob%" structure
 	var/obj/item/weapon/card/id/myid// An ID card if they have one to give them access to stuff.
-	var/list/organ_names = list("body") //Bodyparts that can be shown as hit by projectiles.
-
+	var organ_names = /decl/mob_organ_names //'False' bodyparts that can be shown as hit by projectiles in place of the default humanoid bodyplan.
```
MistakeNot4892: Missing `/`

Cerebulon: Not sure how you managed to comment this in the like 5 minutes before I put up the last push that fixed this!

MistakeNot4892: Maint status is a pathway to abilities many would call... unnatural.

code/modules/projectiles/projectile.dm
```diff
@@ -667,17 +667,23 @@
 			playsound(target_mob, "bullet_miss", 75, 1)
 		return FALSE
 
+	var/impacted_organ = parse_zone(def_zone)
+	if(istype(target_mob, /mob/living/simple_mob))
+		var/mob/living/simple_mob/SM = target_mob
+		var/decl/mob_organ_names/organ_plan = GET_DECL(SM.organ_names)
```
Atermonera, line 673: Should the mob not just replace the path with this reference on initialize? That is, every time you want to ask what's in it, you don't have to fetch it out of the decl repo, thus saving some small overhead?

code/modules/projectiles/projectile.dm
```diff
@@ -670,7 +670,7 @@
 	var/impacted_organ = parse_zone(def_zone)
 	if(istype(target_mob, /mob/living/simple_mob))
 		var/mob/living/simple_mob/SM = target_mob
-		var/decl/mob_organ_names/organ_plan = GET_DECL(SM.organ_names)
+		var/decl/mob_organ_names/organ_plan = SM.organ_names
```
MistakeNot4892, line 673: You shouldn't need to cast it here since the base var is typed appropriately.

PR #8251, opened 2021-08-27
Gives the full toolbelt a multitool

lbnesquik:
## About this pull request
It adds a multitool to the full belts. The one that roboticists and engineers start with.

Having to search for a multitool is a silly little bother when it is a perfectly essential tool. Especially for engineers. Given you end up just getting one anyway, it makes no sense that a wrench is the default gear, but not a multitool.


MistakeNot4892 (approved):
Code's fine.

Schnayy:
Discussion at the staff meeting agreed that the roboticist and engineers starting with multitools is a QoL feature we would like. However, adding them to the regular toolbelts that spawn in tool storage makes them too common and was not wanted. 

Requested change is creating a subtype of the full toolbelt that includes a multitool and having roboticists & engineers start with those.

PR #8252, opened 2021-08-27
Expands the material values of certain items.

lbnesquik:
## About this pull request
This adds some default values to a few extra objects for materials.
This mostly comes into play regarding autolathes and recycling gear.
This also adds a default value of 1 metal to every item which you can reasonably expect to be holding.

What can be recycled or not along with the exact yields of which are really, really not consistent at all. 
But because it would take ages to go through and make a comprehensive update of material costs, i simply added a few to the more outlandish ones, and given the base tiny value of one to everything.

### DIFF
--- a/code/game/objects/items/apc_frame.dm
+++ b/code/game/objects/items/apc_frame.dm
@@ -6,6 +6,7 @@
 	icon = 'icons/obj/apc_repair.dmi'
 	icon_state = "apc_frame"
 	refund_amt = 2
+	matter = list(DEFAULT_WALL_MATERIAL = 100,"glass" = 30)
 
 /obj/item/frame/apc/try_build(turf/on_wall, mob/user as mob)
 	if (get_dist(on_wall, user)>1)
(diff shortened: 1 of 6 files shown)


lbnesquik:
Screwdrivers already have a metal cost so i doubt that.

code/game/objects/items/apc_frame.dm
```diff
@@ -6,6 +6,7 @@
 	icon = 'icons/obj/apc_repair.dmi'
 	icon_state = "apc_frame"
 	refund_amt = 2
+	matter = list(DEFAULT_WALL_MATERIAL = 100,"glass" = 30)
```
MistakeNot4892, line 9: Please use MAT_STEEL and MAT_GLASS for these and the other lines - DEFAULT_WALL_MATERIAL being used for matter lists was really a mistake in the first place and using bare strings for glass is not ideal.

MistakeNot4892 (changes requested):
See comment.

MistakeNot4892 (approved):

PR #8253, opened 2021-08-27
By-species age restrictions

Schnayy:
Allows jobs to have an age restriction based on their species/brain type. Currently adjusts the following:

All command jobs require Unathi to be 70, Teshari to be 20, Positronics to be 10.

Prometheans have the following age restrictions applied:
Junior Officer: 3
Nurse & Chemist: 3
Paramedic: 2
SAR: 2
Roboticist: 2
Xenobio&botany: 3
Scientist: 3
Engineer & Atmos Tech: 2

Disables the "ideal age" vars as well, because it's not balanced for species and I personally think it's unfair to not just pick the round-start character at random for those jobs anyways.

~~This does NOT make any restrictions to jobs these species should not be taking, but that will be intended in a separate PR. Whitelisted species should already know what jobs not to snag in the meantime like they always did.~~ Did it anyways for visual clarity.

ALSO bumps the max prommie age appropriately. They can now legally drive.

Absolutely needs staff discussion, but the age choices were discussed with the proper maintainers.

:cl:
tweak - Age restricted jobs now account for species.
rscadd - Adds age restrictions to a fair few promethean jobs.
rscadd - Jobs will prevent species that are not supposed to be in that role from taking it.
tweak - Prometheans can now be as old as 16.
/:cl:

### DIFF
--- a/code/game/jobs/job/captain.dm
+++ b/code/game/jobs/job/captain.dm
@@ -23,13 +23,15 @@ var/datum/announcement/minor/captain_announcement = new(do_newscast = 1)
 	economic_modifier = 20
 
 	minimum_character_age = 25
-	ideal_character_age = 70 // Old geezer captains ftw
+	min_age_by_species = list(SPECIES_UNATHI = 70, SPECIES_TESHARI = 20, "mechanical" = 10)
+///	ideal_character_age = 70 // Old geezer captains ftw
 
 	outfit_type = /decl/hierarchy/outfit/job/captain
 	job_description = "The Site Manager manages the other Command Staff, and through them the rest of the station. Though they have access to everything, \
 						they do not understand everything, and are expected to delegate tasks to the appropriate crew member. The Site Manager is expected to \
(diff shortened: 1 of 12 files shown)


MistakeNot4892 (changes requested):
Personally I think there's still a space for `ideal_character_age` to be used to weight roles towards a desirable age. Making it use a species-key list like you've now done for minimum age is pretty feasible.

Otherwise it would be preferable to delete the commented lines rather than leaving them there cluttering the place up imo.

Schnayy:
Noting that I intend to adjust to account for #8273 with whatever changes staff meeting asks for.

Schnayy:
Adjusted and readded the ideal age var as requested at the staff meeting. 

Also went ahead and added species job bans so it's not so visually confusing and is related to this PR. All job bans are in accordance with the ones listed on the wiki's rule page.

MistakeNot4892 (approved):

PR #8256, opened 2021-08-27
Reduces the headset noise

lbnesquik:
## About this pull request
Those newly introduced radio sounds are quite loud and given it is something we'll be hearing literally thousands of times, i'd rather have that as a background thing rather than as something quite noticeable.


MistakeNot4892:
Redundant in view of #8254

lbnesquik:
Maybe. You guys make the choice given my PR reduces noise further than this other one.

MistakeNot4892:
Going to make a call on this one in view of the other PR being merged (conflicting this one) and close it - feel free to reopen if the modified noise is too loud and you don't want to use the pref and we can revisit.

PR #8257, opened 2021-08-28
Moves a LOT of sprites around. [IDB IGNORE]

lbnesquik:
## About this pull request
This is like a spring-cleaning, but in August. There are more things which are a little odd in our sorting, but this should get the most obvious out of the way.

Here are the details :

- Moved all the screens dmi files to the screen sub folder
- Moved the 'roaper' file to clothing. It isn't actually used anywhere, but it doesn't belong in /mob
- Deleted amorph and bats, moved the respective sprites into mob.dmi
- Deleted eye.dmi and moved the AI eye into AI.dmi. It is simpler to have everything AI related in the same folder.
- Deleted gimmick_head and pilot_helmet.dmi and moved them into head.dmi. Each of those files had only a couple sprites and nothing particularly unique.


MistakeNot4892:
Oh Jesus, that `roaper.dmi` is a custom item I committed by accident in like 2014.

lbnesquik:
> 
> 
> Oh Jesus, that `roaper.dmi` is a custom item I committed by accident in like 2014.

IT WAS YOU

MistakeNot4892:
Alright so: 
- `/screen` stuff is all good.
- `roaper.dmi` should just be deleted, not kept around, nothing uses it and iirc the sprite ended up in the voidsuit icons anyway.
- bats and pilot helmet being merged into the giant monolithic DMI files - why? Having everything in one file makes conflicts more likely and is ideally something to move away from, not an improvement.

MistakeNot4892 (changes requested):
Will approve after deleting roaper.dmi and explaining rationale for the monolith DMI merges.

lbnesquik:
> 
> 
> Will approve after deleting roaper.dmi and explaining rationale for the monolith DMI merges.

Pilot helmets having their own file for 3 little sprites is silly to me. That's it.

AffectedArc07:
Could you stick `[IDB IGNORE]` in the title to make IconDiffBot ignore this PR? The job took 4 minutes to process, which is quite a waste of CPU time when its just reorganisation stuff.
image

MistakeNot4892:
Done, apologies, should have occurred to me.

code/modules/clothing/head/pilot_helmet.dm
```diff
@@ -4,7 +4,7 @@
 	name = "pilot helmet"
 	desc = "Standard pilot gear. Protects the head from impacts."
 	icon_state = "pilot_helmet1"
-	item_icons = list(slot_head_str = 'icons/mob/pilot_helmet.dmi')
+	item_icons = list(slot_head_str = 'icons/mob/head.dmi')
 	sprite_sheets = list(
 		SPECIES_TESHARI = 'icons/mob/species/teshari/pilot_helmet.dmi'
```
Atermonera, line 9: :+1: on squishing this into the monolith. We have existing infrastructure to split the clothing monolithic .dmi's into multiple (via the `index` var) when they grow too big.

Also, you missed a spot.

code/modules/mob/living/simple_mob/subtypes/animal/space/bats.dm
```diff
@@ -2,7 +2,7 @@
 	name = "space bat swarm"
 	desc = "A swarm of cute little blood sucking bats that looks pretty upset."
 	tt_desc = "N Bestia gregaria" //Nispean swarm bats, because of course Nisp has swarm bats
-	icon = 'icons/mob/bats.dmi'
```
Atermonera, line 5: :-1: on unifying this. Different mobs may have different sets of sprites, so it doesn't necessarily make sense to group all [generic] mob sprites into a single file. We also don't have the infrastructure or expectations of clothing, so the differentiation in icon file comes from the assignment of the icon var itself. This is a lot less intuitive unless there's logical separation in the naming of the .dmi files used.

And I actually think our `icons/` folder hierarchy needs more depth. If you _must_ change this, make an animals dmi, and move this along with other animal simplemob sprites into it. Or create an animals subdir, and split other animal simplemob sprites off of the monolith. (Latter is very much preferred)

Yes, this is a different answer than for clothing, but clothing sprites are much more complicated, and adjusting them in a similar manner according to their existing infrastructure will be a fair bit more involved.

Atermonera (changes requested):

PR #8262, opened 2021-08-29
Allergy severity tweaks

Cerebulon:
Current allergy effects seem best tuned for an entirely opt-in system like downstream where allergies and their severity are based on optional traits instead of being inherent to half of the station species.
So this:

- Reduces damage taken to be based on the amount of harmful reagent ingested. It will still hurt a bit, it just won't send you to surgery after one refreshing sip of poison. You probably won't see 'lasting' oxy damage unless you REALLY couldn't help yourself sipping that refreshing poison.
- Slightly lowers the chance for visible emotes.
- Very slightly lowers the severity of non-lethal disabling effects like weakness/drowsiness so you aren't totally physically immobilized for several minutes after one refreshing sip of poison.


MistakeNot4892 (approved):
This honestly seems like an oversight. Pretty odd that dosage wouldn't change allergy response.

Cerebulon:
I think the idea might have been that you get a flat reaction for as long as its in your system? But yeah it seems like a weird way to have done it when afaik the previous species-specific checks on specific foodstuffs worked off reagent processing instead of "is there an allergen in your system y/n?" so... Shrug.

PR #8264, opened 2021-08-31
Sec Helmet sprite patch

AdmiralDragon:
A naming issue was found for the normal security helmet, in the files there is 2 helmets for humans (Helmet & Helmet_Security) in game the sprite calls for 'Helmet' this patch adds another helmet sprite in the vox head.dmi 'helmet'.


MistakeNot4892 (approved):
@Schnayy

PR #8265, opened 2021-09-01
Adds a base matter cost to mech weapons and equipment.

lbnesquik:
## About this pull request
This PR adds matter values to mech weapons and all tools. 
This is a base value that does not take into account special materials, but at least you can get your steel and glass back.

The values are inspired from the drill and scattershot respectively, but it is a lower average still.

More immersion and something to do with unneeded equipment.


MistakeNot4892:
`matter` not `materials`. 
It's generally recommended to try compiling before you open the PR :P

lbnesquik:
> 
> 
> `matter` not `materials`.
> It's generally recommended to try compiling before you open the PR :P

I never do anything wrong!

MistakeNot4892 (approved):
Needs a squash/squashmerge.

PR #8266, opened 2021-09-01
Converts pretty much all of the mech fab and pros_fab designs to mat defines instead of the other...String thing?

lbnesquik:
## About this pull request 
"gold" became MAT_GOLD.  Same thing for other materials in the mech fab and prothestic fab.

Prettier code. Screw strings like that.

### DIFF
--- a/code/modules/research/mechfab_designs.dm
+++ b/code/modules/research/mechfab_designs.dm
@@ -23,7 +23,7 @@
 	id = "ripley_torso"
 	build_path = /obj/item/mecha_parts/part/ripley_torso
 	time = 20
-	materials = list(MAT_STEEL = 30000, "glass" = 11250)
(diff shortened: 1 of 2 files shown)


MistakeNot4892:
For the sake of clarity, `MAT_GOLD` is basically a compiler marker that is swapped out for `"gold"` when you build. The reason defines are used instead of bare strings is that if you make a typo, the compile will warn you, while making a typo like `"glod"` will not show up as a problem until the matter list breaks or such.

lbnesquik:
> 
> 
> For the sake of clarity, `MAT_GOLD` is basically a compiler marker that is swapped out for `"gold"` when you build. The reason defines are used instead of bare strings is that if you make a typo, the compiler will warn you, while making a typo like `"glod"` will not show up as a problem until the matter list breaks or such.

glod. It's prettier like this anyway.

MistakeNot4892 (approved):

PR #8267, opened 2021-09-01
Removes an empty file. No clue what's the deal with this.

lbnesquik:
## About this pull request
Deleted an empty file. It doesn't seem to contain something in the downstreams either.

Cleaning up our code.


MistakeNot4892 (approved):

Atermonera:
Probably a remnant from whenever power cells got refactored for device cells? Or else moved to the power module? P. sure that's where they live now.

PR #8269, opened 2021-09-02
Mecha Bugfixes

Mechoid:
:cl:
bugfix - Mechas process again.
bugfix - Phasing mechs can phase through walls properly.
tweak - Mecha LS management is more logical.
/:cl:

Mechas now properly process again.

Damage management for life support is, tangentially, fixed to be more logical.

### DIFF
--- a/code/game/mecha/mecha.dm
+++ b/code/game/mecha/mecha.dm
@@ -924,24 +924,22 @@
 	if(istype(obstacle, /mob))//First we check if it is a mob. Mechs mostly shouln't go through them, even while phasing.
 		var/mob/M = obstacle
 		M.Move(get_step(obstacle,src.dir))
+	else if(istype(obstacle, /turf) && phasing && get_charge()>=phasing_energy_drain)//Phazon check. This could use an improvement elsewhere.
+		src.use_power(phasing_energy_drain)
+		phase()
+		. = ..(obstacle)
+		return
 	else if(istype(obstacle, /obj) || istype(obstacle, /turf))//Then we check for regular obstacles.
 		var/obj/O = obstacle
-
-		if(phasing && get_charge()>=phasing_energy_drain)//Phazon check. This could use an improvement elsewhere.
-			src.use_power(phasing_energy_drain)
-			phase()
-			. = ..(obstacle)
-			return
 		if(istype(O, /obj/effect/portal))	//derpfix
 			src.anchored = 0				// Portals can only move unanchored objects.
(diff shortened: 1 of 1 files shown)


code/game/mecha/mecha.dm
```diff
@@ -945,6 +949,20 @@
 		else
 			step(obstacle,src.dir)
 
+	else if(istype(obstacle, /turf))
+		if(phasing && get_charge()>=phasing_energy_drain)
+			spawn()
```
MistakeNot4892, line 954: Should be splitting this out into a `set waitfor = FALSE` proc or using a timer/INVOKE_ASYNC rather than spawning.

code/game/mecha/mecha.dm
```diff
@@ -945,10 +942,29 @@
 		else
 			step(obstacle,src.dir)
 
+	else if(istype(obstacle, /turf))
```
MistakeNot4892, line 945: Should combine this with the phasing checks so that a non-phasing mech still calls `..()` at the end of the chain.

MistakeNot4892 (changes requested):
Fine to approve this if the phasing check is tweaked as commented.

code/game/mecha/mecha.dm
```diff
@@ -924,7 +924,7 @@
 	if(istype(obstacle, /mob))//First we check if it is a mob. Mechs mostly shouln't go through them, even while phasing.
 		var/mob/M = obstacle
 		M.Move(get_step(obstacle,src.dir))
-	else if(istype(obstacle, /obj))//Then we check for regular obstacles.
+	else if(istype(obstacle, /obj) || istype(obstacle, /turf))//Then we check for regular obstacles.
 		var/obj/O = obstacle
```
MistakeNot4892, line 933: Why is this cast?

MistakeNot4892: Looks like you can safely move this cast down to the portal bit.

MistakeNot4892: Never mind - this is a strange cast considering turfs can be passed in and runtime on the `O.anchored` check

Mechoid: For some godforsaken reason this bloc isn't updating.

Mechoid: Oh, it's because I'm a dumbass, got it.

MistakeNot4892 (approved):

PR #8277, opened 2021-09-05
Cynosure map

Neerti:
All the code from #8134, except now it's targeting a newly made Cynosure branch, so that it's easier for other people to collaborate.


Atermonera:
Merging with Disposition: creating dev branch for collaborative Cynosure work from multiple contributors. Moving the submission of those contributions away from Neerti's branch.

PR #8279, opened 2021-09-05
Kills Doorstop

PrismaticGynoid:
Fixes the arrivals announcements not being received by the station by allowing a fallback to non-overmap behavior while not on an overmap sector, not just when the overmap is not being used.

### DIFF
--- a/maps/~map_system/maps.dm
+++ b/maps/~map_system/maps.dm
@@ -194,14 +194,9 @@ var/list/all_maps = list()
 // Get a list of 'nearby' or 'connected' zlevels.
 // You should at least return a list with the given z if nothing else.
 /datum/map/proc/get_map_levels(var/srcz, var/long_range = FALSE, var/om_range = -1)
-	//Overmap behavior
-	if(use_overmap)
-		//Get what sector we're in
-		var/obj/effect/overmap/visitable/O = get_overmap_sector(srcz)
-		if(!istype(O))
-			//Not in a sector, just the passed zlevel
-			return list(srcz)
-
+	//Get what sector we're in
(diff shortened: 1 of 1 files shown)


Schnayy:
you monster

Cerebulon:
Doorstop cockpit plushie or riot

maps/~map_system/maps.dm
```diff
@@ -194,14 +194,9 @@ var/list/all_maps = list()
 // Get a list of 'nearby' or 'connected' zlevels.
 // You should at least return a list with the given z if nothing else.
 /datum/map/proc/get_map_levels(var/srcz, var/long_range = FALSE, var/om_range = -1)
-	//Overmap behavior
-	if(use_overmap)
```
MistakeNot4892, line 198: Slightly concerned that this is going to have knock-on effects for servers that use the overmap but have set up their announcements to work with the current logic (via a shared overmap sector or something like that).

MistakeNot4892: Thinking about it, why not just move the announcement computer onto the player z-levels?

PrismaticGynoid: This does not change any announcements already coming from an overmap sector. It just allows for those not on overmap sectors to fall back on the old behavior.

MistakeNot4892: Yes, which isn't really 1:1 with previous code - so you may end up with offmap Z-levels now announcing on the main levels.
It's a niche issue and not one that applies to Polaris but I figured it was worth raising.

Atermonera: That sounds like a fairly esoteric issue. Think this is fine.

Atermonera (approved):

PR #8281, opened 2021-09-06
Misc runtime fixes

Cerebulon:
- Runtime in vampire.dm where it attempted to affect nearby mobs when there were no nearby mobs.
- Runtime with foods added with the Snix vendor which used an invalid format for the center_of_mass which couldn't be read (which just hadn't come up because we didn't have a Snix vendor mapped in until recently).
- Runtimes in BorgLab PoI where the root _random randomspawner handler item had been placed repeatedly instead of actual spawners.

Not a runtime but kind of important: Throwing items to an adjacent tile no longer keeps a ghost of the item in your hand.

### DIFF
--- a/code/game/objects/random/guns_and_ammo.dm
+++ b/code/game/objects/random/guns_and_ammo.dm
@@ -170,7 +170,7 @@
 	icon_state = "grenade_2"
 
 /obj/random/grenade/item_to_spawn()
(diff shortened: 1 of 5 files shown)


Schnayy (approved):
looks fine to me, throw bug is very frustrating right now

Atermonera (approved):

PR #8283, opened 2021-09-06
Stimulant allergen for Tajaran

Cerebulon:
Replaced Tajaran coffee allergy with a more general caffeine allergy. This applies to coffee, tea, cola and cola-based sodas, dr. gibb, mountain wind, 13loko, and kahlua.
Added decaf versions of tea (and all tea recipes), space cola (also comes in 2L for the bar vendor), and starkist.
Retained the coffee allergy in code for usage with traits n crap, even if there's usually overlap (except decaf coffee!!)

:cl:
rscadd - Added new stimulant allergen for Tajaran instead of coffee. Coffee, tea and certain drinks/sodas contain caffeine and may cause a reaction. Decaf alternatives provided where necessary.
/:cl:

"why isn't it just called caffeine if these are all caffeine"? space reasons. maybe theres other stimulants out there!


MistakeNot4892:
Stimulant implies they should be allergic to hyperzine as well.

Cerebulon:
I used Stimulant because that's what Serithi said the allergy was supposed to be. Personally I think having them react poorly to hyperzine (if they don't already) would be fine but idk if genuine mechanical disadvantages are hugely desired so if it needs re-re-fluffed back to caffeine then I have absolutely no strong opinions because I'm not The Catlord I just play some kittycats.

MistakeNot4892 (approved):

PR #8287, opened 2021-09-08
Fixes Colony Director's Armour

FenodyreeAv:
The CD's armour inherits from clothing/suit/armor/ instead of clothing/spacesuits/, like everything else in this file, so it lost it's space protection when the SPACEPROOF tag got removed.

### DIFF
--- a/code/modules/clothing/spacesuits/miscellaneous.dm
+++ b/code/modules/clothing/spacesuits/miscellaneous.dm
@@ -22,6 +22,8 @@
 	slowdown = 1.5
 	armor = list(melee = 65, bullet = 50, laser = 50, energy = 25, bomb = 50, bio = 100, rad = 50)
 	flags_inv = HIDEGLOVES|HIDESHOES|HIDEJUMPSUIT|HIDETAIL
+	min_pressure_protection = 0 * ONE_ATMOSPHERE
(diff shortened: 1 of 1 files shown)


code/modules/clothing/spacesuits/miscellaneous.dm
```diff
@@ -22,6 +22,8 @@
 	slowdown = 1.5
 	armor = list(melee = 65, bullet = 50, laser = 50, energy = 25, bomb = 50, bio = 100, rad = 50)
 	flags_inv = HIDEGLOVES|HIDESHOES|HIDEJUMPSUIT|HIDETAIL
+	min_pressure_protection = 0 * ONE_ATMOSPHERE
```
MistakeNot4892, line 25: This is pointless, 0 * anything is 0.

FenodyreeAv:
I just copied the missing bit from /obj/item/clothing/suit/space so it would be the same as everything else in the file.
I'm not sure why it isn't just 0, but it's like that for everything in the game, so I assumed it was important.

MistakeNot4892:
The entire rest of the game is wrong in that case. Mathematically there's no purpose to the second part of the value.

MistakeNot4892 (approved):
Needs a squash merge.

PR #8297, opened 2021-09-12
Adds sanity check to overmap temp Z level can_die()

Atermonera:
Should fix spurious can_die runtimes:
```
[2021-09-12T21:11:20] Runtime in spacetravel.dm,24: cannot read from list
   proc name: can die (/obj/effect/overmap/visitable/sector/temporary/proc/can_die)
```

Relevant code is attempting to access index 1, hence list must have at least two elements.

### DIFF
--- a/code/modules/overmap/spacetravel.dm
+++ b/code/modules/overmap/spacetravel.dm
@@ -21,6 +21,8 @@
 	return ..()
 
 /obj/effect/overmap/visitable/sector/temporary/proc/can_die(var/mob/observer)
+	if(LAZYLEN(map_z) <= 1)
+		return 0
 	testing("Checking if sector at [map_z[1]] can die.")
 	for(var/mob/M in global.player_list)
 		if(M != observer && (M.z in map_z))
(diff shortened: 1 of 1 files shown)


MistakeNot4892:
This doesn't seem like a fix so much as a bandaid. Why is there a temporary sector with an invalid z-level?

code/modules/overmap/spacetravel.dm
```diff
@@ -5,31 +5,35 @@
...
-/obj/effect/overmap/visitable/sector/temporary/New(var/nx, var/ny)
+/obj/effect/overmap/visitable/sector/temporary/Initialize(var/nx, var/ny)
 	loc = locate(nx, ny, global.using_map.overmap_z)
 	x = nx
 	y = ny
 	var/emptyz = global.using_map.get_empty_zlevel()
 	map_z += emptyz
 	map_sectors["[emptyz]"] = src
 	testing("Temporary sector at [x],[y] was created, corresponding zlevel is [emptyz].")
+	return ..()
```
MistakeNot4892, line 16: You're going to want to do `. = ..()` on line one so that the base Initialize() proc executes and populates `map_z`.

MistakeNot4892 (approved):

PR #8303, opened 2021-09-16
Fix Hot/Cold Discomfort Message Strings

Rykka-Stormheart:
Cold Discomfort would not display if your body was covered - every /under/ item by default did so.
Heat discomfort would not display if your body was UNcovered.

This is fixed now - once you're uncomfortable, irregardless of what you're wearing, you'll start feeling it.
This should help prevent random deaths to freezing or overheating as you'll have warnings the area is too warm/cold before you start going ouchie.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/species_getters.dm
+++ b/code/modules/mob/living/carbon/human/species/species_getters.dm
@@ -79,9 +79,10 @@
 
 /datum/species/proc/get_environment_discomfort(var/mob/living/carbon/human/H, var/msg_type)
 
-	if(!prob(5))
+	if(!prob(10))
 		return
 
+	/* // Commented out because clothes should not prevent you from feeling cold if your body temperature has already dropped. You can absolutely feel cold through clothing, and feel too warm without clothing. ???
 	var/covered = 0 // Basic coverage can help.
 	for(var/obj/item/clothing/clothes in H)
 		if(H.item_is_in_hands(clothes))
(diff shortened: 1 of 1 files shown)


MistakeNot4892:
The idea behind the coverage check was that these messages are purely flavour and shouldn't really be a primary indicator of the environment being out of whack. You experience these before you hit the temperature warning thresholds that set off the HUD elements. Pretty sure removing this check will mean that Tajarans will be constantly spammed with the message in station temperature, as previously they could bypass it by rolling their undersuit down.

No strong opinion about removing the check, just some rationale. You should really remove the code rather than commenting it out though.

Schnayy:
Is the HUD icon for overheating/freezing not triggering early enough? If there have been temp issues resulting in deaths I'd think that's what needs to be adjusted. The most these alerts indicate to me is that someone adjusted the thermostat out of my comfort range.

Also not against the PR, just make sure it's not *too* spammy if an Unathi and a Taj get in an argument over the thermostat, please.

Rykka-Stormheart:
> Is the HUD icon for overheating/freezing not triggering early enough? If there have been temp issues resulting in deaths I'd think that's what needs to be adjusted. The most these alerts indicate to me is that someone adjusted the thermostat out of my comfort range.
> 
> Also not against the PR, just make sure it's not _too_ spammy if an Unathi and a Taj get in an argument over the thermostat, please.

That icon is intended for when you're starting to take damage. Cold_level_1 is when you start physically taking damage, and as such, is when the alert is thrown.

Rykka-Stormheart:
> I see. Might be a smart idea to make discomfort a HUD display as well, but that's a thought for a different PR probably.

I mildly agree, something like a mini-health-doll, but, eh. It's for a diff PR. This is effectively a bugfix for never getting discomfort messages outside two very specific situations.

MistakeNot4892:
It's not a bugfix. The current behavior is intentional and this change is going to have additional effects as noted. You will need to change the discomfort thresholds to prevent them appearing at standard dtation temperatures.

Cerebulon:
A 'mild discomfort' HUD icon seems far more sensible than just enabling the discomfort messages in situations where you're already doing everything you can to reduce discomfort in a normal environment (Wearing more/less clothes) because this is going to get really spammy.

Atermonera:
I support doing more things with the temp alert HUD element. That's a _much_ more effective means of telling players "Hey you're kinda hot" than spamming chat logs

PR #8304, opened 2021-09-16
Removes suicide verb function

Schnayy:
**Why this is good for the game:** It's only ever abused, and it's outright against our rules.

**Why the verb is staying at all:** It's good for admins to know if someone is trying to do it.

**Why didn't you remove borg verbs:** Wasn't sure on this one. I feel there's probably edge cases where appropriate. They're not the particularly problematic one anyways.

**Why comment it out and not delete it:** In case it needs to come back for an unforeseen edge case, I guess. Felt weird to keep the verb and not keep its function there to at least see.

### DIFF
--- a/code/game/verbs/suicide.dm
+++ b/code/game/verbs/suicide.dm
@@ -1,6 +1,6 @@
 /mob/var/suiciding = 0
 
-/mob/living/carbon/human/verb/suicide()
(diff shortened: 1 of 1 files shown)


MistakeNot4892:
Leaving the verb in and commenting the code out both seem like strange choices to me. Leaving it in means the verb will be present in the verb panel and look like an implicit suggestion, and commented code is generally pointless with git.

MistakeNot4892 (approved):

PR #8306, opened 2021-09-17
Quick age fix

Schnayy:
Accidentally set age for captain vatborn the same as other command, happened to notice like 2 seconds before it was merged. Oops!

Also followed a recommendation to increase command vatborn to 14, not 12.


MistakeNot4892 (approved):

Schnayy:
Stressful morning has made me realize this was a mistake, min age IS the same for all command. Will either be raising that too in another pr, or putting it back to 14.

PR #8308, opened 2021-09-19
Fixes runtime in mech equipment deletion

Verkister:
No need to detach from anything when you're already loose.


MistakeNot4892:
Might be worth gating the proc itself (ie. put the check in detach()) rather than specifically in Destroy, that way it will protect against other calls.

MistakeNot4892 (approved):

PR #8311, opened 2021-09-19
Remove suicide act

Atermonera:
Further cleans up the suicide() verbs because that's mostly copy-pasta.
Removes suicide_act() procs wherever they may be found. Suicide() is the only thing that invokes them, and I remember bumping into suicide code with some frequency and grumbling about how bad it is but not having bandwidth to really do the cleanup it should've gotten. Most of them are silly, nonsensical, or don't really make much sense anyways (Licking the electrodes of a battery?)
**Do not squash this.** Much like with @Schnayy 's previous change, it's possible that downstreams don't want this behaviour. The refactor part may be desirable separate from the bit where I remove a whole bunch of outdated code.


code/game/objects/effects/spawners/bombspawner.dm
```diff
@@ -1,110 +1,3 @@
-/* The old single tank bombs that dont really work anymore
```
Atermonera, line 1: The last time this entire code block was changed was 7y ago, and its been commented out for 10y

code/game/objects/items/toys/mech_toys.dm
```diff
@@ -197,48 +196,6 @@
...
-			in_combat = FALSE
-			return (BRUTELOSS)
-
-	sleep(0.5 SECONDS)
-	user.adjustBruteLoss(450)
-
-	in_combat = FALSE
-	visible_message("AN EASY WIN. MY POWER INCREASES.") // steal a soul, become swole
-	color= "#ff7373"
-	max_combat_health = round(max_combat_health*1.5 + 0.1)
```
MistakeNot4892, line 238: This is bizarre. /tg/ port?

code/game/verbs/suicide.dm
```diff
@@ -1,84 +1,52 @@
...
+
 
-	var/confirm = alert("Are you sure you want to commit suicide?", "Confirm Suicide", "Yes", "No")
+/mob/living/carbon/human/suicide() /// At best, useful for admins to see if it's being called.
+	..()
+	to_chat(src, "<span class='warning'>No. Adminhelp if there is a legitimate reason, and please review our server rules.</span>")
+	message_admins("[ckey] has tried to trigger the suicide verb as human, but it is currently disabled.")
 
-	if(confirm == "Yes")
+/mob/living/carbon/brain/suicide()
```
MistakeNot4892, line 25: Can't they just ghost instead of this weird-ass message and proc? If you're going to hide the verb and forbid it generally it seems odd for it to have a legitimate use.

code/game/verbs/suicide.dm
```diff
@@ -1,84 +1,52 @@
...
-		return
-
-	if (suiciding)
-		to_chat(src, "You're already committing suicide! Be patient!")
-		return
-
-	var/confirm = alert("Are you sure you want to commit suicide?", "Confirm Suicide", "Yes", "No")
-
-	if(confirm == "Yes")
+/mob/living/silicon/ai/suicide()
```
MistakeNot4892, line 33: As above, I would think something like the Wipe Core verb from Bay would be better than a hidden suicide verb here.

code/game/verbs/suicide.dm
```diff
@@ -1,84 +1,52 @@
...
+	if(..())
 		suiciding = 1
-		to_chat(viewers(src),"<span class='danger'>[src] is powering down. It looks like they're trying to commit suicide.</span>")
+		visible_message("<span class='danger'>[src] is powering down. It looks like they're trying to commit suicide.</span>")
 		//put em at -175
 		adjustOxyLoss(max(getMaxHealth() * 2 - getToxLoss() - getFireLoss() - getBruteLoss() - getOxyLoss(), 0))
 		updatehealth()
 
-/mob/living/silicon/pai/verb/suicide()
+/mob/living/silicon/pai/suicide()
```
MistakeNot4892, line 49: This one sort of has to be separate due to the pAI reimplementing the verb categories etc. Personally I'd rename it to Wipe Core or something instead of suicide.

MistakeNot4892 (changes requested):
Generally seems a bit inconsistent, would like some rationale before approving.

code/game/objects/items/toys/mech_toys.dm
Atermonera: Yup. This was backported to us through https://github.com/VOREStation/VOREStation/pull/8665

code/game/verbs/suicide.dm
Atermonera: pAIs also apparently have `/mob/living/silicon/pai/verb/wipe_software()`, which seems to do exactly this.

Atermonera:
The only thing with a real case for leaving something like this is the brain, to provide a visible message to others that you're ghosting, because you don't get that from examine. But that feels to me like more of an issue with examine. AI is known to have a cryo-like way to leave the round, and I've determined that pAIs also have that. Hence, there's no point in not making the warning to read rules the default behaviour.

@Schnayy 's rationale for leaving it in at all is to signal to folks that don't know better to go read the rules, and to signal admins to keep a closer eye on someone (Tho personally I feel like unless they somehow fatfinger `suicide` they're most likely doing it deliberately, in which case we'd also get that message if they were to start banging their head with a crowbar).

code/game/verbs/suicide.dm
MistakeNot4892: I thought they did! Didn't go and check so didn't want to make a false claim.

MistakeNot4892 (approved):

Schnayy:
There has been more than one occasion where someone new joins the server, doesn't read the rules for whatever reason, and attempts suicide like they would on another server. Having a notice to them as to why that's inappropriate gives them the chance to understand before just jumping in disposals, I feel. But yeah, no issue here with this cleanup.

MistakeNot4892:
I more meant a rationale for the perceived inconsistently, Schnayy ran me through the logic for keeping the verb.

PR #8312, opened 2021-09-19
Repaths obj/effect/effect, datum/effect/effect [MDB_IGNORE]

Atermonera:
The `/obj`s that get spawned are now `/obj/effect/vfx` (from `/obj/effect/effect`).
The `/datum`s that manage them are now `/datum/effect_system` (from `/datum/effect/effect/system`).
There was also almost identical code for `/datum/effect/explosion` and `datum/effect/expl_particle`, which I merged into the main system.

`/obj/effect/gibspawner` and `/obj/effect/graffitispawner` were repathed onto `/obj/effect/spawner` (which itself doesn't really do much). These are mapping tools, so there's not a whole lot that actually needed to change with them, just repathing because they have similar behaviour.


code/modules/effects/vfx/explosion.dm
```diff
@@ -0,0 +1,92 @@
...
+	anchored = 1
+	mouse_opacity = 0
+	pixel_x = -32
+	pixel_y = -32
+
+/obj/effect/vfx/explosion/Initialize()
+	. = ..()
+	QDEL_IN(src, 1 SECONDS)
+
+/datum/effect_system/explosion/set_up(n = 0, c = 0;, turf/loc)
```
MistakeNot4892, line 15: Linter issue here.

MistakeNot4892 (commented):
The actual code in the spawners is a nightmare, but I appreciate this is a repath rather than a rewrite. Hard to review it without leaving a billion comments though. Will do a proper review this evening hopefully.

Atermonera:
Yeah that stuff is pretty atrocious, it mostly just caught my eye when I was cleaning up suicide_act stuff and chanced upon the big commented-out block in gibspawner. And then while looking at that stuff I stumbled into the effect_system stuff

PR #8315, opened 2021-09-22
Rootspeak isn't a RESTRICTED language

Atermonera:
Making it WHITELISTED instead achieves the same semantic function, where silly humans and friends can't select it from character select, but diona can. RESTRICTED was interpreted as completely unavailable to players in character select when I implemented the system that sanitizes invalid languages.

It's probably a better solution to make the `species_language` accept a list of languages but that would take a _lot_ more gutwork and it's 1AM, and I got a flight at 10:40AM.

### DIFF
--- a/code/modules/mob/language/station.dm
+++ b/code/modules/mob/language/station.dm
@@ -7,7 +7,7 @@
 	colour = "soghun"
 	key = "q"
 	machine_understands = 0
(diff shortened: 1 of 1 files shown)


MistakeNot4892 (approved):
Bit leery of this but can't see a reason for it to use RESTRICTED in code in the first place.

PR #8316, opened 2021-09-26
Refactors reagent pump code

Atermonera:
It was messy. It is now 82 lines _less_ messy. Cleans up overlay generation, formatting, organization, logic.
Completely untested beyond compilation, but I am confident that this functions no worse than the previous code (And quite a bit more performant, at least in icon operations)


code/modules/reagents/machinery/pump.dm
```diff
@@ -0,0 +1,188 @@
...
+		user.drop_from_inventory(W, src)
+		to_chat(user, "<span class='notice'>You insert the power cell.</span>")
+
+	else
+		. = ..()
+
+	RefreshParts()
+	update_icon()
+
+/obj/machinery/pump/proc/pump_reagents()
```
MistakeNot4892, line 166: This should really be a proc on the turf.

Atermonera: Moved it onto turf. I would nominally like to make it an assoc list on the turf of {reagents : volume mult}, but because there's extra logic involved on water, that feels like an even clunkier solution to split the definition between variables and external environment parameters.
Also, loc.name = "Sea" is pretty bad, but this isn't much better. Put up https://github.com/PolarisSS13/Polaris/issues/8324 to address repathing some of this stuff because that's gonna be pretty big map changes, and probably pretty messy.

Atermonera:
Dropping edge case for turf area being ocean: That area is a map-specific define, and should rely on underlying turf subtypes. Because that case isn't properly implemented, and the new map is not currently planned to feature a salt water body, I can safely drop the case. If someone wants to reimplement it, see #8324

PR #8334, opened 2021-10-09
Adds event regions to GameMaster events

Atermonera:
This lets you define an event as occurring in
```
EVENT_REGION_PLAYER_MAIN_AREA // The main areas that players are expected to be in
EVENT_REGION_SUBTERRANEAN // The mines
EVENT_REGION_PLANETSIDE   // Somewhere on a planetary surface
EVENT_REGION_SPACESTATION // Somewhere in space that players likely have ready access to
EVENT_REGION_DEEPSPACE    // Somewhere "far away" on the overmap
EVENT_REGION_UNIVERSAL    // Affects all players, everywhere (E.g., comms blackout)
```
Event selection will pick a random region, weighted by the activity of players in those regions. The universal region is the total activity of all players everywhere, and so is multiplied by [ARBITRARY FACTOR] (1/2) to favor events that target specific regions. This number is straight out of my ass.

This is useful, because it means engineers on a planetary map won't cause meteor storms for people in space, but we _could_ set up an event that does something similar to _planetary_ z levels.


MistakeNot4892:
Structurally looks fine, only really skimmed the functional logic.

Atermonera:
We did the test back in november, didn't find any major issues.

PR #8338, opened 2021-10-11
Fixes emote targetting.

MistakeNot4892:
- Emotes can now be targeted again.
- Emotes will look for the closest atom that has your target string as a subpattern in the atom name.


MistakeNot4892:
Very strange issues with the target loop, need to fix after work.

Atermonera (approved):

Atermonera:
Sorry this took so long. Work has been extremely busy lately.

PR #8343, opened 2021-11-11
Fixes runtime with ducks and turrets

Verkister:
Fixes ducks trying to figure out whether the turret that just shot them shot bullets or breadcrumbs.

### DIFF
--- a/code/modules/mob/living/simple_mob/subtypes/animal/sif/duck.dm
+++ b/code/modules/mob/living/simple_mob/subtypes/animal/sif/duck.dm
@@ -57,9 +57,10 @@
 	. = ..()
 
 	var/has_food = FALSE
-	for(var/obj/item/I in L.get_contents())	// Do they have food?
-		if(istype(I, /obj/item/weapon/reagent_containers/food))
-			has_food = TRUE
-			break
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/simple_mob/subtypes/animal/sif/duck.dm
```diff
@@ -57,9 +57,10 @@
 	. = ..()
 
 	var/has_food = FALSE
```
MistakeNot4892, line 59: This chunk could be condensed to `var/has_food = isliving(L) && (locate(/obj/item/weapon/reagent_containers/food) in L.get_contents()))` unless I am misunderstanding.

MistakeNot4892: More generally:
````
/mob/living/simple_mob/animal/sif/duck/IIsAlly(mob/living/L)
	. = ..() && (!istype(L) || !(locate(/obj/item/weapon/reagent_containers/food) in L.get_contents()))
````

MistakeNot4892 (approved):
Fix in general is fine but the original code is a bit wonky.

PR #8344, opened 2021-11-11
Fixes rocket launcher

Verkister:
The launcher type guns use throw_at instead of the fancy bullet projectile features, so the fancy bullet projectile mechanics were unable to make things go boom by having several dead-end returns happen before the projectiles could even reach the on_hit, which already bonks into a return in the ..()'d parent proc anyway.

### DIFF
--- a/code/modules/projectiles/projectile/explosive.dm
+++ b/code/modules/projectiles/projectile/explosive.dm
@@ -8,22 +8,20 @@
 	damage = 30	//Meaty whack. *Chuckles*
 	does_spin = 0
 
-/obj/item/projectile/bullet/srmrocket/on_hit(atom/target, blocked=0)
-	..()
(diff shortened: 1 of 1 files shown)


code/modules/projectiles/projectile/explosive.dm
```diff
@@ -9,22 +9,30 @@
 	does_spin = 0
 
 /obj/item/projectile/bullet/srmrocket/on_hit(atom/target, blocked=0)
-	..()
```
MistakeNot4892, line 12: Removing the parent call seems a bit odd. The super call is where various affects are applied and it shouldn't impact the following code.

Verkister: As far as I know the only effect this projectile needs to apply is the explosion.

Atermonera (approved):

PR #8356, opened 2021-11-25
Fixes non-growing spiderlings growing.

Verkister:
They had functionally 0 difference whatsoever to the regular type here.

### DIFF
--- a/code/game/objects/effects/spiders.dm
+++ b/code/game/objects/effects/spiders.dm
@@ -156,7 +156,7 @@
 	pixel_y = rand(6,-6)
 	START_PROCESSING(SSobj, src)
 	//50% chance to grow up
-	if(amount_grow != -1 && prob(50))
+	if(amount_grown != -1 && prob(50))
 		amount_grown = 1
 	get_light_and_color(parent)
 	. = ..(ml)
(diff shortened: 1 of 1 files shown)


code/game/objects/effects/spiders.dm
```diff
@@ -156,7 +156,7 @@
 	pixel_y = rand(6,-6)
 	START_PROCESSING(SSobj, src)
 	//50% chance to grow up
-	if(prob(50))
+	if(!istype(src, /obj/effect/spider/spiderling/non_growing) && prob(50))
```
MistakeNot4892, line 159: Generally speaking if you are having to check the type of `src` then you should be doing something else. In this case it looks like you want `amount_grown != -1` instead, as `non_growing` and various other 'do not grow up' procs like the vermin event use `-1` to indicate non-growing spiderlings.

MistakeNot4892 (changes requested):
See comment.

code/game/objects/effects/spiders.dm
```diff
@@ -156,7 +156,7 @@
 	pixel_y = rand(6,-6)
 	START_PROCESSING(SSobj, src)
 	//50% chance to grow up
-	if(prob(50))
+	if(amount_grow != -1 && prob(50))
```
Atermonera, line 159:
```suggestion
	if(amount_grown != -1 && prob(50))
```

PR #8357, opened 2021-11-28
Fixes static PoI loading for Cynosure Map

Neerti:
Makes static PoI loading not be done by landmarks, but instead centralizes it by using `SSmapping`. The idea of having an object that just handles everything by itself when you place it on the map sounded really nice but I guess it was too clever for me, since somehow it made init take forever, so this time lets do it the boring way.

It will no longer add 300+ seconds to init, now it only adds about 0.4 seconds, ~~which is technically a 750x improvement, so... yay?~~

Other things:
* Also moves a large amount of map generation logic from the map datum to `SSmapping`.
* Unifies the engine and static PoI loading systems into one general static submap loading system.
  * Note that there is now no longer any protection against mapping two engine submap markers, since I imagine that will be a requirement for multi-Z engines, which Cynosure will need to be able to have random engines again.
* Adds some world log output to make it easier to know how long things are taking to load inside of map gen.
* Re-adds the static PoI markers on the Cynosure basement.
* Replaces the Southern Cross engine loader marker with a new static engine submap marker, as the engine loader no longer exists.

Tested on both Southern Cross and Cynosure.

Interesting to note: On Cynosure, 90% of `SSmapping`'s init time is taken up by generating cave tunnels, which might be a good optimization target down the road.


code/controllers/subsystems/mapping.dm
```diff
@@ -22,10 +27,15 @@ SUBSYSTEM_DEF(mapping)
...
 
 	if(config.generate_map)
+		// Load the engine and other submaps.
+		load_static_submaps(static_submap_landmarks)
+
 		// Map-gen is still very specific to the map, however putting it here should ensure it loads in the correct order.
 		using_map.perform_map_generation()
-	
-	loadEngine()
+	else
```
MistakeNot4892, line 35: Why not just
````
load_static_submaps(engine_submap_landmarks)
if(config.generate_map)
	using_map.perform_map_generation()
````

Comment (MistakeNot4892) [reply code/controllers/subsystems/mapping.dm:35]: Oh nvm just saw the args are different.

File: code/controllers/subsystems/mapping.dm [hunk 2]
```diff
@@ -44,37 +57,192 @@ SUBSYSTEM_DEF(mapping)
 		map_template_types[template.type] = template
 	return TRUE
 
-/datum/controller/subsystem/mapping/proc/loadEngine()
-	if(!engine_loader)
-		return // Seems this map doesn't need an engine loaded.
+/// Loads submaps at a specific point on the main map, using landmarks.
+/datum/controller/subsystem/mapping/proc/load_static_submaps(list/submap_paths)
+	log_mapload("Going to load [submap_paths.len] static PoI\s.")
+	for(var/thing in submap_paths)
```
MistakeNot4892, line 63: `for(var/obj/effect/landmark/submap_position/P as anything in submap_paths)` will give you an untyped loop without needing to cast.

code/controllers/subsystems/mapping.dm
```diff
@@ -44,37 +57,192 @@ SUBSYSTEM_DEF(mapping)
...
+			log_mapload("[log_info_line(src)] was not given a submap template, but was instead given [log_info_line(template)].")
+			return
+		
+		log_mapload("[P.name] chose '[template.name]' to load.")
+		admin_notice("[P.name] chose '[template.name]' to load.", R_DEBUG)
+		
+		P.moveToNullspace()
+		var/time_started = REALTIMEOFDAY
+		if(P.annihilate_bounds)
+			template.annihilate_bounds(T)
```
MistakeNot4892, line 82: This appears to be handled in load() if the template has `annihilate` set.

code/controllers/subsystems/mapping.dm
```diff
@@ -44,37 +57,192 @@ SUBSYSTEM_DEF(mapping)
...
+			chosen_template.load(T, centered = TRUE, orientation=orientation) // This is run before the main map's initialization routine, so that can initilize our submaps for us instead.
+
+			var/time_ended_submap = (REALTIMEOFDAY - time_started_submap) / 10
+			if(verbose_loading_time_logs)
+				log_mapload("Loaded submap '[chosen_template.name]' in [time_ended_submap] second\s.")
+
+			CHECK_TICK
+
+			// For pretty maploading statistics.
+			if(loaded_submap_names[chosen_template.name])
```
MistakeNot4892, line 208: `loaded_submap_names[chosen_template.name] = loaded_submap_names[chosen_template.name] + 1` will do this as well.

MistakeNot4892 (approved):
This seems fine to me. Only really skimmed the submap loader logic though.

PR #8365, opened 2022-01-10
Fixes Cynosure Shuttle Issue

Neerti:
The shuttles that create roofs will no longer cause what was thought of to be a crash, but was actually an hour long lagspike. This  makes the new map playable once more..
Also give thanks to @Atermonera, @PrismaticGynoid, and @Woodratt for their help, too.

### DIFF
--- a/code/modules/multiz/turf.dm
+++ b/code/modules/multiz/turf.dm
@@ -39,6 +39,7 @@
 	if(T)
 		GLOB.turf_entered_event.unregister(T, src, .proc/BelowOpenUpdated)
 		GLOB.turf_exited_event.unregister(T, src, .proc/BelowOpenUpdated)
+		turf_changed_event.unregister(T, src, /atom/proc/update_icon)
 	. = ..()
 
 /turf/simulated/open/Initialize()


Atermonera:
Profiler image for context. This is what happens _after_ that 1h+ hang:
image
For a bit more explanation, when the turf is qdel'd, it triggers a `/decl/observ/deleted` callback via
```
/datum/Destroy()
	if(GLOB.destroyed_event)
		GLOB.destroyed_event.raise_event(src)
	. = ..()
```
```
/decl/observ/destroyed/raise_event()
	. = ..()
	if(!.)
		return
	var/source = args[1]

	if(GLOB.global_listen_count[source])
		cleanup_global_listener(source, GLOB.global_listen_count[source])
	if(GLOB.event_sources_count[source])
		cleanup_source_listeners(source, GLOB.event_sources_count[source])
	if(GLOB.event_listen_count[source])
		cleanup_event_listener(source, GLOB.event_listen_count[source])
```
Because the code wasn't cleaning up the third listener, that meant that there was an extra listener set up for the turf. On the southern cross, this is actually fine! Because the list of things listening for the turf_changed event is very short (~141 things, which is what clued me into this in the first place once I saw the profiler logs and compared to the server on the SC).
However, the Cynosure has an _entire Z-level_ of basically just open_turfs. So when Neerti checked the number of listeners, there were ~57000 of them, or roughly on the order of 256^2 (Approx. the bounds of the map).
From the second code block, the profiler identified the third function as being egregious:
```
/decl/observ/destroyed/proc/cleanup_event_listener(listener, listener_count)
	GLOB.event_listen_count -= listener
	for(var/entry in GLOB.all_observable_events.events)
		var/decl/observ/event = entry
		for(var/event_source in event.event_sources)
			if(event.unregister(event_source, listener))
			//	log_debug("[event] - [listener] was deleted while still listening to [event_source].") // TODO: Apply axe, reimplement with datum component listeners
				if(!(--listener_count))
					return
```
For `event in all_observable_events` covers not just turf_changed, but also the other two listeners that, by the way, _also have 57k entries_. Without even looking at the other event sources, just those three alone account for an extraordinary computational burden, so it really makes sense how the server is hanging.

This change _fixes_ the issue by cleaning up that third listener before the Destroy() chain has a chance to process it, so we don't need to consider that loop at all.

PR #8369, opened 2022-01-17
Adds a roof to indoor rain sounds

Cerebulon:
Hopefully this should sound less like it's literally just raining indoors only quieter.
Basically just added a layer of building-roof-beating sound to a much quieter version of the existing sounds so it remains consistant.

The only issue is that it takes a few seconds for the looping sound system to register when you've changed areas since it has to complete its current loop so it will sound indoorsy/outdoorsy when you're outdoors/indoors for up to 10 seconds, but this is already a problem with the existing sound and just means you don't quite get the slap in the face/relief from stepping out/in from a storm.


Cerebulon:
The runtimes being caught are ones I distinctly noticed during the map test last night, so I assume they're present in master.

Atermonera:
Yeah, issue seems to be a one-off thing when pois generate and put mercs too close to targets. AI subsystem fires for the first time and stuff hasn't been set up yet.

PR #8371, opened 2022-01-18
Fixes literally everything dismantling walls

Atermonera:
Oops. It's the `if(dismantle_toolspeed)` check, and erroneously defaulting that value to 1.


Atermonera:
Failing unit test is an active edge with railroading, flake.

PR #8378, opened 2022-01-23
Fixes assorted mining toolspeed uses

Atermonera:


Atermonera:
CI is failing to mercs spawning in a POI, should be fine

PR #8381, opened 2022-01-24
Properly filters the genemod whitelisting

Atermonera:
Reviewed with @Cerebulon and @PrismaticGynoid 
cc @Schnayy for approval on unathi getting drake ears.

Note: I removed a bunch of the bad sprites, and basically all of the explicitly colored sprites.
Inclusion or removal of particular sprites is not a guarantee that they will be permitted by the whitelist at any time in the future.
The hard and fast rules are:
* No taurs for regular use
* No wings for regular use
* Tails that are too large/obstructive aren't permitted.

but every ear and tail was ultimately decided on a case-by-case basis.


Atermonera:
Update from Schnayy: drake ears are fine for regular use by unathi

Schnayy:
Would kind of like them renamed from drake ears to something more descriptive and ambiguous. They're just cool new frills.

Atermonera:
Renamed to just `frills`

PR #8408, opened 2022-02-22
Ingestion Actually Happens

Mechoid:
:cl:
tweak - Ingestion is now handled primarily by the stomach and intestine, instead of them only determining processing rates.
tweak - Stomach and Intestine form a "queue", wherein the stomach is the intake, and the intestine is the medium between the stomach and metabolism.
tweak - If your stomach produces an acid, you cannot be harmed from _ingesting_ that acid. All other processes with that acid occur as normal. Touch, injection, inhaling.
tweak - Ingestion metabolism speed is now also influenced by bloodflow. As, no bloodflow = no chemical movement through the body.
tweak - If you are meant to have a stomach, and you do not have one, you cannot drink or eat large volumes of material at once, and cannot swallow pills without vomiting. Liquids are key.
addition - Pills take time to crack when taken, and largely skip the stomach in processing "queue", unless the intestine is full.
addition - Vomiting will eject pills from the processing cycle, and obscure their names. Appearance is unchanged.
/:cl:

In non-changelog, ingestion is limited by the stomach and intestines' own processing speeds and individual volumes. Large diversity of reagents in the "queue" means less of each move per cycle, and pills can temporarily block it if multiple crack at once, in FIFO order.

If you have a heart, and it stops for some reason, ingestion will also not save you anymore. Get the heart pumping, it's what literally makes the chems go round.


Sypsoti:
While changes to ingestion are not *undesired*, it is my understanding Ater has been clear that he is working on GooseMed and planned changes to ingestion are included. Implementing this would cause notable changes to mechanics that would require an adjustment period... only to be followed with a whole medical rewrite that changes it all over again. For the sake of player experience and other current coding projects, I will be closing this.

PR #8416, opened 2022-03-04
Improves RD access on Cynosure

N-Komaeda:
As discussed in Discord, gives access to Maintenance, EVA and Exploration to the RD. Should only be valid on Cynosure.

First PR, Mechoid helped out with the actual meat of the thing, so lmk if it is proper or not and if things need changing.


Atermonera:
Known CI flake

Atermonera (approved):

PR #8419, opened 2022-03-05
New Changeling Menu

Sypsoti:
Hi, I have a new github account now.

Makes a new and improved changeling evolution menu split into categories as well as an in-game introduction to the game mode's basic mechanics. Also fixes removing changeling antag status not... actually doing anything! Removing it should now remove all verbs and abilities associated.

The menu refreshes each time it's reopened to avoid incorrect data being displayed after a readapt or the like. It updates immediately after clicking evolve so you can't accidentally spend your points twice.

This is my first time fiddling with menus like this, I think? I'd still like to touch up formatting a little but the code itself is ready to review. This is a first little babystep towards me reworking changeling some, as the current menu would not support my plans.

Previews of the menu:
image
image
image


Spookerton (approved):
does the thing

PR #8423, opened 2022-03-07
Fixes buns having the power to break food inits

Verkister:
Fixes a bug that caused a bad index runtime on mapped-in buns during map init that somehow apparently makes all food cause nausea and vomiting or something.


Spookerton (approved):
does what it says

PR #8425, opened 2022-03-07
fixes puke coating stealing nutriment reagent's id

Verkister:
How do you even miss this and how in the world is the code itself able to miss it too until something messes up a bunch of buns?


Spookerton (approved):
does what it says
nit; may be preferable to subtype coating for data-sourced coatings to keep the base implementation pure

PR #8426, opened 2022-03-07
Prometheans now do not clean items through gloves.

Mechoid:
:cl:
fix - Prometheans do not clean items through gloves.
/:cl:


Spookerton (approved):
does the thing

PR #8427, opened 2022-03-08
Remove evening gloves temp. protection

KillianKirilenko:
Elbow-length evening gloves are very fancy and all, but I'm reasonably certain they don't protect your delicate fingies from temperatures ranging from as low as 2K (-271.15C/-456.1F) to as high as 1500K (1226C/2240F).

So they won't any more.

:cl:
tweak - evening gloves no longer have amazing thermal protection values
/:cl:


Spookerton (approved):
does the thing

PR #8428, opened 2022-03-11
Makes the chirp emote available to Zaddat

listerla:
See title; no comment beyond that the Wiki mentions they chirp.

> Zaddat tend to chirp like crickets under high stress, which, Zaddat being Zaddat, is a fairly frequent occurrence.


Spookerton (approved):
code do thing if lore 👍

Cerebulon:
lore's fine go for it

PR #8429, opened 2022-03-12
Breathing Sounds While in Space

Rykka-Stormheart:
:cl:
rscadd - Breathing with a suit and internals on in low pressure areas has inhale/exhale sounds.
/:cl:

Upstream port of https://github.com/CHOMPStation2/CHOMPStation2/pull/3831

Plays a breathing sound akin to Bay's breathing sound while you're in space.

Major difference (codewise) is the ability to set an inhale + exhale sound, and it is currently designed SOLELY to be heard while in space.
Keep inhale + exhale combined to 7 seconds or less if you add custom sounds.
The code DOES support custom breathing per-species, if you see fit to add that for w/e reason.

Any internals playing breathing sounds later on may be looked at, but for now, this gets breathing working.

Does not come with a pref currently, as it's quiet - and intended to be an audible indicator of if you're breathing or not.
(IE if you don't hear it, be concerned.)


KillianKirilenko:
Atmospheric!

Why no SFX for synthetic lungs, though? FBPs don't have lungs at all to my knowledge so the only people who can have them are organics with prosthetics, aka the same people who benefit from the presence/lack of these cues if they have regular lungs.

Rykka-Stormheart:
> Atmospheric!
> 
> Why no SFX for synthetic lungs, though? FBPs don't have lungs at all to my knowledge so the only people who can have them are organics with prosthetics, aka the same people who benefit from the presence/lack of these cues if they have regular lungs.

That's a fair point, hm.
I was under the impression that setting full body selection to prosthetic would still include lungs, hrm.

KillianKirilenko:
I'm 99% sure that choosing to be FBP replaces your organs with a set of synth-specific organs, most of which don't cover regular organ functions because they're suspended for FBPs. I remember using a test server to experimenting with installing a synthetic heart in an FBP one time and it ended up causing runtimes because the game didn't know how to handle the situation.

You can easily check with a cyborg analyzer.

code/modules/mob/living/carbon/human/life.dm
```diff
@@ -629,6 +638,26 @@
...
 	return 1
+	
+/mob/living/carbon/human/proc/play_inhale(var/mob/living/M, var/exhale)
+	var/suit_inhale_sound
+	if(species.suit_inhale_sound)
+		suit_inhale_sound = species.suit_inhale_sound
+	else // Failsafe
+		suit_inhale_sound = 'sound/effects/mob_effects/suit_breathe_in.ogg'
+	
+	M << sound(suit_inhale_sound,0,0,0,100)
```
Spookerton, line 649:
```suggestion
	playsound_local(get_turf(src), suit_inhale_sound, 100, volume_channel = VOLUME_CHANNEL_AMBIENCE)
```

code/modules/mob/living/carbon/human/life.dm
```diff
@@ -629,6 +638,26 @@
...
+		addtimer(CALLBACK(src, .proc/play_exhale, M), 5 SECONDS)
+	
+/mob/living/carbon/human/proc/play_exhale(var/mob/living/M)
+	var/suit_exhale_sound
+	if(species.suit_exhale_sound)
+		suit_exhale_sound = species.suit_exhale_sound
+	else // Failsafe
+		suit_exhale_sound = 'sound/effects/mob_effects/suit_breathe_out.ogg'
+	
+	M << sound(suit_exhale_sound,0,0,0,100)
```
Spookerton, line 660:
```suggestion
	playsound_local(get_turf(src), suit_exhale_sound, 100, volume_channel = VOLUME_CHANNEL_AMBIENCE)

```

Spookerton (changes requested):

code/modules/mob/living/carbon/human/life.dm
```diff
@@ -629,6 +638,26 @@
...
+/mob/living/carbon/human/proc/play_inhale(var/mob/living/M, var/exhale)
+	var/suit_inhale_sound
+	if(species.suit_inhale_sound)
+		suit_inhale_sound = species.suit_inhale_sound
+	else // Failsafe
+		suit_inhale_sound = 'sound/effects/mob_effects/suit_breathe_in.ogg'
+	
+	M << sound(suit_inhale_sound,0,0,0,100)
+	if(!exhale) // Did we fail exhale? If no, play it after inhale finishes.
+		addtimer(CALLBACK(src, .proc/play_exhale, M), 5 SECONDS)
```
Spookerton, line 651: may be worth making this TIMER_STOPPABLE and holding the id so that the pending exhale can be cancelled when the mask/suit is removed

Rykka-Stormheart: Apologies for the super-late reply, but we do actually check if exhale can be played, hence the !exhale, and then in play_exhale, we simply play it, because timing it after the inhale was easiest, and by the time we reach this check, we've already tried to breathe (and therefore know if our breath is valid or not)

code/modules/mob/living/carbon/human/life.dm
Rykka-Stormheart: Done, pushing momentarily.

code/modules/mob/living/carbon/human/life.dm
Rykka-Stormheart: Done, pushing momentarily.

code/modules/mob/living/carbon/human/life.dm
Spookerton: The point was that stuff can happen in the intervening five seconds, but it's a small issue either way.

Spookerton (approved):

PR #8443, opened 2022-03-18
Drill quality now affects mining speed

Sypsoti:
Drills were all drilling the same speed. Now they do not.

Tested. Ater found the issue, I just typed the words.


Spookerton (approved):
does the thing

worth changing the arg name to `tool` instead of `quality` since tool_qualities is a `(TOOL_NAME = TOOL_QUALITY, ...)` assoc, but out of scope here

PR #8448, opened 2022-03-18
Robs You In Public

Cerebulon:
Adds a bunch of variants of mugs, and a randomspawner for mapping (not done yet because big maptouch PR is up)

They're also in loadout and function like hot drink flasks.


code/game/objects/random/misc.dm
```diff
@@ -932,3 +932,39 @@
 				prob(3);/obj/item/clothing/accessory/poncho/thermal/purple,
 				prob(3);/obj/item/clothing/accessory/poncho/thermal/blue)
 
+/obj/random/mug
+	name = "Random Mug"
+	desc = "This is a random coffee mug."
+	icon = 'icons/obj/drinks_mugs.dmi'
+	icon_state = "coffeecup_spawner"
+
+/obj/random/mug/item_to_spawn()
```
Spookerton, line 941: May be preferable here to simply `return pick(typesof(/obj/item/weapon/reagent_containers/food/drinks/glass2/coffeemug))` since it will mean not having to maintain the list later.

code/modules/client/preference_setup/loadout/loadout_general.dm
```diff
@@ -120,6 +120,52 @@
 	gear_tweaks += new/datum/gear_tweak/path(sortAssoc(lunchboxes))
 	gear_tweaks += new/datum/gear_tweak/contents(lunchables_lunches(), lunchables_snacks(), lunchables_drinks())
 
+/datum/gear/coffeemug
+	display_name = "coffee mugs"
+	description = "A coffee mug in various designs."
+	cost = 1
+	path = /obj/item/weapon/reagent_containers/food/drinks/glass2/coffeemug
+
+/datum/gear/coffeemug/New()
```
Spookerton, line 129: same as on `/obj/random/mug/item_to_spawn()`

Cerebulon: Not sure about using it here since it would include tall mugs and the glass mug which imo are a pretty majorly different vibe from normal run of the mill ceramic mugs.

Spookerton: Fair; is only for thought

Spookerton (approved):

PR #8449, opened 2022-03-18
skathari disappear with a flash when killed

Spookerton:
also adds a heavily processed effect from a korg koassilator for the actual teleport effect


Cerebulon:
There's already a code block for this on 146

Spookerton:
Oh woop, totally missed that after the in round chatter about how they don't do the thing. Will unify them.

Atermonera (approved):

PR #8451, opened 2022-03-19
Fixes universal_understand keeping mobs illiterate

Verkister:
Fixes paper text contents being censored to simplemobs with universal_understand enabled.


Spookerton (approved):
placing the var check *before* the istypes would be a little micro optimization, but it do the thing

PR #8317, opened 2021-09-27
Removes /obj/item/weapon, /obj/item/device [MDB_IGNORE]

Atermonera:
I'm gonna regret putting this up now because there's other big PRs that are gonna conflict with this massively but here goes:

Mechanical changes:
Any /obj/item can cleave. Previously just /obj/item/weapon.
Icons should all be in working order. I went through searching for every icon state in icons/obj/weapons.dmi and icons/obj/device.dmi.

PR #8458, opened 2022-03-21
/weapon/ redux [MDB_IGNORE]

Spookerton:
partial alt of #8317 -- removes /weapon/ from paths

Builds and runs happily, draft push to run CI. The icon changes in the other PR seem to have already been done at some point, so this *should* be fine as this `minimal` changeset.


Spookerton:
NB: It's impossible for this to pass mapdiff, really, because of the size of the change - the job will always be killed before completion. The outcome of that particular workflow should be ignored here. There isn't actually a change anyway, just a huge mess of updated paths.

Atermonera:
Original PR fixed up and merged.

PR #8463, opened 2022-03-22
Ports Vaurca autohiss from Aurora to Zaddat

listerla:
> Zaddat Shrouds also assist in muting the natural buzzing accent Zaddat have when speaking Galactic Common, though cybernetic modification of the larynx for clearer speech has become popular in the last few decades.

I thought it might be some fun to actually use that natural buzzing accent - I have not checked with lore yet, however, so this will require a thumbs up or down.


Spookerton (approved):
approval for code quality; needs staff input for lore

Atermonera:
:+1: from staff meeting

PR #8464, opened 2022-03-23
Reviews all POIs for loot/balance + adds random universal anomalies [MDB_IGNORE]

Cerebulon:
Makes the Anomalous Region actually Anomalous.

- Adds the ability for mapped in items to easily have anomaly datums on init (rather than manually adding the component to each)
- Fixed the shotgun random spawner.
- Rebalanced many PoIs to have fairer loot based on their difficulty. No more fighting 30 mobs for like a single autoinjector.
- More reliable 'starter' equipment spawns in Plains PoIs.
- Adjusts basically every PoI to have more random elements, and many to contain random anomalous objects.
- Removed a lot of incredibly bad sound emitters, improved detailing in particularly bland PoIs, made variants of PoIs look/feel more distinct
- Replaces some cases where mobs were set to fight each other on spawn and the AI would get caught in a loop, with pre-spawned corpses. Worst offender being Xenohive.

About 90% of these will not generate to activate under neutral atmospheric conditions without intervention, so most will go unnoticed until they're actively sought out or randomly generate as active in another round. Most of these are random spawned random items, some are specific items, but either way it will not be easy to keep track of where these spawn or what they are from round to round.
To avoid going into too much detail and doing Spoilers, I will post a complete list of changes staff-side.


Spookerton (changes requested):
Needs running map merge / resaving with strongdmm in tg mode for those maps where that was already the case.

Avoid using DM wherever possible -- it mangles dmm files into an output style that's hard to modify and review, and creates out-sized diffs.

<https://github.com/SpaiR/StrongDMM> is very friend shaped.

Cerebulon:
> Needs running map merge / resaving with strongdmm in tg mode for those maps where that was already the case.
> 
> Avoid using DM wherever possible -- it mangles dmm files into an output style that's hard to modify and review, and creates out-sized diffs.
> 
> https://github.com/SpaiR/StrongDMM is very friend shaped.

These were all edited with StrongDMM except for like 1 which caused StrongDMM to instantly crash any time I tried to open it with it. Which ones are you referring to?

Spookerton:
I could've sworn I saw a TGM -> DM reformat in there somewhere.


Either way it's strongly preferable to use strongdmm in explicit TGM mode:

This will cause some initial big diffs since a lot of polaris' stuff seems to be in DM mode, so actually **ignore me for this one**. I'll do a big conversion and ignore-rev PR.

Atermonera (approved):

PR #8470, opened 2022-03-24
Some admin log "usr" fixes

Verkister:
"usr" is only meant for manually operated verbs and only causes issues when used in automated procs.

Fixes adminlog messages occasionally slapping the first completely unrelated mob as a "prefix" for the message itself.
Fixes emergency shuttle overrides also blaming the adminlog messages on the randomly picked hapless "usr" instead of the actual "user" operating the things.

### DIFF
--- a/code/modules/admin/admin_attack_log.dm
+++ b/code/modules/admin/admin_attack_log.dm
@@ -3,7 +3,7 @@
 /mob/var/attack_log = list( )
 /mob/var/dialogue_log = list( )
 
-/proc/log_and_message_admins(var/message as text, var/mob/user = usr)
+/proc/log_and_message_admins(var/message as text, var/mob/user)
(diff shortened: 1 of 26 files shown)


code/modules/admin/admin_attack_log.dm
```diff
@@ -3,7 +3,7 @@
 /mob/var/attack_log = list( )
 /mob/var/dialogue_log = list( )
 
-/proc/log_and_message_admins(var/message as text, var/mob/user = usr)
+/proc/log_and_message_admins(var/message as text, var/mob/user)
```
Spookerton, line 6: Offhand I think this will break a bunch of existing uses where usr is expected to be available from being called in a verb or proc as verb's tree.

Spookerton (approved):

Spookerton:
although gosh, could use a squash now, lol

PR #8480, opened 2022-03-27
Gives mech autocannon projectiles their own entry with unique sprite and lighting

Serithi:
:cl:
tweak - Added unique mech autocannon projectile sprites and lighting.
/:cl:

Also makes it a bit louder and changes the makeshift version to use the pistol sound matching its projectile. Suggesting buffing the AC's damage while lowering the fire rate so it can actually be a proper autocannon (by definition the smallest is a 20mm gun, you aint walking away from that impact) with some balancing, maybe even putting it as a higher-tier weapon and adding a proper LMG to take its current place, but that's for another PR, this is just aesthetics.


Spookerton (approved):
does the thiiiiing

PR #8489, opened 2022-03-28
Adds Shiny New Lobby Screen Art

Cerebulon:
New high-detail title art commissioned for Polaris from Adrien Perrier, whose portfolio can be found here: www.artstation.com/gasgano

PR #8490, opened 2022-03-28
Adds Shiny New Lobby Screen Art

Cerebulon:
New high-detail title art commissioned for Polaris from Adrien Perrier, whose portfolio can be found here: [www.artstation.com/gasgano](http://www.artstation.com/gasgano)

#8489 but not large. That PR was an even larger version than I'd originally even intended to upload (11MB vs 7MB, new version is 1.5MB), reposted to avoid having an extra pointless ~20MB in the git


maps/cynosure/cynosure_defines.dm
```diff
@@ -13,7 +13,7 @@
 	full_name = "Cynosure Station"
 	path = "cynosure"
 
-	lobby_screens = list('icons/default_lobby.png', 'icons/lobby_old.gif') // New lobby screen if possible.
+	lobby_screens = list('maps/cynosure/title_cynosure.png') //Title screen created for Polaris by Adrien Perrier - www.artstation.com/gasgano
```
Spookerton, line 16: It's probably preferable to use https://github.com/PolarisSS13/Polaris/blob/master/ATTRIBUTIONS.md for this task, instead of a comment

Cerebulon: Yeah good catch, I'd actually looked for that file in just about every subfolder but am a dumbass and didn't look at the literal root folder very hard.

Spookerton (approved):
code/asset size approval - might need some more back and forth in `#assets` before everyone's happy with the exact presentation though, idk

PR #8499, opened 2022-04-01
Fixes configuration datum ignoring the TICKLAG setting.

Raeschen:
Noticed this upstream. 'TICKLAG' in config/config.txt is ignored due to some historical mixup in this file. This fixes that issue.


Raeschen:
This might not necessarily be a 'historical mixup', i'm just a little confused at the ditching of ticklag, and apparent swapping to fps. Are they one and the same? Server fps and server tickrate. Byond seem to not document 'world.fps'. I may of got this wrong.

Raeschen:
Closing this until i understand it better.

PR #8504, opened 2022-04-03
C4 still explodes if the turf its placed upon changes

### LINKED ISSUE #8466: C4 disappears when the rock it's on is drilled
Drilling away a rock with C4/a seismic mining charge on it removes the explosive.

For it to drop OR for it to explode.

#### What actually happened
It disappears. Still pending if it will actually explode.

#### Steps to reproduce
Put C4/seismic mining charge from the mining vendor on a rock, then drill the rock.

#### Additional info:
 - **Server Revision**: Found using the "Show Server Revision" verb under the OOC tab. master - 2022-03-20 6054da8b3d1ba7df2950ccbb42880c39fd95144e

Atermonera:
fixes #8466

### DIFF
--- a/code/datums/wires/explosive.dm
+++ b/code/datums/wires/explosive.dm
@@ -31,4 +31,5 @@
 
 /datum/wires/explosive/c4/explode()
 	var/obj/item/weapon/plastique/P = holder
-	P.detonate(get_turf(P))
+	P.set_target(get_turf(P))
+	P.detonate()
--- a/code/datums/wires/explosive.dm
+++ b/code/datums/wires/explosive.dm
@@ -31,4 +31,4 @@
 
 /datum/wires/explosive/c4/explode()
 	var/obj/item/weapon/plastique/P = holder
-	P.explode(get_turf(P))
+	P.detonate(get_turf(P))
--- a/code/game/objects/items/weapons/explosives.dm
+++ b/code/game/objects/items/weapons/explosives.dm
@@ -83,15 +83,10 @@
 	if(!target)
 		target = src
 	
-	var/turf/T = null
-	if(isturf(get_turf(target))
-		T = get_turf(target)
-	if(LAZYLEN(location) == 3)
+	var/turf/T = get_turf(target)
+	if (!T && length(location))
 		T = locate(location[1], location[2], location[3])
-	else
-		T = get_turf(target)
-
-	if(isturf(T))
+	if (T)
 		explosion(T, blast_dev, blast_heavy, blast_light, blast_flash)
 
 	if(target)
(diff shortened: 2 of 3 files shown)


code/datums/wires/explosive.dm
```diff
@@ -31,4 +31,4 @@
 
 /datum/wires/explosive/c4/explode()
 	var/obj/item/weapon/plastique/P = holder
-	P.explode(get_turf(P))
+	P.detonate(get_turf(P))
```
Spookerton, line 34: Parameter was removed from signature below.

code/game/objects/items/weapons/explosives.dm
```diff
@@ -68,16 +71,19 @@
...
-/obj/item/weapon/plastique/proc/explode(var/location)
+/obj/item/weapon/plastique/proc/detonate()
 	if(!target)
 		target = get_atom_on_turf(src)
 	if(!target)
 		target = src
-	if(location)
-		explosion(location, blast_dev, blast_heavy, blast_light, blast_flash)
+	var/turf/T = get_turf(src)
+	if(!istype(T) && location.len == 3)
```
Spookerton, line 82: General: prefer core procs (isturf) where available for those teeny tiny gains.

Not sure this is an ideal behavior since it makes the C4 able to detonate from null loc among other things. It would surely also be preferable to give up on other stuff and always use the x/y/z either way, if using this approach?

Spookerton (changes requested):

code/game/objects/items/weapons/explosives.dm
```diff
@@ -68,16 +69,30 @@
...
+	var/turf/T = null
+	if(isturf(get_turf(target))
+		T = get_turf(target)
+	if(LAZYLEN(location) == 3)
+		T = locate(location[1], location[2], location[3])
+	else
+		T = get_turf(target)
+
+	if(isturf(T))
+		explosion(T, blast_dev, blast_heavy, blast_light, blast_flash)
```
Spookerton, line 95:
```suggestion
	var/turf/T = get_turf(target)
	if (!T && length(location))
		T = locate(location[1], location[2], location[3])
	if (T)
		explosion(T, blast_dev, blast_heavy, blast_light, blast_flash)
```

Spookerton (changes requested):

Spookerton (approved):

Spookerton:
Putting this in but I think it deserves a followup for more nuanced behavior regarding the difference between "changed" and "destroyed" turfs; stuff like planting two C4 on a wall will destroy the wall and send the second C4 to null -- from where it will still cause a second explosion at the end of its timer.

PR #8511, opened 2022-04-05
updated get_atom_on_turf

Spookerton:
nufc

improves behavior of /proc/get_atom_on_turf
- accounts for bad cases (non-movable subject, subject with a bad loc)
- gently improves loop test performance


Spookerton:
fork lied to me

Atermonera (approved):

PR #8516, opened 2022-04-08
Station's Message Monitor Functions Again

### LINKED ISSUE #8404: Message Monitoring console cannot be accessed with password
It is not possible to authorize yourself at the Message Monitoring console with the password.

To login and have the ability to observe everyone's text messages.

#### What actually happened
No authorisation.

#### Steps to reproduce
Get the password, then try to enter it. Spam enter to no avail.

#### Additional info:

Sypsoti:
After the Great New() Massacre, this was initializing at the same time as the server and therefore not finding one. 

fixes #8404

### DIFF
--- a/code/game/machinery/computer/message.dm
+++ b/code/game/machinery/computer/message.dm
@@ -69,11 +69,15 @@
 	..()
 
 /obj/machinery/computer/message_monitor/Initialize()
+	..()
+	return INITIALIZE_HINT_LATELOAD
+
+/obj/machinery/computer/message_monitor/LateInitialize()
+	. = ..()
 	//Is the server isn't linked to a server, and there's a server available, default it to the first one in the list.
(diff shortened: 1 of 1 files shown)


code/game/machinery/computer/message.dm
```diff
@@ -69,11 +69,15 @@
 	..()
 
 /obj/machinery/computer/message_monitor/Initialize()
+	..()
+	return INITIALIZE_HINT_LATELOAD
+
+/obj/machinery/computer/message_monitor/LateInitialize()
+	. = ..()
```
Spookerton, line 76: LateInitialize doesn't actually do initialization in the same sense as Initialize so its name is a little misleading; "PostInitialize" would probably be better semantically. Regardless, there's no need to return a hint from super here.

Spookerton (approved):

PR #8521, opened 2022-04-11
Skathari buff & AI bugfix

Sypsoti:
Adds skathari-specific AI, placed datums in separate folder for cleanliness and potential future enhancements by someone smarter ~(or me when I understand AI code better.)~ (I understand code better than I thought I did!)

Skathari now teleport like bluespace slimes.
Skathari drones attempt to maintain an ideal distance from their target bc squishy and ranged.
Skathari drones also should switch off of stunned targets.
~Still working on skathari mama a bit.~

Unsure if issue report is up on it, but fixes issue of mobs CONTINUOUSLY searching for their lost target.

TO DO:
~Fix offset of mama skathari's sprite~
~Make drones dodge around without it being obscenely fast.~
~Poke at unseen targets more because a check APPEARS to not be working right?~

First time touching AI code much so please let me know any suggestions on what to test for that I might've overlooked.

EDIT: Went in and looked at the Adv. Dark Gygax and implemented cleaner, cooler code and a new mechanic for mama skathari. She now teleports to targets, just like the others, but ALSO will displace enemies if she's surrounded! I'm confident it does what it's supposed to, now.

### DIFF
--- a/code/game/objects/effects/overlays.dm
+++ b/code/game/objects/effects/overlays.dm
@@ -111,6 +111,16 @@
 	pixel_x = -32
 	pixel_y = -32
 
+// For skathari empress' telegrab warning, could replace down the line. 
+/obj/effect/overlay/skathari_telegrab
+	name = "bluespace displacement"
+	desc = "An eerie field of bluespace energy, you might want to run!"
+	icon = 'icons/effects/96x96.dmi'
+	icon_state = "emfield_s3"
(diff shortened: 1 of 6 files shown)


Spookerton (approved):
Making the teleport action proc on the same input as the spit action, but only when it's ready, is a bit odd for player-controlled use. But presumably that's rare.

Otherwise, do what say.

PR #8524, opened 2022-04-11
Adds Preference for Graffiti Engraving

Cerebulon:
Meant to do this a while ago. Saw downstream's half-fixes and compensations for those half fixes and just... Did the sensible thing and added a toggle.

The pref is on by default.
When on, it behaves as it currently does, engraving with sharp objects on hard surfaces (floor/wall) happens on help intent.
When off, sharp object + help intent = nothing, no accidental clickies.
~~Regardless of preference, you can alt click on a hard surface (floor/wall) with a sharp object and do an engraving that way.~~

It's also extremely funny that you can engrave with syringes but that's something for another day.

### DIFF
--- a/code/game/turfs/simulated/wall_attacks.dm
+++ b/code/game/turfs/simulated/wall_attacks.dm
@@ -406,11 +406,3 @@
 
 	else if(!istype(W,/obj/item/weapon/rcd) && !istype(W, /obj/item/weapon/reagent_containers))
 		return attack_hand(user)
-
-
-/turf/simulated/wall/AltClick(mob/user)
-	if(isliving(user))
-		var/mob/living/livingUser = user
-		if(try_graffiti(livingUser, livingUser.get_active_hand()))
-			return
-	. = ..()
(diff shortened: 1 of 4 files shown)


Sypsoti:
As voiced to you already, I don't think Alt click functions should be included. I used alt-click to view tiles and I don't want to start engraving from that. Unless anyone has a strong reason that should remain, I'd like to see it removed. The toggle is fine.

code/game/turfs/simulated/wall_attacks.dm
```diff
@@ -407,3 +408,9 @@
 		return attack_hand(user)
 
 
+/turf/simulated/wall/AltClick(mob/user)
+	if(isliving(user))
+		var/mob/living/livingUser = user
+		if(try_graffiti(livingUser, livingUser.get_active_hand()))
+			return
+	. = ..()
```
Spookerton, line 416:
```suggestion
```

Alt-clicking to view tile contents is a long standing and dramatically more useful behavior. Intercepting it this way is a mistake.

Spookerton (changes requested):

Spookerton (approved):

PR #8527, opened 2022-04-12
Big Pharma's Blessing: Snorting Mechanics

Sypsoti:
Ports snorting and pill chopping mechanics. You can snort with a straw or rolling paper. You can chop with an ID or anything sharp.

Renames space drugs to ~ecstasy~ bliss, along with every code instance mentioning "space_drugs". Creates ambrosia extract and replaces instances of space drugs in ambrosia to that.

~Adds cocaine and crack. I'd rather implement the recipes for these separately so I can keep my sanity and port more wanted drugs. So for now, you'd need an admin to spawn the stuff. ~ Will be resolved in separate PR.

Relocates most? instances of drugs to this reagent subtype. Focused on ones that sent messages. Included psychiatric drugs. Also changes some flavors, specifically nicotine, because everything tasting like "bitterness" is stupid.

~The message sending mechanic is... a little confusing to get right and at least when I tested snorting, I kept it from spamming. Probably still needs tweaks. I'm tired.~


Sypsoti:
Messages now proc every 90 seconds or so, keeping it from spamming but still noticeable. Smoking now actually gives you enough reagents for a noticeable effect! This may or may not require adjustments later, but now you can actually get high off your blunts.

Ingestion also shouldn't suffer the old message spam issue due to the raised ingestion metabolism.

I removed toxin from ambrosia, too, because that was stupid.

Spookerton (commented):
@Sypsoti @Atermonera -- code approval but I'm hesitant to agree to *real* drug names on the basis that it causes the repo to be conceptually more fragile versus the [AUP](https://docs.github.com/en/site-policy/acceptable-use-policies/github-acceptable-use-policies#2-user-safety), which could provide an avenue for malicious users to cause service restrictions by submitting reports about gamifying substance abuse.

Sypsoti:
Removed cocaine and crack, the latter unnecessary and the former's name up for debate. Renames ecstasy to bliss regardless of what the pending discussion above decides. Secondary PR to add drugs was going to be a staff discussion anyways, will include name alternatives.

Spookerton (approved):
Code do thing.

@Sypsoti You might consider adding the pill behavior for dissolving in other containers to powder, and maybe also removing it from pills. Seems like a neat extra step to me.

PR #8530, opened 2022-04-12
alt of 8336

Spookerton:
closes #8336

depends on #8528 for OnNew / etc -- that PR should be handled separately and first.

The original PR included riders, like adding a sound to clicking buttons, which I've left out for sanity's sake.

The first commit is simply adding get_overlays, which is unused, and moving a list's definition from one file to another.

Everything in the second commit is the 1:1s of the original.

The third commit is a rewrite of the behavior of SSoverlays and the related atom procs. This one's hairy but a fairly significant improvement in maintainability terms (and I'd like to think competitive on performance).

The fourth commit cleans up remaining bare uses of overlays. I've left the two in /sync because of the comment, and in a handful of other places where it would not make sense to use the subsystem.


Atermonera (approved):

MistakeNot4892:
This PR appears to have completely cooked overlays on the live server. They are being mixed between atoms, showing/not showing, not being cleared, etc. Hydro trays and GPS are two specific things that are busted.

PR #8531, opened 2022-04-13
updates all bare globals to use /global/ semantics, and all shared members to /static/

Spookerton:
This doesn't add `global.` to all *uses* of bare globals because that's ~800 things to check and I do not currently have the energy. Being entirely flavor anyway, it doesn't affect anything.


Spookerton:
Doing this on bay, I realized I missed all the globals starting with /var. Will fix.

Atermonera (approved):

PR #8543, opened 2022-04-15
Adds Snowmobiles + Minor Quadhancement

Cerebulon:
Snowmobiles are very similar to ATVs, except they're cooler. Like ice.

A Snowmobile is:

- Slightly slower on most terrain than a quadbike.
- Slightly faster on snow and ice than a quadbike.
- Not able to pull a trailer, but can carry 2 people.
- Able to be built very similarly to a quadbike, marginally cheaper.
- Thematic.

Also more generally:

- Added ignition sound to vehicles with keys.
- Added an engine sound loop to the quad and snowmobile (loops neatly with a distinct 'shutdown' sound when it ends.)
- Quadbike terrain speed checks now account for some turfs types added since they were implemented.

Sprites made by me based on a /tg/ motorcycle.
Yes there's a lot of duplicate code from cargo trains/quadbikes and space bikes but each of those had too much snowflake bs for a subtype to be much of a meaningful parent type. Vehicle code is, shall we say, a bit of a mess.

https://cdn.discordapp.com/attachments/412152289219051521/964461564159742013/2022-04-15_10-43-26.mov


Spookerton (dismissed):
partial - issues that've shown up immediately:

- snowmobile mousedrop behavior breaks being able to exit or move the snowmobile

- removing snowmobile key does not turn off the engine

Cerebulon:
> partial - issues that've shown up immediately:
> 
>     * snowmobile mousedrop behavior breaks being able to exit or move the snowmobile
> 
>     * removing snowmobile key does not turn off the engine

The issue actually seemed to be that removing the key turned off the engine, but did not stop the engine sound. I wasn't able to recreate the mousedrop issue, are you sure the inability to move wasn't just a stopped engine with sound still playing?

Cerebulon:
Note there still is an issue where a passenger can't unbuckle themselves using the 'You are buckled' UI element (They can still click on the vehicle), but that's a problem inherited from existing vehicle multi-buckling code and not something new.

Spookerton:
There's a lot in the splatted in code for the snowmobile type that does nothing or circumvents movement implementations entirely. I'm going to take a look at making it act according to intent.

Aside - Local vehicle code is a window into hell. 😨

Spookerton:
> I'm going to take a look at making it act according to intent.

I don't currently have the time to do this from the ground up, sorry. The easiest way you can make this work is redo snowmobiles to be based off cargo trains, then remove the behavior you don't want.

PR #8546, opened 2022-04-16
Deletes mob/living/carbon

Atermonera:
Re-using this PR to accomplish the second part of the groundwork: deleting `/mob/living/carbon`. #8561 accomplishes the groundwork for this groundwork, by removing all of the other dependencies upon carbonmobs, such that the only thing left on carbon is now human mobs, and therefore, all carbon mobs _are_ humans. Thus, we can do some simple find/replaces to replace references to carbons with references to humans. There's probably a lot of typecasting from human->human now, that... will just have to be scrubbed out long-term. Because this is 16 _thousand_ lines already.


Atermonera:
Actually, I'm going to close this and make a test branch on master, so that I can split PRs into things that are too cumbersome for review with those that are not. Everything so far is in the former category because I kinda worked on 3 different threads at once. (I will do so tomorrow because it is Late)

PR #8548, opened 2022-04-16
Fixes Persistence Issues

### LINKED ISSUE #8414: Mining sheet storage behaving strangely across rounds.
The SheetStorage at Cargo seems to save resources improperly. It is no longer capped at 50 sheets to be stored across rounds, but it seems as if the sheets inside randomly disappear throughout the rounds.

To put in materials, and then keep a max of 50 sheets per resource over the rounds.

#### What actually happened
Weird amounts of various resources being in the machine, resources just... flat out disappearing.

#### Steps to reproduce
Put sheets into the storage, then check in later rounds. They don't match up to the normal cap and some just disappear.

#### Additional info:
 - **Anything else you may wish to add** (Location if it's a mapping issue, etc) This seems to be since the transfer to Cynosure Station.

### LINKED ISSUE #8503: Persistent Notes/Stickies Load Incorrectly
Persistent items which include HTML elements, particularly noticeboard notes and stickynotes, fail to load up correctly in subsequent rounds. They are placed in the correct location, but the other data becomes scrambled so they do not have a name (So cannot be clicked or examined normally) and their contents data spills into the BYOND UI and populates the entire right pane somehow.
This does not affect other persistent items like trash and dirt, which are saved separately.
This also does NOT affect graffiti, which uses a simple text input rather than being a subtype of paper with all the formatting that entails.

Write on sticky note, place, end round, start new round, be able to read the same note.

#### What actually happened
Sticky is in correct location but functionally blank and bugs out the UI when examined.

#### Steps to reproduce
Write on sticky note, place, end round, start new round, attempt to interact with note.

#### Additional info:
 - **Anything else you may wish to add** Almost certainly an issue with loading HTML as a result of the New/Init refactor, as HTML has failed to load correctly in several other instances.

Cerebulon:
Fixes persistent paper/notes being utterly busted, was a misnamed tag from when it was converted to JSON format. Was loading a different tag from the one it saved and resulting in items that didn't interact good.

Produce Storage now caps per item (at a much lower rate) instead of only saving sequentially until it hit the global cap.

Replaced slightly wonky switch statement for material sheets to a flat rate (as was done downstream) on request.

Fixes #8503 
Fixes #8414

### DIFF
--- a/code/modules/persistence/storage/smartfridge.dm
+++ b/code/modules/persistence/storage/smartfridge.dm
@@ -48,14 +48,8 @@
 
 		// Delete some stacks if we want
 		if(stacks_go_missing)
-			var/fuzzy = rand(-5,5)
-			switch(count / max_amount)
-				if(0 to 1)
-					count -= 10+fuzzy // 1 stack or less, lose 10
-				if(1 to 10)
-					count -= max_amount+fuzzy // 1 to 10 stacks, lose a stack
-				if(10 to INFINITY)
-					count -= max_amount*3+fuzzy // 10+ stacks, lose 3 stacks
+			var/fuzzy = rand(55,65)
+			count = round(count*0.01*fuzzy) // loss of 35-45% with rounding down
 			if(count <= 0)
 				continue
 
--- a/code/modules/persistence/storage/smartfridge.dm
+++ b/code/modules/persistence/storage/smartfridge.dm
@@ -16,9 +16,6 @@
 			qdel(A) // Should clean this up here, it couldn't be stocked
 
 
-
-
-
 /datum/persistent/storage/smartfridge/sheet_storage
 	name = "sheet storage"
 	max_storage = 50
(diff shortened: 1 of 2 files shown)


code/modules/persistence/storage/smartfridge.dm
```diff
@@ -79,6 +70,16 @@
 	name = "fruit storage lossy"
 	go_missing_chance = 12.5 // 10% loss between rounds
 
+/datum/persistent/storage/smartfridge/produce/generate_items(var/list/L)			// Mostly same as storage/generate_items() but without converting string to path
```
Spookerton, line 73: I haven't looked at the rest. Cute as this is-

- Persistent storage is a terrible mistake.
- The base implementation does text=type checking for a reason.
- The base implementation does not check *count* on write or read, but *really* should.
- ***Persistent storage is a terrible mistake***.

Cerebulon:
Tested: Does not save mutant weirdo plants, it saves them as a generated ID name which then fails to load on new round and notifies via debug log that it failed to spawn that imaginary fruit type and was discarded. So. Works.

Sypsoti (approved):
A type check won't work for fruits as they're all snack/grown as far as I understand. Persistent storage is currently a desired feature and the discussion of if it should be kept is irrelevant to fixing its current state. Code looks fine & my concerns have been tested.

PR #8557, opened 2022-04-17
New Rollerbeds... Again!

Cerebulon:
New rollerbed sprites via https://github.com/Baystation12/Baystation12/pull/31914
Technically they were upgrading from the OLD old sprites, but I reckon this one is more visually interesting than the flat one we have right now. Also a recolour for the advanced rollerbed which the original did not have, matching the old colour scheme.

Also technically matches bluemed.


Spookerton (approved):
It do the thing

PR #8558, opened 2022-04-18
Ponchos are compatible with custom items

Sypsoti:
i've walked through hellfire and it still wasn't enough because my master branch decided to be fucky, too. just ignore that change to xeno ai.

I'll adjust the custom items in our other repo to use the kit, just poke me after merging this.

i never want to see another line of clothing code again.


Sypsoti:
Replaced with #8562

PR #8560, opened 2022-04-20
Deletes the instruments of war

Atermonera:
... on my eardrums.
Also known as the violin and piano.


Cerebulon:
Is there any particular reason for this other than it sometimes sounding a bit jank? Because there's a preference for that, surely.

Atermonera:
The intent of the piano and violin is to play discrete notes in a coherent fashion (I.e., music, but you're actually defining the music instead of playing e.g., a MIDI). The problem with them is that the delivery of each note is dependent upon a players' connection. They can sound alright to someone with a good connection, they will always sound like shit to someone with a spotty connection. I feel like this limitation makes the instruments fundamentally flawed with regards to their intended purpose, and that the jukebox is wholly better suited, because while its list of songs is limited, it plays entire songs coherently.

I'm not opposed to keeping them as props, (Nor do I feel very strongly about this change in the first place, given how often the instruments are used). TG has purportedly done a lot of work with instruments, but I do not have the time or interest to look into porting that behaviour. The bugfix and QoL labels are only slightly sarcastic.

Atermonera:
Not going to touch this anytime soon, and the proposed work would require assets I don't have time/inclination to make.

PR #8570, opened 2022-04-23
Fixes ALL random tech components having a chance of being anomalous

Cerebulon:
That's no subtype!
Surprised it took so long to come up.


Sypsoti:
High priority bc as-is, the station can spawn very bad things in maintenance.

Sypsoti (approved):

PR #8572, opened 2022-04-23
Adjusts frequency of anomalies and weights effect rarity

Cerebulon:
- Random anomaly spawns are now only a 50% chance of actually being anomalous. Should prevent anomalies being in practically ever PoI.
- Added a rarity rating to effects, so that more disruptive and/or useful effects are rarer.

Weighting will not function correctly until #8547 is merged, so do not advise merging until that PR is fixed/merged otherwise may result in DNA Fucker anomalies being more common instead of much, much rarer.

### DIFF
--- a/code/game/objects/random/_random.dm
+++ b/code/game/objects/random/_random.dm
@@ -32,9 +32,8 @@
 		A.pixel_x = pixel_x
 		A.pixel_y = pixel_y
 		A.set_dir(dir)
-		if(start_anomalous)
-			if(prob(50))
-				A.become_anomalous()
+		if(start_anomalous && prob(50))
+			A.become_anomalous()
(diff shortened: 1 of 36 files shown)


code/game/objects/random/_random.dm
```diff
@@ -33,7 +33,8 @@
 		A.pixel_y = pixel_y
 		A.set_dir(dir)
 		if(start_anomalous)
-			A.become_anomalous()
+			if(prob(50))
+				A.become_anomalous()
```
Atermonera, line 37:
```suggestion
		if(start_anomalous && prob(50))
			A.become_anomalous()
```

Atermonera (approved):

PR #8526, opened 2022-04-11
Ports TG Lighting... Again. With none of the extras.

Cerebulon:
https://github.com/PolarisSS13/Polaris/pull/8294 but only the primary PR.

Current issues:

~~- [Entirely unlit exterior planet tiles are literally pitch black (Can't see more than the tile you're standing on) unless manually lit by placing a light or shining a light source.](https://i.gyazo.com/31b1a5bd6bf6c53d9e80e8f348f8500a.png)~~
~~- [Ghosts CAN see those tiles with proper lighting applied.](https://i.gyazo.com/8eddca6e554f2626609a0ee2e2f7f180.png)~~
~~- Similarly, openspace tiles indoors are 'fullbright' until 'new' light is cast on them (e.g. turning a light casting on those tiles off and on, or night lighting toggling for the entire station)~~
- Lots of values need tweaking but that's to be expected.
- 'Sharp' lighting edges where overhangs meet outdoor tiles, and on some frames of flashlights.

PR #8582, opened 2022-04-26
Atmospheric Retention Fields Port

KillianKirilenko:
Just a little downstream project of mine that I figure I might as well push upstream for compatibility's sake.

> A high-tech alternative to the humble firedoor, atmospheric retention fields activate when an area's air or fire alarms are triggered, generating a shimmering forcefield that prevents airflow across it. They are fairly power-hungry, but more convenient than firedoors as they prevent airflow without preventing movement. Alternately a more power-efficient impassable variant also exists. Their activation may not be instant, but it's quick enough that it can help save lives.
> 
> image
> 
> Adjacent fields link together to form more visually-pleasing barriers in any shape or size imaginable.
> image
> 
> If shut down by an EMP, fields will come back online with a slight randomized delay.
> staggered_repower
>
> ARFs run off the Environmental circuit, so they should be the last things to shut down when an APC starts to run flat. They can also be tinkered with using various tools to adjust their behaviour, or just disassembled entirely: ex. you can pop one open using a crowbar and either cut the wires to sabotage it or toggle it into always-active mode using a multitool.

Their functionality hasn't changed significantly since they were initially added, though since you guys don't have /tg/lights (yet) I've commented out the associated lighting settings. Once #8526 is merged, you can just uncomment that block and it should look nicer.

These are 110% Coder Art and could _really_ do with some nicer sprites for both the emitter module and the field itself, and maybe some SFX for shutdown/powerup, but otherwise are fully functional.

## Todo?
- [ ] Nicer Field/emitter sprites
- [ ] Power up/down & ambient SFX
- [ ] Wall-mounted subtype?

## Changelog
:cl:
add: Added atmospheric retention field generators, which can be built after science has done some research.
/:cl:


code/game/area/areas.dm
```diff
@@ -170,6 +174,25 @@
 					spawn(0)
 						E.open()
 
+// Activate all retention fields!
+/area/proc/arfgs_activate()
+	if(!arfgs_active)
+		arfgs_active = TRUE
+		if(!all_arfgs)
```
Atermonera, line 181: Should this be lazylen?

code/game/machinery/atm_ret_field.dm
```diff
@@ -0,0 +1,246 @@
...
+		to_chat(user, "<span class='warning'>You can't open the ARF-G whilst it's running!</span>")
+		return
+	if(W.is_crowbar() && !isactive)
+		if(!src) return
+		to_chat(user, "<span class='notice'>You [hatch_open? "close" : "open"] \the [src]'s access hatch.</span>")
+		hatch_open = !hatch_open
+		update_icon()
+		if(alwaysactive && wires_intact)
+			generate_field()
+		return
```
Atermonera, line 56:
```suggestion
	if(!src) return
	if(W.is_crowbar())
		if(isactive)
			to_chat(user, "<span class='warning'>You can't open the ARF-G whilst it's running!</span>")
			return
		to_chat(user, "<span class='notice'>You [hatch_open? "close" : "open"] \the [src]'s access hatch.</span>")
		hatch_open = !hatch_open
		update_icon()
		if(alwaysactive && wires_intact)
			generate_field()
		return
```

code/game/machinery/atm_ret_field.dm
```diff
@@ -0,0 +1,246 @@
...
+	if(hatch_open && W.is_multitool())
+		if(!src) return
```
Atermonera, line 58: I covered this in the suggestion above, but if you're doing this at this in every single if statement, you should just do it at the top of the function.

code/game/machinery/atm_ret_field.dm
```diff
@@ -0,0 +1,246 @@
...
+		update_icon()
+		return
+	if(hatch_open && istype(W,/obj/item/weldingtool))
+		if(!src) return
+		var/obj/item/weldingtool/WT = W
+		if(!WT.isOn())
+			return
+		if(!WT.remove_fuel(0,user))
+			to_chat(user, "<span class='warning'>You need more fuel to complete this task.</span>")
+			return
```
Atermonera, line 76: Is this not covered by `isOn()` and the actual `remove_fuel()` call below?

code/game/machinery/atm_ret_field.dm
```diff
@@ -0,0 +1,246 @@
...
+		playsound(src, WT.usesound, 50, 1)
+		if(do_after(user,15 * W.toolspeed))
+			if(!src || !user || !WT.remove_fuel(5, user)) return
+			to_chat(user, "<span class='notice'>You fully disassemble \the [src]. There were no salvageable parts.</span>")
+			qdel(src)
+		return
+
+/obj/machinery/atmospheric_field_generator/perma/Initialize()
+	. = ..()
+	generate_field()
```
Atermonera, line 87: ... Is it powered during initialize? Or does this actually get handled by `power_change()` when the round actually starts electrical things doing electricity?

Also, what';s stopping an enterprising engineer from stabbing it with a multitool to swap it from alwaysactive to emergency?

code/game/machinery/atm_ret_field.dm
```diff
@@ -0,0 +1,246 @@
...
+	else if(isactive)
+		icon_state = "arfg_on"
+	else
+		icon_state = "arfg_off"
+
+/obj/machinery/atmospheric_field_generator/power_change()
+	var/oldstat
+	. = ..()
+	if(!(stat & NOPOWER))
+		ispowered = 1
```
Atermonera, line 105: Nit: TRUE/FALSE

code/game/machinery/atm_ret_field.dm
```diff
@@ -0,0 +1,246 @@
...
+		update_icon()
+		if(alwaysactive || wasactive)	//reboot our field if we were on or are supposed to be always-on
+			generate_field()
+	if(stat != oldstat && isactive && (stat & NOPOWER))
+		ispowered = 0
+		disable_field()
+		update_icon()
```
Atermonera, line 112: Are you calling `update_icon()` every time you `generate_field()` or `disable_field()`? Should those procs be the thing that calls `update_icon()`?

code/game/machinery/atm_ret_field.dm
```diff
@@ -0,0 +1,246 @@
...
+
+/obj/machinery/atmospheric_field_generator/power_change()
+	var/oldstat
+	. = ..()
+	if(!(stat & NOPOWER))
+		ispowered = 1
+		update_icon()
+		if(alwaysactive || wasactive)	//reboot our field if we were on or are supposed to be always-on
+			generate_field()
+	if(stat != oldstat && isactive && (stat & NOPOWER))
```
Atermonera, line 109: Er, where does `oldstat` get populated? Is it not just always `null`?

code/game/machinery/atm_ret_field.dm
```diff
@@ -0,0 +1,246 @@
...
+		if(3)
+			emp_act()
+	return
+
+/obj/machinery/atmospheric_field_generator/proc/generate_field()
+	if(!ispowered || hatch_open || !wires_intact || isactive) //if it's not powered, the hatch is open, the wires are busted, or it's already on, don't do anything
+		return
+	else
+		isactive = 1
+		icon_state = "arfg_on"
```
Atermonera, line 141: `update_icon()`

code/game/machinery/atm_ret_field.dm
```diff
@@ -0,0 +1,246 @@
...
+			src.visible_message("The ARF-G cracks and shatters!","You hear an uncomfortable metallic crunch.")
+			disable_field()
+		if(3)
+			emp_act()
+	return
+
+/obj/machinery/atmospheric_field_generator/proc/generate_field()
+	if(!ispowered || hatch_open || !wires_intact || isactive) //if it's not powered, the hatch is open, the wires are busted, or it's already on, don't do anything
+		return
+	else
```
Atermonera, line 139: Early return means you don't really need this else

code/game/machinery/atm_ret_field.dm
```diff
@@ -0,0 +1,246 @@
...
+		isactive = 1
+		icon_state = "arfg_on"
+		new field_type (src.loc)
+		src.visible_message("<span class='warning'>The ARF-G crackles to life!</span>","<span class='warning'>You hear an ARF-G coming online!</span>")
+		update_use_power(USE_POWER_ACTIVE)
+	return
+
+/obj/machinery/atmospheric_field_generator/proc/disable_field()
+	if(isactive)
```
Atermonera, line 148: Nit: `if(!isactive) return` instead

code/game/machinery/atm_ret_field.dm
```diff
@@ -0,0 +1,246 @@
...
+	ASSERT(istype(A))
+
+	LAZYADD(A.all_arfgs, src)
+	areas_added = list(A)
+
+	for(var/direction in cardinal)
+		A = get_area(get_step(src,direction))
+		if(istype(A) && !(A in areas_added))
+			LAZYADD(A.all_arfgs, src)
+			areas_added += A
```
Atermonera, line 175: Should this be a `LAZYADD()` too? Do we not care about duplicates in the list? (Suppose it's on a corner)

code/game/area/areas.dm
```diff
@@ -170,6 +174,25 @@
...
 
+// Activate all retention fields!
+/area/proc/arfgs_activate()
+	if(!arfgs_active)
+		arfgs_active = TRUE
+		if(!all_arfgs)
+			return
+		for(var/obj/machinery/atmospheric_field_generator/E in all_arfgs)
+			E.generate_field() //don't need to check powered state like doors, the arfgs handles it on its end
+			E.wasactive = TRUE
```
Atermonera, line 185: Should this get set every time `generate_field()` is called?

Atermonera (dismissed):
Oh hey I remember this.
I'm a little dubious of throwing `arfg` around everywhere because acronyms breed tribal knowledge, but I can also leverage it as a reason to call them "arfig"s ICly.

Review comments are a bunch of mostly pedantic nits, it looks like it should in fact do the thing. Up to you @KillianKirilenko if you want to address them.

code/game/machinery/atm_ret_field.dm
KillianKirilenko: When the field's active, you *can't* open the hatch to fiddle with it. You'd need to power enviro off at the APC first, in which case that's fine (in my book).

code/game/area/areas.dm
KillianKirilenko: `wasactive` is a slight hack that allows them to autoreboot if they're knocked out by an EMP or a low-power explosion. without it they stay offline.

code/game/machinery/atm_ret_field.dm
KillianKirilenko: This is a bit old so I think it predates the remove_fuel rework to have that notification hooked into the base proc itself.

Might need to rework that chunk a bit in general.

code/game/machinery/atm_ret_field.dm
KillianKirilenko: Up in `/machinery/proc/power_change()` by the looks of things? not sure why it was redefined here. dreammaker should be yelling at it being redefined too.

Atermonera (approved):

PR #8588, opened 2022-04-29
Adds Sjorgen Inertial Shotgun (by H-H)

Doctress:
Semi-automatic, 5 round, tube magazine fed
Lootable, or buyable from both cargo and tator uplinks.
Balancing intended to lie between combat shotgun and not!AA-12
100 TC each (between combat shotgun and fuel rod cannon)
100 Req Points for a box of two (same price as a burst laser pair)


Atermonera (approved):

Atermonera:
Was discussed at the 1-May staff meeting, with general approval. If it's found to be unbalanced, will apply hammer as necessary.

PR #8593, opened 2022-05-01
SSOverlays cuts queue from start to count, not count to end

Atermonera:
Maybe also fixes other overlays issues? But this is mostly an issue at high queue volumes, like after SSOpenspace initializes.


Atermonera:
Tested, no manual steps taken to produce this:
image

PR #8597, opened 2022-05-01
 Changes alien_whitelist to an assoc list of ckeys

Atermonera:
This makes me sad:
image

This change maybe addresses that? It can't be slower.

### DIFF
--- a/code/game/jobs/whitelist.dm
+++ b/code/game/jobs/whitelist.dm
@@ -25,8 +25,12 @@ var/global/list/alien_whitelist = list()
 	var/text = file2text("config/alienwhitelist.txt")
 	if (!text)
 		log_misc("Failed to load config/alienwhitelist.txt")
-	else
-		alien_whitelist = splittext(text, "\n")
+		return
+
+	var/list/whitelist_lines = splittext(text, "\n")
+	for(line in whitelist_lines)
+		var/list/entry = splittext(line, " - ")
+		LAZYADD(alien_whitelist[entry[1]], entry[2])
 
 /proc/is_alien_whitelisted(mob/M, var/datum/species/species)
 	//They are admin or the whitelist isn't in use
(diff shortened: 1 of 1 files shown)


code/game/jobs/whitelist.dm
```diff
@@ -42,12 +41,7 @@ var/global/list/alien_whitelist = list()
 		return TRUE
 
 	//If we have a loaded file, search it
-	if(alien_whitelist)
-		for (var/s in alien_whitelist)
-			if(findtext(s,"[M.ckey] - [species.name]"))
-				return TRUE
-			if(findtext(s,"[M.ckey] - All"))
-				return TRUE
+	return findtext(alien_whitelist, regex("[M.ckey] - [species.name]", "g"))
```
Spookerton, line 44: I don't think this will be dangerous in practical use and it's the easiest/fastest way to go about it as mentioned in code without pre-parsing - but `ckey(M.ckey)` may be preferable for paranoia since the ckey of a mob is mutable and stuff like the @ symbol is introduced to ckeys for some admin stuff like aghosting.

The second part of the regex is harder. We can't be sure someone won't put a `.` (for example) in a species (or language) name later, or some other symbol with special regex meaning that should be escaped but won't be when templated in verbatim.

code/game/jobs/whitelist.dm
```diff
@@ -14,19 +14,18 @@ var/global/list/whitelist = list()
...
+var/global/alien_whitelist = null
 
 /hook/startup/proc/loadAlienWhitelist()
 	if(config.usealienwhitelist)
 		load_alienwhitelist()
 	return TRUE
 
 /proc/load_alienwhitelist()
-	var/text = file2text("config/alienwhitelist.txt")
+	var/alien_whitelist = file2text("config/alienwhitelist.txt")
```
Spookerton, line 25:
```suggestion
	alien_whitelist = file2text("config/alienwhitelist.txt")

```

code/game/jobs/whitelist.dm
```diff
@@ -14,19 +14,18 @@ var/global/list/whitelist = list()
...
 	if (!text)
```
Spookerton, line 26:
```suggestion
	if (!alien_whitelist)
```

Spookerton (changes requested):

Spookerton:
Parsing it all once at load to a structure like `list(ckey = list(name, name, ...), ...)` may be preferable in the end, since then the matter of is_alien_whitelisted reduces to just
```
.../is_alien_whitelisted(mob/M, species/S)
  return (S.name in alien_whitelist[ckey(M.ckey)])
```

Spookerton:
nb: *because* this uses the single monolithic file approach, the `regex the whole thing` or `pre-parse the whole thing` approaches are necessary. Were it sqlite, json-per-user in player_saves, or some other oob store, you could load the data per-client, reducing work and making in-game editing of allow lists more reasonable.

PR #8604, opened 2022-05-02
Starts adding Wounds to Goosemed

Atermonera:
TODO, actually write down what this does


Atermonera:
Hmmm... nope, think this needs merge commits to keep the history clean. Will sort out tomorrow.

PR #8605, opened 2022-05-02
Starts adding Wounds to Goosemed

Atermonera:
Now without all the cruft of an attempted rebase!
TODO: Actually write out a desc here.


Atermonera:
This needs a different approach, the branch is way too out of date at this point

PR #8607, opened 2022-05-03
Expands Loadout Contraband

Sypsoti:
Does the following:
- A selection of 2 gasmasks + the 2 plague doctor masks added to the masks loadout tab.
- A selection of steel knives in the contraband tab, including the boxcutter.
- A phase pistol to the contraband tab.
- A zip gun to the contraband tab.
- A lesser pepperspray with 2 smaller sprays of 5 units, compared to the security standard 4 sprays of 10 units. Has a unique sprite.
- Changes the cost of the neural web to be cheaper, as it has a mechanical *disadvantage* if anything.

The zip gun works as it always has, it holds a single round of ammunition that is randomized when the gun initializes. You could get a decent single shot... or a beanbag.


Spookerton (approved):
Code's fine. Note that there are new loadout costs (and a couple of tweaks) involved. Not sure if discussed / bears particular attention.

Sypsoti:
Was discussed and approved at staff meeting for the record, @Spookerton so safe to merge

PR #8589, opened 2022-04-29
Map Tweaks - End of April 2022

Cerebulon:
Repathed cassette tapes: Fixes #8584 
Corrected Chapel Office Access: Fixes #8574 
Fixed first aid cabinets with empty kits.
Adjusted Anomaly Lab atmospherics computer to match cell alarms only. (Cells were missing corresponding alarms): Fixes #8553 
Fixed missing descriptions and names on holodeck stuff: Fixes #8566 and Fixes #8565 
Renamed poolside lockers.
Added two winter coats to the locker at the front door.
Changed the Anomaly Isolation Room doors to ones with windows.
Added two additional anomaly suits to Xenoarch for those bigger investigations:

PR #8617, opened 2022-05-06
added two coats to the science front door locker

Spookerton:
from #8589 by Cerebulon

Added two winter coats to the science front door locker


Sypsoti (approved):
Think you just meant the front door to the station, not science

PR #8622, opened 2022-05-07
Fixes directional overlays on padded comfy and rounded chairs

Cerebulon:
Rounded chairs were invisible when facing north because I didn't add the overlay-adding code to them apparently, and comfy chairs lacked their material colour when facing north for similar reasons. This didn't actually crop up until recently because nobody makes weird chairs much!

Also repathed roundeds as a subtype as comfies because they use exactly the same system and weren't mapped in anywhere anyway.


code/game/objects/structures/stool_bed_chair_nest/chairs.dm
```diff
@@ -93,12 +93,26 @@
...
 
-// Leaving this in for the sake of compilation.
 /obj/structure/bed/chair/comfy
+	name = "comfy chair"
 	desc = "It's a chair. It looks comfy."
 	icon_state = "comfychair"
 	base_icon = "comfychair"
 
+/obj/structure/bed/chair/comfy/Initialize()
+	. = ..()
```
Spookerton, line 103:
```suggestion
/obj/structure/bed/chair/comfy/update_icon()
	..()
```

Setting these in initialize only lasts until the next cut and rebuild which makes the north-facing state still bad when rotating.

Spookerton (changes requested):

Spookerton (approved):

PR #8626, opened 2022-05-08
Adds singularity artifact effect

Atermonera:


Atermonera:
Apparently this already exists. As an uncommon effect. With a 1/7 chance of affecting somewhere between twice world.view and the whole z-level...

Spookerton (changes requested):
Exist or not, I can't convince this to actually function. Touch triggered, 7 range, 7 strength (and higher, just to try). Doesn't drag anything around as-is *or* with `src` switched for `holder`.

(while we're at it, `for (var/atom/A `**`as anything`**` in range(...))` is probably preferable for this kind of thing in future - you can't not get atoms back, and it's 514)

Spookerton (approved):
Works.

@Atermonera this moves stuff toward the person activating the effect, rather than the artifact itself. Intended?

PR #8627, opened 2022-05-08
db flex tape

Spookerton:
**A**. removes unused forum sql configuration

**B**. adds some up-front config guards to save a little work on spamming db connect attempts even when sql is disabled

**C**. Adds SSmisc_slow, a 5 minute wait SS.

The only behavior of SSmisc_slow at the moment is DoReconnects, which handles connection maintenance for dbcon and dbcon_old by:
- checking if they're connected and sending a tiny safe query to keep them open.
- or if not, making sure they're closed, then reopening them with new handles.
- and if it fails this too many times, giving up softly.

With logging at every step.

This should remove the need for someone to push the fix database connection button by simply not allowing valid connections to time out when the lobby's been sitting empty for a half day stretch.

~~Draft because I want to make sure of DoReconnects locally first.~~


Spookerton:
For what it's worth, this is a bandaid if it works. Properly redoing mysql/maria handling, squashing the two db's together, and having a subsystem specific to connection maintenance and query interface is more appropriate for the role, but that would mean coordinating live data changes as well as writing it and I don't have time right now.

Atermonera (approved):

PR #8631, opened 2022-05-09
New blender sound

Serithi:
Don't know why the commits on my fork are all weird, just needs the blender one.


Spookerton (approved):
Audio is not balanced to the same peak/avg as before. Will be very quiet.

Audio was output probably as ogg-vorbis -q9. Can be q0 with no ear loss, and be 35kib instead of 175.

Audio has some trimmable silence.

eg:
[blender.zip](https://github.com/PolarisSS13/Polaris/files/8649231/blender.zip)

Aaand mangled commits. But this does work as-is.

Serithi:
@Spookerton

[blender.zip](https://github.com/PolarisSS13/Polaris/files/8666114/blender.zip)

Serithi:
Goddamn this do be messy commits though. Want me to just re-do the whole PR?

PR #8639, opened 2022-05-22
allow lists

Spookerton:
replaces the various file whitelist approaches with a commonly styled set and provides performance benefits over existing approaches by using GLOB.list("someckey" = list("allowed", "things", ...), ...) format instead of traversing and splitting the list matter every time.


Spookerton:
Closing since Ater's got an alternative behavior in.

PR #8663, opened 2022-07-10
Nerfs Fennec

Cerebulon:
Replaces fennec foxes with Fun Aliens instead of somethin 1. from earth (gross) 2. that's a bit of a meme and 3. real life exotic pet trade clickbait epidemic.
BUT still retains the spirit of the original PR (cute lil fluffy dude who is sometimes a pet) because I'm not trying to be a killjoy here.

It's still the same sprite because the sprite is fine and isn't like hugely obviously a 'real' fennec unless you label it as such. Now its an alien thats fennec adajent, but also kind of arctic foxy... But not white.

Catlore Approved
Also the zoom emote for these lil dudes is visible instead of audible because the zoomies is a state of mind and not a noise.


Ketrai:
Dear god. They ayylyums now. Hide your children-

Atermonera (approved):

PR #8666, opened 2022-07-11
Xenoarchaeology sprites n sounds

Cerebulon:
A selection of xenoarch-related sprites for:

- Hand picks and case
- Measuring tape
- Depth scanner
- Alden-Saraspova scanner
- Suspension field generator (+ panel and wrenched sprites)
- Anomaly container

And some sounds added to those devices. And microwaves because its the same sound (Will do actual wider-scale implementation of those sounds in a separate PR).
Also adds pick sets as a low level RnD product.

Via https://github.com/Baystation12/Baystation12/pull/32248

Did not port the adjustable-depth drills which are still under review on Bay, as we already have something extremely similar but with a very different look.

Some of these were just updated a couple of months ago but! Hey! These are nicer. Thanks spriter.


Atermonera (approved):

Atermonera (approved):
CI stinky, Ater approve!

PR #8671, opened 2022-07-26
conk rete

Greenjoe12345:
Adds concrete as a material, somewhat ported from bay/nebula (at least, in terms of the colour value and the texture overlay for it), obtainable via cargo or the stacks found in engineering, or by crafting it with 2 water, 2 silicate and 2 calcium per brick.
Cynosure and a number of POIs have been edited to have concrete walls, except in some special locations such as the toxins burn chamber and the supermatter core.
There is also a new type of flooring for mapping and POI use.


Atermonera:
Marking staff input: specifically re the find/replace of concrete on main station walls (It happens to be tomorrow so this isn't a huge delay)

Atermonera (approved):

PR #8675, opened 2022-08-07
Removes eject from the mecha radial.

MistakeNot4892:
Accidentally ejecting yourself during melee combat because the mech sprites are unnecessarily scaled up, and then dying because there's no channel on eject but there is on getting back in, sucks.


MistakeNot4892:
Unrelated CI fail.

Atermonera (approved):

PR #8684, opened 2022-08-21
Removes irrelevant gluttony regen code segment.

Cameron653:
This fires every life tick so it's impossible to increase your nutrition in the middle of this firing off. 

Additionally, it's impossible to have this ever return a negative value in any manner, since adjust_nutrition has a minimum cap of 0, meaning you won't ever go into a negative. 
Healing_amount will always return 10 (if above 9 nutrition), 1-9 (if your starting nutrition is 1-9) or 0 (if nutrition when this fired was 0)

Additionally, prevents it from trying to heal you if your healing_amount is 0. 
-0 multiplied by 0.025 is still zero.

### DIFF
--- a/code/modules/mob/_modifiers/unholy.dm
+++ b/code/modules/mob/_modifiers/unholy.dm
@@ -185,16 +185,14 @@
 			var/mob/living/carbon/human/H = holder
 			var/starting_nutrition = H.nutrition
 			H.adjust_nutrition(-10)
-			var/healing_amount = starting_nutrition - H.nutrition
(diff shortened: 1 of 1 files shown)


Cameron653:
I doubt this'll ever be used in the next 5 years since cult is practically nonexistent (**for good reasons**) but I just saw this segment of code in there and went "what" while trying to spend the next thirty minutes trying to figure out how in the seven hells you can get your nutrition to increase in the middle of this and be blessed with double regen. (Spoiler: You can't).

Also I wanted to optimize this some more but I didn't really see any good way of doing so, so I just left it as it was except for actually checking to make sure it actually has a healing_amount that isn't equal to 0.

Spookerton (approved):

PR #8685, opened 2022-08-21
cellcharge DoEffectAura sends a message per robot (and friends)

Spookerton:
post #8672 the unused turf var generated a warning; discovered that only the first robot recharged would get a message while checking it.

also gentle refactor for minimizing depth and readable var names


Atermonera (approved):

Spookerton:
Since I needed to rebase, I updated the PR to include cleaning up all the artifact effect files.

Atermonera (approved):

PR #8693, opened 2022-08-27
Even more drake stuff.

MistakeNot4892:
- Drake AI refinement. They won't consider cross-faction drakes enemies but also pack alphas.
- `IWasAttackedBy` moved into AI logic and generalized to `/mob/living` so drakes won't ignore other drakes biting their faces off.
- `trans_to_mob` has been split up into a mob proc so it can be overridden by drakes.
- Trained drakes can get some gear installed in their harness (radio, GPS, armour plate), though the GPS/radio need to be configured by a human. They start with a radio.
- Drakes can be injected with reagents and scanned with a health scanner.
- Reagents now have an `affect_animal` proc, though only drakes invoke it in Life().
- Bicaridine will heal drake scarring and brute damage.
- Generalized emote cooldown and sound broadcasting, made it use playsound_local and be directional.
- Summon whistle can be heard across the z-level.
- Drakes will slowly regenerate sap up to 10u.


Spookerton (approved):
Well there's a lot of it, but it all seems to work as intended.

PR #8703, opened 2022-09-11
Fixes cleanbot runtime spam

Verkister:
No need to bother looking for an obstacle in a nonexistent path

### DIFF
--- a/code/modules/mob/living/bot/bot.dm
+++ b/code/modules/mob/living/bot/bot.dm
@@ -250,7 +250,8 @@
 	return
 
 /mob/living/bot/proc/handleFrustrated(var/targ)
-	obstacle = targ ? target_path[1] : patrol_path[1]
+	if((targ && LAZYLEN(target_path)) || LAZYLEN(patrol_path))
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/bot/bot.dm
```diff
@@ -250,7 +250,8 @@
 	return
 
 /mob/living/bot/proc/handleFrustrated(var/targ)
-	obstacle = targ ? target_path[1] : patrol_path[1]
+	if((targ && LAZYLEN(target_path)) || LAZYLEN(patrol_path))
+		obstacle = targ ? target_path[1] : patrol_path[1]
```
Spookerton, line 254:
```suggestion
	if (targ)
		if (length(target_path))
			obstacle = target_path[1]
	else if (length(patrol_path))
		obstacle = patrol_path[1]
```

Spookerton (changes requested):

code/modules/mob/living/bot/bot.dm
Spookerton: @Verkister ?

PR #8707, opened 2022-09-25
Typing say into the input control will activate typing indicators

Atermonera:
For the folks like me that don't hit f3 every time they want to say something, but still would appreciate being able to use the typing indicator: https://puu.sh/Jn4ar/6e18129a22.mp4

Putting a `winget()` in Life() is probably a mistake, but that's why I wrapped it in a spawn. It does give a little bit of lag before it'll update, compared to using the `.say()` verb, but wyd. I had to add a latch around the `.say()` and `.me()` verbs to make sure the input checker wouldn't just disable the overlay immediately, because those only apply/remove it once, where the checker has to constantly watch the input control.

I made an attempt at checking to see if the input windows for `.say()` and `.me()` exist, and was utterly foiled by the available methods to determine if a control element exists (I guess `input()` does something else under the hood?). Honestly, that would've required another `winexists()` call or similar, which'd require another round-trip message to the client and back, so this implementation is probably for the best.

I'm admittedly wary of the `winget()` so kinda tempted to testmerge, but if other folks think it's fine as is, there's always the revert button (Plus trying to get a round going can be hard)

### DIFF
--- a/code/modules/mob/dead/observer/observer.dm
+++ b/code/modules/mob/dead/observer/observer.dm
@@ -125,7 +125,7 @@
 			add_overlay(H.overlays_standing)
 		default_pixel_x = body.default_pixel_x
 		default_pixel_y = body.default_pixel_y
-	if(!T && length(latejoin))	
+	if(!T && length(latejoin))
 		T = pick(latejoin)			//Safety in case we cannot find the body's position
 	if(T)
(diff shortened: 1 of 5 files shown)


Atermonera:
https://github.com/PolarisSS13/Polaris/pull/8722

PR #8708, opened 2022-09-26
A milieu of misc sprites

Cerebulon:
Some of these may be slightly controversial but generally an effort to replace some of the oldest most garish random stuff still poking around

New sprites for:

- Bomb assembly parts
- Medals
- Armbands
- Crowbars
- Wirecutters
- Wrenches
- Safes
- Gifts and giftwraps
- Essential clown crap
- Cheap lighters
- Paint buckets
- Pickaxes
- Drills
- Fire extinguishers
- Fire extinguisher cabinets
- Console screens
- Riot Shields
- Fireaxes
- Fireaxe cabinets
- Butterfly knives
- Bats (as in baseball)
- Tactical Knives
- Missiles
- Floor igniters
- Gas sensors
- Telescreens
- Trashcans (these aren't actually used)
- Noticeboards
- Retro TV
- Morgue Trays
- Captain's Display Case
- Most grenades

I think the most likely ones to cause alarm because they're so idk... Ingrained might be:
Wrenches, cutters, fireaxes and morgue trays.

New sprites from all over mostly hand-picked from CM where the art style doesn't clash hard with generally more suitable new-bay-stuff (and if Bay artists redo the same stuff they might get swapped out again if the level of detail/perspective feels off).
A handful of additions and edits by me tee hee


Cerebulon:
> My only criticism is that the new fireaxe does not match the new fireaxe closet, which better matches the old one.

Agreed and tweaked, it also didn't really match the inhands and honestly the desaturation didn't make a not of sense for an explicitly high-vis emergency tool.

Greenjoe12345:
Some feedback I have:
image
the open floor safe's perspective seems odd, it's looking directly down at it, which doesn't feel right, the old floor safe's open spite had a better angle for the open door.

image
the tacknife could do with being rotated -90 degrees to make it the same position the old one was

Spookerton:
@Cerebulon beep ^

Cerebulon:
> @Cerebulon beep ^

I saw greenjoe's comment but don't really agree. Majority of weapons ARE at a side angle and switching it 'back' to vertical is entirety arbitrary. The floor safe is still a significant upgrade over the existing and is visually consistent.

Spookerton (approved):
looks about right to me

PR #8713, opened 2022-09-28
Morgan Trading prosthetic expansion

Greenjoe12345:
Adds a sifwood version of the Morgan Trading (wood) prosthetic
Adds Teshari versions of the Morgan Trading prosthetics, both normal and sif wood.
image
image


Spookerton (approved):
wooden torsos and heads feels questionable, but it do thing

PR #8719, opened 2022-09-30
Admin Build Mode Agony Removal

Cerebulon:
Finally, you'll never know you're being meddled with by the sudden presence of random lockers!

> - Makes Basic build mode warn you before you delete the ground and space everyone
> - Makes Basic build mode turn the turf you deleted into the base turf for the area. If area base turf is not available, it goes with the base turf of the Z level.
> - Makes Advanced build mode start with Null. No more spawning lockers and qdel()ing players!
> - Fixes a longstanding bug where if you cancelled selection in build mode and then tried to select something from the list, it defaulted to a locker. Instead, if you cancel it, it defaults back to null and lists work like normal.

via https://github.com/VOREStation/VOREStation/pull/13820

### DIFF
--- a/code/modules/admin/verbs/buildmode.dm
+++ b/code/modules/admin/verbs/buildmode.dm
@@ -189,6 +189,7 @@
 	var/atom/movable/throw_atom = null
 	var/list/selected_mobs = list()
 	var/copied_faction = null
+	var/warned = 0
 
 /obj/effect/bmode/buildholder/Destroy()
 	qdel(builddir)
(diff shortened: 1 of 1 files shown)


Spookerton:
Probably better off looking at bay for build mode changes tbh. Polaris' and vore's modes are primitive and largely unmaintained for years.

Atermonera (approved):

Cameron653:
Oh hey, y'all ported and merged my PR before it even got merged into the branch I coded it for, ha.

There's another PR that deals with buildmode areas made via the room builder mode and lets you assign an actual area to the room you're creating if you want, and if you do, it fixes the lighting (for the most part) in the room instead of being vantablack.

If wanted, I can port it up here as well. It's part of me porting over /tg/ blueprints.

PR #8723, opened 2022-10-02
Fixes a tiny maybe-snafu with sstyping

Spookerton:
I thought I'd pushed this amendment, whoops. Re-adds mimic mobs having typing indicators if they do not already.

Should see how it looks live before merging or closing if not necessary.


Spookerton:
The last couple of meetings, people have said they haven't noticed anything wrong with mimic talking popups. Closing this since that's the case.

PR #8235, opened 2021-08-23
Firefighting Additions

Cerebulon:
New firefighting equipment and sprites, some minor buffs to make them more viable as firefighting equipment. Including non-voidsuit stuff for planetside firefighting.

Extinguishers now contain firefighting foam and are refilled using extinguisher refill tanks or single-use refill canisters. Foam is more effective at putting out fires than water and will suppress fire on tiles where it is spread.

Foam remains on tiles for a brief time but is not slippery so firefighters will not own themselves constantly.
Now we just need to actually have fires ;)

:cl:
rscadd - New firefighting gear sprites.
rscadd - Added enhanced firefighting gear for atmospheric techs and chief engineer.
tweak - Fire extinguishers now use foam instead of water.
/:cl:

PR #8739, opened 2022-10-09
Firefighting Gear

Cerebulon:
Smaller version of #8235 minus the foam stuff because there's a lot that needs to be changed around that and I haven't the time!
I think downstream earlyported these sprites anyway.
Basically some new sprites, and a fancy upgraded version of the gear for CE/AT


Spookerton:
Is the deletion of the `surgical-2` state from `icons/obj/clothing/suits.dmi` intended?

Cerebulon:
> Is the deletion of the `surgical-2` state from `icons/obj/clothing/suits.dmi` intended? Unmentioned is also the added 10'000K and 15'000K buffs to various atmos clothing _over_ the default fire gear temperature thresholds; this makes them functionally invulnerable to temperature.

Yes, [x]-2 states are duplicates

Spookerton (approved):

PR #8740, opened 2022-10-09
Adds Parkas

Cerebulon:
Adds 7 parkas.
5 basic colour variants, one slightly different 'vintage' one, and an NT branded variant.
Functionally identical to winter coats. Will be in loadout as part of the winter coat selection list.
Also added the NT variants to the outdoors lockers.
And winter gear in general to the back door locker which seems to have been an oversight.

Body sprites from CM13, hoods and hooded states by me because what's the point of a hoodless parka??


Spookerton (approved):

Spookerton (changes requested):
needs conflict resolution

Spookerton (changes requested):
aaand again

Spookerton (approved):

PR #8743, opened 2022-10-10
Basic Uniform Polish: Engineering

Cerebulon:
...And some assistants. Doing this in stages because it's a lot and some of the default uniforms might actually need more significant redesigns than this.

Mechanical tweaks:

- Added engineering overalls. Equivalent to webbing vests.
- Hazard vests can be toggled open/closed.

Sprites:

- Replacements for the basic uniform sprites for Engineer, Atmos Tech and CE.
- Less garish hard hat and hazard vests. But still bright!
- New uniforms are NT branded.

Default loadouts:

- Atmos techs now start with a blue hazard vest and hard hat.
- Engineers now start with an orange hazard vest.
- CE starts with white overalls attached to a jumpsuit.
- Assistants start with an 'assistant jumpsuit', which is far more visually similar to the old grey jumpsuit than the recently added dark-grey.
- Technical assistants start with a version of the assistant jumpsuit with small yellow accents.
- Both types of engineering lockers start with engineer overalls.

The visual differences are ultimately pretty minor, just colour and shading changes for the mostpart. 90% of people replace the defaults anyway, but there's no reason for the defaults to be ancient primary-coloured sprites!
Hard hats, accessory-overalls, and some hazard vests via Aurora. Adjustments to those, and other sprites by me.

Having slept on it, some revisions need doing:

- [x] General contrast re-do, they feel a little washed out right now because I toned down the shading from the base-colour suits but now that I do side by sides they lack depth compared to the old sprites.
- [x] Greenjoe's helmet thing.


Greenjoe12345:
the tesh version of the hardhats will want a slight edit so they're not wonky on teshari heads, due to the funkyness of teshari sprites (like how the old ones had two pixels in the middle vs one the new ones have)

Spookerton (approved):

Spookerton (changes requested):
needs conflict resolution; suggest waiting on #8740 since this will conflict with that as well.

Spookerton (approved):

PR #8747, opened 2022-10-13
Added a fairly important break.

MistakeNot4892:
Prevents the pod from always spawning 100 instances of itself when a crash survivor tries to join with no preexisting pod. :(


Spookerton (approved):
fear, terror even

PR #8750, opened 2022-10-14
Adding vox reagents, removing Vox Delight.

MistakeNot4892:
- Removed Vox Delight.
- Made vox 65% resistant to almost all toxins, medicines, alcohol and painkillers.
- Added vox-specific meds and alcohol, which are included in loadout and are excepted from the above.
- Removed special phoron chem interactions with vox.
- Added oxyloss (voxyloss) recovery to ammonia for vox.
- Refactored ethanol to respect species alcohol mod properly.
- Fixed an incidental problem where injected booze was applying the mod exponentially.


code/modules/reagents/reactions/instant/vox.dm
```diff
@@ -0,0 +1,67 @@
...
+
+/decl/chemical_reaction/vox_medication
+	name = "Vox Medication"
+	id = "vox_medication"
+	result = "voxmeds"
+	result_amount = 2
+	required_reagents = list(
+		"voxslurry" = 3,
+		"voxenzyme" = 1
+	)
```
Atermonera, line 21: Won't the slurry produced by the first reaction immediately react with the catalyzing enzyme to produce voxmeds?

code/modules/reagents/reactions/instant/vox.dm
```diff
@@ -0,0 +1,67 @@
+/decl/chemical_reaction/protoslurry
```
Atermonera, line 1: Do these recipes actually get instantiated? As I recall, I repathed the regular reactions as subtypes of /instant either to organizationally differentiate them from distillery recipes, _or_ because distillery recipes were causing an issue and doing so was prudent.

code/modules/reagents/reagents/vox.dm
```diff
@@ -0,0 +1,106 @@
...
+	description = "A complex serum composed of vox-manufactured nanomachines, complex protein chains and microscopic bundles of metallic fibers."
+	strength = 8
+	scannable = FALSE
+
+/datum/reagent/toxin/voxmeds/affect_touch(var/mob/living/carbon/M, var/alien, var/removed)
+	affect_blood(M, alien, removed)
+
+/datum/reagent/toxin/voxmeds/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
+
+	if(alien != IS_VOX)
```
Atermonera, line 46: Dubious about the ordering using a negative condition where everything else asserts the truthful comparison of `alien`, but probably fine here.

Atermonera (commented):
Looks mostly good, couple concerns.

Atermonera (approved):

PR #8761, opened 2022-10-17
A Short Investigation

Cerebulon:
...A brief case, if you will.

Adds a few briefcase variants to loadout. Unique sprite for the director's medal box.

Sprites via Aurora

### DIFF
--- a/code/game/objects/items/weapons/storage/lockbox.dm
+++ b/code/game/objects/items/weapons/storage/lockbox.dm
@@ -95,7 +95,7 @@
 	icon_locked = "medalbox+l"
 	icon_closed = "medalbox"
 	icon_broken = "medalbox+b"
-	item_state_slots = list(slot_r_hand_str = "briefcase", slot_l_hand_str = "briefcase)
+	item_state_slots = list(slot_r_hand_str = "briefcase", slot_l_hand_str = "briefcase")
 	req_access = list(access_heads)
 	storage_slots = 7
 	starts_with = list(


code/modules/client/preference_setup/loadout/loadout_utility.dm
```diff
@@ -1,9 +1,21 @@
 // "Useful" items - I'm guessing things that might be used at work?
 /datum/gear/utility
-	display_name = "briefcase"
-	path = /obj/item/storage/briefcase
+	display_name = "camera"
+	path = /obj/item/camera
```
Spookerton, line 4: camera?

code/modules/client/preference_setup/loadout/loadout_utility.dm
```diff
@@ -26,10 +38,6 @@
 		communicators[initial(communicator_type.name)] = communicator_type
 	gear_tweaks += new/datum/gear_tweak/path(sortAssoc(communicators))
 
-/datum/gear/utility/camera
-	display_name = "camera"
-	path = /obj/item/camera
-
```
Spookerton, line 32: no camera?

code/modules/client/preference_setup/loadout/loadout_utility.dm
listerla: changing base /datum/gear/utility to be camera so that the briefcase selection can be a child of it, otherwise every utility entry would inherit the selection

Spookerton: A filthy antipattern, but permissible since I don't want to unbreak it right now.

Spookerton (approved):

PR #8767, opened 2022-10-18
Adds Sivian Creole Language

Cerebulon:
Adds Sivian Creole.
Lore:

- Local language derived from a blend of GalCom and Scandinavian languages and spoken almost exclusively in Vir, particularly on Sif where the majority of original colonists settled and developed the local culture.
- No longer an "official language" that anybody is _expected_ to speak on a national scale, but still holding on among Vir Natives and probably something VirGov sticks on road signs in a smaller font to keep it alive.
- Hundreds of such languages probably exist in the galaxy, but it only makes sense to individually represent The Local One for people who want to play Locals or we'd be here all day to no mechanical benefit.

Mechanically:

- Language key is ,7
- Stok language has been moved to upper-case S (when has anybody spoken Stok ever anyway)
- About 40% understandable to speakers of GalCom or SolCom, as would be expected from a creole.
- Sounds like a mix of Swedish, Norwegian and Icelandic, with elements from Chinese and French, and a hint of Skrellian.
- Most borgs don't speak it by default.
- A desaturated cold blue colour that should show up well on light and dark modes. Geddit, because ice.

Spookerton pelase don't speedmerge people may have input on this one.

### DIFF
--- a/code/__defines/species_languages.dm
+++ b/code/__defines/species_languages.dm
@@ -96,6 +96,7 @@
 #define LANGUAGE_ZADDAT "Vedahq"
 #define LANGUAGE_PROMETHEAN "Promethean Biolinguistics"
 #define LANGUAGE_BLOB "Chemosense Transmission"
+#define LANGUAGE_SIVIAN "Sivian Creole"
 #define LANGUAGE_GIBBERISH "Babel"
 
 // Language flags.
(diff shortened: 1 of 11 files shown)


Spookerton (approved):
code approve etc

Cerebulon:
> code approve etc

have asked for feedback in staff channels and nobody has any issues with the addition so can prolly take the input tag off

Atermonera:
Can confirm, staff input is "if it doesn't break things then yeah sure"

PR #8772, opened 2022-10-19
Fixes a bug with deleted disposal outlets

Verkister:
Fixes deleted disposal outlets not unlinking their trunks, causing anything that the bare trunk tries to eject to runtime and get qdel'd instead.


Spookerton (approved):
It's vaguely horrifying that these aren't aware of their pipes.

PR #8776, opened 2022-10-21
Snowmobiles and Quadhancements V2

Cerebulon:
#8543 without the absolute mountain of redundant code.

Riding datums for passengers are still a teensy bit wonky, but that isn't unique to this by any means.


code/modules/vehicles/quad.dm
```diff
@@ -154,6 +167,18 @@
...
+	..()
+	if(on)
+		src.visible_message("\The [src] rumbles to life.", "You hear something rumble deeply.")
+		soundloop.start()
+
+/obj/vehicle/train/engine/quadbike/turn_off()
+	if(on)
+		src.visible_message("\The [src] putters before turning off.", "You hear something putter slowly.")
+		soundloop.stop()
+	..()
```
Cerebulon, line 180: If this seems nonsense, it runs before on is set to 0 so that the putter message doesn't display on init

code/modules/vehicles/construction.dm
```diff
@@ -272,3 +273,103 @@
...
+					playsound(src, W.usesound, 30, 1)
+					var/obj/vehicle/train/engine/quadbike/snowmobile/built/product = new(src)
+					to_chat(user, "<span class='notice'>You finish \the [product]</span>")
+					product.loc = get_turf(src)
+					product.cell = cell
+					cell.forceMove(product)
+					cell = null
+					user.drop_from_inventory(src)
+					qdel(src)
+				return
```
Atermonera, line 375: Missing `. = ..()` to handle whacking it with other things?

code/modules/vehicles/quad.dm
```diff
@@ -49,12 +57,16 @@
 
 /obj/vehicle/train/engine/quadbike/Moved(atom/old_loc, direction, forced = FALSE)
 	. = ..() //Move it move it, so we can test it test it.
-	if(!istype(loc, old_loc.type) && !istype(old_loc, loc.type))	//Did we move at all, and are we changing turf types?
+	get_turf_speeds(old_loc)
+	handle_vehicle_icon()
+
+/obj/vehicle/train/engine/quadbike/proc/get_turf_speeds(atom/prev_loc)
+	if(!istype(loc, prev_loc.type) && !istype(prev_loc, loc.type))	//Did we move at all, and are we changing turf types?
```
Atermonera, line 64:
```suggestion
	// Same speed if turf type doesn't change
	if(istype(loc, prev_loc.type) || istype(prev_loc, loc.type)
		return
```
And then dedent the rest of this proc.
Low-key wishing there was some way to attach this to the turf type but then snowmobiles are supposed to do better on snow so that's ultimately tied to the mover more than the terrain

code/modules/vehicles/quad.dm
Atermonera: As dumb as I find the logic that's happening on the parent call.... I think it's out of scope to ask that you make it not-dumb for the rest of vehicles/trains.

code/modules/vehicles/quad.dm
```diff
@@ -28,8 +30,10 @@
 /obj/vehicle/train/engine/quadbike/Initialize()
 	cell = new /obj/item/cell/high(src)
 	key = new key_type(src)
+	soundloop = new(list(src), FALSE)
 	. = ..()
```
Atermonera, line 34: As a train, this parent call _appears_ to invoke the following:
```
/obj/vehicle/train/Initialize()
	. = ..()
	for(var/obj/vehicle/train/T in orange(1, src))
		latch(T, null)
```
Can you test what happens if a quadbike is spawned in adjacent to another? (Or to other train-type things?)

Cerebulon: It does indeed attach stuff that's spawned in such that they meet the latch proc requirements (/train type, facing the same way) which seems intentional with quads/trailer, but have added a check so it can not do that for train subtypes, and won't for snowmobiles.

code/modules/vehicles/construction.dm
Cerebulon: Added at top instead, in keeping with the quad

code/modules/vehicles/snowmobile.dm
```diff
@@ -0,0 +1,80 @@
...
+/obj/vehicle/train/engine/quadbike/snowmobile/Initialize()
+	. = ..()
+	riding_datum = new riding_datum_type(src)
+
+/obj/vehicle/train/engine/quadbike/snowmobile/built/Initialize()
+	dir = 2 //To match the under construction frame
+	. = ..()
+
+/obj/vehicle/train/engine/quadbike/snowmobile/get_turf_speeds(atom/prev_loc)
+	if(!istype(loc, prev_loc.type) && !istype(prev_loc, loc.type))	//Did we move at all, and are we changing turf types?
```
Atermonera, line 42: Here too with the dedent stufff

Atermonera (changes requested):

Atermonera (approved):

PR #8784, opened 2022-10-23
Nudges sif's resting temperature

Atermonera:
I previously made it so that pressure from planetary railroading turfs would scale inversely with temperature.
I did not account for the part where sif's default temperature was _-30C_. This means that all of our weather is _warmer_ than this value, such that the pressure is expected to _decrease_.

Let's move the base temperature to 0C and see if that helps to fix the issue of poplung.


Spookerton (dismissed):
Pretty big difference, but easily returned to a lower value for lore raisins later

Spookerton (approved):

PR #8785, opened 2022-10-23
size traits don't have slowdowns

Spookerton:
Having a wider mob sprite no longer makes you arbitrarily slower.

Pretty simple- no other physical trait does this, and you'll already get faster slowdowns from being hungry more often. The "1.1" and "1.2" values also just get rounded to 1, which *halves* human speed.

### DIFF
--- a/code/modules/mob/_modifiers/traits.dm
+++ b/code/modules/mob/_modifiers/traits.dm
@@ -50,15 +50,15 @@
 	name = "High Metabolsim"
 	desc = "Your body's metabolism is faster than average."
 
-	metabolism_percent = 2.0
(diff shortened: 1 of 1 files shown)


Cerebulon:
Maybe want to take the health bonus off too so it isn't a straight positive?

Spookerton:
the actual bonus tends to be less than a single hit from a small weapon, but I can do that for sure.

Spookerton:
addressed that by removing the health and metabolism differences from size modifiers entirely. If you want to be hungrier but get more healing, you can with high metabolism - or the opposite with slow.

Atermonera (approved):

PR #8786, opened 2022-10-23
Standardized area flag defines, added vendor rampancy flag check.

MistakeNot4892:
Should prevent vendors in PoIs or other inaccessible sites being infected for the event.


Spookerton (approved):
pass do thing. The only reason area flags should fail is if it's being improperly used elsewhere, and that's not this PRs problem.

PR #8787, opened 2022-10-24
Deletes All Bots

Atermonera:
Every now and again the server locks up. And after spending 10-20 minutes of trying to figure out what the issue is each time, it _always_ ends up being a bot, somewhere, frantically trying to call A* to path to something that it can no longer get to.
I refuse to fix bots. They are a relic of bygone eras of code. If they are to be reintroduced, it must be as simple_mobs.


Cerebulon:
Has there been an instance of this in the past several years that wasn't **just** beepsky? Nuking an entire, regularly used system (esp medbots and farmbots) seems entirely unnecessary for a once in a blue moon bug which is almost universally one specific bot which uses hard waypoints.

Atermonera:
We've definitely had issues with cleanbots in the past, but it's much rarer because a) people don't make them and b) they don't move around as much as beepsky

PR #8788, opened 2022-10-24
Nerfs mech armor

Atermonera:
Staff have noticed over the past several combat-focused events that mechs invalidate all of the simple-mob enemies that staff favor for events.
Let's try to do something about that.


Atermonera:
Staff input from the last meeting was: go for it

PR #8789, opened 2022-10-24
Enforcing no sleep during mob/Life(), tick checking AStar().

MistakeNot4892:
Alternative to #8787. Opening for diff, still need to test bot pathfinding changes.

- Adds linter instruction to mob/Life() to issue a warning on sleep. Sleeping during Life() sleeps the entire SSmobs subsystem and is undesirable in the first place.
- Removes a `spawn(0)` from `bot/Life()` that was preventing it from being tick checked properly at all.
- Adds various `set waitfor = FALSE` alternatives to sleeps in Life().
- Removes a check against `radio_controller` being null - this should no longer be relevant as `New()` isn't firing any code and SSatoms will initialize long after the startup hook.
- Adds a sanity counter and tick checking to A* as it can't reasonably be factored into bot pathing due to the spaghetti.

One possible issue that will need testing will be whether blobs now spread too fast - when we fixed the sleep() issue in blobcode on Neb, they started expanding more aggressively than intended since they were previously limited by their own sleep() proc (which was also limiting _every other mob in the game_).

The bot pathfinding solution is absolutely horrible, but a more reasonable implementation using movement callbacks or `walk_to` is not something I am interested in committing to for Polaris. This should (still need to test) at least prevent it from locking the server up.

ed:
closes #8787 (for now, if accepted)

### DIFF
--- a/code/defines/procs/AStar.dm
+++ b/code/defines/procs/AStar.dm
@@ -134,7 +134,10 @@ length to avoid portals or something i guess?? Not that they're counted right no
 
 	open.Enqueue(new /PathNode(start, null, 0, call(start, dist)(end), 0))
 
-	while(!open.IsEmpty() && !path)
+	var/how_many_iter = 0
+	var/sanity = 500 // From some local testing, AStar() tends to hit tick check or succeed around the 450 mark.
+	while(!open.IsEmpty() && !path && sanity)
+		sanity--
 		var/PathNode/current = open.Dequeue()
 		closed.Add(current.position)
 
@@ -177,5 +180,8 @@ length to avoid portals or something i guess?? Not that they're counted right no
 
 			if(max_nodes && open.Length() > max_nodes)
 				open.Remove(open.Length())
+
+		if(TICK_CHECK)
+			return // Took too long, give up.
 
 	return path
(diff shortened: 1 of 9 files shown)


code/defines/procs/AStar.dm
```diff
@@ -134,7 +134,10 @@ length to avoid portals or something i guess?? Not that they're counted right no
 
 	open.Enqueue(new /PathNode(start, null, 0, call(start, dist)(end), 0))
 
-	while(!open.IsEmpty() && !path)
+	var/how_many_iter = 0
+	var/sanity = 500 // From some local testing, AStar() tends to hit tick check or succeed around the 450 mark.
```
Spookerton, line 138: commentary only- very dependent on architecture and load of the host device. I live in fear of the day of byond cpu measures. DUNGMIPS perhaps.

Spookerton (approved):

Spookerton:
For what it's worth (ie, very little, since this is mostly about a failure case) it is actually very possible to dramatically improve the caching and performance of the A* implementation widely used in ss13, including here, with nearly no structural changes. It would not resolve this issue though - it would just thrash more efficiently.

I have approved this for code but have made no attempt to stare at a spreading blob since it's 2240 and I'm quite drunk.

Spookerton:
I'm marking this as staff input since ultimately whether to slash and burn or try a potentially cheaper option is on ater. Code's fine.

Atermonera:
Ater's input is: This literally bites us like, once every 6mo and the next time we'll run into this we'll probably have forgotten about it again. If this makes that longer, then hooray. :+1:

PR #8792, opened 2022-10-24
Ports Aurora hijab sprite

listerla:
Credit: https://github.com/Aurorastation/Aurora.3/pull/11433

I found this much easier to color properly because they're actually white, and I didn't like how the current ones become much darker in the side views - thus making it feel a lot harder to match with outfits.

Whether or not this is a good change should probably be considered, because they're actually fairly different.


Spookerton (approved):
the difference in shading style adds some noise, but that's probably a good thing considering it's supposed to be wrapped cloth and not a plastic orb.

PR #8803, opened 2022-10-30
fix atom init behavior

Spookerton:


Spookerton:
We're just at that full refactor level again, huh

PR #8804, opened 2022-10-30
Craftable poncho and cloaks.

MistakeNot4892:
- Ponchos/cloaks are now storage items with two slots.
- You can mash food up in buckets. This is primarily so you can get sifsap out in the wilds via a bucket and a rock.
- You can trim wood and chitin armour plate inserts down with a sharp object rather than a welding torch.
- You can craft ponchos and cloaks from leather/cloth and plate them with armour plates.

image
image
image


code/modules/clothing/under/accessories/crafted/poncho.dm
```diff
@@ -0,0 +1,174 @@
+/obj/item/clothing/accessory/storage/poncho/crafted
+	name = "poncho"
+	desc = "A handmade poncho, little more than a rough sheet with a neckhole in it."
+	icon_state = "poncho_base"
+	item_state = "poncho_base"
+	default_material = MAT_CLOTH
```
Spookerton, line 6:
```suggestion
	default_material = MAT_CLOTH
	material_slowdown_multiplier = 0
```

Spookerton (changes requested):

Spookerton (approved):

PR #8811, opened 2022-10-31
add atom/Topic & atom/CreateAtomTopic + examine

Spookerton:
I was originally going to do a suggestion but ended up writing an alternative since it would've been a very big suggestion.

atom/Topic and atom/CreateAtomTopic are provided as a pair with basic source verification built in. *This verification is not currently applied for subtypes of atom because their own topic links do not pass the appropriate reference.*

adds the AtomTopicExamine behavior & related code in atom examine. Examined atoms that do not display all of their available information in the chat box will indicate that they have more information available and provide a link to switch to the examine tab of the stat panel. This is applied for both preferences where the examine tab is not opened automatically.

fixes a couple of display issues with examine behavior, including a blank line showing when usage information should be included but none was available.

~~closes 8808~~
reverts #8808

### DIFF
--- a/code/__defines/misc.dm
+++ b/code/__defines/misc.dm
@@ -317,6 +317,8 @@ var/global/list/##LIST_NAME = list();\
 
 #define SPAN_OCCULT(X) "<span class='cult'>[X]</span>"
 
+#define SPAN_GOOD(X) "<span class='green'>[X]</span>"
+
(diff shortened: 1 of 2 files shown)


Spookerton:
need to handle resolution before this is viable again; I'll get to it tomorrow

Atermonera (approved):

PR #8816, opened 2022-11-05
Reduces Teshari speed to Tajara/Vox equivalent.

MistakeNot4892:
Current speed is ridiculous and their frailty is compensated for by their ability to bypass tables/grilles, their small icon, and their gliding.


Cerebulon:
Teshhead thumbs up. They were fine before the base speed was increased but with the baseline bumped up they're just impossible to control and deal with. Plus diagonal movement enabling them to move across tabley areas like the bar in half a second flat.

Spookerton (approved):

PR #8780, opened 2022-10-22
Basic Uniform Polish: Civilian

Cerebulon:
...And some civilian-adjacent concepts.
Same deal as #8743

Mechanical Tweaks:

- Nothing really, the basic space suit no longer hides your head or face because it's a big ol' bowl.

Sprites:

- New branded basic uniforms for Botanist and Janitor (sprites by me)
- Slightly upgraded uniform sprites via Fulp for Chef and Chaplain.
- Slightly upgraded chef suit and apron sprites also via Fulp.
- New emergency softsuit and 'basic' voidsuit sprites via Fulp (suits) and Bay (Space helmet)
- Replacements for several suits. This is civilian because librarians and IAA wear these a lot. Sprites via Fulp again.

Default loadouts:

- Librarian starts in a tan suit, because the red suit sucks.
- Survivalists start in grey jumpsuit now that this is both not the "assistant uniform" and is a different shade from classic greytiders. Previously they were in new-age-cult-garb which was weird.
- Cooks start with an apron and hairnet instead of the fancy-pants chef gear.

Other:

- A couple more suit options.
- Removed the scratch suit because it was _really_ bad.
- Did you know that apparently no server (which has botanists) has done anything with the blue apron sprite?

PR #8818, opened 2022-11-06
Basic Uniform Polish: Cargo

Cerebulon:
...And their friends, the truck.

Same deal as #8743 and #8780 but just a wee one.

Sprites:
- New redesigned uniforms for Cargo Technician, Quartermaster and Shaft Miner by me. Cargo is themed Brown-ish Orange and now their uniforms reflect that.
- New workboot sprites (and variants) via Aurora.
- Slight edit to the previously added mining overalls, added a high-vis strip to the legs.
- Mulebot and cargo tug sprites from... uh. Somewhere? Fulp maybe? A few servers have em.

Loadouts:
- Miners start with high-vis-striped dark overalls.
- Cargo staff start with workboots instead of generic/brown shoes.
- Added overalls to cargo lockers, and swapped out generic black gloves for duty gloves.
- Removed some of the alt uniforms from lockers because there's no need for that.
- Added all cargo alt uniforms (jeans, skirt, shorts) to loadout under consolidated lists. Shorts = the old cargo tech uniform.
- Workboot variants (all access) and utility overalls (cargo, engineering) added to loadout.
- Old mining uniform is also available under loadout if you really liked it for some reason.

All other sprite deletions are intentional and were unused. The tesh rename fixes a sprite broken in the last uniform polish thing.


code/datums/outfits/jobs/cargo.dm
```diff
@@ -19,7 +19,7 @@
 
 /decl/hierarchy/outfit/job/cargo/mining
 	name = OUTFIT_JOB_NAME("Shaft miner")
-	uniform = /obj/item/clothing/under/rank/miner
+	uniform = /obj/item/clothing/under/rank/miner{ starting_accessories=list(/obj/item/clothing/accessory/storage/overalls) }
```
Spookerton, line 22: Anonymous subtyping ("modified types" in dm ref) should be avoided where possible (ie, everywhere that is not in a dmm). It's very unclean for maintenance and creates hidden unique things.

Spookerton (approved):

code/datums/outfits/jobs/cargo.dm
Cerebulon: So what would be preferable? Creating a subtype of under/rank/miner (in this case) called like under/rank/miner/startgear that has the starting accessory gear defined, exclusively for using in the job outfit decls? I did it this was because there were a couple of preexisting outfits that used this quick tag-on.

PR #8819, opened 2022-11-07
Replaces whitelisting by direct datum path with decls stored by uid

Atermonera:
From a sample unversioned whitelist of
```
{"/datum/species/tajaran":1,"/datum/species/unathi":1,"/datum/language/unathi":1,"/datum/language/tajaran":1,"/whitelist/genemod":1,"/datum/species/zaddat":1,"/datum/language/skrell":1,"/datum/language/tajsign":1,"/datum/species/skrell":1}
```
this was translated to
```
{"version":1.1, "whitelists": ["species_taj", "species_unathi", "lang_unathi", "lang_taj", "genemod", "species_zaddat", "lang_skrell_common", "lang_taj_sign", "species_skrell"]}
```
And printing it out
```
You are whitelisted for: 

Tajara
Unathi
Sinta'Unathi
Siik
Genemods
Zaddat
Common Skrellian
Alai
Skrell 
```

I also tweaked the input selection for adding/removing whitelists to be more human-readable:
image
(Tho the lookup back and forth is a little ugly, and I refuse to make the display_name into the uid because the idea that sparked all of this was handling special characters in whitelisted-thing-names)


Atermonera:
@Spookerton does this close #8639?

code/modules/whitelist/whitelist_decl.dm
```diff
@@ -0,0 +1,79 @@
...
+
+/decl/whitelist/species/promethean
+	display_name = "Promethean"
+	uid = "species_promethean"
+
+/decl/whitelist/species/zaddat
+	display_name = "Zaddat"
+	uid = "species_zaddat"
+
+/decl/whitelist/species/vox
```
Atermonera, line 77: Calling out explicit support for vox here since things have to be deliberately enumerated in code now

Spookerton:
> @Spookerton does this close #8639?

It does if you want this behavior going forward! It's a little less flexible, I think, but given that I haven't had the time/energy to finish actually tying allow lists into where they'd be used, what you have works and what I was working on is not complete. :P

Atermonera:
The constraint of using decls as an intermediate between text strings stored in a json file and actual paths of things being whitelisted still feels clunky to me. At least, compared to my initial implementation whereby it was storing the paths-whitelised directly. But this is definitely a more extensible framework to account for things being repathed than that original impl.
At least with a proper versioning system that I've sorta-implemented here, it should be much more reasonable to translate from one implementation to another moving forwards.

Atermonera:
Closing this, will take a separate stab at the system

PR #8821, opened 2022-11-07
Cleans up the storage accessory attackby path and prevents double insertion of items into storage accessories.

MistakeNot4892:
`attackby()` should always return `TRUE` if it has handled the attackby.


Spookerton (approved):
Seems functional to me!

PR #8823, opened 2022-11-09
Cleaning up species alcohol mod.

MistakeNot4892:
- Alcohol mod replaces a block of reagent tag checking.
- Alcohol mod applied consistently (no more exponential mod application).


Spookerton (approved):
Alcohol effects seem slow to me, but apparently this was because of previous effect duplication. Does the thing.

PR #8828, opened 2022-11-11
GPS has a discrete SetTag proc

Spookerton:


Atermonera (approved):

Atermonera:
For a-few-lines-changed PRs, generally not. It's the several-hundred-lines-changed PRs that tend to hang me up since I need to sit down and properly review them, requiring time and proper focus.

PR #8835, opened 2022-11-15
The dead can't rename food.

Mechoid:
The tin.

### DIFF
--- a/code/modules/food/food.dm
+++ b/code/modules/food/food.dm
@@ -21,7 +21,7 @@
 
 /obj/item/reagent_containers/food/proc/handle_name_change(var/mob/living/user)
 	if(!isliving(user) || user.stat == DEAD)
-		to_chat(user, SPAN_NOTICE("The dead can't cook!"))
+		to_chat(user, SPAN_WARNING("The dead can't cook!"))
 		return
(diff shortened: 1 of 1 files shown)


code/modules/food/food.dm
```diff
@@ -20,6 +20,9 @@
 	handle_name_change(usr)
 
 /obj/item/reagent_containers/food/proc/handle_name_change(var/mob/living/user)
+	if(!isliving(user) || user.stat == DEAD)
```
Spookerton, line 23: I would say this should be limited to humans and silicons, and so prefer
`user.stat == DEAD || !(ishuman(user) || isrobot(user))`
just to eke out that tiny stat-first and short circuit benefit

code/modules/food/food.dm
```diff
@@ -20,6 +20,9 @@
...
+		to_chat(user, SPAN_NOTICE("The dead can't cook!"))
```
Spookerton, line 24:
```suggestion
		to_chat(user, SPAN_WARNING("The dead can't cook!"))
```

Spookerton (changes requested):

Spookerton (approved):

PR #8837, opened 2022-11-16
Copied the platform access overrides to cynosure.

MistakeNot4892:
These were missed in the initial map inclusion I think.

### DIFF
--- a/maps/cynosure/cynosure_overrides.dm
+++ b/maps/cynosure/cynosure_overrides.dm
@@ -1,0 +1,9 @@
+/mob/living/silicon/robot/platform/explorer
+	req_access = list(access_explorer)
+
+/mob/living/silicon/robot/platform/cargo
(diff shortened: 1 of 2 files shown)


maps/cynosure/cynosure_overrides.dm
```diff
@@ -0,0 +1,10 @@
+/mob/living/silicon/robot/platform/explorer
+	req_access = list(access_explorer)
+
+/mob/living/silicon/robot/platform/cargo
+	req_access = list(access_cargo_bot)
+
+/obj/item/card/id/platform/Initialize()
+	. = ..()
+	access |= access_explorer
+	access |= access_pilot
```
Spookerton, line 10: access_pilot appears to only exist in southern cross

Spookerton (approved):

PR #8846, opened 2022-11-26
Adds more generic clothing paintkit.

Atermonera:
Why yes, this is nearly a carbon copy of the poncho kits. Those have weird hardcoding and the typepath implies the intended usage. I didn't modify the poncho kits, nor their path, to avoid cross-coupled changes in the custom item repo. We may eventually migrate all the poncho kits to the clothing kits.

These use the base kit type's function of specifying in the custom item entry the paths that can be modified.

### DIFF
--- a/code/game/objects/items/paintkit.dm
+++ b/code/game/objects/items/paintkit.dm
@@ -60,8 +60,8 @@
 	..()
 
 /// Ponchos specifically, but other accessories may eventually need this.
-/// Consequence of snowflake-y teshari code. 
-/obj/item/kit/accessory 
+/// Consequence of snowflake-y teshari code.
+/obj/item/kit/accessory
 	name = "accessory modification kit"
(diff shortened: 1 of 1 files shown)


code/game/objects/items/paintkit.dm
```diff
@@ -78,6 +78,17 @@
 			to_chat(user, "You set about modifying the poncho into [new_name].")
 		return ..()
 
+/// General clothing items, that need to set item_state..
+/obj/item/kit/clothing
+	name = "clothing modification kit"
+	desc = "A kit for modifying clothing."
+
+/obj/item/kit/accessory/customize(var/obj/item/clothing/I, var/mob/user)
```
Sypsoti, line 86: clothing, not accessory.

Sypsoti (changes requested):
The curse of the copypasta.

Sypsoti (approved):

PR #8853, opened 2022-12-03
Fixes shards having unlimited afterattack range

Verkister:
Fixes glass shards and such being able to be smacked into anything on the screen regardless of the targets location to shatter and mess up your hands.

### DIFF
--- a/code/game/objects/items/weapons/material/shards.dm
+++ b/code/game/objects/items/weapons/material/shards.dm
@@ -55,7 +55,9 @@
 			return
 	return ..()
 
-/obj/item/material/shard/afterattack(var/atom/target, mob/living/carbon/human/user as mob)
+/obj/item/weapon/material/shard/afterattack(var/atom/target, mob/living/carbon/human/user as mob, proximity)
+	if(!proximity)
+		return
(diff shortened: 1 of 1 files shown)


Spookerton:
This causes the problem to be hidden, not fixed, by disinheriting the proc override from shards so that only obj/item/afterattack is run. There is no /obj/item/weapon/ here. The proc logic is the problem, not the path.

Spookerton:
Misfire, lol.

code/game/objects/items/weapons/material/shards.dm
```diff
@@ -55,7 +55,9 @@
 			return
 	return ..()
 
-/obj/item/material/shard/afterattack(var/atom/target, mob/living/carbon/human/user as mob)
+/obj/item/weapon/material/shard/afterattack(var/atom/target, mob/living/carbon/human/user as mob, proximity)
```
Spookerton, line 58:
```suggestion
/obj/item/material/shard/afterattack(atom/target, mob/living/carbon/human/user, proximity)
```

Spookerton (changes requested):

Spookerton (approved):

PR #8857, opened 2022-12-12
Xmas Expansion: Snowy Cynosure

Greenjoe12345:
Adds Christmas decorations to Cynosure station and the accessible area of the Northern Star, in addition it adjusts the area around the station to be fully snowy, The river and lake have mostly frozen over.


Spookerton (commented):
@Greenjoe12345 

Could possibly stand to have some touch-ups where fake-snow is used; the lack of transitions is visually offputting and throwing down some decals wouldn't be a huge amount extra.
eg 

If doing anything about the above, resetting and making the change to christmas.dmi its own commit would save a tiny fraction of revert effort later.

Spookerton (dismissed):

PR #8865, opened 2022-12-24
Markings expansion 2: Teshari boogaloo

Greenjoe12345:
Ports a bunch of Teshari markings from downstream, adjusting them to work here.
Teshari also no longer by default have colored feathers on their hands/feet, those are now added as markings.
This is due to the fact that those organ-defined feather markings displayed above any others, making it very hard to actually have markings on your limbs as Tesh.


code/modules/mob/new_player/sprite_accessories_tail.dm
```diff
@@ -262,6 +262,20 @@
 	extra_overlay = "seromitail_feathers_s"
 	species_allowed = list(SPECIES_TESHARI, SPECIES_EVENT1, SPECIES_EVENT2, SPECIES_EVENT3)
 
+/datum/sprite_accessory/tail/special/tesharitail_patern_1
```
Atermonera, line 265: `patern_1`

Atermonera (approved):

PR #8882, opened 2023-01-13
Refactoring mines and simplying disarm.

MistakeNot4892:
- `/obj/item/mine` and `/obj/effect/mine` have been unified.
- `/datum/mine_payload` now handles mine effects.
- Mines now have a chance of cloaking themselves when primed on an appropriate turf, rather than based on mine subtype.
- Mines no longer have wires and are disarmed by opening the panel with a screwdriver for 5 seconds and then using a wirecutter on them for 10 seconds. Messing up either action (or clicking an armed mine with the wrong tool) will detonate the mine.

I still need to test the assembly mine so DNM for now, opening for diff and review.


code/game/objects/items/mines/_mine.dm
```diff
@@ -0,0 +1,186 @@
...
+		visible_message(SPAN_NOTICE("\The [src] beeps as the priming sequence completes."))
+		user.drop_from_inventory(src, get_turf(user))
+		add_fingerprint(user)
+
+	anchored = TRUE
+	armed = TRUE
+
+	if(isturf(loc) && prob(65))
+		for(var/turftype in concealable_turfs)
+			if(istype(loc, turftype))
```
Atermonera, line 132: `if(is_type_in_list(loc.type, concealable_turfs))`?

Atermonera (approved):

PR #8887, opened 2023-01-18
Another take at a new whitelist system

Atermonera:
There's extra work to throw together a python script or smth on the server box to convert the current format.
I considered using decls but it seems like more dev overhead than just maintaining a list of things that can be whitelisted for something that'll hardly ever be touched.

I tested this locally, both using admin buttons to modify the whitelist in-game (Which now uses a drop-down of whitelistable things so no more spelling errors are possible unless people give bad ckeys), and using the included python script to scrape player save dirs for whitelists and assemble the unified list in config.


Atermonera:
@Spookerton this is ready for review. Update blue-line to run the python script on the server box.

PR #8892, opened 2023-01-21
Cleans up drop_from_inventory(), fixes drake harness population.

MistakeNot4892:
- Cleans up `drop_from_inventory` overrides.
- Removes owner set from drake harness init since it was not being set by the time equipment was created.


Spookerton (approved):

Spookerton:
I think moving owner = loc above . = .. would've done the trick, but this does provide some extra reference security either way

PR #8902, opened 2023-01-26
Hydroponics - Braincaps, two* new Traits.

Mechoid:
Plant item and mob products reworked to use a proper gene, TRAIT_UNIQUE_PRODUCT.

Vines can rarely spawn up to 3 mobs in sightline of eachother, on the incredibly rare chance they have a mob product. These will always be hostile, in comparison to the player-controlled ones spawned from harvest.

Braincaps added, which use the speaking gene, TRAIT_SPEAKING. TRAIT_SPEAKING makes plants babble about their status; starving, dehydrated, attacked, etcetera; and causes the product, if it is not a mob, to rarely become talkative, akin to xenoarch baubles. Except these ones you can kill.
Braincaps also (will when PR finished momentarily) contain a (illegal? it is a talking plant with some sense of pain and thought) psychoactive drug due to the method they store their memories.

TRAIT_UNIQUE_PRODUCT and TRAIT_SPEAKING are both on the Special allele, making TRAIT_TELEPORTING no longer alone.

Braincaps, Sporeshrooms, and a Destroying Angel seedpack, have been added to cargo in a contraband crate.

### DIFF
--- a/code/datums/supplypacks/contraband.dm
+++ b/code/datums/supplypacks/contraband.dm
@@ -44,15 +44,28 @@
 	contraband = 1
 
 /datum/supply_pack/munitions/bolt_rifles_militia
- 	name = "Weapon - Surplus militia rifles"
- 	contains = list(
- 			/obj/item/gun/projectile/shotgun/pump/rifle = 3,
- 			/obj/item/ammo_magazine/clip/c762 = 6
- 			)
- 	cost = 50
- 	contraband = 1
- 	containertype = /obj/structure/closet/crate/hedberg
- 	containername = "Ballistic weapons crate"
+	name = "Weapon - Surplus militia rifles"
+	contains = list(
+			/obj/item/gun/projectile/shotgun/pump/rifle = 3,
+			/obj/item/ammo_magazine/clip/c762 = 6
+			)
+	cost = 50
+	contraband = 1
+	containertype = /obj/structure/closet/crate/hedberg
+	containername = "Ballistic weapons crate"
+
+/datum/supply_pack/hydro/fungus
+	name = "Fungal Crate"
+	contains = list(
+			/obj/item/seeds/braincapmycelium = 3,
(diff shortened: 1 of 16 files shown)


Sypsoti:
A talking plant is not sapient. Why would the drug be illegal for that? I'm not sure how I feel about plants just outright being able to speak. If it's psychoactive, why not make it emit spores/gas with the drug that instead makes you think the plant is talking?

... They're not ACTUALLY sapient, right?

Spookerton:
I have to question the conversion of extra products to a trait. It represents decreased function, increased complexity, and extra memory spent simply to adhere to a pattern.

Mechoid:
> A talking plant is not sapient. Why would the drug be illegal for that? I'm not sure how I feel about plants just outright being able to speak. If it's psychoactive, why not make it emit spores/gas with the drug that instead makes you think the plant is talking?
> 
> ... They're not ACTUALLY sapient, right?

They are sapient in the manner they are able to react to their environment, similar but less-so than Slimes (They don't react to speech, slimes do, and can even take orders).

> I have to question the conversion of extra products to a trait. It represents decreased function, increased complexity, and extra memory spent simply to adhere to a pattern.

Decreased function I question, as the only thing it prevents outright is having a plant with a unique item and unique mob product at the same time, unless I'm missing something. Not certain how it increases complexity as said, since it prevents having weird situations with 2 unique products. I'm not an authority on memory but having 2 vars down to one theoretical list component seems wiser?

Atermonera:
Staff input: We're not comfortable merging this. The speaking plants stuff is really disconcerting and has a lot of bad connotations about lore direction, and I'm unconvinced of the value added by the unique product trait. I'm fine with the contraband crate, but that's such a small addition to the PR that it seems more practical to break that out into its own PR, if you're willing.
I'll close this tomorrow, to allow time for  final discussion. (Doesn't stop comments after closing)

PR #8915, opened 2023-01-31
allows soap and showers to clean toes, and space cleaner to actually clean synthetic toes

Seris02:
title

so, synths don't process reagents so they don't get affect_touch called on them, only touch_mob
but touch mob doesn't have washshoes set to yes, so they never get cleaned
(I think that's why affect_touch isn't called anyway, haven't actually looked into the code too much, just know this fixes it)

could probably remove soap from this list and make just showers and space cleaner do it instead but it's really fucking annoying having to find space cleaner just to clean my goddamn blood stained feet


Spookerton (approved):
i despise this title

Seris02:
> i despise this title

ahahahahaha
perfect

PR #8921, opened 2023-02-02
clothes port January-February 23

Greenjoe12345:
Ports a number of clothing items from skyrat, aurora, bay and urist.
Some new utility jumpsuits
New headphone variations, plus replacing the sprites of the old headphones ~~with recolorable ones~~
Some winter gear; mittens, wooly hats, ear warmers
Waist bags that can be worn on the belt slot
A hide oversuit that offers protection against the cold


Cerebulon:
I'd very much suggest at least re-describing the "utility uniforms" because:
The old old old "utility uniform" that the description is based on is actually a turtleneck utility thing <s>which is probably based on Mission Impossible or something.</s> Frankly it shouldn't be called a utility uniform but 'tactical turtleneck' was taken.
These on the other hand, are not turtlenecks with utility pants. These are jumpsuits/coveralls (SS13 calls coveralls 'jumpsuits' for 1970s reasons, but I digress). The slightly less old "utility uniforms" (blue and grey) are jumpsuits, and should be what you're basing the descriptions on.
_Personally_ I think they look less "utility" than the newer basic jumpsuit sprites so would call (and path) them something like "[department] accented jumpsuit" and that the old "utility" uniforms are only called that because they were meant to be like military flight suits or something.

I'm also a little meh about the earphones/earbuds on the basis of them being ultra-2020s and the existing headphones being the same retro-bulky concept as Jumpsuits but apparently that's a losing battle on my part and not up 2 me ultimately.

code/modules/clothing/ears/ears.dm
```diff
@@ -175,3 +175,13 @@
...
+
+/obj/item/clothing/ears/earwarmers
+	name = "ear warmers"
+	desc ="A pair of fuzzy ear warmers."
+	icon = 'icons/obj/clothing/ears.dmi'
+	icon_state = "ear_warmers"
+	gender = PLURAL
+	slot_flags = SLOT_EARS | SLOT_TWOEARS
+	min_cold_protection_temperature = SPACE_SUIT_MIN_COLD_PROTECTION_TEMPERATURE
+	volume_multiplier = 0.7 //not designed to block sound like earmuffs but still covers over the ears, so blocks sound a bit
```
Sypsoti, line 187: Could probably just be a child of earmuffs at this point.

code/modules/clothing/under/miscellaneous.dm
```diff
@@ -1197,3 +1197,39 @@ Uniforms and such
 	name = "hazard cohesion suit"
 	desc = "An orange cohesion suit with yellow hazard stripes intended to assist Prometheans in maintaining their form and prevent direct skin exposure."
 	icon_state = "cohesionsuit_hazard"
+
+/obj/item/clothing/under/utility/supply
+	name = "supply utility uniform"
+	desc = "A comfortable turtleneck and black utility trousers. This one has accents in the supply department's color."
```
Sypsoti, line 1203: I believe cerebul had words about this description being odd paired with the sprite, so please correct that as well.

Sypsoti (changes requested):
A few changes needed, and I'm kind of wanting to keep the colored musical notes for the on-mob sprites for headphones. I specifically changed it in the past due to complaints about readability.

code/modules/client/preference_setup/loadout/loadout_ears.dm
```diff
@@ -6,13 +6,30 @@
...
 	path = /obj/item/clothing/head/headphones
 
+/datum/gear/ears/headphones/New()
+	..()
+	var/headphones = list()
+	headphones["headphones"] = /obj/item/clothing/head/headphones
+	headphones["earphones"] = /obj/item/clothing/head/headphones/earphones
+	headphones["earbuds"] = /obj/item/clothing/head/headphones/earbuds
+	gear_tweaks += new/datum/gear_tweak/path(headphones)
+	gear_tweaks += gear_tweak_free_color_choice
```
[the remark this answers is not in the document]
Greenjoe12345: image
image
works just fine

Cerebulon:
Imo the belts should definitely still cost 2 and 3 **minimum**, not 1.
2 is the relatively standard cost for a thing with mechanical advantages and these are literally Universal Belts with potentially more capacity than simple webbing. Which costs 2. Given half the station don't always have job belts (which have a restricted list of things you can put on them!!) there needs to be literally any reason not to take these and have 4+ extra pockets which is some potentially dumb loadout power creep.
<s>I'm not going to get into huge arguments about how it should be higher and the big ones should be maint/wilderness spawns because loadout balance is a mess so the standard 2 is probably all I'm going to get.</s>

Spookerton:
Drafted. Needs CR and conflict resolution.

Cerebulon:
From the staff meeting on 5th Feb:
Waist packs are a little too power creep-y, so:

1. Small waist pack and fannypack should cost 2 points
2. Large waist pack should be removed from loadout and added to maintenance random-spawners alongside dufflebags.

Atermonera:
@Greenjoe12345 merge conflicts

Greenjoe12345:
Yeah gonna see about getting to fixing when I can

PR #8904, opened 2023-01-27
Drake harnesses are removable.

MistakeNot4892:
- Drake harnesses can be removed and replaced, and exist as items in the world.
- Various drake logic has been merged down to either base animal harness level, or base drake level.

DNM, still getting harness removal/replacement to work.

Depends on #8922 #8923

PR #8922, opened 2023-02-03
Added more drake emotes and sounds, thanks Spook.

MistakeNot4892:
Splitting this out of #8904.
- Adds several new emotes with sounds courtesy of @Spookerton.
- Condenses the drake friendly interaction emote list down to 'sniffs', use `me` or an emote for the more specific stuff.
- `*dhowl` is now `*droar`.
- Headbutt emote has a five second cooldown and a similar proning chance to human disarm.

Giving this a quick local test then should be ready to go.


code/modules/emotes/definitions/slimes.dm
```diff
@@ -2,11 +2,12 @@
 	key = "nomood"
 	var/mood
 
-/decl/emote/slime/do_extra(var/mob/living/simple_mob/slime/user)
+/decl/emote/slime/do_extra(mob/user, atom/target)
```
Spookerton, line 5: I have a preference worked into me for parameters to be typed for their expected input because it saves specifying the path twice and creating an extra reference, but I only add this as a comment

Spookerton (approved):

PR #8927, opened 2023-02-05
Child organs can't robotize() if they're already robotic

Atermonera:
I looked at `robotize()` when I reviewd #8911 to make sure the changes made sense. I missed a very important line:
https://github.com/PolarisSS13/Polaris/blob/master/code/modules/organs/organ_external.dm#L1101-L1102

If the child limb is already robotic, it will not update the brand. Hence, the issue stated as fixed was actually _introduced_ by the changes implemented.


Spookerton (approved):

Seris02:
MY MISTAKE
It needed to be PR'd downstream, for ORGAN_NANOFORM, which is what I was testing it against

PR #8931, opened 2023-02-06
Axes carbon aliums, makes nymphs direct subtype of carbon

Atermonera:
Pretty sure all the procs/features required for nymphs are retained, I had to clean up ~most of that code.
We don't use the carbon aliens for anything, even now that they're skathari, just the simple mob variants that actually have AI.


code/modules/mob/living/carbon/human/human_attackhand.dm
```diff
@@ -117,7 +117,6 @@
 				return
 			M.put_in_active_hand(G)
 			G.synch()
-			LAssailant = M
```
Atermonera, line 120: Unused.

code/modules/mob/living/living.dm
```diff
@@ -81,11 +81,7 @@
 		to_chat(src, "<font color='blue'>You are not injured enough to succumb to death!</font>")
 
 /mob/living/proc/updatehealth()
-	if(status_flags & GODMODE)
```
Atermonera, line 84: GODMODE is unmaintained and therefore deprecated, so when I happen upon references I axe them.

code/modules/mob/living/carbon/alien/life.dm
```diff
@@ -1,161 +0,0 @@
...
-		else if(eye_blurry)
-			eye_blurry = max(eye_blurry-1, 0)
-
-		update_icons()
-
-	return 1
-
-/mob/living/carbon/alien/handle_regular_hud_updates()
-
-	if (stat == 2 || (XRAY in src.mutations))
```
Atermonera, line 91: Gradual removal of references to genetic mutations in hopes of someday being able to rip the system out

PR #8938, opened 2023-02-09
moves us out of kara

elgeonmb:
(Fixes a stray mention of Kara in communicator code)

the northern star was five years ago. she's gone. you have to move on.


Sypsoti (approved):
but i can't stop thinking about her............................

PR #8941, opened 2023-02-10
Fix various anomaly bugs and oversights. 

Mechoid:
Cellcharge and Celldrain have been blasted into compliance. No checking global range() anymore, what the fuck dude.

Various oversights for effect paths being moved to rarities fixed.

### DIFF
--- a/code/modules/xenoarcheaology/effects/cellcharge.dm
+++ b/code/modules/xenoarcheaology/effects/cellcharge.dm
@@ -4,86 +4,41 @@
 	effect_color = "#ffee06"
 	var/last_message
 
+/datum/artifact_effect/uncommon/cellcharge/proc/charge_cells(var/amount = 25)
+	var/atom/holder = get_master_holder()
+	if (!holder)
+		return
+	var/turf/turf = get_turf(holder)
+	if (!turf)
+		return
+
+	var/messaged_robots
+	for(var/atom/movable/AM in range(effectrange, turf))
+		if(isliving(AM))
+			var/mob/living/L = AM
+			var/obj/item/cell/C = L.get_cell()
+
+			if(C)
+				if(issilicon(L) && ((last_message + (1 MINUTE)) < world.time))
+					messaged_robots = TRUE
+					to_chat(L, SPAN_NOTICE("SYSTEM ALERT: Energy boost detected!"))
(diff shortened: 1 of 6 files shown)


code/modules/xenoarcheaology/effects/cellcharge.dm
```diff
@@ -21,39 +41,49 @@
...
+			C.charge = min(C.maxcharge, C.charge + 25)
+
+	if(messaged_robots)
 		last_message = world.time
 
+	for(var/obj/item/I in range(effectrange,turf))
+		var/obj/item/cell/C = I.get_cell()
+
+		if(C)
+			C.charge = min(C.maxcharge, C.charge + 25)
```
Atermonera, line 62: All of this is copy-pasted thrice between touch, pulse, and aura, but for some numbers and a text string - can we squash this into a single proc?

code/modules/xenoarcheaology/effects/cellcharge.dm
```diff
@@ -6,13 +6,33 @@
...
+	var/atom/holder = get_master_holder()
+	if (!holder)
+		return
+	var/turf/turf = get_turf(user)
+	if (!turf)
+		return
+
+	var/messaged_robots
+	for(var/mob/living/L in range(effectrange, turf))
+		var/obj/item/cell/C = L.get_cell()
```
Atermonera, line 20: `get_cell()` is a proc on `/atom/movable` - can we iterate over those and do our own typechecking for the messaging stuff?

polaris.dme
```diff
@@ -3342,7 +3342,7 @@
 #include "code\ZAS\Zone.dm"
 #include "interface\interface.dm"
 #include "interface\skin.dmf"
-#include "maps\cynosure\cynosure.dm"
+#include "maps\example\example.dm"
```
Atermonera, line 3345: Oops!
```suggestion
#include "maps\cynosure\cynosure.dm"
```

Atermonera (changes requested):
Sorry for the delay. The code itself looks fine (I'm a little dubious of iterating over effectrange because that _can_ get large but it's probably a rare case that won't come up too often), but I think we can clean it up a bit.

Atermonera (approved):

PR #8943, opened 2023-02-11
Synthetics can have antennas (also humans if they want?)

tiramisuapimancer:
Enables the dual, left, and right synth antenna sprite accessories for the human, vatborn, and event species (don't know what those are, just copied the line from the elf ears for parity with other universally accessible eartypes). There was a conversation on the Discord the other day that I could whitelist to get these, but it was also mentioned that there's no real reason that these aren't universally accessible anyways (both for the use of synthetics using the Human/Vatborn species as a baseline, and just "as augments" for those who are _actual_ humans). So instead of playing enough that I'm trusted for a whitelist I just PR'ed them. This was ostensibly the better option anyways.

image

Have I mentioned how disconcerting it is that this codebase doesn't seem to have a PR body text format?

### DIFF
--- a/code/modules/mob/new_player/sprite_accessories_ear.dm
+++ b/code/modules/mob/new_player/sprite_accessories_ear.dm
@@ -47,15 +47,17 @@
 /datum/sprite_accessory/ears/dual_robot
 	name = "synth antennae"
 	icon_state = "dual_robot_antennae"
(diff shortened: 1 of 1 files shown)


Atermonera (approved):

Atermonera:
Event species are custom species for events that staff can play with the numbers on, without having to modify the numbers of a species that other players are using (They're singleton datums here)
As for the PR template... I can't say I've really noticed an issue with not having one, because people usually include the information that should be included in PR descriptions?

tiramisuapimancer:
It's not really an issue, I'm just super used to other servers that have one.

PR #8950, opened 2023-02-14
Fixing Tape Recorder Voices

Sharkmare:
Carbon/Human has a voice variable for voice changers and to not go unknown when the face is covered without ID, tape recorder now respects that.

### DIFF
--- a/code/game/objects/items/devices/taperecorder.dm
+++ b/code/game/objects/items/devices/taperecorder.dm
@@ -89,9 +89,9 @@
 	update_icon()
 
 
-/obj/item/taperecorder/hear_talk(mob/M, list/message_pieces, verb)
+/obj/item/taperecorder/hear_talk(mob/living/M, list/message_pieces, verb)
 	var/msg = multilingual_to_message(message_pieces, requires_machine_understands = TRUE, with_capitalization = TRUE)
-	var/voice = M.voice_name //This seems to already be defined on Mob/
+	var/voice = M.GetVoice() //Defined on living, returns name for normal mobs/
 	if(mytape && recording)
 		mytape.record_speech("[voice] [verb], \"[msg]\"")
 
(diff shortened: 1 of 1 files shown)


code/game/objects/items/devices/taperecorder.dm
```diff
@@ -91,8 +91,16 @@
...
 /obj/item/taperecorder/hear_talk(mob/M, list/message_pieces, verb)
 	var/msg = multilingual_to_message(message_pieces, requires_machine_understands = TRUE, with_capitalization = TRUE)
+	var/voice = "Unknown"
+	if (M.type == /mob/living/carbon/human)
+	{
+		var/mob/living/carbon/human/H = M
+		voice = H.voice
+	}
+	else
+		voice = M.name
```
Atermonera, line 101:
```suggestion
	var/voice = M.name ? M.name : "Unknown"
	if (ishuman(M))
		var/mob/living/carbon/human/H = M
		voice = H.voice
```
DM is non-context-free, you don't need the curly braces.

Atermonera: @Sharkmare bumping this, you gonna get back to this?

Sharkmare: for some reason i only got the bump as a notification odd

Atermonera (dismissed):

Sharkmare:
Apparently i made a mistake somewhere down the road and also found that mobs do have a voice var we can just use that is also use on carbons according to say code

Atermonera:
I don't think `voice_name` is the var you want to use here. A cursory search of code shows that it's just set to "unkown" for all human mobs, meaning everyone will show up as "unknown" in the transcript.

Sharkmare:
> I don't think `voice_name` is the var you want to use here. A cursory search of code shows that it's just set to "unkown" for all human mobs, meaning everyone will show up as "unknown" in the transcript.

Good point, GetVoice() is probably the correct way to handle this.

Atermonera (approved):

PR #8952, opened 2023-02-14
Flesh Repair respects Blood Pump

Mechoid:
Flesh repair surgery now works on corpses with the bloodpump modifier, typically from the patient stabilizer.

A surgeon with enough time can stitch a corpse back together to try and defib them, as long as the corpse's blood is being pumped.


Sypsoti:
2/19 meeting consensus: Too overpowered as-is, a single-use operation per-limb to lower the damage is an acceptable change.

Mechoid:
Closing due to age, integrating fixes for mort and necro into another PR

PR #8953, opened 2023-02-14
Refluff space bats and space bears

Cerebulon:
Space bats are regularly used in PoIs of all sorts so its a little odd that they have a goofy joke name from a decade ago. Also: Holy echolocation batman, it's batlore!

Space bears aren't used anywhere where they show up For Realises because they're very silly, but renamed them for the rare occasion they do show up.

### DIFF
--- a/code/modules/mob/living/simple_mob/subtypes/animal/space/bats.dm
+++ b/code/modules/mob/living/simple_mob/subtypes/animal/space/bats.dm
@@ -10,22 +10,6 @@
 
 	faction = "scarybat"
 
-/datum/category_item/catalogue/fauna/razor_bat
-	name = "Nispean Fauna - Razor Bat"
-	desc = "These sharp-toothed flying mammalian animals are adapted to the oxygen-deficient \
-	and hostile jungles of Nisp, and as a result are frightening capable of surviving in harsh, \
-	even near-airless environments. In higher-oxygen environments, these bats can reach the size \
-	of a human head, though they struggle to thrive in low temperatures without proper shelter.\
-	<br><br>\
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/simple_mob/subtypes/animal/space/bats.dm
```diff
@@ -57,3 +57,19 @@
 	maxHealth = 60
 	health = 60
 	melee_damage_upper = 10
+
+/datum/category_item/catalogue/fauna/razor_bat
+	name = "Nispean Fauna - Razor Bat"
+	desc = "These sharp-toothed flying mammalian animals are adapted to the oxygen-deficient \
+	and hostile jungles of Nisp, and as a result are frightening capable of surviving in harsh, \
```
elgeonmb, line 64: frightening_ly_?

code/modules/mob/living/simple_mob/subtypes/animal/space/bats.dm
```diff
@@ -57,3 +57,20 @@
 	maxHealth = 60
 	health = 60
 	melee_damage_upper = 10
+
+/datum/category_item/catalogue/fauna/razor_bat
+	name = "Nispean Fauna - Razor Bat"
+	desc = "These sharp-toothed flying mammalian animals are adapted to the oxygen-deficient \
```
Atermonera, line 63: I wouldn't say Nisp is especially oxygen deficient... although the percentage of oxygen is 10.5%, the pressure is 2.5atm, so the partial pressure of oxygen is actually a little bit higher than terrestrial air.

Atermonera (approved):

PR #8960, opened 2023-02-17
Floodlight refactor, new battery_powered component

Ilysen:
#### purpose

for floodlights: floodlights are among some of the oldest code in basically every bay-derivative server that I know of (including bay itself!) and the poor things could really use some TLC

for the component: it turns out that every item that requires a power cell currently implements its own logic from scratch. while machines have a base-level `get_cell()` proc, subtypes must override this proc and implement their own logic on a per-type basis. this doesn't get rid of that everywhere, but it does lay down a groundwork to do that next.

#### details

* Implements a new component, `/datum/component/battery_powered`. It features attackby() and attack_hand() signals to handle swapping out cells manually as well as providing a proc to drain charge within the cell. the whole thing is lightweight and generally fairly extensible - I'd like to put it onto other battery-powered things once it gets signed off on. name is also subject to change.
* Floodlights have quite a few cleanliness changes:
  * Name is now lowercase; description revamped
  * They now use a battery_powered component to track their cell
  * Icon updating is done through overlays, instead of having multiple various icon states
  * User feedback has been improved with sounds and proper message spans
  * Extraneous vars and procs have been consolidated or replaced where possible

#### testing

I spawned in a floodlight and tried fiddling with it as a human - turning it on and off, replacing the cell, and so on. no issues - I also did it as an engiborg to make sure it worked properly, and then went and found a naturally spawned floodlight to repeat the tests. while I was writing this PR, I kept the screen on my other monitor to make sure the new flickering looked proper (it does).

#### extra

* there are some minor kinks that come out of me not knowing if it's possible to have a signal prevented from firing by having the base proc return something - for stuff like attackby() and removing the panel, right now it falls through to the component proc and is then prevented through the logic therein, but it'd be cleaner to let the base attackby() prevent the component from being triggered at all
* I'm curious if a component is the right way to go here - it *feels* like it is, because there are also things like items that use power cells, but I'd love to hear some input
* the process of replacing cells is currently kind of involved - it takes a screwdriver and a crowbar, instead of just one or the other. I maintained this functionality but I am open to changing it
* this is the now I think the second small bay-derivative server I've refactored floodlights for. with any luck it'll go better this time ;p


code/datums/components/battery_power.dm
```diff
@@ -0,0 +1,51 @@
+/// Generalized framework allowing something to draw power from an inserted power cell.
+/// Contains attack_hand() and attackby() signals allowing cells to be added and removed by hand.
+/// This base type supports only standard cells; device cells won't fit.
+/datum/component/battery_power
+	var/obj/item/cell/cell
```
Atermonera, line 5: ![image](https://imgs.xkcd.com/comics/standards.png
I think the alternative would be to have the cell on /obj, which potentially complicates objects that use _device_ cells, which iirc are a completely distinct type.

code/datums/components/battery_power.dm
```diff
@@ -0,0 +1,51 @@
...
+
+/datum/component/battery_power/Initialize()
+	if (istype(parent, /obj/machinery))
```
Atermonera, line 8: I suspect we'll eventually want to do something about this for not-machinery types.

code/datums/components/battery_power.dm
```diff
@@ -0,0 +1,51 @@
...
+		RegisterSignal(parent, COMSIG_ATOM_ATTACK_HAND, .proc/on_attack_hand)
+		RegisterSignal(parent, COMSIG_PARENT_ATTACKBY, .proc/on_attack_by)
+	else
+		return COMPONENT_INCOMPATIBLE
```
Atermonera, line 12: Do component initialize()'s need to call super?

code/datums/components/battery_power.dm
```diff
@@ -0,0 +1,51 @@
...
+
+/// Checks if a provided item can be used as a power cell for this component.
+/// By default, the component checks for standard-sized cells.
+/datum/component/battery_power/proc/is_valid_cell(obj/item/I)
+	return istype(I, /obj/item/cell) && !istype(I, /obj/item/cell/device)
```
Atermonera, line 17: Oh that's right, this was the awful solution for device cells. God, but I wish this were cleaner. Out of scope, but maybe if we make the current cells a separate subtype of an abstract parent class that has to core cell logic, and machinery is just allowed to use the machinery subtype and items are just allowed to use the device subtype

code/datums/components/battery_power.dm
```diff
@@ -0,0 +1,51 @@
...
+
+/datum/component/battery_power/proc/on_attack_hand(obj/parent_atom, mob/living/user)
+	var/obj/machinery/M = parent_atom
+	if (!M.panel_open)
+		return
+	if (!cell)
+		to_chat(user, SPAN_WARNING("\The [M] has no cell to remove."))
+		return
+	cell.dropInto(M.loc)
```
Atermonera, line 26: Do we want a getturf() here? This smells risky...

code/game/machinery/floodlight.dm
```diff
@@ -1,128 +1,108 @@
...
+	var/panel_locked = TRUE
+	var/brightness_on = 8
+	var/datum/component/battery_power/BP
 
 /obj/machinery/floodlight/Initialize()
 	. = ..()
-	cell = new(src)
+	BP = AddComponent(/datum/component/battery_power)
+	if (BP)
+		BP.cell = new(src)
```
Atermonera, line 16: This will 100% need an argument for the type of cell, because we've got a smattering of objects that spawn with particular cells.

Atermonera: Also, is the object responsible for spawning its cell or should the component be?
Also also, do we want a getter function for the cell? Even though the cell _lives_ in the object, it's a _var_ in the component, which feels... just slightly obtuse.

code/game/machinery/floodlight.dm
```diff
@@ -1,128 +1,108 @@
...
 
 /obj/machinery/floodlight/process()
-	if(!on)
+	if (use_power != USE_POWER_ACTIVE)
 		return
-
-	if(!cell || (cell.charge < (use * CELLRATE)))
-		turn_off(1)
+	else if (!BP?.draw_power(get_power_usage() * CELLRATE))
+		visible_message(SPAN_WARNING("\The [src]'s power cell dies!"))
```
Atermonera, line 38:
```suggestion
		visible_message(SPAN_WARNING("\The [src] shuts off abruptly!"))
```
Don't want to give people _too_ much of an idea why.

code/datums/components/battery_power.dm
```diff
@@ -0,0 +1,51 @@
...
+	user.put_in_active_hand(cell)
+	cell.add_fingerprint(user)
+	user.visible_message(SPAN_NOTICE("\The [user] removes \the [cell] from \the [M]."), SPAN_NOTICE("You remove \the [cell] from \the [M]."))
+	M.update_icon()
+	cell.update_icon()
+	cell = null
+
+/datum/component/battery_power/proc/on_attack_by(obj/parent_atom, obj/item/I, mob/living/user)
+	var/obj/machinery/M = parent_atom
+	if (!M.panel_open || !is_valid_cell(I))
```
Atermonera, line 36: `!M.panel_open`
I feel like we'll want a proc on M that tells us whether the cell should be accessible.

code/datums/components/battery_power.dm
```diff
@@ -0,0 +1,51 @@
...
+		to_chat(user, SPAN_WARNING("\The [src] already has \a [cell] installed."))
+		return
+	user.drop_item(I)
+	I.forceMove(M)
+	cell = I
+	user.visible_message(SPAN_NOTICE("\The [user] inserts \the [I] into \the [M]."), SPAN_NOTICE("You slot \the [I] into \the [M]."))
+	M.update_icon()
+
+/// Attempts to use power from a connected cell if one exists.
+/// You generally want to multiply the required amount by CELLRATE beforehand,
```
Atermonera, line 48: I wish that were just handled by the cell but that's a HUGE refactor, maybe if this component sees more widespread use...

code/game/machinery/floodlight.dm
```diff
@@ -1,128 +1,108 @@
...
+	set waitfor = FALSE
+	set_light(brightness_on / 2, brightness_on / 4)
+	sleep(0.5 SECONDS)
+	if (use_power == USE_POWER_ACTIVE)
+		set_light(brightness_on, brightness_on / 2)
+
+/obj/machinery/floodlight/proc/toggle(mob/living/user, to_enable, loud)
+	if (to_enable)
+		if (!BP?.cell?.check_charge(active_power_usage * CELLRATE))
+			to_chat(user, SPAN_WARNING("You flick \the [src]'s power switch, but nothing happens."))
```
Atermonera, line 55: User check?

code/game/machinery/floodlight.dm
```diff
@@ -1,128 +1,108 @@
...
+
+/obj/machinery/floodlight/proc/flicker()
+	set waitfor = FALSE
+	set_light(brightness_on / 2, brightness_on / 4)
+	sleep(0.5 SECONDS)
+	if (use_power == USE_POWER_ACTIVE)
+		set_light(brightness_on, brightness_on / 2)
+
+/obj/machinery/floodlight/proc/toggle(mob/living/user, to_enable, loud)
+	if (to_enable)
```
Atermonera, line 53:
```suggestion
/obj/machinery/floodlight/proc/toggle(mob/living/user, toggle_to, loud)
	if (toggle_to)
```
`to_enable` sounds kinda awkward.

code/game/machinery/floodlight.dm
```diff
@@ -1,128 +1,108 @@
...
-			cell.loc = src.loc
+			user.visible_message(
+				SPAN_NOTICE("\The [user] turns [use_power == USE_POWER_ACTIVE ? "on" : "off"] \the [src]."),
+				SPAN_NOTICE("You turn [use_power == USE_POWER_ACTIVE ? "on" : "off"] \the [src]."))
+	return TRUE
 
-		cell.add_fingerprint(user)
-		cell.update_icon()
+/obj/machinery/floodlight/attack_ai(mob/user)
+	if (istype(user, /mob/living/silicon/robot) && Adjacent(user))
```
Atermonera, line 73: isrobot()

Atermonera (commented):
You have the misfortune of catching me with both time and energy to do a thorough review.

code/datums/components/battery_power.dm
Ilysen: I felt like it'd be unclean to have an obj-level variable even though not everything uses cells. would a component (or extension) not make the interactions cleaner?

code/datums/components/battery_power.dm
Ilysen: yeye; I figured that the component could be expanded in the future, at which point this kinda thing could be amended

code/game/machinery/floodlight.dm
Ilysen: it could go either way. do you have any particular preference?

Ilysen:
I don't mind my PRs being picked apart. as long as you think the core concept has merit, I can refine it however much I need to :p

Ilysen:
discussed this a bunch with ater on discord and we covered a few different takes on how to handle this, the upsides and downsides of them, etc. I've got a lot to chew on now, so imma close this for now and take another stab at it once I've had some time to redo the design

PR #8971, opened 2023-02-20
Revert "fixes a memory leak with mannequins and the issue that then comes with fixing it"

Atermonera:
Reverts PolarisSS13/Polaris#8942

@Seris02 multiple complaints from players after updating to the fixed behaviour. I tried looking a bit last night, and a fix did not seem trivial or obvious, so moving to revert and reattempt the fixes later, rather than forward-fixing.
> Okay, really weird bug. I was rolling a new character. I set one of the arms as amputated to see what it was like, but it won't reset and the arms are amputated on all the characters now.

> and Latee's showing as having cenilimi prosthetics, despite having unbranded tesh ones

> oh this happened with remi but instead it was displaying half his limbs with unbranded prosthetics while still listing them as cremilimi...which is less cursed than this. did correct after switching to another character then back again 

> first of all those aren't even spectre's prosthetics and second of all WHY DOES IT HAVE HUMAN LEGS
image

### DIFF
--- a/code/modules/client/preference_setup/general/03_body.dm
+++ b/code/modules/client/preference_setup/general/03_body.dm
@@ -373,9 +373,7 @@ var/global/list/valid_bloodtypes = list("A+", "A-", "B+", "B-", "AB+", "AB-", "O
 		var/status = pref.organ_data[name]
 		var/obj/item/organ/external/O = character.organs_by_name[name]
 		if(O)
-			if(status == null)
-				O.derobotize()
-			else if(status == "amputated")
+			if(status == "amputated")
 				O.remove_rejuv()
 			else if(status == "cyborg")
 				if(pref.rlimb_data[name])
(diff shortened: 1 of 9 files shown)


Seris02:
What the actual fuck

Atermonera:
@Seris02 Still getting issues reported with this.
I joined the server, with a human slot loaded, all limbs were correct.
I loaded an FBP character, limbs were all robotized.
I reloaded teh original human, limbs were all de-robotized.
I reloaded the FBP, limbs were re-robotized.
I loaded a _different_ human, got this:
!(this)[https://puu.sh/JAh7e/6949c0b185.png]
I joined the game, and that hodge-podge was applied to spawned human mob.

code/modules/client/preference_setup/general/03_body.dm
```diff
@@ -373,9 +373,7 @@ var/global/list/valid_bloodtypes = list("A+", "A-", "B+", "B-", "AB+", "AB-", "O
 		var/status = pref.organ_data[name]
 		var/obj/item/organ/external/O = character.organs_by_name[name]
 		if(O)
-			if(status == null)
-				O.derobotize()
```
Atermonera, line 377: Keeping the derobotize proc, but reintroducing the memory leak means we don't worry about this.

Atermonera:
ftr, moving forward with this revert because it _is_ affecting players in-game, not just character setup anymore.

Atermonera:
Extended testing with a new round while waiting for PR builds to finish, it took several tries of "load FBP then load human" but it did eventually happen again

Seris02:
> Extended testing with a new round while waiting for PR builds to finish, it took several tries of "load FBP then load human" but it did eventually happen again

fair enough, do you still want the memory leak fix, though

PR #8972, opened 2023-02-21
few fixes for the derobotizing fix

Seris02:
so uh. PR #8971 was a thing
forgot to test a few things apparently, weh.

may want to testmerge this first, maybe

(an alternative fix, as someone suggested, is just forcing the set_species call on copy_to, which would qdel all the organs and limbs and restore them every time it's called)


Atermonera:
APPARENTLY it's nigh-impossible for me to catch times that are good to run test merges at so uh, fuck it, doing it live.

Atermonera (approved):

PR #8976, opened 2023-02-21
Tweaks the positioning and icons of the eye slot and ear slots on the inventory UI

Frenjo:
Adds new individual left and right ear UI icons for the left and right ear slots respectively.

Repositions the left and right ear slots so that they're on different sides of the head slot as opposed to being on the same side.

All other icons remain in their current positions.

image

🆑
Tweak: Adjusted the positioning of the inventory's left and right ear slots so that they're on either side of the mask slot instead of bundled together on one side.
Tweak: Moved the inventory's eye slot upwards so it's next to the helmet slot.
Tweak: The inventory's left and right ear slots now have separate HUD icons so it's easier to tell them apart at a glance.
/🆑


Frenjo:
I've just swapped the ear slots and the eye slot to avoid having a gap in the UI following feedback by Cerebulon.

The new layout looks like this:
image

Atermonera:
I kinda want to collect opinions about this.

Frenjo:
If people aren't happy with it, I can always strip out the positional changes and just keep the icon swap so you can still tell which ear is which by the way it's facing in the slot itself.

Atermonera:
Staff input: Sure why not.
~~Ater input: hiss change bad~~

Atermonera (approved):

PR #8978, opened 2023-02-21
Swaps Air Tanks for Oxygen Tanks

Doctress:
New Perri-Air
Canned in Druidia
Naturally Sparkling
Salt Free Air
All Natural Druidia-- okay I'm done with this bit now.


Doctress:
Check with Schnayy as NT Union Meeting Person before merging.

PR #8979, opened 2023-02-21
Menthol doesn't kill you at 7 units

Cerebulon:
You would start getting poisoned from a couple of menthol cigarettes. Overdose is now 30 (the baseline) instead of a quarter of the baseline.


Sypsoti (approved):
Smoking kills, you know.

PR #8984, opened 2023-02-23
Removes BluSD and DJOutpost4 from spawning

Sypsoti:
Reason for BluSD: The IC implications of killing literal government officials is really weird and bad and shouldn't be encouraged. It just eats a spawn chance for something to logically clear out.

Reason for DJOutpost4: Replicants have weird IC implications as well and are very, VERY strange to rp around. 

Neither of these PoIs have a place in HRP on our server. Both should still be spawnable if an EM wants them for w/e reason.


Sypsoti:
Added spatial anomaly to the removal due to portals being broken with no ETA for a fix.

Atermonera (approved):

PR #8986, opened 2023-02-24
Generalizing sleeping updates and adding it to simple mobs.

### LINKED ISSUE #8936: Reviving a dead drake with paddles makes them permamently deaf, "blind", and they cannot move
Title. But to extrapolate. My vision is fully unobscured by inspecting any element gives the "Something is there but you can't see it" message. Might be related to sleep not clearing on revival. Eventually setting sleep var to 0 fixed the issue.
Revived drakes should likely be able to do things.
#### What actually happened
Stuck unable to move, could only interact with self and objects immediately around self. Resting and unresting, *collapse did nothing to alleviate the situation
#### Steps to reproduce
Die, get resurrected via paddles
#### Additional info:
Server revision: master - 2023-02-07 

[781fe82a78ac84c887e627789330c8d6a1fc36ca](http://127.0.0.1:61590/tmp19104/%3Ca%20href=%22https://github.com/PolarisSS13/Polaris/commit/781fe82a78ac84c887e627789330c8d6a1fc36ca%22%3Ehttps://github.com/PolarisSS13/Polaris/commit/781fe82a78ac84c887e627789330c8d6a1fc36ca%3C/a%3E)

MistakeNot4892:
Should fix drakes having Big Sleep when revived from the Eternal Sleep.
Fixes #8936

### DIFF
--- a/code/modules/mob/living/life.dm
+++ b/code/modules/mob/living/life.dm
@@ -41,10 +41,10 @@
 
 	//Check if we're on fire
 	handle_fire()
-	
+
 	if(client && !(client.prefs.ambience_freq == 0))	// Handle re-running ambience to mobs if they've remained in an area, AND have an active client assigned to them, and do not have repeating ambience disabled.
 		handle_ambience()
-	
+
 	//stuff in the stomach
 	handle_stomach()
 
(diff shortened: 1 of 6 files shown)


code/modules/mob/living/life.dm
```diff
@@ -122,6 +122,14 @@
 	handle_slurring()
 	handle_confused()
 
+/mob/living/proc/handle_sleeping()
+	if(sleeping)
```
Atermonera, line 126: Do we need to check if sleeping is positive?

PR #8992, opened 2023-02-25
R&D Design Updates

Mechoid:
Mining rig vision module added to Robotics.

### DIFF
--- a/code/modules/research/designs/HUDs.dm
+++ b/code/modules/research/designs/HUDs.dm
@@ -42,6 +42,22 @@
 	name = "graviton visor"
 	id = "graviton_goggles"
 	req_tech = list(TECH_MAGNET = 5, TECH_ENGINEERING = 3, TECH_BLUESPACE = 3, TECH_PHORON = 3)
-	materials = list(MAT_PLASTEEL = 2000, "glass" = 3000, MAT_PHORON = 1500)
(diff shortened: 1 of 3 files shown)


Woodratt:
This is going to need to go to a staff review, but I am, hard against this PR since it is going to give balance an absolute headache.

Sypsoti:
Against thermals and spike cells, personally. Aren't these antag items? What's the point of antag uplink if we give all of them to the station as well?

Mechoid:
Thermals at TECH_ILLEGAL 7 will rarely be acquirable, namely if xenoarch or explorers bring back Alien Objects that happen to have illegal tech 6 or higher on them. Right now, the highest you can get without them is 5 IIRC.

The cell isn't overtly dangerous, and so isn't a great Threat for an illegal item in the protolathe. It could theoretically be useful for explorers, or for security dealing with blob spores.

I expect TECH_ILLEGAL to have some crossover with antagonist items just by the nature of it.

Atermonera:
Staff input: The rig vision module is fine. The rest is too much power creep.
NVGs are technically available today via the tcomm sat, and should not be more available. Thermals should remain a tool of antags, because antags need advantages to overcome the crew's numbers. Spike cell... doesn't belong in research, and has no use outside of antaggery.

Atermonera (approved):

PR #8993, opened 2023-02-25
Converts move intents to /decl/move_intent

Frenjo:
I'll be honest straight out: This is a dodgy Hypatia port.
Ideally I want someone experienced with `/decl` to take a look over it because it's _very_ dodgy.

Converts hardcoded move intents to use a modular `/decl/move_intent` system so that more can potentially be added down the line. This did originally stem out of my idea to add the creep intent, but I couldn't because of the hardcoded nature of Polaris current move intent system.

Also adds the `IS_RUNNING(X)` and `IS_WALKING(X)` macros so the code isn't littered with `if(X.move_intent.name == "walk")` and other such variations all over the place.

Also included is a refactor of `/mob/living/proc/set_m_intent` into `/mob/living/proc/set_move_intent(decl/move_intent/intent)`, with handling of setting HUD icon states all merged into one.

Finally, this PR also removes the useless `/mob/living/var/m_int` and associated infrastructure as it wasn't really used for anything and was just taking up space.

No changelog for this one since it's backend, and I don't think those need a changelog?


Frenjo:
I've marked this as ready for review after looking over the code a few more times... It seems to work and seems stable on local tests too, which is also good.

code/datums/move_intent.dm
```diff
@@ -0,0 +1,22 @@
+/decl/move_intent
+	var/name
+	var/move_delay = 1
+	var/hud_icon_state
+
+// Walking
+/decl/move_intent/walk
```
Atermonera, line 7: Is this supposed to have a different move_delay?

code/modules/projectiles/targeting/targeting_mob.dm
```diff
@@ -57,10 +57,15 @@
...
+	var/new_intent = GET_DECL(intent)
+	if(isnull(new_intent))
+		return
+
+	move_intent = new_intent
 	if(hud_used)
-		if (hud_used.move_intent)
-			hud_used.move_intent.icon_state = intent == "walk" ? "walking" : "running"
\ No newline at end of file
+		if(hud_used.move_intent)
```
Atermonera:
```suggestion
	if(hud_used?.move_intent)
```

Atermonera (approved):

PR #8994, opened 2023-02-25
big heckin chonker ultragigamega-ears!!!!

Varlaisvea:
Boosts the range at which certain races can hear whispers.
Default whisper hearing range is 1 (audible) / 5 (semi-audible). This is unchanged for most races... 

.. but Tajaran and Teshari now have 2/6 (a boost of +1) - this means they can hear your whispers from several tiles away. You cannot hide secrets from the big heckin chonker ultragigamega-ears.

unashamed teshari buff, teshari hiding in backpacks now make for fucking incredible spies.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/species.dm
+++ b/code/modules/mob/living/carbon/human/species/species.dm
@@ -53,6 +53,7 @@
 	var/blood_level_fatal = 0.4								//"Fatal" blood level; below this, you take extremely high oxydamage
 	var/hunger_factor = 0.05								// Multiplier for hunger.
 	var/active_regen_mult = 1								// Multiplier for 'Regenerate' power speed, in human_powers.dm
+	var/hearboost = 0 										//Whisper hearing sensitivity. Default is 1 (adjacent).
 
 	var/taste_sensitivity = TASTE_NORMAL							// How sensitive the species is to minute tastes.
 	var/allergens = null									// Things that will make this species very sick
--- a/code/modules/mob/living/carbon/human/species/species.dm
+++ b/code/modules/mob/living/carbon/human/species/species.dm
@@ -53,7 +53,7 @@
 	var/blood_level_fatal = 0.4								//"Fatal" blood level; below this, you take extremely high oxydamage
 	var/hunger_factor = 0.05								// Multiplier for hunger.
 	var/active_regen_mult = 1								// Multiplier for 'Regenerate' power speed, in human_powers.dm
-	var/hearboost = 0 										//Whisper hearing sensitivity. Default is 1 (adjacent).
+	var/hearboost = 0 										//Extra ranger on whisper hearing. 0 is adjacent.
 
 	var/taste_sensitivity = TASTE_NORMAL							// How sensitive the species is to minute tastes.
 	var/allergens = null									// Things that will make this species very sick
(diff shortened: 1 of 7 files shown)


Atermonera:
Staff input: Tune teshari to just +1 (same as taj) and this is fine. As a number, 3 tile range seems small, but in practice, that's actually a pretty significant distance to get away from the wallflowers.

Varlaisvea:
tesh now have two less ears and one less boosted whisper hearing range

Greenjoe12345:
shame it has to be lowered down to the same level as taj :(

code/modules/mob/living/carbon/human/species/species.dm
```diff
@@ -53,6 +53,7 @@
 	var/blood_level_fatal = 0.4								//"Fatal" blood level; below this, you take extremely high oxydamage
 	var/hunger_factor = 0.05								// Multiplier for hunger.
 	var/active_regen_mult = 1								// Multiplier for 'Regenerate' power speed, in human_powers.dm
+	var/hearboost = 0 										//Whisper hearing sensitivity. Default is 1 (adjacent).
```
Atermonera, line 56:
```suggestion
	var/hearboost = 0 										//Extra range on whisper hearing. 0 is adjacent.
```

code/modules/mob/living/say.dm
```diff
@@ -331,20 +331,27 @@ var/global/list/channel_to_radio_key = new
...
-					if(dst > message_range && dst <= w_scramble_range) //Inside whisper scramble range
+					if(dst > message_range_boosted && dst <= w_scramble_range_boosted) //Inside whisper scramble range
 						if(M.client && !runechat_enabled)
 							var/image/I2 = listening[M] || speech_bubble
 							images_to_clients[I2] |= M.client
 							M << I2
 						M.hear_say(stars_all(message_pieces), verb, italics, src, speech_sound, sound_vol*0.2)
-					if(dst > w_scramble_range && dst <= world.view) //Inside whisper 'visible' range
+					if(dst > w_scramble_range_boosted && dst <= world.view) //Inside whisper 'visible' range
 						M.show_message("<span class='game say'><span class='name'>[name]</span> [w_not_heard].</span>", 2)
```
Atermonera, line 355:
```suggestion
				if(dst <= min(message_range + M.hearing_boost_range(), world.view) || (M.stat == DEAD && !forbid_seeing_deadchat)) //Inside normal message range, boosted whisper range, or dead with ears (handled in the view proc)
					if(M.client && !runechat_enabled)
						var/image/I1 = listening[M] || speech_bubble
						images_to_clients[I1] |= M.client
						M << I1
					M.hear_say(message_pieces, verb, italics, src, speech_sound, sound_vol)
				if(whispering && !isobserver(M)) //Don't even bother with these unless whispering
					if(between, dst message_range + M.hearing_boost_range(), w_scramble_range + M.hearing_boost_range()) //Inside whisper scramble range
						if(M.client && !runechat_enabled)
							var/image/I2 = listening[M] || speech_bubble
							images_to_clients[I2] |= M.client
							M << I2
						M.hear_say(stars_all(message_pieces), verb, italics, src, speech_sound, sound_vol*0.2)
					if(between(dst, w_scramble_range + M.hearing_boost_range(), world.view) //Inside whisper 'visible' range
						M.show_message("<span class='game say'><span class='name'>[name]</span> [w_not_heard].</span>", 2)
```
This diff renders really messily but I'm basically proposing you make a proc on /mob to get the "hearing boost range", and override that on human mobs to get the species boost range. Drops the mess of `message_range_boosted` and `w_scramble_range_boosted`.
Check my syntax on `between()`, the order of arguments is out of my ass.

Atermonera (changes requested):

Atermonera:
It's `between(lower, middle, upper)` ftr, so the value that you're clamping is the middle arg. Which is really weird! And imo poor usability.

Atermonera (approved):

PR #8999, opened 2023-02-26
Adds departmental cloaks to lockers, cleans closet list formatting

Frenjo:
Cloaks are good, so why not make prefab departmental ones available within their respective department lockers?
They can be worn over winter coats too, so you can really keep warm by putting on those extra layers!

I originally intended to just add cloaks to the Head of Staff closets, but once I realised there were other departmental prefabs I figured they might as well be added too.

Adds departmental cloaks to the following lockers:
- Site Manager
- Head of Personnel
- Research Director
- Chief Medical Officer
- Chief Engineer
- Quartermaster
- Atmospherics
- Cargo / Mining
- Medical / Paramedic / Psychiatrist

Also fixes some weird list formatting across closets for consistency.

🆑
rscadd: Departmental cloaks have been added to their respective lockers.
/🆑

### DIFF
--- a/code/game/objects/structures/crates_lockers/closets/secure/bar.dm
+++ b/code/game/objects/structures/crates_lockers/closets/secure/bar.dm
@@ -6,4 +6,5 @@
 	close_sound = 'sound/machines/closet/closet_wood_close.ogg'
 
 	starts_with = list(
-		/obj/item/reagent_containers/food/drinks/bottle/small/beer = 10)
+		/obj/item/reagent_containers/food/drinks/bottle/small/beer = 10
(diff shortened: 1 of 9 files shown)


Woodratt:
Kind of against this since several of the lockers are already pretty well cluttered and I personally feel as though departmental cloaks ought to stay as something in the loadout.

Frenjo:
That's a fair assessment.

If not closets, I definitely think there should be a way to access the departmental cloaks in-game, perhaps via cargo ordering?

That's out of scope for this PR so I'm closing pending better ideas.

PR #9019, opened 2023-03-03
meson/nvg/thermal etc overlays can now be toggled

Varlaisvea:
title

alt-click to toggle the cosmetic overlay on mesons, nvgs, thermals, gravitons, material scanners, science goggles
gameplay functionality is unchanged but now you can disable the overlays separately to the functionality if you get eye strain

### DIFF
--- a/code/modules/clothing/glasses/glasses.dm
+++ b/code/modules/clothing/glasses/glasses.dm
@@ -26,6 +26,9 @@ BLIND     // can't see anything
 	var/active = 1
 	var/activation_sound = 'sound/items/goggles_charge.ogg'
 	var/obj/screen/overlay = null
+	var/obj/screen/stored_overlay = null //Stores the overlay when not in use.
+	var/has_overlay = FALSE
+	var/overlay_enabled = FALSE
 	var/list/away_planes //Holder for disabled planes
 	drop_sound = 'sound/items/drop/accessory.ogg'
 	pickup_sound = 'sound/items/pickup/accessory.ogg'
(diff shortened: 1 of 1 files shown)


Atermonera:
That eye-searing overlay is a balance measure, to give some disadvantage to just always running around with them on. Do you have a proposed alternative to discourage that behaviour?

Atermonera:
~~So what I'm hearing is we should make the goggles give eye damage to the _character_...~~
Also honestly I wear a wool jacket as a middle layer, and just walking around with an outerlayer on top of it it'll build up a decent static charge, to the point where every time I walk into work and take off my jacket to sit down at my desk, I'll touch my workstation to deliberately try to shock myself to discharge the static before I accidentally shock myself somewhere else...

I think I need to pitch this to more staff, now that I've had some time to actually think about this. It was brought up in passing at the last staff meeting as like a footnote at the very end of a particularly long meeting so there wasn't really an official discussion, just peoples' immediate takes. I might try to get opinions before the next staff meeting but can't make guarantees.

Atermonera:
Oh and while I'm lying here in bed, I dug through the commit history. The change that appears to have first I triduced the overlays (which does date to bay) didn't have much conversation at all, which means it'll be very hard to procure any original discussion for whatever insights it may offer

Atermonera:
Talked w/ some staff. Concerns are that...
* If you _can_ turn off the overlays, why would you use them at all?
* If people turn off the overlays, they will forget that they're wearing material scanners or NVGs and then go to weld and take a bunch of eye damage.
* It's nice diegetic effect of wearing something over your eyes and seeing something affect your vision.
Proposed alternatives are...
* Give mesons the same overlay as NVGs (Or otherwise recolor the mesons to still be _indicative_ without being _eyesearing_).
* Make the overlay a circular gradient, sorta like how the screen fades to white when you're near-death.
* Just remove the overlays wholesale.

I'm most keen on trying out the circular gradient, because that seems like the best overall solution to particular colors are bad. If you're not up to that, I'm much more partial to recoloring the meson overlay than to removing them. And if you _have_ to remove them, possibly just the removing the overlay for the mesons, since that's the most offensive.
I don't see any particular value in adding a toggle over just removing the overlays outright, because both options have the same practical effect.

Varlaisvea:
a circular gradient would honestly be really neat, i'm not sure how we'd go about doing it though tbh
as far as the fear of people forgetting they're wearing them and then welding their eyes goes... skill issue

Atermonera (changes requested):
To be clear the staff input is that this will not be merged as-is, still waiting on you to propose another solution or implement one of the suggestions.

Varlaisvea:
will be faster to close this and remake it with the changes, stay tuned!

PR #9021, opened 2023-03-04
Bioprinter refactor and nanoUI-ification

Ilysen:
fully refactors bioprinters with cleaner and documented code, and supplants its interface (which currently works via popup chains) with a new and shiny tgui window.

key backend changes:
* unified the base type and "flesh" type, since the robotic version of the organ printer is obsolete
* changed path from `/obj/machinery/organ_printer` to `/obj/machinery/bioprinter`
* now makes use of a cached timer to track whether or not it's printing something
* general cleanliness; raw deciseconds -> `SECONDS`, biomass reagent ID is changeable per type, and so on

key frontend changes:
* brand new nanoui interface
* rewritten descriptions and messages
* general polish - now updates the draw mode of syringes after using them to inject a blood sample, prints can be canceled/discarded early, blood sample lists off its species and dna hash, etc

I'm aware that the cost display isn't properly centered, but I can't figure out how to fix it :p

<details><summary>screenshots, in no particular order</summary>

dreamseeker_jQCuNQxohk
50XuqvHagH
NHuE1SSx2Q
nCZSRsWGBM
FZjBIHisAK

testing:
ongoing. need to test swapping stock parts, among other stuff. need to figure out how to make the printing section scrollable


Ilysen:
I don't *think* this has any problems, although in light of recent discussions it's possible that ater or one of the other devs might want to opt out of the tgui in favor of a simpler HTML interface

Atermonera:
I find it just a little ironic that, looking at the tgui template, this looks like it'd be pretty easy to transplant to nanoui.

Ilysen:
is nano stable for more users? I seem to have hiccups with nano locally during testing, but if it's more stable on live deployment then I'm totally okay biting the bullet

Atermonera:
I need to run some tests to verify the state of tgui so lets hold on that until I can actually confidently say whether there _are_ any issues with using tgui to begin with (Since the issues haven't ever been well documented and we don't seem to have players actively using linux rn)

Ilysen:
Marking this as a draft 'til I can replace it with a nano interface, since ater's wantin' to move away from tgui

Ilysen:
tgui stuff is now nano!

code/game/machinery/bioprinter.dm
```diff
@@ -44,336 +59,245 @@
...
+
+	data["biomassContainer"] = container
+	data["biomassVolume"] = get_biomass_volume()
+	data["biomassMax"] = container?.reagents.maximum_volume
+
+	data["printTime"] = print_delay
+	data["isPrinting"] = print_timer != null
+
+	data["dna"] = loaded_dna
+	data["dnaHash"] = loaded_dna?["blood_DNA"]
```
Atermonera, line 189: :o TIL

Atermonera (approved):

PR #9034, opened 2023-03-07
Flavour Enhancement: Flavour

Cerebulon:
Adds unique flavour strings to **all food**.

More recipes containing explicit raw ingredients are now "cleared" on production and fully set up in the food init, fixing the "taste of raw dough" and similar issues where the cooked flavour should be totally different from the ingredients. This was previously very inconsistent.
Generic "nutriment" may now contain allergens, depending on the food it is in. Especially useful for above cases where the recipe is cleared but the result should still contain e.g. eggs or cheese. This prevents the need to add units of "generic protein" to trigger dairy allergies.
- When adding meat specifically this is usually done with the "protein" reagent (with custom flavour string applied) since that has special species code for carnivorous-leaning species beyond the allergen concerns.
- Sensible allergens added to foods where this isn't already handled by explicit reagent inclusions (e.g. protein, egg, fruit juice), which could serve a proper optional allergen system later.
- Cases where the recipe is NOT cleared, duplicate reagent adds removed and/or rolled into nutriment quantities. Sometimes nutriment_amt is still buffed so cooked food is more nutri-tasty than guzzling raw onions.
- Cases where "nutriment" was added in init moved, because nutriment_amt literally does that, that's what it exists for.
- In theory you could make a food and extract the 'custom' nutriments/proteins to add weird flavourings to other foods, if that's your jam. Not really intended, but that's just how the condimaster works (The nutri-allergen change does mean you can't just magic allergens out of most food though).
- Subtypes of nutriment now accept custom taste strings on init. This was a bug left over from the original implementation of taste.

While I tried to account for them where possible, there may be cases where certain foods have significantly more or less bites/nutritional value than before. If any of these seem insane, please let me know.

Some foods that contained egg by nature of the way dough is made no longer contain egg if it makes no god damned sense. As far as I'm concerned, basic dough being made with egg is primarily a game mechanics consideration because flour+water would fuck with too many other systems, and shouldn't make all bread potentially deadly to skrell. Anything containing dough now clears on init so egg is applied as needed.

Renamed some foods that were excessively tg-era jokes or weirdly inconsistent.

Food renames (Will need wiki change):
Space fries -> skinny fries
Wing Fang Chu -> sticky alien wings
Omelette Du Fromage -> cheese omlette
Tossed salad -> garden salad
Roffle waffles -> fungal waffles (no longer contains egg!)
Spacy liberty duff -> psilocybin duff
Cheese cake->cheesecake
Vanilla cake -> spongecake
Spesslaw -> beef stroganoff (+ 5 flour to recipe as it was overriden by spag meatball anyway)

All food names changed to lower case (instead of a weird mixture).

To do **some another time**: holy shit a lot of food descriptions suck.

Misc:
- Fluff tweaks for some vendor stuff.
- Remove unused lore_fluff side-file for drinks cans. This stuff is already written in the items in their own files.
- Protein now tastes of "umami" by default instead of "some sort of meat"
- Capsaicin tastes of "spiciness" by default.
- Roro lives.


Atermonera:
@Cerebulon merge conflicts

Cerebulon:
> @Cerebulon merge conflicts

fixed

Cerebulon:
> @Cerebulon merge conflicts

fixed pls

code/modules/food/food/snacks.dm
```diff
@@ -4354,7 +4308,7 @@
 	if (!flat_icon)
 		flat_icon = getFlatIcon(src)
 	var/icon/I = flat_icon
-	color = "#FFFFFF" //Some fruits use the color var. Reset this so it doesn't tint the batter
+	color = "#FFFFFF" //Some fruits use the color var. Reset this so it doesnt tint the batter
```
Atermonera, line 4311: `doesn't`

Atermonera (commented):
godsdamn that file is way too big

code/modules/food/food/snacks.dm
Cerebulon: oh i think this is reverting a correction from the big spell check PR which conflicted with this. it's literally just a code comment though. will address other stuff when you're done.

Atermonera: Me posting review comments means I've finished the review.

Atermonera (approved):

PR #9038, opened 2023-03-07
Hotfix for moving pulled objects

### LINKED ISSUE #9037: Cannot click tiles to move objects
When pulling an item, like a crate, you were able to click a tile to move it there. It does not seem to work anymore.
Move a crate from tile A to tile B by clicking on tile B while pulling the crate.
#### What actually happened
Absolutely nothing.
#### Steps to reproduce
Pull an object. Click a turf. Nothing happens.
#### Additional info:

Ilysen:
Closes #9037

#8970 accidentally didn't include a `..()` call on floor's `attack_hand`, and so the base `/turf` call was being ignored. this prevented pulled objects from being moved with clicking. I'm sorry! this is a hotfix for that.

quickly tested on the example map; I used click-dragging to move stuff around both before and after setting all the turfs to be snowy. also now tested on cynosure as a survivalist by dragging around an umbrella in the wilderness.

### DIFF
--- a/code/game/turfs/simulated/floor_attackby.dm
+++ b/code/game/turfs/simulated/floor_attackby.dm
@@ -136,12 +136,13 @@
 						to_chat(user, "<span class='warning'>You need more welding fuel to complete this task.</span>")
 
 /turf/simulated/floor/attack_hand(mob/user)
-	if (has_snow())
+	if (!user.pulling && has_snow())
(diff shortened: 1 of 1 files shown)


Atermonera (approved):
I might consider making `if(user_pulling) return ..()` as an early return but this is fine.

PR #9041, opened 2023-03-08
Union Proposal: Emitter Safety Changes

Varlaisvea:
emitters now have targeting lasers! unless they're disabled (via multitool or emag), they'll always be active when the emitter is live, and on a dormant emitter, can be toggled on/off by alt-clicking.

handy guide to emitters:
- Is the targeting software disabled (multitool/emag)? If yes, then there's no laser. If no, then...
- Is the emitter active right now? If yes, then there will always be a laser (unless it's disabled). If no, then...
- The laser _can_ be active if someone enables it via alt-click.

the targeting laser is purely visual, and doesn't damage anything. since it's light-based, it passes through windows and machinery, but not walls or opaque objects (like mechs).

emagging emitters now also randomises their burst size and fire delays, making them far more unpredictable - and if you try and reenable the safeties on an emagged emitter, you'll end up just turning the emitter on instead. but thankfully that'll never happen!

remember: **the emitter is always live, even when it isn't.**


code/modules/power/singularity/emitter.dm
```diff
@@ -102,9 +120,26 @@
...
+				if(T.opacity) //Is it an opaque turf (i.e. a wall)? Stop here.
+					break beammeupscotty
+				else if(T && T.contents)
+					for(var/atom/A in T.contents) //Otherwise, check the tile for any opaque atoms, like mecha or curtains
+						if(A.opacity) //We don't care about density. This means the targeting laser will still pass through dense, non-opaque atoms (like windows and blobs).
+							break beammeupscotty //You know kirk never actually said this?
+						else
+							beamtarget = T
+				T = get_step(T, src.dir)
+		src.Beam(beamtarget, icon_state = "sniper_beam", time = 2 SECONDS, maxdistance = 50) //Refreshes on process() so short duration is fine.
```
Atermonera, line 141:
```suggestion
	if(safety) // Being emagged will prevent the safety from being enabled
		var/list/hit = check_trajectory(get_step(src, dir), src)
		src.Beam(hit[0], icon_state = "sniper_beam", time = 2 SECONDS, maxdistance = 50) //Refreshes on process() so short duration is fine.
```
Double check this `hit` indexing because that might make the beam 0 distance

code/modules/power/singularity/emitter.dm
```diff
@@ -147,6 +182,27 @@
...
+		else if(safety_override == FALSE)
+			safety_override = TRUE
+			safety = FALSE
+			to_chat(user, "You disable [src]'s targeting software. It was just getting in the way anyway.")
+			return
+		else
+			safety_override = FALSE
+			to_chat(user, "In a brief moment of wisdom, you reenable [src]'s targeting software.")
+			return
+
```
Atermonera, line 205:
```suggestion
		try_toggle_safety(user)
		return

```

code/modules/power/singularity/emitter.dm
```diff
@@ -288,4 +349,34 @@
...
+		return
+	if(safety_override)
+		to_chat(user, "<span class='danger'>\The [src]'s targeting laser has been manually disabled!</span>")
+		return
+	if(emagged) //You weren't standing in front of it, right?
+		to_chat(user, "<span class='warning'>The targeting laser on [src] short-circuits!</span>")
+		activate(user)
+		return
+	else
+		toggle_safety(user)
```
Atermonera, line 378:
```suggestion
	try_toggle_safety(user)
	
/obj/machinery/power/emitter/proc/try_toggle_safety(mob/user)
	safety_override = !safety_override
	if(safety_override)
		to_chat(user, "In a brief moment of wisdom, you reenable [src]'s targeting software.")
	else
		to_chat(user, "You disable [src]'s targeting software. It was just getting in the way anyway.")

	toggle_safety()
	return
```

code/modules/power/singularity/emitter.dm
```diff
@@ -288,4 +349,34 @@
...
+
+/obj/machinery/power/emitter/proc/toggle_safety(mob/user as mob)
+	safety = !safety
+	user.visible_message("[user.name] [safety ? "enables" : "disables"] \the [src]'s targeting laser.","<span class='notice'>You [safety ? "enable" : "disable"] \the [src]'s targeting laser.</span>")
```
Atermonera, line 382:
```suggestion
/obj/machinery/power/emitter/proc/toggle_safety()
	safety = !emagged && (powernet || !active_power_usage) && (!safety_override || active)
```
Truth table
```
Emag | Override | Active | Has Power | Safety
1    | X        | X      | X         | Off
X    | X        | X      | 1         | Off
0    | 0        | 0      | 0         | On
0    | 1        | 0      | 0         | Off
0    | X        | 1      | 0         | On
```

Atermonera (changes requested):

code/modules/power/singularity/emitter.dm
```diff
@@ -45,6 +47,18 @@
 	. = ..()
 	if(state == 2 && anchored)
 		connect_to_network()
+	if(emagged) //So they can start emagged, if for some ungodly reason you want to torment the engineers.
+		locked = FALSE
+		safety = FALSE
```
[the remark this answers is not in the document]
Varlaisvea: the default value is meant to be false which makes these entirely and completely redundant now i think about it,,

code/modules/power/singularity/emitter.dm
```diff
@@ -102,9 +120,26 @@
 	if(stat & (BROKEN))
 		return
 	if(src.state != 2 || (!powernet && active_power_usage))
-		src.active = 0
+		src.active = FALSE
+		src.safety = FALSE //Don't bother with the targeting laser if we're unpowered.
```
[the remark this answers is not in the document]
Varlaisvea: wouldn't that mean if the safety is _off_, and then it loses power, the safety becomes _on_ (and then flicks on and off every process() tick)?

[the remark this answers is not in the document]
Atermonera: `toggle_safety()` is... maybe not the best term anymore. Per the truth table below I'm basically positing that it update the state based on several factors rather than explicitly toggle

code/modules/power/singularity/emitter.dm
```diff
@@ -63,31 +75,52 @@
...
+		if(!safety && !emagged && !safety_override)
+			toggle_safety(user) //Enable safeties, unless it's been tampered with or they're already active.
+		src.active = TRUE
+		to_chat(user, "You turn on [src].")
+		src.shot_number = 0
+		src.fire_delay = get_initial_fire_delay()
+		message_admins("Emitter turned on by [key_name(user, user.client)](<A HREF='?_src_=holder;adminmoreinfo=\ref[user]'>?</A>) in ([x],[y],[z] - <A HREF='?_src_=holder;adminplayerobservecoodjump=1;X=[x];Y=[y];Z=[z]'>JMP</a>)",0,1)
+		log_game("EMITTER([x],[y],[z]) ON by [key_name(user)]")
+		investigate_log("turned <font color='green'>on</font> by [user.key]","singulo")
+	update_icon()
```
Atermonera, line 123: snip

code/modules/power/singularity/emitter.dm
```diff
@@ -288,4 +335,57 @@
...
+		to_chat(user, "<span class='warning'>The targeting laser on [src] short-circuits!</span>")
+		activate(user)
+		return
+
+	safety_override = !safety_override
+	if(safety_override)
+		to_chat(user, "You disable [src]'s targeting software. It was just getting in the way anyway.")
+		safety = FALSE
+	else
+		to_chat(user, "In a brief moment of wisdom, you reenable [src]'s targeting software.")
```
Atermonera, line 391:
```suggestion
/obj/machinery/power/emitter/proc/update_safety()
	safety = !emagged && (powernet || !active_power_usage) && (!safety_override || active) //Emagged, overridden or depowered = no safety. Otherwise, toggle.

/obj/machinery/power/emitter/proc/can_toggle_safety(mob/user)
	if(state != STATE_WELDED)
		to_chat(user, "<span class='notice'>You can't adjust the emitter's targeting laser without securing it first.</span>")
		return FALSE
	if(active)
		to_chat(user, "<span class='danger'>You can't adjust \the [src]'s targeting laser while it's firing!</span>")
		return FALSE
	if(safety_override)
		to_chat(user, "<span class='danger'>\The [src]'s targeting laser has been manually disabled!</span>")
		return FALSE
	if(emagged) //You weren't standing in front of it, right?
		to_chat(user, "<span class='warning'>The targeting laser on [src] short-circuits!</span>")
		activate(user)
		return FALSE
	return TRUE

/obj/machinery/power/emitter/proc/try_toggle_safety(mob/user) //Enable or disable the targeting laser when it's dormant
	if(!can_toggle_safety(user))
		return
	to_chat(user, "You adjust \the [src]'s targeting laser.") //give feedback to user!
	safety = !safety
	return

/obj/machinery/power/emitter/proc/toggle_safety_override(mob/user) //Prevent the targeting laser from showing up at all
	if(!can_toggle_safety(user))
		return
	safety_override = !safety_override
	if(safety_override)
		to_chat(user, "You disable [src]'s targeting software. It was just getting in the way anyway.")
	else
		to_chat(user, "In a brief moment of wisdom, you reenable [src]'s targeting software.")
	update_safety()
```

Atermonera (changes requested):

Varlaisvea:
OKAY i think i've finally got it working again. time to test.

code/modules/power/singularity/emitter.dm
```diff
@@ -63,31 +75,33 @@
...
-			to_chat(user, "<span class='warning'>The controls are locked!</span>")
-	else
-		to_chat(user, "<span class='warning'>\The [src] needs to be firmly secured to the floor first.</span>")
-		return 1
+	if(state != STATE_WELDED)
+		to_chat(user, SPAN_WARNING("\The [src] needs to be firmly secured to the floor first.</span>"))
+		return TRUE
+
+	if(!powernet && active_power_usage)
+		to_chat(user, SPAN_WARNING("\The [src] has no power!"))
```
Atermonera, line 87: Missing return

code/modules/power/singularity/emitter.dm
```diff
@@ -43,8 +48,15 @@
 
 /obj/machinery/power/emitter/Initialize()
 	. = ..()
-	if(state == 2 && anchored)
+	if(state == STATE_WELDED && anchored)
 		connect_to_network()
+	if(emagged) //So they can start emagged, if for some ungodly reason you want to torment the engineers.
+		emag_act()
```
Atermonera, line 54: This doesn't actually work b/c there's an `if(!emagged)` guarding the logic in `emag_act()`

Varlaisvea: we're setting emagged = TRUE on line 59, that shouuuuld mean emag_act() works? will test

code/modules/power/singularity/emitter.dm
Varlaisvea: fix'd...

code/modules/power/singularity/emitter.dm
Varlaisvea: can confirm it does work! this is specifically for emitters that start emagged (e.g. for asshole PoIs - see 56-59 ominous emitter as a template for those), rather than emitters that get emagged in-round (which also works, but because those call emag_act() directly rather than checking if they're emagged on initialize)

Varlaisvea:
## UNIT_TEST: !! FAILURE !! [MAP: Active edges (all maps)]: Maps contained 1 active edges at round-start.

ci pls i didn't even TOUCH the map

code/modules/power/singularity/emitter.dm
Atermonera: Did you check that burst count and firerate were updated? Line 59 is in a class definition, meaning an instance of that class will set variables to their proscribed values (`emagged = TRUE`) _before_ called `Initialize()`.
`Initialize()` will correctly call `emag_act()`, but `emag_act()` will return immediately because of the `if(!emagged)` check on line 267 guarding all of the logic.

Atermonera: @Varlaisvea Still waiting for a response.

Atermonera: @Varlaisvea ping

Spookerton:
coauthor to resolve conflict, rebase, run CI, etc

Spookerton:
Aaaalrighty. This didn't compile when it was opened and it doesn't compile now. Considering it adds a big mess of snowflake procs for no actually good reason and the author's been gone for five months, dead.

PR #9048, opened 2023-03-10
skybridge lounge!!!!

Varlaisvea:
image

western skybridge now has a lounge
here's how it looks
it's comfy as fuck


Atermonera (dismissed):

Atermonera:
The code itself looks fine; why do we need this? Sorta detracts from the "skybridge" by stapling things onto the side of it...

Varlaisvea:
do we _need_ it? probably not
is it still a comfy aesthetic improvement? hell yeah

Woodratt:
Not keen on it on a personal level, but that's just me. Otherwise fine with it sans one thing. There's no air alarm in the closed off area. It'd probably be better off if it was not all single windowed/windoored off from the main hall.

Atermonera:
> a comfy aesthetic improvement

Does it need to be glued onto the side of the skybridge? What about this space is unique from other extant areas of the station?

Varlaisvea:
doesn't technically _Need_ to be where it is but it wouldn't be a skybridge lounge if it wasn't on the skybridge

(and also it's there because the engineering team built it like Several Times in-round in that exact position in various different configurations)

as for what it offers that we don't already have: it's a comfy chill space with a nice view where people can sit and rp and watch the snow fall, and weirdly enough we have very few of those that aren't department-locked!

Atermonera:
> because the engineering team built it like Several Times in-round

Several people doing this independently or one person doing it over and over and over again? Was it actually used? Neil Patterson building a sauna every round for _months_ wasn't used as justification for mapping it in, ftr.

Atermonera:
Chatted about this at the staff meeting.
There's apparently a seating area already at the western skybridge, could you just expand that out a little bit?
As-is, there's already a fair number of secluded spots you can sneak off to to RP in private, I don't much see the value in adding another. The map was actively designed to try to draw characters together into the larger common spaces to, y'know, RP together. And if you want to RP in a skybridge, you... can just stand in the hallway? Anyways it sorta reads to me like the aesthetic you're trying to go for here is akin to the Space Needle, but... it's a 2-story building.

Atermonera:
Closing due to inactivity. Please let me know if you'd like this reopened.

PR #9053, opened 2023-03-12
Full cult gamemode refactor

Ilysen:
this was originally a project I mothballed. then I had an actual real life dream that I was a nar-sie cultist, though, and at that point it kind of became a legal requirement that I finish it

#### purpose

cult's code is among the oldest in the repo, and it shows. the amount of bugs and overhead is ludicrous for what it is, the code quality is in general pretty hideous, and the age shows abundantly. cult never really gets played, and I'm fairly sure that the bad code quality is a pretty big reason for why. even if it's just bringing it up to the same functionality as current (more or less), having the code be modernized paves the way for making it not suck or to have it be reworked.

the goal here is to ensure roughly full feature parity with the current code, barring deliberate adjustments and removals. although the code is being completely remade, having the gameplay start more or less the same seems like a good start towards potential feature updates in the future.

##### key front-end changes
* Blood Boil rune removed (massive AoE damage with a chance to gib; not HRP-friendly)
* Manifest Ghost rune removed (create hordes of invincible ghost-controlled humans; not HRP-friendly. could replace it with a Seance rune if the goal is to speak with the dead)
* the arcane tome now has tgui, including in-game documentation
* runes and talismans can now be examined to get a summary of what they do
* general description and text consistency across the board

##### key back-end changes
where to begin?
* completely refactored `/obj/effect/rune` from scratch. logic is now dependent on various subtypes, rather than being base-level procs. a bunch of logic like required invokers and such can now be easily overridden (and in some cases varedited) on a per-type basis.
* completely refactored `/obj/item/paper/talisman` from scratch in much the same way as runes - logic has been generalized and moved onto subtypes as relevant
* completely refactored arcane tomes. they're now way cleaner, less snowflakey, and have a spooky new ui.
* removed the vestiges of the cult word research system. word meanings are now always fixed, and only used for drawing rune sprites.

#### media/extra info

the alchemical symbol used in `bg-cult.svg` is in the public domain and is taken from [this wikimedia commons page](https://commons.wikimedia.org/wiki/File:Black_sulfur_symbol.svg)

#### testing

wip. there's a lot to do.


Ilysen:
Looking at the code for soul stones, it turns out that it'll be a pretty big undertaking unto itself; as a result, to make the maintainers at least *somewhat* less horrified, I'll defer it for a future PR so that this one can focus on the main stuff.

Ilysen:
There's still testing in progress and general touchups here and there, but I am tentatively ready to say that I think this might be ready for review. I'm sure there's lots to do and plenty of mistakes I've made, but my (new) goal of getting the runes, talismans, and tomes refactored seems to be mostly complete. If this gets merged, follow-up stuff would include work on constructs.

PR #9059, opened 2023-03-14
Ports extensions from Nebula

Atermonera:
First pass at porting extensions, as on tin. Provided without verification but with warranty.
* Grabs what appears to be the basic underlying things.
* Updates hand labeler to use the label extension, tries to tie it into other places but this is probably incomplete and will most likely require further effort on my part.
* Swipes, but does not tick, some extensions that I'm particular interested in trying to use.
  * /eye and /on_click are unticked.
* Makes extensions a _thing that exists_, such that the lack of extensions outright is no longer a non-starter for a number of other potential ports.


Atermonera:
I agree that the two-implementation solution is messy (See: nanoui v. tgui, among probably numerous other examples), but there actually isn't that much that uses components right now.
I much more often see shiny things that we _could have_, but for lacking extensions, than I do shiny things that we're _getting_ because we have components.

Spookerton:
I would want this to come with a commitment to transition existing component use to extensions, which I think defeats the "I much more often see shiny things that we could have" sentiment which, I assume, carries the caveat of still wanting the shiny things from components without the effort of translation. I agree with MistakeNot4892 - both is a recipe for overhead, interoperability issues, and dev headaches.

I'm biased in favor of extensions because they're what I use already elsewhere. I've found DSC a tad less approachable, largely due to the rote use of its own signals system creating a lot of async funnies and signal comprehension instead of direct calls.

Buuuut it is what currently exists. Translating extensions to it is just as possible.

PR #9063, opened 2023-03-15
Consistency Pass #1

enbyfriend:
- Power cells of all types are now considered stock parts
- Went through materials.dm and did some copy+paste work make material definitions consistent across the codebase (eg "steel" = 50 -> MAT_STEEL = 50)


Atermonera:
What's the mechanical advantage of subtyping cells under stock_parts? If it's purely organizational I can't say I'm in favor atm.

Atermonera:
@enbyfriend pong!

enbyfriend:
In theory, the advantages of setting cells as stock parts as opposed to what they were before would allow you to
1) carry cells and the like inside an RPED and
2) swap out cells in machines using the RPED
Of course, I'll have to take a look at sci code to see how it all works, but that's the theory in any case

Atermonera:
tl;dr: Your find/replace is a little flawed because it touches chemistry, and I'm opposed to making typepaths longer where it confounds the overall purpose of a type.
Sorry it's taken me so long to get back to this. I don't think a theoretical benefit is worth the change, and I don't think that's the right pattern to approach - cells _can_ be upgraded in the same manner as stock parts, but the functions are different. Stock parts impact actual mechanical effects on the device - more storage, do things faster, etc. Power cells basically just govern "You can use it for longer before you have to recharge."
For items where "Time until recharge" is an important factor, you generally need a simpler/faster system for removing/replacing the cell than "Something you can only get from science" (See guns, for example).

I'm not _wholly_ opposed to the idea of cells fitting into the RPED, but I think the approach should be to make it more generalized, perhaps by an extension or datum.

I appreciate the effort you went to here, but I think that's enough of a nonstarter that further attempts should be in a separate PR (Also, this has been open forever, and there's merge conflicts) And the find-replace should be in a separate PR because that's much more digestible and I'm less likely to take issue with it.

PR #9064, opened 2023-03-15
Slimes can eat loose food, decals, dirt, and so on

Ilysen:
#### purpose

slimes right now can only eat live monkeys, which is both needlessly cruel (though I recognize it's likely just a vestige of old code) and doesn't align with how prometheans can absorb basically literally anything. this makes them more alike to their sapient cousins and is also cute.

might be good to get some lore feedback from cere; there was some discussion a while ago but I don't think she mentioned anything solid in either direction on if this was something normal slimes did

#### details

slimes will now recognize things designated as edible - including snack items, decals on the ground, or turfs with enough dirt on them - as valid food, and will move over to them and do various things like schlorping them up or cleaning the dirt off of the ground, gaining a token amount of nutrition in the process. slimes will prioritize the closest item, and will always prioritize living targets compared to inanimate ones (so if given the choice between a bunch of pizza and a living monkey, it'll always choose the monkey.)

valid new types of slime food includes all snacks, most decals, turfs with enough dirt on them, and organs. meaty food (i.e. food with the protein reagent) is more nutritious for slimes than other kinds, although they can still gain nutrition from any food-type reagents. offal is very nutritious! just maybe, yknow, keep them away from anywhere where brains are stored

backend changes;
* atom-level `is_slime_food()` proc added, defaulting to null
* atom-level `slime_chomp(slime)` proc added, used to handle the logic of a slime "eating" this atom
* xenobio slimes' AI holder now has extra logic in its `find_target()` - if they're not fixated on a living thing, rabid, etc. then they'll search for any reachable atom whose `is_slime_food()` returns a nonzero value, and if they find one, they'll roll over to it and try to eat it.
* note that human/monkey consumption has not been changed and still uses the old system

#### testing

spawn in some slimes or let 'em out of their cages, watch as they vacuum up everything they can eat like the adorable things they are

#### media

video demo - note how the slime immediately prioritizes a monkey once it sees one

https://user-images.githubusercontent.com/47678781/225287918-69696a06-e154-40d0-ba06-2d37c07ecd44.mp4

### DIFF
--- a/code/modules/ai/ai_holder_subtypes/slime_xenobio_ai.dm
+++ b/code/modules/ai/ai_holder_subtypes/slime_xenobio_ai.dm
@@ -112,8 +112,34 @@
 			adjust_discipline(-1)
 			last_discipline_decay = world.time
 
+/datum/ai_holder/simple_mob/xenobio_slime/find_target(list/possible_targets, has_targets_list)
+	if (hostile)
+		. = ..()
+	if (target || leader || rabid) // Always prioritize mobs, if we can. Otherwise, check for loose things to eat if we're not otherwise busy
+		return
+	var/list/yummies
+	for (var/atom/M in view(world.view, holder))
+		if (M.is_slime_food())
+			// Make sure that we can reach the object first, but don't try too hard at it and make sure that the thing is edible beforehand
+			var/list/path = AStar(holder.loc, M.loc, astar_adjacent_proc, /turf/proc/Distance, min_target_dist = 1, max_node_depth = world.view * 2, id = holder.IGetID(), exclude = obstacles)
+			if (path?.len)
+				LAZYDISTINCTADD(yummies, M)
+	if (!LAZYLEN(yummies))
+		return
+	var/atom/yummy = pick(yummies)
+	var/closest_distance = get_dist(holder, yummy)
+	for (var/atom/M in yummies) // Find the closest snack!
+		var/dist = get_dist(holder, M)
+		if (dist < closest_distance)
+			yummy = M
+			closest_distance = dist
+	give_target(yummy)
+	give_destination(yummy.loc)
+	return yummy
+
 /datum/ai_holder/simple_mob/xenobio_slime/handle_special_tactic()
 	evolve_and_reproduce()
+	nom()
 
 // Hit the correct verbs to keep the slime species going.
 /datum/ai_holder/simple_mob/xenobio_slime/proc/evolve_and_reproduce()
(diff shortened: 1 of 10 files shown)


Cerebulon:
Not sure about this applying to all food/dirt/etc. myself. I think it kind of defangs the concept of slimes if you can just feed them normal human food.

I've always felt the intent with monkeys is that they only recognise moving/respiring/warm things as food in the way that many animals won't eat something wierd and unfamiliar even if it's "good for them", or how some snakes will only eat a mouse if its unconcious and/or warm rather than cold and/or dead. And part of what makes prometheans _prometheans_ is that they can recognise that hey wait, no, that's unnecessary. I really don't like the idea of xenobio just being able to jump a bunch of cold, dead meat in there and get the same effect.

Maybe this could be a trained behaviour for docile slimes?

Ilysen:
I am open to different ideas if the concept itself has merit. what do you mean by docile slimes - ones that have become obedient through repeated disciplining, or ones that have been pacified through light pink extract like Kendrick?

Ilysen:
Cere and I talked on discord, and given the current numbers (of slimes feeding on monkeys being vastly more effective than inanimate food - it takes a dozen meat pizzas to grow one to adulthood, for instance, let alone to let it reproduce), she seems to be okay with this, although naturally correct me if I'm wrong.

code/modules/ai/ai_holder_subtypes/slime_xenobio_ai.dm
```diff
@@ -112,8 +112,34 @@
...
+	var/list/yummies
+	for (var/atom/M in view(world.view, holder))
+		if (M.is_slime_food())
+			// Make sure that we can reach the object first, but don't try too hard at it and make sure that the thing is edible beforehand
+			var/list/path = AStar(holder.loc, M.loc, astar_adjacent_proc, /turf/proc/Distance, min_target_dist = 1, max_node_depth = world.view * 2, id = holder.IGetID(), exclude = obstacles)
+			if (path?.len)
+				LAZYDISTINCTADD(yummies, M)
+	if (!LAZYLEN(yummies))
+		return
+	var/atom/yummy = pick(yummies)
```
Atermonera, line 129: What's the purpose of the pick() if you're just going to seek the closest one?

code/modules/ai/ai_holder_subtypes/slime_xenobio_ai.dm
```diff
@@ -126,7 +126,7 @@
 				LAZYDISTINCTADD(yummies, M)
 	if (!LAZYLEN(yummies))
 		return
-	var/atom/yummy = pick(yummies)
+	var/atom/yummy
 	var/closest_distance = get_dist(holder, yummy)
```
Atermonera: What's the distance between `holder` and `NULL`?

Ilysen: gah, I missed that as part of putting in the other commit. it still seems to *work* in the code, so I'm guessing it's something serviceable, but I'll change it to something better regardless

Atermonera: Not sure if this actually applies given it seemed to work, but per the byond reference:
> At this time, get_dist() never returns a value greater than 127.
For a distance in pixels, use bounds_dist().
get_dist() will return -1 in error conditions, such as when Loc1 and Loc2 are the same object.

Atermonera (approved):

PR #9068, opened 2023-03-16
Smol botany update

enbyfriend:
- Adds the botany belt, capable of holding botany related tools, beakers & bottles as well as the floral somatoray
- Gives spades the ability to dig up plants so we don't have to go full texas chainsaw massacre on them whenever we want to remove one
- Adds botany belts to biogenerators
- Adds 2 and 3 spades to xenobotany and botany, respectively


code/game/objects/items/weapons/storage/belt.dm
```diff
@@ -464,3 +464,21 @@
...
+		/obj/item/reagent_containers/glass/bottle,
+		/obj/item/reagent_containers/glass/beaker,
+		/obj/item/material/minihoe,
+		/obj/item/shovel/spade,
+		/obj/item/material/knife/machete/hatchet,
+		/obj/item/analyzer/plant_analyzer,
+		/obj/item/gun/energy/floragun,
+		/obj/item/tool/wirecutters,
+		/obj/item/tool/wrench,
+		/obj/item/multitool,
```
Atermonera, line 483: What are these tools for?

code/modules/hydroponics/trays/tray.dm
```diff
@@ -574,6 +574,29 @@
...
 		to_chat(user, "<span class='filter_notice'>You [anchored ? "wrench" : "unwrench"] \the [src].</span>")
 
+	else if(istype(O, /obj/item/shovel/spade))
+		if(!weedlevel && !seed) // Don't do anything if the tray is truly empty
+			to_chat(user, SPAN_NOTICE("[src] doesn't have any plants or weeds!"))
+			return
+		user.visible_message(SPAN_NOTICE("[user] starts digging out [src]'s plants..."))
+		to_chat(user, SPAN_NOTICE("You start digging out [src]'s plants..."))
+		playsound(src, O.usesound, 50, 1)
+		if(do_after(user, 40 * O.toolspeed) || (!seed && !weedlevel))
```
Atermonera, line 584: This is already checked above. Putting it here will be at the start of the do_after afaik, not the end.

code/game/objects/items/weapons/storage/belt.dm
enbyfriend: The floragun is the floral somatoray, for xenobotany, wirecutters function as improvised hedge trimmers, wrenches are used to move trays around, should you want to, and multitools are used to toggle cryogenic freezing on a plant

code/modules/hydroponics/trays/tray.dm
enbyfriend: Having a little trouble understanding what you mean, but I hope I'm interpreting this correctly and updated it

code/game/objects/items/weapons/storage/belt.dm
Atermonera: Floragun I'm fine with (And don't think was a line I selected for the comment, just auto-shown).
"Improvised" means it doesn't need to be held by a belt dedicated to storing proper tools for a job. E.g., a surgery belt shouldn't be able to hold screwdrivers or welding tools.
I'm also dubious about whether the "move trays around" or "toggle cryogenic freezing" are something that'd be needed regularly enough to really be warranted. Mind, these tools have a _lot_ of other uses, compared to the botany tools listed.

code/modules/hydroponics/trays/tray.dm
Atermonera: This specific condition is checked earlier in the function, so checking it here is redundant.
If it's meant to be checked _after_ the do_after() finishes, then I believe you need to put it inside of the scope of the if statement containing the do_after().
So yes, that is a way to resolve it, if you didn't intend to check that condition after the do_after().

Atermonera:
Chatted about this at the staff meeting, results were, "Eh, it's probably fine. You can already print a toolbelt from the printer anywyas."

Atermonera (approved):

PR #9082, opened 2023-03-21
Updates text sanitization

Varlaisvea:
ok i have
absolutely 0 guarantees this won't like, turbofuck something if someone comes up with a html exploit but ??? i have tried my own html exploits and they don't seem to be working so i'm like 80% sure this is safe ,, ,

amway this pr does the following things:

- lets you use apostrophes in your records and in labellers
- lets you use new lines in traitor fake command announcements, and in request console announcements, and in exploitables
- fixes the bug where if a request console is damaged or deconned and you put it back together you can't use it any more
- adds a check to prevent double-encoding in sanitization

basically the most important part is that it unfucks records tbh
at what cost???

### DIFF
--- a/code/_helpers/text.dm
+++ b/code/_helpers/text.dm
@@ -54,6 +54,9 @@
 	if(extra)
 		input = replace_characters(input, list("\n"=" ","\t"=" "))
 
+	input = html_decode(input) 	//this prevents double-encoding, but does nothing if it's not encoded already.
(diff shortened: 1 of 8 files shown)


Varlaisvea:
[Definitely Requires Code Review Holy Shit]

Atermonera:
Please keep PR titles at least a little serious/on-topic.

Atermonera (approved):

PR #9085, opened 2023-03-27
makes overlays less eyesearing (v2)

Varlaisvea:
so the last version allowed for a toggle but that was spaghetticode

this version just... reduces the alpha so they're less obnoxious, and gives meson-likes the nvg hud (green) instead of the meson hud (sickly green)

coincidentally it's also much smaller of a PR!

### DIFF
--- a/code/_onclick/hud/hud.dm
+++ b/code/_onclick/hud/hud.dm
@@ -41,7 +41,7 @@ var/global/list/global_huds = list(
 
 /datum/global_hud/proc/setup_overlay(var/icon_state)
 	var/obj/screen/screen = new /obj/screen()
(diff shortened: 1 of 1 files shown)


code/_onclick/hud/hud.dm
```diff
@@ -84,7 +84,7 @@ var/global/list/global_huds = list(
 
 	nvg = setup_overlay("nvg_hud")
 	thermal = setup_overlay("thermal_hud")
-	meson = setup_overlay("meson_hud")
+	meson = setup_overlay("nvg_hud") //2023-03-27, formerly meson_hud but nvg_hud is a little less painful to look at tbh
```
Atermonera, line 87: Nit: don't datestamp comments. That's what git blame is for.

Atermonera (approved):

PR #9087, opened 2023-04-01
Refactor of object spawners (of the /obj/effect/spawner variety)

Ilysen:
only the most important changes here from github dot com slash ilysen

#### purpose

object spawners are old code! mostly, though, it annoyed me when I saw that gib spawners lingered forever after being created. once I saw the code, I knew I couldn't leave the poor things as they were.

#### details

* object spawners (`/obj/effect/spawner`) now have their logic handled in a per-type `do_spawn()` proc that's called on Initialize, inheriting the arguments passed to the Initialize call itself.
* removed some unused spawner types.
* revamped gib spawners; they now store their data as a list of associative lists, instead of through three separate variables.

right now, the code uses both these and instances of things like `/obj/random` and `/obj/random_multi`. at some point, it's likely possible to merge the two together, but right now I just wanted to make these particular ones not suck as much.

#### testing

jump to some away sites/secret areas/etc with graffiti spawners and behold as they appear normally. gib something (via admin commands, butcherin' an animal, or so on) and see that the gibs spawn correctly while the spawner itself is deleted immediately.


maps/submaps/surface_submaps/wilderness/demonpool.dmm
```diff
@@ -440,7 +440,6 @@
 	},
 /area/submap/DemonPool)
 "IP" = (
-/obj/effect/spawner/lootdrop,
```
Atermonera, line 443: Er, is this just outright removing the loot from this POI?

Ilysen: I was worried about that, but looking at the code, I don't actually think those particular spawners dropped any loot. it's been a hot minute since I booted up and compared (since this pr has been open a while) but I seem to recall that when I compared the before and after, the stuff the PoI spawns was handled in a different way. that being said, I can make some time to check again if you like

Atermonera: Ah, yeah that code was dependent upon mapping to set the loot list.

Atermonera (approved):

PR #9113, opened 2023-06-10
Sanity check for shredding ID's in the Autolathe

hex37:
Prompts the user if they are sure about feeding ID-bearing items into the Autolathe. Prevents a simple miss-click from locking someone in a room until they can break out.

### DIFF
--- a/code/game/machinery/autolathe.dm
+++ b/code/game/machinery/autolathe.dm
@@ -186,6 +186,16 @@
 			to_chat(user, "\The [mag] is too hazardous to put back into the autolathe while there's ammunition inside of it!")
 			return*/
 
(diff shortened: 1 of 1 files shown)


hex37:
It prevents a miss-click from ruining your access until you can get a new ID printed it is literally a baby gate yes-

That is not a reason it should not be added, however.

Atermonera:
If you misclick and accidentally discard your ID, I kinda feel like that's an interesting scenario? You can certainly break out, but it also more generally requires you to interact with other people. The only case where I think it's downright bad is... when you're the only player, because there's no one else to interact with. And I'm of the opinion that's not a situation we should design around.

PR #9114, opened 2023-06-10
Printable Mops

hex37:
Adds mops to the list of things that the autolathe can print. This allows for the easy construction of ED-CLN Cleaning Robot in robotics.

### DIFF
--- a/code/datums/autolathe/general.dm
+++ b/code/datums/autolathe/general.dm
@@ -2,6 +2,10 @@
 	name = "bucket"
 	path =/obj/item/reagent_containers/glass/bucket
 
+/datum/category_item/autolathe/general/mop
+	name = "mop"
+	path =/obj/item/mop
+
 /datum/category_item/autolathe/general/cooler_bottle
 	name = "water-cooler bottle"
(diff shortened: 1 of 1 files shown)


Cerebulon:
I think the need to ask janitorial or cargo for this specifically to make the "high end fancy nonsense" cleanbot is fine, honestly?

Atermonera:
... But why though?
It does also make it really easy for a griefer to get access to a mop, which is probably one of the best ways to wet a large floor area. I'm inclined to agree with Cere, but I'll give it some more thought.

hex37:
> ... But why though? It does also make it really easy for a griefer to get access to a mop, which is probably one of the best ways to wet a large floor area. I'm inclined to agree with Cere, but I'll give it some more thought.

No real reason other than the one I gave. If that isn't enough to justify it, that is fine.

Atermonera:
Oh this has been up 3 weeks, huh.
I've discussed this with a few other members of staff, and they were also generally hesitant. There's not really a shortage of ways to get mops if you go and talk to people, designing around lowpop is not my priority, and the ED-CLN is kinda high-end.

PR is denied.

PR #9128, opened 2023-06-29
Faction Update

elgeonmb:
- Citizenship now points at what country you hold citizenship in. Default is SCG.
- Home System updated to include major human population  / cultural centers. I resisted the mighty urge to add Zhu Que. EUM removed because nobody is actually from there.
- Some "new" corporations from the brainstorming session like 6 years ago added to factions, a selector whose very existence I continue to protest. Default faction is now NT.
- Starlit Path and the Unity added to religion selector.
- "Antag Factions" populated with our list of criminals + Vystholm. Resisted the urge to add foreign governments and shady state departments despite many characters covertly working for them.

Changes to the default or the selector list will not impact existing character slots.

### DIFF
--- a/code/modules/client/preference_setup/general/05_background.dm
+++ b/code/modules/client/preference_setup/general/05_background.dm
@@ -24,7 +24,7 @@
 
 /datum/category_item/player_setup_item/general/background/sanitize_character()
 	if(!pref.home_system) pref.home_system = "Unset"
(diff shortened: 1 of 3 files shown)


elgeonmb:
oh this might have failed wildly because i did not check it lmfao.

Atermonera (approved):

PR #9129, opened 2023-06-29
Plains POI - Rest Stop

ephemeralstoryteller:
Adds a relatively simple POI to the plains pool: an ambitious 'rest stop' that once resupplied explorers and travelers before being ultimately abandoned. The goal was for this to not only be a fun location to explore and scavenge, but its multiple modes of entry and exit make it an interesting base for antags, drakes, and survivalists alike.

My first map! Assistance and feedback is appreciated.

image

### DIFF
--- a/maps/submaps/surface_submaps/plains/plains.dm
+++ b/maps/submaps/surface_submaps/plains/plains.dm
@@ -288,3 +288,9 @@
 	desc = "A temporarily abandoned VR den, still functional."
 	mappath = 'maps/submaps/surface_submaps/plains/VRDen.dmm'
 	cost = 10
+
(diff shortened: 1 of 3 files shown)


Atermonera (approved):

Cerebulon:
I know I'm too late now (i was away all week) but there are shelving units in addition to racks for future reference, I think this might have been a spot where they'd have fit. O well, no biggie!

ephemeralstoryteller:
> I know I'm too late now (i was away all week) but there are shelving units in addition to racks for future reference, I think this might have been a spot where they'd have fit. O well, no biggie!

ty for the comment regardless! I actually used those in one of the more recent builds!
image

PR #9142, opened 2023-07-04
Fixing Nanopaste for prosthetic organs

hex37:
Fixes a bug that caused prosthetic organs in the lower body to be impossible to heal with nanopaste.

### DIFF
--- a/code/modules/surgery/robotics.dm
+++ b/code/modules/surgery/robotics.dm
@@ -273,7 +273,7 @@
 		if(I.damage > 0 && (I.robotic >= ORGAN_ROBOT))
 			is_organ_damaged = 1
 			break
-	return affected.open == (affected.encased ? 3 : 2)) && is_organ_damaged
+	return affected.open == (affected.encased ? 3 : 2) && is_organ_damaged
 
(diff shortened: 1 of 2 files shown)


code/modules/surgery/robotics.dm
```diff
@@ -8,17 +8,35 @@
 
 /datum/surgery_step/robotics/can_use(mob/living/user, mob/living/carbon/human/target, target_zone, obj/item/tool)
 	if (isslime(target))
+		#ifdef SURGERYDEBUG
+		log_debug("surgery_step/robotics/can_use Surgery step returned 0; target is slime")
```
Atermonera, line 12: Do we need to leave these debug logs in? They're kinda... a lot.

hex37: I recommend leaving it in because stuff is legit about the only way you're going to debug surgery code, but I can remove them if asked.

hex37: Updated to remove all debugging.

Atermonera: Thanks. I appreciate that debug text messages are frequently the way to go in ss13, I think leaving them lying around like this hurts code readability overall.

Atermonera (approved):

PR #9143, opened 2023-07-04
Blob Chunks 2

### LINKED ISSUE #8903: Blob core chunks do not spawn (/exist?) anymore.
When killing a blob, it is supposed to drop a core chunk with a specific effect. It doesn't do so anymore.

For the blob to drop a core chunk related to the type of blob.

#### What actually happened
There is no core chunk.

#### Steps to reproduce
Kill a blob core, then see how there's no core chunk.

#### Additional info:
 - **Anything else you may wish to add:** (Location if it's a mapping issue, etc) Talked about this briefly with Mechoid; It is assumed that this is the result of #8783.

Mechoid:
Restores blob chunks, utilizing the artifact system for effect control, instead of the blob types themselves.

WIP

Ater edit: fixes #8903


Mechoid:
From my previous conversation with @Atermonera blob chunks were removed on their side almost solely for backend reasons, the only 'frontend' reason mentioned being they were unreliable to acquire.

Atermonera:
It's been long enough that I don't rightly remember specifics of any bugged chunks that I was disgruntled by. Mechoid has the right of my primary motivations.

Atermonera (approved):

PR #9145, opened 2023-07-06
POI - Collapsed Mine

ephemeralstoryteller:
Squashed and merged edition!

This POI is an old Grayson-funded mining operation that found itself in hot water when they discovered they had built atop a spider nest. Now the facility's in chaos, and it doesn't help that its former occupants ordered their drones to contain the problem however possible before they all expired...

Main features

- Many rooms are blocked off by rubble, with plenty of tools for players to carve their own paths... so long as they're careful where they're drilling.
- Two hostile factions, spiders and hostile drones. With some clever manipulation of the terrain, the two can even be made to do some of the fighting themselves.
- Several anomaly spawn points, to make going here actually lucrative. Guns, ammo, and tools remain in the clutches of their former owners or piled in former strongpoints that are now dens of the enemy, and plenty of heaps of scrap wait for the clever scavenger to exploit.

overview image, more detailed looks available at request:
image

### DIFF
--- a/maps/submaps/surface_submaps/wilderness/collapsedmine.dmm
+++ b/maps/submaps/surface_submaps/wilderness/collapsedmine.dmm
@@ -1,0 +1,6950 @@
+//MAP CONVERTED BY dmm2tgm.py THIS HEADER COMMENT PREVENTS RECONVERSION, DO NOT REMOVE
+"ae" = (
+/obj/structure/ore_box,
+/obj/effect/floor_decal/corner/yellow/border,
+/turf/simulated/floor/tiled/techfloor,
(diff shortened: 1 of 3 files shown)


ephemeralstoryteller:
Not sure how to resolve 'active map edges' - which was occurring with another PR of mine too.

Atermonera:
That's a flake in CI - basically there's a chance for cave generation to make caves next to the railroading (Planetary wall turfs that maintain the weather-based atmosphere), and when such caves occur, their air mix is different than the railroading so it's considered an active edge.

Atermonera (approved):

PR #9147, opened 2023-07-07
cold lizard hours

hex37:
- Unathi now take cold damage at the same thresholds as humans

- Unathi experience movement slowdown from the cold earlier and more intensely than other races

- Unathi are weakened when they experience cold that is too great

Totes fine if it is decided that this PR should not go through, but I needed a distraction and thought it was funny, so.


hex37:
closed due to a lack of interest from staff/playerbase

PR #9150, opened 2023-07-09
POI - add more warning to chasm trap

ephemeralstoryteller:
Spoilers ahead for an existing POI!

The chasm POI remains one of the most instantly lethal in the rotation, as it has a set of loose rock stairs within that can catapult you over a cliff and into a pit of spiders and radiation - a trap unique to this POI and basically guaranteed death. Short of reworking the entire POI, I've added a custom description to the loose rock stairs to further clarify that they are dangerous, as 'loose stairs' could easily be mistaken for a bit of harmless flavor. 

The goal of this PR is to give a little more of a fighting chance to those who do not have the meta knowledge that the stair trap exists, because it is so lethal and so easily avoided if the gimmick is known.

I have also added a 'battered wet floor sign' near the stairs that could've plausibly been catapulted out of the blast zone, just a bit of the SAR's dark humor and an extra clue to the intrepid.

The wet floor sign is absolutely up for negotiation, as it might make the trap 'too easy'. Its purpose is to make dying to the trap a bit funnier in the worst case (janitor points to the sign), and in the best case the clever will notice the unique text and be better primed to see the stairs.


Greenjoe12345:
I thought the slippery stairs were removed?

ephemeralstoryteller:
> I thought the slippery stairs were removed?

As far as I can tell (unless done by an open PR I didn't see), they have not been

Atermonera (approved):

PR #9154, opened 2023-07-14
Addition of a Research Testing site

Woodratt:
- Adds a small dilapidated testing site for Research to tinker with that's somewhat removed from the rest of the station in the underground area (slightly off the path to xenoarch) In short a small control room with a SMES facing a pad.

- Added a couple concrete pillars to the underground (underneath the corners of the pool and library) to help give a vague idea where the station is.

[Image Link](https://i.imgur.com/lEcNif3.png)

### DIFF
--- a/maps/cynosure/cynosure-1.dmm
+++ b/maps/cynosure/cynosure-1.dmm
@@ -171,6 +171,13 @@
 	},
 /turf/simulated/floor/tiled/techfloor/grid,
 /area/surface/station/rnd/mixing)
+"av" = (
+/obj/effect/floor_decal/rust,
+/obj/machinery/door/airlock/maintenance/rnd{
+	name = "Research Testing Site"
+	},
+/turf/simulated/floor/tiled/techfloor,
+/area/surface/station/rnd/testingsite)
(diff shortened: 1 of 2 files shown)


hex37:
Do you think maybe some pipes could be ran out there to help with temperature regulation in the observation room?

Woodratt:
Then it would require an airlock system, as there is only currently one door between it and the exterior. Otherwise, every time you open the door it will bleed heat to space. Somewhat out of scope for a place that is supposed to be somewhat disused and rundown and most likely will not be used that often. Best I can suggest is to snag a space heater.

N-Komaeda:
When pitching the research testing site, an important factor was that it had to be as barebones as possible. Yes, there's going to be logistical issues - I preview one with gas harvesting, where we'll either need to make it pump into canisters, or we need to hook it up to the scrubbers line. But that is very much a part of the charm, I feel. Having this room as is leaves open a lot of possibilities.
That, and if the cold air is your most pressing concern, you've got things sorted out pretty well.

Atermonera (approved):

PR #9157, opened 2023-07-14
Changes Phoron Borer SFX

Doctress:
No more confusing it for a Skath jumping on you.


Cerebulon:
If this sound is literally from Half Life Opposing Force we cannot use it.

PR #9162, opened 2023-07-18
Introducing the New Bing

Cerebulon:
Allows FBPs to bing again.


Atermonera (approved):

Atermonera:
It seems that the old bing was not good enough.

PR #9170, opened 2023-07-23
Ports several Unathi and Skrell hair from Neb and Bay

MistakeNot4892:
- Ports a split tendrils state for Skrell from Nebula, which I believe came from a downstream fork. I'm not sure of the original source.
- Ports several horn and frill styles from Bay, with conversion to match Polaris' colour ramp. ~~I believe most of these were done by Azlan.~~ These are Pawn's work originally.


Sypsoti:
Thumbs up from Unathi maintainer.

Atermonera (approved):

PR #9177, opened 2023-07-27
Adds a volume channel for system sounds (roundstart, votes, etc)

Ilysen:
#### purpose

I've found always found the roundstart sound to be uncomfortably loud, and this is generally because most sounds in the game use 50 volume as a baseline and that one, like almost all global sounds, uses 100 volume - double that of most sounds use.

#### details

adds a new volume channel, System, to the volume menu. roundstart sounds, pregame lobby sounds, and vote start sounds are all tied to this channel, and its volume can be controlled like any other channel, allowing them to be made quieter

this is just one way of doing it - it's also possible to adjust the default or to skip the extra channel in favor of just reducing the default volume, but as a stickler for customization I liked the idea of letting folks have fine-grained control over what is purely a clientside thing

byond appears to cap sound volume at 100, so the roundstart sound (mercifully) cannot be made any louder - just quieter

#### testing

started a buncha rounds on my local server using various settings on the slider. the default should sound unchanged from the current version, but it can be changed in either direction now. I also tried with a guest account having it muted and another account having it normally to make sure the sounds played correctly, but admittedly I don't really have a way of testing it on a large scale, sadly

#### media

100% volume, the current default:

https://github.com/PolarisSS13/Polaris/assets/47678781/aeb40b56-680c-484a-9778-be9b300920c7

50% volume:

https://github.com/PolarisSS13/Polaris/assets/47678781/0b7f9721-708d-4f1d-8493-240912522db4

### DIFF
--- a/code/__defines/misc.dm
+++ b/code/__defines/misc.dm
@@ -335,6 +335,7 @@ var/global/list/##LIST_NAME = list();\
 #define VOLUME_CHANNEL_AMBIENCE "Ambience"
 #define VOLUME_CHANNEL_ALARMS "Alarms"
 #define VOLUME_CHANNEL_DOORS "Doors"
+#define VOLUME_CHANNEL_SYSTEM "System"
 
 // Make sure you update this or clients won't be able to adjust the channel
 GLOBAL_LIST_INIT(all_volume_channels, list(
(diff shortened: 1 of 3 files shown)


Atermonera (approved):
I'm a little wary of the vote volume, since people have historically complained that it was too quiet.
But I guess this does give people a way to actually change it, so maybe that can be less of an issue in the future.

Ilysen:
if it tends up bein iffy, we can always change it to be on the master channel or cranking up the default volume. either or

PR #9191, opened 2023-08-02
Refluffs spellbooks as dubious augment applicators

Cerebulon:
So wizard was essentially removed but the code for it is still in place because it's used... Occasionally, for some cult stuff, I think? But nobody ever sees 90% of it.

Except for spellbooks which show up once in a blue moon because they're in at least one PoI, and have a novel lil function... But it makes no god damned sense in-universe! So:

- Spellbooks are "Nanomachine Hives", which rapidly apply some very experimental very illegal tech to your body. Which is why they're incredibly temperamental and weird.
- Also changes most of the text around spellbooks, even though the non-single-use version is no longer used anywhere because Wizard isn't an active mode. Just in case.
- Also fucks with some text on the fireball spell so you're not obviously casting a literal magic spell, but are still obviously doing something a little fucked up that people can RP however they like.
- Would make the Wizard mode a little wonk, but Wizard was replaced by Technomancer by Neerti literally over 7 years ago, so fuckem.
- New sprites are a weird tech-box but still vaguely evoke the shape of a book and use the old cover images.

The only ones which are currently available anywhere are still fireball and forcewall which is probably for the best. But the other can be spawned, in theory.


code/modules/spells/spellbook.dm
```diff
@@ -1,8 +1,8 @@
 /obj/item/spellbook
-	name = "spell book"
-	desc = "The legendary book of spells of the wizard."
-	icon = 'icons/obj/library.dmi'
-	icon_state ="spellbook"
+	name = "nanomachine hive" //"Wizard" may be long deprecated but since these wizard-based items still show up on rare occasion for gimmicks...
+	desc = "A secure delivery system for highly experimental nanomachine swarms capable of rapidly installing (often illegal) biotech augments to those both educated in its usage, and reckness enough to try."
+	icon = 'icons/obj/technomancer.dmi'
```
Spookerton, line 4: These aren't related to technomancer - use a new dmi

Spookerton (changes requested):

Spookerton (approved):
i feex dmi

Atermonera (approved):

PR #9192, opened 2023-08-02
Fix records being wiped when hitting 'cancel' on input

Cerebulon:
Introduced by #9082 

Record inputs were doing html_decode twice, since it's built into sanitize now too? And if cancelled then the input had nothing to decode but saved anyway etc. etc.

I question the need for html_decode to be _built in_ to sanitize() as a whole but that may bea  whole thing and this is just a fix for the main issue which is annoying everybody including myself, while maintaining the intended change of "apostraphes work in records". The double html_decode isn't really an issue for most textboxes like this where hitting cancel probably DOES mean you don't give a shit about what was in the box but people sure would like to not accidentally wipe their lengthly character pieces!!!


Atermonera:
Oh it's because sanitize is turning null into "". And `"" != null`, so it gets into the save check.

Atermonera (approved):

PR #9212, opened 2023-08-10
Digging update.

MistakeNot4892:
- Digging on help intent will not dig a plot or grave, but will dig for buried stuff like worms/rocks.
- Digging on harm intent will dig a grave.
- Digging on any other intent will dig a soil plot.
- Filling in a grave with nothing inside it will delete the grave.

### DIFF
--- a/code/game/turfs/simulated/outdoors/outdoors.dm
+++ b/code/game/turfs/simulated/outdoors/outdoors.dm
@@ -27,6 +27,11 @@ var/global/list/turf_edge_cache = list()
 	var/can_dig = FALSE
 	var/loot_count
 
+/turf/simulated/floor/outdoors/examine(var/mob/user)
+	. = ..()
+	if(!has_snow())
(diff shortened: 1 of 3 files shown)


Spookerton (approved):

code/game/turfs/simulated/outdoors/outdoors.dm
```diff
@@ -27,6 +27,11 @@ var/global/list/turf_edge_cache = list()
 	var/can_dig = FALSE
 	var/loot_count
 
+/turf/simulated/floor/outdoors/examine(var/mob/user)
+	. = ..()
+	if(!has_snow())
+		. += "A shovel could be used on help intent to dig up worms or rocks, harm intent to dig a grave, or any other intent to dig a plot for planting."
```
Atermonera, line 33: For what good it normally does, do you think this should also go in the shovel's `extended_desc`?

PR #9214, opened 2023-08-13
Fix robobags icons, synthmorph->synthbody

Cerebulon:
Fixed icons for "synthmorph bags" being broken, and also lacking any code to use the 'visibly a body inside' version that existed.

But also renamed them to "synthbody bags" because the term "synthmorph" was added by a primarily-downstream coder offhandedly years ago and was inconsistent with every other reference to synthetic bodies in the entire game and setting. And as far as I can tell, it's almost entirely an Eclipse Phase term which has strong ties to contrasting with "biomorphs" specifically in the context of Resleeving and Bodies As Something Which Is Merely Inhabited. Which is not a thing in the setting....... Anyway 90% of the usage of the term was in code comments referencing these bodybags. Even ctrl+ synthmorph in the discord is entirely people talking about eclipse phase years ago!


code/game/objects/items/robobag.dm
```diff
@@ -32,8 +32,15 @@
 		. += "<span class='notice'>[src] has a [corptag] attached to it.</span>"
 
 /obj/structure/closet/body_bag/cryobag/robobag/update_icon()
+	if(opened)
+		icon_state = "open"
+	else if (contents.len)
+		icon_state = "closed_occupied"
+	else
+		icon_state = "closed_unlocked"
```
Spookerton, line 40:
```suggestion
		icon_state = "closed"
```
In addition, the state this aims at in the DMI does not have a name and needs correcting

code/game/objects/items/robobag.dm
```diff
@@ -32,8 +32,15 @@
 		. += "<span class='notice'>[src] has a [corptag] attached to it.</span>"
 
 /obj/structure/closet/body_bag/cryobag/robobag/update_icon()
+	if(opened)
+		icon_state = "open"
+	else if (contents.len)
```
Spookerton, line 37:
```suggestion
	else if (length(contents))
```
always avoid `.len`

Spookerton (changes requested):

code/game/objects/items/robobag.dm
Cerebulon: the unnamed icon state + closed_unlocked (presumably a carryover from lockers in general when lockers were redone a few years ago) is the same for all types of bodybag, so it's consistent(ly weird and fucky) right now so I'll correct this for all of them in that case.

Spookerton: freaky, but for another PR

Spookerton (approved):

PR #9218, opened 2023-08-14
respawn timers are correctly applied to ghosting and observing

Spookerton:
causes the `Allow player to respawn` admin button to consequently work correctly as well

also bugfixed GLOB.respawn_timers having a null entry whenever a set_respawn_timer() call couldn't resolve a ckey


Greenjoe12345:
What are the respawn timers set to on polaris anyway?

Spookerton:
config.txt doesn't set anything on the server, so the default of 5 minutes

Atermonera (approved):

PR #9230, opened 2023-08-29
removes tesharii

Greenjoe12345:
but not teshari
removes the improper usage of tesharii, which isn't a thing, and also replaces instances of Sirisaian with the proper Sirisaii


Cerebulon:
Critical bird maintenance.

Spookerton (approved):

PR #9236, opened 2023-09-06
Adding survivalist and drake options to jobban panel.

MistakeNot4892:


Atermonera:
Does this fix explorer being missing from the jobban panel as well? That's also been reported missing.

Atermonera (approved):

PR #9238, opened 2023-09-08
Poker chips for playing with cards

woody41:
Hi,

this is PR where I would like to add new poker chips. This PR contains few new textures for tokens and one new texture for box as a holder of chips. cynosure-2 has change of spawn location of case with poker chips - library on table under d20. 

This is my first DM code ever and I don't know best practice, so if something is bad, let me know.

:cl: 
rscadd: Added poker chips for playing games and betting. They can be split and joined like money. A box of them spawns in the library.
/:cl:


code/modules/economy/chip.dm
```diff
@@ -0,0 +1,200 @@
+/obj/item/chip
+	name = "0 Chip"
+	var/initial_name = "Chip"
+	desc = "It's worth 0 Chips."
```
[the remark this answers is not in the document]
woody41: it's just placeholder for init. You cannot have 0 value chip.

maps/cynosure/cynosure-2.dmm
Spookerton, line 1: The dream maker editor isn't good and produces a format that's bad for git diffs -  mapping changes should be done with strongdmm instead:
https://github.com/SpaiR/StrongDMM/releases/tag/v2.8.1.alpha
or if that crashes:
https://github.com/SpaiR/StrongDMM/releases/tag/v1.9.1

code/modules/economy/chip.dm
```diff
@@ -0,0 +1,200 @@
...
+				var/mob/living/carbon/human/H = user
+				var/obj/item/organ/external/temp = H.organs_by_name["r_hand"]
+
+				if (H.hand)
+					temp = H.organs_by_name["l_hand"]
+				if(temp && !temp.is_usable())
+					to_chat(user,"<span class='notice'>You try to move your [temp.name], but cannot!</span>")
+					return
+
+
```
Spookerton, line 111: avoid random empty lines

code/modules/economy/chip.dm
```diff
@@ -0,0 +1,200 @@
...
+	icon_state = "chip1"
+	opacity = 0
+	density = 0
+	anchored = 0.0
+	force = 1.0
+	throwforce = 1.0
+	throw_speed = 3
+	throw_range = 10
+	w_class = ITEMSIZE_SMALL
+	var/access = list()
```
Spookerton, line 16: unused var

code/modules/economy/chip.dm
```diff
@@ -0,0 +1,200 @@
...
+	var/worth = 0
```
Spookerton, line 17: I'm inclined to say this should be done with types, since these chips do not actually have a worth

Spookerton (changes requested):
also, needs squash

Spookerton:
Regarding the above, I'll fix it up.

Spookerton (approved):

woody41:
@Spookerton When you fixed the code you added a bug. Now you cannot take specific amount of poker chips, but you have to take each chip one by one. My code: When you click on chips you can select amount of worth what you want to take. No you have to select a specific value from predefined list. That's not good. I think it is caused by removing worth variable.

PR #9248, opened 2023-09-15
POI - Remap CaveS

ephemeralstoryteller:
CaveS, the spider cave, suffers from an overabundance of cloaking spiders and loot somewhat haphazardly scattered about its chambers. It has some nice touches, like the implication that this is a criminal den of some sort that was overrun, which I've sought to maintain and expand on while making it feel both more like a part of the natural ecosystem and an interesting dungeon.

Features
- Has multiple points of entry
- Separates spiders into arenas
- Adds some viscerators to the interior chambers, for variety
- Generally livens the place up a bit

Before:

After:


Spookerton (approved):

Spookerton (approved):
squashed

PR #9255, opened 2023-09-26
Graffiti centers on where you click.

Mechoid:
Tin.
Persisting graffiti will also save its pixel shift.


Cerebulon:
You're doing half of this in two different places.
 
/obj/effect/decal/cleanable/crayon (Crayon doodles), where you've applied the "appears where you click" feature is a distinct thing from /obj/effect/decal/writing (Engraved graffiti).

Crayon doodles are handled by the Filth persistence system, so you'll need to include the pixel x/y save/load there.
Engraved graffiti (the Graffiti persistence system) will need the on-click positioning where it's handled, probably on turf/proc/try_graffiti

Mechoid:
Overlapping engravings aren't too much of an issue in my opinion, as if they get to the point of being obstructive somehow, Engineering should already have been notified of their presence.

Atermonera (approved):

PR #9256, opened 2023-09-27
Adhesive posters

Greenjoe12345:
Adds adhesive posters, functioning like sticky notes, they can be written on and save between rounds. There are some in the art supplies room, a yellow stack in engineering, a blue stack in medical, a special command stack in the bridge and a stack in the union office.


elgeonmb:
in the grim darkness of the 26th century, we still use the same standard paper sizes

Spookerton (approved):

PR #9264, opened 2023-10-07
Demolishing Mobs

Mechoid:
Space Worm given a maintenance pass.
- Able to charge toward a target, devouring objects or mobs in their path.
- Able to target mobs through walls, and damage many types of obstructions normally ignored by simple mobs. (Both are toggles on the AI datum.)

Thermic spiders given a "tactical" role for giant spiders, acting as breachers.
- Able to demolish walls and other common obstructions, similar to the worm.
- Variant for PoI or adminspawn capable of targeting through walls similar to the worm.

Phorogenic Spiders
- Able to demolish walls & other common obstacles given the fact they are massive and unlikely to care about a few twigs in the way. Still slow though.

Demolishing behavior does not instigate ally-calling as normal targeting does, unless the ally _also_ shares the ability to demolish.


Cerebulon:
I have some balance concerns about a mod which afaik is in the Random Spider Rotation for PoIs, being able to target and attack you through walls which it can also destroy? It seems like a thing which should be a special kind of mob and not just something which can totally alter a poi by randomly spawning?

Spookerton:
Doubt the 10% reduction addresses the concern; tagged for contention.

Atermonera (changes requested):
Requesting subtypes of phorogenic & thermic that get demolishing, and are explicitly _not_ in the random spider spawn rotation.

code/modules/ai/ai_holder_targeting.dm
```diff
@@ -29,18 +29,19 @@
...
-	for(var/HM in typecache_filter_list(range(vision_range, holder), hostile_machines))
-		if((ignore_opacity && (HM in range(vision_range, holder))) || can_see(holder, HM, vision_range))
-			. += HM
+	for(var/atom/movable/AM in range(vision_range, holder))
+		// Ignoring opacity makes this a slightly hungrier proc as we're now not just checking range for specific machine types, but also living things outside hearing but inside range.
+		if(ignore_opacity && (isliving(AM) || ((ispath(preferred_target) && istype(AM, preferred_target)) || (istype(AM) && AM == preferred_target))))
+			. |= AM
+
+		if(hostile_machines[AM.type])	// Is the atom in the cache of machines we care about
+			if(ignore_opacity || can_see(holder, HM, vision_range))
```
Atermonera: Do you mean AM?

Atermonera (approved):

PR #9267, opened 2023-10-14
Small shuttle additions

Greenjoe12345:
Alters the arrival and departure shuttles slightly, adding spots for wheelchair users, as well as arrow markings by the doors.


Cerebulon:
I don't think there's really need for _two_ spots, especially since spots are already kind of at a premium at higher pops and I've never seen two wheelchair users at once arriving/departing. Maybe should have got feedback on this in suggestions?

Greenjoe12345:
> I don't think there's really need for _two_ spots, especially since spots are already kind of at a premium at higher pops and I've never seen two wheelchair users at once arriving/departing. Maybe should have got feedback on this in suggestions?

I haven't really seen the shuttle be completely filled before which is why I thought two spots was okay.

Atermonera (approved):

PR #9278, opened 2023-10-29
Fixing some grammar in plant strings.

MistakeNot4892:


out-of-phaze:
thank you now i can rest in peace knowing that this typo on a server i don't play on and only found by chance is fixed

Spookerton (approved):

PR #9283, opened 2023-11-27
area blurbs

Greenjoe12345:
This adds a system for area blurbs/descriptions. These automatically show upon entering the area, once per person, and can be shown again via a verb.
ported from https://github.com/Aurorastation/Aurora.3/pull/17741 , which ported from https://github.com/Occulus-Server/Occulus-Eris/pull/596, with a fix from https://github.com/NebulaSS13/Nebula/pull/3533


Greenjoe12345:
Currently has an issue with the blurb showing repeatedly on different subtypes of the area, which is unintended.

Greenjoe12345:
This should be good to check over now.

Atermonera (approved):

PR #9293, opened 2023-12-24
Scratches up the bar communicator take 2

Greenjoe12345:
post-event fluff


Cerebulon:
It was fully pushed off a table and rolled :P 
The description I used for it during the gala was:
A static communicator unit, seen in homes, offices and the receptions of businesses across the galaxy. It looks badly dented?";

So if we're do this, may I suggest:
"A static communicator unit, seen in homes, offices and the receptions of businesses across the galaxy. The casing on this one is dented and covered in scuff marks.";
So the question becomes "wtf happened to this, make your own theories"

Atermonera (approved):

PR #9300, opened 2023-12-29
remove tcomms scripting

Spookerton:
this ancient thing allows various sorts of malicious behavior and has long since been removed from many other codebases

technically a user facing change, but the actual machinery to use it isn't mapped anywhere

Also comsumed tcomms housekeeping and remove /tmp/ since then they're not conflicting/dependent.


Atermonera (approved):

Spookerton:
conflict resolution, no changes

Atermonera (approved):

PR #9302, opened 2023-12-29
examine signature changed to include distance

Spookerton:
- distance is INFINITE, 0, or a positive integer based on get_dist() between the examiner and examinee's turfs when they are on the same Z. When the examiner is dead or a ghost, distance is always 0.
- also updates /examine( implementations to use it!
- adds w_class_to_name proc for reusing "It is a X item." examine behavior.

nufcy


Atermonera (dismissed):

Spookerton:
conflicts handled

Atermonera (approved):

PR #9303, opened 2023-12-29
replace legacy proc/between with clamp builtin

Spookerton:


Atermonera (approved):

Spookerton:
except that this would destroy everything because between is L V H and clamp is V L H
Time for a redo 🙃

Atermonera (dismissed):

Spookerton:
more conflict resolution

Atermonera (approved):

PR #9307, opened 2024-01-01
teshari meat looks like chicken

Spookerton:
by request 😉


Greenjoe12345:
Maybe the color might want adjusting cause of teshari blood color being different?

Atermonera (approved):

PR #9313, opened 2024-01-05
charming winter clothes

Greenjoe12345:
Ports some winter gear and charms from Aurorastation
Wooly hats, ear warmers, and mittens, all recolorable.


Spookerton (approved):
author credits would be nice, as usual

PR #9315, opened 2024-01-08
Adds nugget share packs.

MistakeNot4892:
- Adds nugget boxes.
- Lowers nugget protein to 3 from 4.
- Makes the nugget recipe produce 4 nuggets (6 protein from chicken plus 5 flour).


Spookerton (approved):
nugét

PR #9320, opened 2024-01-18
Generalizing drying racks and adding jerky.

MistakeNot4892:
# Map changes
- Added campfire to the fork in the road between Cynosure, the wilderness and the checkpoint.
- Expanded the wilderness shack to add some staple foods, a stove, some matches, some logs, some survival gear, and a latrine.

# Drying changes
- Generalizing the drying backend to `/obj/item` and tied it into the grilling system.
- Renames/repaths drying racks to drying ovens.
- Renames/repaths tanning racks to drying racks.
- Drying racks can be dismantled with a wrench.
- Adds growns, meat, spidermeat and fish to the drying rack.
- Adds jerky items made by drying meat, spidermeat, fish and cutlets. Sprites by @evreeone!
- Increased the time taken for leather to dry.

# Bonfire changes
- Repathed logs and wood to a shared `/fuel` type.
- Refactored/collapsed the bonfire backend and added fireplaces, campfires and stoves (icons from Nebula, unsure of the original source, will edit when LFP gets back to me hopefully).
- Removed the permanent bonfire types and reworked existing types to use the material of the fuel instead of the crafting fuel.
- Buffed fuel times for logs and wood, and added a dried wood type with a higher fuel value.
- Added grilling noise and fire crackle to the bonfire.

# Other food changes
- Added grilled fish (1 fish fillet, in the oven). Icons also by @evreeone.
- Increased the time taken for grilling from 20 seconds to 1 minute for meat and 40 seconds for fruit.


code/__defines/materials.dm
```diff
@@ -14,7 +14,7 @@
 #define MAT_WOOD			"wood"
 #define MAT_LOG				"log"
 #define MAT_SIFWOOD			"alien wood"
```
Greenjoe12345, line 16: With changing alien log to sifwood log, wouldn't this want to be changed to sivian wood?

Spookerton (approved):
Seems operable!

code/__defines/materials.dm
Spookerton: this can be a followup tbh

PR #9324, opened 2024-01-21
Revised Sivian Fauna Lists

ephemeralstoryteller:
- Audits the Sivian creature tables and removes some of the more inappropriate entries like normal crabs, geese, and tamaskans, swapping them with more relevant creatures like the Sivian shelf crab, sakiim, or frostducks.
- Makes a new spawn pool for aquatic creatures for PoIs so that mappers can prevent shelf crabs from spawning far inland, also implements this in the Swamp Den PoI.
- Also normalizes the wander distance, setting it to 10 for the peaceful creatures to prevent them from getting too far away from their dens.

Siffets would've made it onto these lists but their ability to retaliate against non-small aggressors seems to be currently broken.
Karik could also probably use the ability to retaliate but I was unable to fix this with the limited knowledge I have.

### DIFF
--- a/code/game/objects/random/mob.dm
+++ b/code/game/objects/random/mob.dm
@@ -73,16 +73,17 @@
 	mob_wander_distance = 10
 
 /obj/random/mob/sif/item_to_spawn()
-	return pick(prob(30);/mob/living/simple_mob/animal/sif/diyaab,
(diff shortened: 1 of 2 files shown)


Spookerton (approved):
cod

Spookerton:
ah yes, autofocused input boxes

PR #9131, opened 2023-06-30
Bidons and Chemistry QOL

hex37:
BIDONS

- Ports BIDON containers, portable and refillable chemical tanks, from Eris

- Adds the various BIDON Containers to R&D's protolathe

  - The Basic Bidon has a large storage supply
  - The Advanced Bidon functions as a giant Cryostasis beaker
  - The Trigger Bidon, a potential antag tool, is an advanced bidon that can be set to deactivate its cryostasis component after 2 minutes elapse.

- Beakers will now attempt to totally empty themselves if used with Disarm intent

PR #9326, opened 2024-01-31
add a cut down version of eris bidons

Spookerton:
closes #9131

The other PR's non functional (even pre-conflict) and littered with bad port debris, so I rewrote it largely from scratch. I think this is a bad thing to add, but people seemed to care about it last august.

A bidon is a reagent container structure (like a fuel tank) that you can put stuff into but can't attach to chem sprayer hoses. Clicking a bidon with an open hand adjusts whether its lid is on or off.

A stasis bidon is a bidon where the contents don't react. You can attach a timer assembly to a stasis bidon that causes its contents to react at the end of the timer period. A multitool toggles the timer on or off. A screwdriver removes the timer again.

You can add to bidons by clicking them with a reagent container (like a beaker) on disarm intent, which empties the container into the bidon so long both are open. This behavior carries over to all other reagent container interactions, which is horrible but previously got a green tick.

examine behavior expects #9302 to already be in.


Spookerton:
author request

